PR #109, opened 2012-09-17
Add Python 3.x support for scripting

dmeranda:
This patch makes it possible for FontForge to use either Python 2.x or Python 3.x; though FontForge has to be compiled against only one Python version of your choosing.  For complete details see the commit messages.

There is little end-user visible changes, except of course that Python 3 can be used for writing scripts.  The bulk of the changes significantly overhaul the code, especially related to all the python initialization functions.

The visible differences include:
- Initialization scripts are now searched for in a python-specific config dir first
- The fontforge and psMat modules are now automatically imported in embeded-python mode
- The fontforge module has two new methods: scriptPath() and runInitScripts()
- The names of some secondary python classes have changed to be more consistent. Mostly this affects classes that are transient and can not be instantiated, like iterators, so the change would only matter if you attempted to use isinstance() for example.


latsami:
Hello, this is an old post, but Arch Linux just changed its installation package to build Fontforge with python3. I'm adapting a script to work with python3, but it seems I cannot import both fontforge and psMat modules:

    ➜  glyphtracer-1.4 git:(master) ✗ python
    Python 3.4.3 (default, Feb 26 2015, 23:01:07) 
    [GCC 4.9.2 20150204 (prerelease)] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> import fontforge
    >>> import psMat
      File "<stdin>", line 1, in <module>
    ImportError: 'psMat' is not a built-in module
    >>> 
    ➜  glyphtracer-1.4 git:(master) ✗ python                        
    Python 3.4.3 (default, Feb 26 2015, 23:01:07) 
    [GCC 4.9.2 20150204 (prerelease)] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> import psMat
    >>> import fontforge
      File "<stdin>", line 1, in <module>
    ImportError: 'fontforge' is not a built-in module

Weird error. How do I use psMat with python3 then?

PR #1287, opened 2014-03-12
Added support for named points in the CharView.

blipvert:
1. Added "Name Point..." item to the Point menu of the Charview.
2. Added "Name Point..." item to the selection popup menu for a point.
3. Show point names in CharView, similar to what is done for contours.
4. Added an icon to the tango theme, with a placeholder for the 2012 theme.


frank-trampe:
Uh, it's standard practice (I hope) to copy me on these sorts of pull requests before merging, but that seems not to have happened. I'm just seeing this one now.

I think that it's important to be able to toggle visibility of point names. It might also be good to enforce uniqueness. And this new code leaks the old point name.

davelab6:
@frank-trampe Reviewing https://github.com/fontforge/fontforge/graphs/contributors I believe I merged this when you were less active

davelab6:
But yes, I think the CONTRIBUTING guidelines are clear about having you review stuff, eg https://github.com/fontforge/fontforge/blob/master/CONTRIBUTING.md#contributing-code

frank-trampe:
I always check my e-mail.

PR #1625, opened 2014-08-06
tottf: accept exactly 65535 glyphs as in the sfnt specification

### LINKED ISSUE #1581: There's a font using 65535 glyphs in sfnt format.
Well, Adobe did that in Source Han Sans. To generate the font I need to lift this limit.

In `fontforge/tottf.c`, around line 5679:

``` C
    if ( at->gi.gcnt>=65535 ) {
        /* You might think we could use GID 65535, but it is used as a "No Glyph" */
        /*  mark in many places (cmap tables, mac substitutions to delete a glyph */
        ff_post_error(_("Too many glyphs"), _("The 'sfnt' format is currently limited to 65535 glyphs, and your font has %d of them."),
                at->gi.gcnt );
        AbortTTF(at,sf);
return( false );
    }
```

adrientetar:
Closes #1581.

r? @frank-trampe

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -5676,13 +5676,15 @@ return( false );
 	}
 	AssignTTFBitGlyph(&at->gi,sf,at->map,bsizes);
     }
-    if ( at->gi.gcnt>=65535 ) {
-	/* You might think we could use GID 65535, but it is used as a "No Glyph" */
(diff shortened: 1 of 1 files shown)


Rameshdaspam:
I need a Source Han Sans HW SC of ttf font and .eot font

PR #1700, opened 2014-08-31
1629: cv allow paging through chars without a unicode point.

monkeyiq:


fontforgeexe/wordlistparser.c
```diff
@@ -351,6 +361,177 @@ unichar_t* WordlistEscapedInputStringToRealStringBasic(
...
+
+    // truncate insanely long lines rather than crash
+    if( strlen(input) > PATH_MAX )
+	input[PATH_MAX] = '\0';
+
+    WordListChar* ret = calloc( WordListLineSz, sizeof(WordListChar));
+    WordListChar* out = ret;
+    char* in = input;
+    char* in_end = input + strlen(input);
+    // trim comment and beyond from input
```
adrientetar: @monkeyiq Why would there be comments in the input field? If `# sth` an accepted syntax for comments in a wordlist?  
Asking this because otherwise we could stop forbidding \# in its raw form.

davelab6: To document what the line is for

Issue #1934, opened 2014-11-16
Fonts saved using master version cannot be opened in older versions


santhoshtr:
I am using fontforge built from git(specifically my build is from a7edc6db40aec54f95ab28c28fa25d2be83eec4c). The sfd files generated by this version of fontforge has lot of additions to the sfd created by fontforge shipped by distros-example: 2012 versions. When the new sfd is opened by 2012 fontforge, all glyphs are shown blank. 

To reproduce try opening an sfd saved by fontforge from git master in older versions. All glyphs will be blank. It is also noted that fontforge 201408xx version cannot open SFD generated from 201410xx versions.

This is a serious problems since the distros like Fedora build TTF from sfds using packaged fontforge versions and resulting ttf will not have any glyphs.

(Sample SFD generated using latest fontforge is available at https://github.com/smc/Chilanka )

ctrlcctrlv:
Superseded by #3800.

PR #1940, opened 2014-11-20
Skip through unrecognized tags in an S. F. D..

frank-trampe:
This will enhance the ability of present FontForge to parse files from future versions that add one-line tags to S. F. D..

cc #1934.

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -562,6 +562,28 @@ return( NULL );
 return( pt );
 }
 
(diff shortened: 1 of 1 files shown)


frank-trampe:
Hmm. After all that, it turns out that there is already logic for skipping unrecognized lines elsewhere. It does not currently support multi-line entities, but it does what I was trying to do here, so any work would be an improvement to that.

Issue #1622, opened 2014-08-06
UI: Metrics window annoyances


adrientetar:
Outstanding issues:

- [ ] if you select an unencoded glyph from font view and click open metrics window the glyph shows up fine but the input field has a random unicoding and an empty glyph with the said encoding is created in the font view
- [ ] let's say I have a glyph variant (unencoded) with a dotted suffix &ndash; if the base glyph is in the `uniXXXX` form the alt glyph will not appear when `/uniXXXX.alt` is typed (just shows the base uni glyph and `.alt` afterwards)  
The uniXXXX code uses special rooting to call any Unicode value even if named otherwise but I think we should nonetheless check for suffixes there  
The bug also occurs when opening the glyph in the char view so in fact this makes it hard to edit it
- [x] kerning does not rewind when updating a kern value with keyboards &ndash; works when using arrows
- [x] if you call unencoded glyphs through their entity (e.g. `/f_i`) they don't display

---

- [x] glyph kerning does not properly refresh (upon change on the kern
cell on the bottom, I have to delete the concerned chars and write them again &ndash;
the drawing seems to refresh but keeps the old kerning value).  
Actually, spacing refreshes properly but not kerning.
- [x] \(lexer) In the input field, add an entity like /comma, then a space. You'll
need a second space for it to actually show up.  (cc @monkeyiq, I think you wrote the lexer)

Another lexer bug:

- [x] \(lexer) Char window: In the input field, try to add a slash after the current letter,
it disappears. So how do you add an entity (e.g. /eacute)?

(Generally speaking, I think that the lexer should be refactored with more consistency
between what you input and what gets rewinded in the input field.)

A lookup menu bug:

- [x] The lookup listings (under GPOS and GSUB) never add scrollbars, and don't either when you open up a table into subtables.  
(Source Sans Pro is an example that has enough lookups to repro it)

adrientetar:
Added another bug.

davelab6:
> if you select an unencoded glyph from font view and click open metrics window the glyph shows up fine but the input field has a random unicoding and an empty glyph with the said encoding is created in the font view

So, actions to fix are
- [ ] prevent creation of new empty glyphs with the said encoding
- [ ] show the glyph name in the input field, eg `/dagger`

frank-trampe:
For item 1, I plan to have the unencoded glyph appear with its escaped glyph name in the metrics view rather than creating an encoding from an uninitialized string (as seems from the description to be happening).

[...]

PR #2018, opened 2014-12-30
Handle all char up events in the metrics view.

frank-trampe:
Some char_up events were getting discarded, leading to a failure to update the values and preview as in #1622.

This supersedes #2016 and #2017.


andreaslarsen:
With the latest HEAD on mac adjusting bearing updates the view correctly but when adjusting kern I have to update the view by adding or removing a character before the view is updated.

PR #2021, opened 2014-12-31
Stop doubling keyboard actions in the Metrics View.

frank-trampe:
Stop triggering MVChar from the keyboard down events since we now trigger it on the keyboard up event and want to avoid duplication.

This addresses a problem from #2018.


adrientetar:
@frank-trampe I was thinking, why do we use charup instead of chardown? It seems charup has a slight delay to it.

PR #2036, opened 2015-01-08
Fix #2034 knife tool not cutting at the tip

### LINKED ISSUE #2034: Knife tool not cutting at the tip of the cursor
Open a font, open outline view, select knife tool and make a cut. 

The cut isn't made at the upper left corner - especially evident when you make vertical cuts.

andreaslarsen:
I'm no programmer and haven't been able to build+test but I can't see why it shouldn't work.


davelab6:
I think this might be related to the changes made to allow point handle
sizes to change....

I would prefer to get this tested before merging :)

adrientetar:
We do not use the GTK UI, it is a stub written in 2008 or so.

davelab6:
We should remove that stub

Issue #2050, opened 2015-01-18
Allow moving a point without touching its handles


adrientetar:
i.e. moving an anchor point along the path of its handles.

This is *very* useful when creating high-quality outlines so that you can fine-tune the location of the extrema without disrupting the course of the handles.

davelab6:
How do other editors do it? A chord key?

adrientetar:
> How do other editors do it? A chord key?

Glyphs uses Option + drag on the Mac. http://vimeo.com/43552747

davelab6:
I'm not sure if that's good for XQuartz; `Cmd + Shift + Drag` seems to be identical to `Shift + Drag` so that seems like a good candidate.

Issue #1397, opened 2014-06-06
Drawing tools issues


adrientetar:
- [ ] Let's say I have an anchor selected, then if I click on a cyan/magenta handle of another anchor, both aforementioned anchors will be selected. It should deselect the previous anchor.
  ![rec](https://cloud.githubusercontent.com/assets/1250894/3486944/f4bc2f9e-045a-11e4-83e4-2c954d304243.gif)
  This is annoying when switching between different anchors' handles.
- [ ] Then, when zooming with a handle selected, the selection changes to the parent point. It should stick to the handle.
  ![rec](https://cloud.githubusercontent.com/assets/1250894/3486947/5a181fd8-045b-11e4-809a-e185dc2b8432.gif)
  This means that if you move a handle, then zoom, then try to move again, it will move the point instead of its handle. This is annoying.
- [ ] When moving corner points, those at the edge of an open path have their handles stick at the departure point  
  ![1](https://cloud.githubusercontent.com/assets/1250894/5791795/9457f926-9ef1-11e4-8552-2f0b8519d36b.png)  
  ![2](https://cloud.githubusercontent.com/assets/1250894/5791796/99b0b5ca-9ef1-11e4-9fbe-c5a28edaa2d8.png)
- [ ] Coercing a corner point to a curve point changes moves the handles even if the handles path is straight (same weirdness when going HVCurve &rarr; Curve). 
  I noticed that paths drawn using the circle tool do not exhibit this behavior, instead the handles path's length changes (??). The code path here probably needs to be scrutinized.

adrientetar:
@davelab6 @frank-trampe Updated with another drawing tools issue.

adrientetar:
@davelab6 @frank-trampe Updated with another drawing tools issue.

PR #2060, opened 2015-01-31
Add experimental support for fine point adjustment.

frank-trampe:
Allow for moving of points on their control point lines without moving the control points (by holding down the control key).

This is in response to #2050.

We also need to settle on a key combination. Alt is not available for drag operations because some window managers use that for moving the window. Shift is not available since that would cause the click beginning the drag operation to deselect the point to be moved.

### DIFF
--- a/fontforge/views.h
+++ b/fontforge/views.h
@@ -1018,6 +1018,7 @@ extern Undoes *CVPreserveMaybeState(CharView *cv, int isTState );
 extern void CVRestoreTOriginalState(CharView *cv);
 extern void CVUndoCleanup(CharView *cv);
 
+extern void AdjustControls(SplinePoint *sp);
 extern void CVAdjustPoint(CharView *cv, SplinePoint *sp);
 extern void CVMergeSplineSets(CharView *cv, SplinePoint *active, SplineSet *activess,
 	SplinePoint *merge, SplineSet *mergess);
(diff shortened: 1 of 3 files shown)


frank-trampe:
@adrientetar, take a look.

adrientetar:
I'll have a look this evening. Ctrl key is okay I think, the real oops imo is that Ctrl and Shift are inverted in the Char View w.r.t the usual conventions &ndash; I say in the Char View because it is (confusingly) not the case e.g. with the OpenType panel in Metrics View.

adrientetar:
It works and it's great to have. Caveat: with tangent/corner points you cannot go "back" once you start going along the path &ndash; that is not too bothering for me as the balancing feature is mostly for curve points. Also if you have a corner point with two handles, it will go along a path that's tangent to its handles.  
Is fine to merge imo, just want to make sure you are aware of these two points beforehand.

frank-trampe:
Since we are on the topic, I will add another caveat.

If one selects multiple points and then drags a point with the control modifier, the control points are only fixed for the point that he drags, and the other selected points and control points move with the dragged point (thus on or parallel to the control point line). Is there a case in which this is useful?

adrientetar:
Not for me but it may well be&hellip; I do not know.

adrientetar:
@frank-trampe It seems if you have a handle at the edge of an open path that handle sticks when moving the point (see the third bullet point of #1397) &ndash; it seems this bug always trigger the adjustment code of this PR so it makes it a nightmare to move a point at the edge of a path that has handles. cc @davelab6

frank-trampe:
I don't entirely understand what you're saying. Can you provide a file with a path that exhibits this problem? And are you saying that the tangential adjustment happens even when one is not pressing control?

Issue #2005, opened 2014-12-19
Ubuntu PPA really old


pathumego:
Official PPA gives me a really old version of FontForge. 

https://launchpad.net/~fontforge/+archive/ubuntu/fontforge

Can we update this?

pathumego:
@frank-trampe I followed the process from  `http://designwithfontforge.com/en-US/Installing_Fontforge.html ` I think I was wrong. Its the latest. But the UI is really different to the Mac version (looks old) and icons are different. That's why I thought Its old.

frank-trampe:
Can you check the about box and confirm the git hash or the compile date?

pathumego:
![screenshot from 2015-01-08 20 02 01](https://cloud.githubusercontent.com/assets/6846609/5663883/b1c7f164-9771-11e4-83fa-01fce661cdab.png)

pathumego:
collab server is not there, this is what I see
![screenshot from 2015-01-08 09 46 03](https://cloud.githubusercontent.com/assets/6846609/5663936/22be224e-9772-11e4-832c-2d384acfa9d0.png)

davelab6:
It can't find the theme file, that's all. Where is it? :)

pathumego:
I cant find it either :(

frank-trampe:
We're also missing the hotkeys file, it seems; I plan to fix both of these
issues soon.

sungsit:
For Ubuntu PPA package, most directories & files are missing from `/usr/share/fontforge/`.

I have FontForge on ArchLinux so I tested by copying `/usr/share/fontforge/` from Arch to Ubuntu then the interface works fine. This is the message when FontForge get resources file.

<pre>
Copyright (c) 2000-2014 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Based on sources from 00:50 UTC 31-Dec-2014-ML-D.
 Based on source from git with hash: 6bfa9ed50dbea615f54c31cef7550906a47d15b1
no xdefs_filename!
TESTING: getPixmapDir:/usr/share/fontforge/pixmaps
TESTING: getShareDir:/usr/share/fontforge
TESTING: GResourceProgramDir:/usr/bin
trying default theme:/usr/share/fontforge/pixmaps/resources
fontforge.hasUserInterface!
[...]

PR #2061, opened 2015-02-01
Update fontforge-common.dirs

n7s:
fix path to restore access to shortcuts and theme


frank-trampe:
Well that was easy!

adrientetar:
Does these PRs fix #2005 as well?

Issue #12, opened 2012-08-18
Metric window glitch


MihailJP:
When typing something into the textbox on the metric window, the window goes into a rubbish.
On my machine with Cairo 1.10.2-14.1.3 (in OpenSUSE 12.1 x86_64 repository), looks like http://twitpic.com/akvrir.
Once this occured, other windows (at least the save-before-closing confirmation dialog) will be affected, looking like http://twitpic.com/akvs0x.
I was able to detect by which commit this issue was introduced with git-bisect tools; this was 3ebbf03c52827093c9e32f3c797e5befc5040423 "Use cairo drawing for all windows".

Issue #11, opened 2012-08-17
Bounds checking not done on all layers


jhps:
When bounds are found, not all layers are considered.
In the  View->Fit function this results in different behavior depending on the shape of the Fore layer or other layers, whose dimensions are ignored even if they are active.
In the preview function, if freetype is not used, rendering is also corrupted in that only the parts of an active layer that fit in the bounds of the Fore layer are rendered properly and the parts that fall outside the boundaries on just the X axis are rendered shifted inside the boundaries. This last may be a separate issue with the built-in renderer.

Issue #10, opened 2012-08-17
timezone use not portable


jhps:
In cvexport.c the following code:

``` C
#ifdef _NO_TZSET
fprintf( pdf, "Z)\n" );
#else
tzset();
if ( timezone==0 )
fprintf( pdf, "Z)\n" );
else 
fprintf( pdf, "%+02d')\n", (int) timezone/3600 ); /* doesn't handle half-hour zones */
#endif
```

Assumes that a system that has tzset() has the global timezone as an integer value. In fact on FreeBSD (10)  there is tzset but timezone is a function returning char \* for the timezone abbreviation.

Perhaps new autotools changes will address this.

PR #2084, opened 2015-02-19
mview: stop creating fake unicode characters when importing fview select...

### LINKED ISSUE #1622: UI: Metrics window annoyances

adrientetar:
...ion

Closes #1622.

r? @frank-trampe

### DIFF
--- a/fontforgeexe/wordlistparser.c
+++ b/fontforgeexe/wordlistparser.c
@@ -284,15 +284,21 @@ u_WordlistEscapedInputStringToRealString_readGlyphName(
 	    {
 		unichar_t* endptr = 0;
 		long unicodepoint = u_strtoul( glyphname+3, &endptr, 16 );
+                SplineChar* tmp = 0;
 		TRACE("uni prefix, codepoint: %ld\n", unicodepoint );
 		sc = SFGetChar( sf, unicodepoint, 0 );
-		if( sc && endptr )
-		{
-		    unichar_t* endofglyphname = glyphname + u_strlen(glyphname);
-//		    printf("endptr:%p endofglyphname:%p\n", endptr, endofglyphname );
-		    for( ; endptr < endofglyphname; endptr++ )
-			--endpos;
-		}
+                if (tmp = SFGetChar( sf, -1, u_to_c(glyphname) )) {
+		    TRACE("have subst. char: %s\n", tmp->name );
+                    sc = tmp;
+                } else {
+		    if( sc && endptr )
+		    {
+		        unichar_t* endofglyphname = glyphname + u_strlen(glyphname);
+//		        printf("endptr:%p endofglyphname:%p\n", endptr, endofglyphname );
+		        for( ; endptr < endofglyphname; endptr++ )
+                            --endpos;
+		    }
+                }
 	    }
 	    
 	    if( firstLookup && glyphname[0] == '#' )
--- a/gdraw/gmenu.c
+++ b/gdraw/gmenu.c
@@ -1907,23 +1907,23 @@ int GMenuBarCheckKey(GWindow top, GGadget *g, GEvent *event) {
     }
 #endif
 
-    TRACE("about to look for hotkey in new system...state:%d keysym:%d\n", event->u.chr.state, event->u.chr.keysym );
-    TRACE("     has ksm_control:%d\n", (event->u.chr.state & ksm_control ));
-    TRACE("     has ksm_meta:%d\n",    (event->u.chr.state & ksm_meta ));
-    TRACE("     has ksm_shift:%d\n",   (event->u.chr.state & ksm_shift ));
+//    TRACE("about to look for hotkey in new system...state:%d keysym:%d\n", event->u.chr.state, event->u.chr.keysym );
+//    TRACE("     has ksm_control:%d\n", (event->u.chr.state & ksm_control ));
+//    TRACE("     has ksm_meta:%d\n",    (event->u.chr.state & ksm_meta ));
+//    TRACE("     has ksm_shift:%d\n",   (event->u.chr.state & ksm_shift ));
 
     event->u.chr.state |= ksm_numlock;
-    TRACE("about2 to look for hotkey in new system...state:%d keysym:%d\n", event->u.chr.state, event->u.chr.keysym );
+//    TRACE("about2 to look for hotkey in new system...state:%d keysym:%d\n", event->u.chr.state, event->u.chr.keysym );
 
     /**
      * Mask off the parts we don't explicitly care about
      */
     event->u.chr.state &= ( ksm_control | ksm_meta | ksm_shift | ksm_option );
 
-    TRACE("about3 to look for hotkey in new system...state:%d keysym:%d\n", event->u.chr.state, event->u.chr.keysym );
-    TRACE("     has ksm_control:%d\n", (event->u.chr.state & ksm_control ));
-    TRACE("     has ksm_meta:%d\n",    (event->u.chr.state & ksm_meta ));
-    TRACE("     has ksm_shift:%d\n",   (event->u.chr.state & ksm_shift ));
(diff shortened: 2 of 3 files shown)


adrientetar:
@frank-trampe The `buf` chars ptr is initialized with 120 cells, is there a risk of overflowing it eventually?

adrientetar:
Well it just occured to me:

~~~
0x00007ffff71e3cc9 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56	../nptl/sysdeps/unix/sysv/linux/raise.c: Aucun fichier ou dossier de ce type.
(gdb) where
#0  0x00007ffff71e3cc9 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff71e70d8 in __GI_abort () at abort.c:89
#2  0x00007ffff7220f24 in __libc_message (do_abort=do_abort@entry=1, 
    fmt=fmt@entry=0x7ffff732d0cb "*** %s ***: %s terminated\n")
    at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007ffff72b882c in __GI___fortify_fail (msg=<optimized out>, 
    msg@entry=0x7ffff732d0b3 "stack smashing detected") at fortify_fail.c:37
#4  0x00007ffff72b87d0 in __stack_chk_fail () at stack_chk_fail.c:28
#5  0x00007ffff79b4548 in MetricsViewCreate (fv=0x9051b0, sc=0x0, bdf=0x0)
    at metricsview.c:5407
#6  0x00007ffff79309d0 in FVMenuOpenMetrics (gw=0xcbf820, UNUSED_mi=0xced330, 
    UNUSED_e=0x7fffffffb750) at fontview.c:1191
#7  0x00007ffff68641b7 in gmenu_mouse (m=0x1931f20, event=0x7fffffffb750)
    at gmenu.c:901
#8  0x00007ffff6865c6b in gmenu_eh (w=0x1932030, ge=0x7fffffffb750)
    at gmenu.c:1336
#9  0x00007ffff68b1705 in dispatchEvent (gdisp=0x69db50, event=0x7fffffffb900)
    at gxdraw.c:3478
#10 0x00007ffff68b1c9d in GXDrawEventLoop (gd=0x69db50) at gxdraw.c:3586
#11 0x00007ffff68387cb in GDrawEventLoop (gdisp=0x69db50) at gdraw.c:791
#12 0x00007ffff7a58491 in fontforge_main (argc=1, argv=0x7fffffffdc38)
    at startui.c:1367
~~~

adrientetar:
Bumping it to 200 fixed my problem. Thinking 400 may be a reasonable value generally speaking&hellip;

davelab6:
400 seems good

frank-trampe:
I'm not sure about assigning a value to pt. Later code checks buf for a value, and it seems a problem that it is not updated.

I think that this function would benefit from a refactoring that moves away from static allocation of space for data of unknown size or at least checks for buffer over-runs.

adrientetar:
Otherwise using strncpy() with some ptr arithmetic could make it overflow-safe. It would be better to prevent the overrun rather than checking for it, esp. since guard pages will likely abort the program anyway.

> Later code checks buf for a value, and it seems a problem that it is not updated.

Not sure I understand this, pt is the ptr that walks through the storage assigned to buf. What do you mean later code checks buf for a value? It seems only an assignment of buf is made after that passage. But even the assignments made before are concerned since they reference buf.

> that moves away from static allocation of space 

Yea this seems to me like how it should be done... should it realloc on every loop turn? Also one must look at how resizes are handled post-initialization.

adrientetar:
There is a limit in the number of glyphs you can "carry" from fontview to metricsview. If we multiply it by the max number of chars in a glyph name (is there one?), we can possibly initialize statically to the worst case.

frank-trampe:
Oops. I did not bother to look at how utf8_idpb worked; assigning to pt there is probably okay.

adrientetar:
@frank-trampe Updated. If the buffer okay or should I make sth with realloc()?

frank-trampe:
It wouldn't hurt to let me review it.

adrientetar:
Mmh yes sorry, I wanted to have this done. I can do a followup if you have remarks.

gdraw/gmenu.c
```diff
@@ -1907,23 +1907,23 @@ int GMenuBarCheckKey(GWindow top, GGadget *g, GEvent *event) {
     }
 #endif
 
-    TRACE("about to look for hotkey in new system...state:%d keysym:%d\n", event->u.chr.state, event->u.chr.keysym );
-    TRACE("     has ksm_control:%d\n", (event->u.chr.state & ksm_control ));
-    TRACE("     has ksm_meta:%d\n",    (event->u.chr.state & ksm_meta ));
-    TRACE("     has ksm_shift:%d\n",   (event->u.chr.state & ksm_shift ));
+//    TRACE("about to look for hotkey in new system...state:%d keysym:%d\n", event->u.chr.state, event->u.chr.keysym );
```
frank-trampe: Why are we disabling these traces? Were they producing annoying output somewhere?

fontforgeexe/wordlistparser.c
```diff
@@ -284,15 +284,21 @@ u_WordlistEscapedInputStringToRealString_readGlyphName(
 	    {
 		unichar_t* endptr = 0;
 		long unicodepoint = u_strtoul( glyphname+3, &endptr, 16 );
+                SplineChar* tmp = 0;
```
frank-trampe: This pointer ought to be null. The existing pointer before probably ought to be null as well.

gdraw/gmenu.c
adrientetar: Yes they produced noise in the console each time you press a key.

fontforgeexe/wordlistparser.c
adrientetar: Well when it comes to ptrs, `0 == NULL`.

frank-trampe: Sort of. But `NULL` exists for a reason, and I think that some compilers complain about type.

adrientetar: Mmh... it will only matter on arch where NULL is not 0x0 and these [are rather old](http://stackoverflow.com/a/2597232). If it matters we should audit the code and try to remove all spots where that is (surely it's not only present there?).

PR #2092, opened 2015-02-28
Add detection for Python library collisions.

frank-trampe:
This outputs a message when installing the Python library if the Python environment detects another FontForge Python library in another location. It is not an error or a warning since there are contexts (such as packaging) in which the presence of another FontForge Python library in another location is not incorrect.


davelab6:
@n7s will be pleased :)

PR #2093, opened 2015-02-28
ufo: fix a flipped OS/2 descender value

adrientetar:
Commit f06cb470c77c8f5073ff146057de54ebaf938716 followup.

Fixes http://doc.robofont.com/forums/topic/bug-on-generating/.


davelab6:
Great catch! :)

Issue #2095, opened 2015-03-02
Crash when opening any Android Roboto font


gwgill:
Using FontForgeSetup-31-12-2014-r2.exe on WinXP
Tried:
Roboto-Regular.ttf
Roboto-Italic.ttf
Roboto-Bold.ttf
Roboto-BoldItalic.ttf
All crash at LIBGDRAW-5! 6eff5c3c()

davelab6:
@gwgill where can we download the same files? cc @jtanx

gwgill:
Dave Crossland wrote:
> @gwgill where can we download the same files? cc @jtanx 

Google makes various versions of its Android Roboto fonts available
(see <http://developer.android.com/design/style/typography.html>
     <https://developer.android.com/downloads/design/roboto-1.2.zip>
     <http://material-design.storage.googleapis.com/downloads/RobotoTTF.zip>
 etc.),
but to avoid confusion as to the exact version, here is a copy of
one of the ones I tried to open:

<http://www.argyllcms.com/Roboto-Regular.ttf>

Graeme Gill.

jtanx:
I've just tried on Windows 8.1 64 bit and it doesn't crash. I'll try on XP some time later today.

gwgill:
Jeremy Tan wrote:
> I've just tried on Windows 8.1 64 bit and it doesn't crash. I'll try on XP some time later today.

I don't know if there is any connection, but the UI looked rather strange -
not the usual XP UI style. Deleting the FontForge preferences did not
cure this problem. Perhaps it is not locating the installation properly -
all my installs get directed to d:/bin by default, and in this case
FontForge was installed in d:/bin/FontForge.

Graeme Gill.

jtanx:
I've just tried on Windows XP (32-bit SP3/latest) and it still doesn't crash. Strange. When does it crash for you? As you open the font? Perhaps try updating to the latest release and see how it goes from there (it should be released soon; no more than 30 minutes from now).

gwgill:
Jeremy Tan wrote:
> I've just tried on Windows XP (32-bit SP3/latest) and it still doesn't crash. Strange.
> When does it crash for you? As you open the font?

Yes. I get an empty window in the upper right titled with the font name
and a "Warnings" window at the bottom right with a number of warnings -
see the attached screen capture, and a crash report.

> Perhaps try updating to the latest
> release and see how it goes from there (it should be released soon; no more than 30
> minutes from now).

[...]

PR #2104, opened 2015-03-03
Fix share folder logic to be relative to current program dir

jtanx:
Previous logic searched for first instance of "/bin". This could give the
wrong result if there is more than one bin folder in the path, e.g.:
"Y:/bin/FontForge/bin" gives a share folder of "Y:/share"

With the new scheme, the share folder is relative to the current path, e.g.:
"Y:/bin/FontForge/bin" gives "Y:/bin/FontForge/share"

See also #2095.

### DIFF
--- a/gutils/fsys.c
+++ b/gutils/fsys.c
@@ -784,7 +784,8 @@ char *getShareDir(void) {
 
     set = true;
 
-    pt = strstr(GResourceProgramDir,"/bin");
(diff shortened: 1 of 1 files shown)


frank-trampe:
It looks good to me. Have you tested?

jtanx:
I've tested it on Windows and it works okay. I don't know if it works on other platforms or not though.

Issue #2073, opened 2015-02-09
UFO: Improve handling of malformed UFOs


davelab6:
It is common that PS type contours are sometimes presented in UIs as having a (TT style) single control point, which in fact means the nearby control point is co-located on the on-curve point, but the on-curve point takes precedence in the UI. 

It seems there is a buggy UFO editor out there that produces GLIF data that is expected to be PS flavor, but which FF parses as a contour of mixed TT and PS and thus (silently?) drops **part** of the contour on import. Yikes! :)

The UFO spec says:

> **point:** Draw a cubic bezier curve from the last non-offcurve point to this point. If the number of offcurve points is zero, a straight line is drawn. If it is one, a quadratic curve is drawn. If it is two, a regular cubic bezier is drawn. If it is larger than 2, a series of cubic bezier segments are drawn, as defined by the Super Bezieralgorithm.
– http://unifiedfontobject.org/versions/ufo2/glif.html

I guess supporting mixed PS/TT curves is a huge changeset and thus out of the question :) 

@adrientetar says, 
> robofont itself bugs when I try to open this glyph. Glyphs seems to open it just fine

I think we should all the User what to do on that case. I would want to demote quadratic to cubic :)

@frank-trampe says,

> A quadratic curve on an otherwise cubic contour is probably not what the designer intended. I thought that we would promote the cubic to a quadratic in this case, but that is not happening.

I think we should demote a quadratic to cubic in this case, and place a 2nd cp at the position of its on-curve point, such that it has a zero length.

adrientetar:
Well, in fact, (as per my most recent email) I suspect a codegen bug in FontForge that adds a stray point when there is an on-curve corner point with a zero-length handle, hence the "quadratic part" of the glyph that is just an additional point inserted by error.  
That's just a guess, I haven't verified it but I don't think it comes from foreign data. I had suspected so initially because I did not know what in the chain caused it since I used multiple editors and tools. But robofont too errs on that faulty contour, as does the AFDKO.

Speaking of Glyphs as you did it seems it retains the extra point but doesn't use it for drawing or sth. At least I roundtripped the faulty glyph through it and it looked fine in its "char view".

davelab6:
Okay cool. In a telephone call chat, @frank-trampe agreed with you that FontForge is at fault, but he says its that FF is _dropping_ a control point when its zero length, thus the segment appears as TT as there's only 1 offcurve point. 

He says he'll look at the UFO exporter again and fix it :)

[...]

PR #2105, opened 2015-03-03
Improve conversion between Python and native control points.

frank-trampe:
Make native control points whenever there are corresponding Python control points.

This addresses #2073.


frank-trampe:
@adrientetar, take a look.

adrientetar:
Interesting&hellip; I'll try it.

frank-trampe:
Ping.

Issue #2055, opened 2015-01-23
Various first-use issues encountered on Max OS X 


davelab6:
### Quick things
- [x] ~~The default value of `Element, OS/2, Metrics, X Height` is 0 but should be 500.~~ Adrien didn't think this was wise, see below, so striking it out
- [x] ~~Turn off CharView, View, Show, Control Points (Always) by default~~ Adrien didn't think this was wise, see below, so striking it out
- [x] ~~Crash, https://github.com/fontforge/fontforge/issues/2056~~ spurious so punting
### Mac integration
- [ ] After installing FontForge on a Mac, the first time it is run, it runs `fc-cache` over all the `/LIbrary/Fonts` which takes about 3 minutes on a Macbook Air. This is excruciating, and I wonder if this is really necessary, because I believe the reason this is run is to be able to register the fonts used to render the UI, and I believe those fonts are all in the `FontForge.app` bundle. So, can this scan be only scanning the bundle, and take just a second or two max..?
- [x] Weirdly the first time the program is run, the dock icon goes away after the `fc-cache` process completes, and XQuartz starts and the FF window appears. Quitting FontForge and launching it from the Finder again solves this. I don't think anything can be done about that, so checking this mark, but, it sucks :) Here's an image:

![screen shot 2015-01-22 at 7 51 03 pm](https://cloud.githubusercontent.com/assets/261579/5868663/d815c904-a27a-11e4-806a-8510c8639943.png)
- [x] The second time the program is run, the dock icon bounces forever. This really sucks and I hope something can be done! really distracting :( Here it is bouncing after the program is fully loaded:

![screen shot 2015-01-22 at 7 51 19 pm](https://cloud.githubusercontent.com/assets/261579/5868667/dcb6abcc-a27a-11e4-8819-4b9424706981.png)
- [x] File, Open Font, now opens in `/` instead of `~` which is very confusing. Change it to open in `~` :)

![screen shot 2015-01-22 at 8 13 14 pm](https://cloud.githubusercontent.com/assets/261579/5868670/eec9a724-a27a-11e4-8e79-40157bd65eea.png)
### Metrics View weirdness
- [ ] In the metrics view left side bar, clicking the items should toggle them; currently clicking activates and shift clicking deactivates...
- [ ] In metrics view, after going View, Set Point Size, the hotkeys make no sense. Fortunately the fix will be to just redefine the hotkeys to something sensible.

[...]

PR #2108, opened 2015-03-04
Update osx/FontForge.app/Contents/MacOS/FontForge

davelab6:
In https://github.com/fontforge/fontforge/issues/2055 I reported a dock bounce issue. This resolves this, by adding a `&` to the final command. The effect is that now when run the FF dock icon disappears after loading. This is helpful because the XQuartz icon is actually what users must click to bring FF windows to the foreground; and if they Force Quit XQuartz then FF will also exit. 

`/Applications/FontForge.app/Contents/MacOS/FontForge --debug` works with this change.

### DIFF
--- a/osx/FontForge.app/Contents/MacOS/FontForge
+++ b/osx/FontForge.app/Contents/MacOS/FontForge
@@ -118,4 +118,4 @@ $scriptdir/fcprogress.pl >|/tmp/zz 2>&1
 
 export PATH="$PATH:$bundle_bin"
 echo $WRAPPER $bundle_bin/fontforge -new "$@" > /tmp/oo
-$WRAPPER $bundle_bin/fontforge -new "$@"
+$WRAPPER $bundle_bin/fontforge -new "$@" &


frank-trampe:
Although I will not oppose this pull request, I would like to be on record stating why I think that it is a bad idea. It is standard practice on the Macintosh that the executable in an application bundle not return control until it is finished running, and this breaks that convention. On my Macintosh, it also causes inconsistent appearance and disappearance of the dock icon.

davelab6:
FontForge is so far from Mac standard practices, I am not concerned about this :)

davelab6:
cc @monkeyiq @pathumego

pathumego:
Well, I agree with @davelab6 about the annoying behavior of the FF icon. I always click on the FF icon on the dock and then click on the XQuartz icon.

How will this affect the Alt+Tab app switching? ( I am not on OSX anymore) 

Adding a note about this behavior to the Mac Download page would help new users. @davelab6 what would be the best place to add a note?

davelab6:
> How will this affect the Alt+Tab app switching? ( I am not on OSX anymore) 

You also have to alt+tab to the XQuartz icon

Issue #1742, opened 2014-09-21
Increment side bearings does not work as expected


Kelvinsong:
When you run Increment Side Bearings By, it doesn’t increment the side bearings, it just sets the width of the glyph to be whatever you typed into it.

frank-trampe:
@davelab6, why did you unassign me?

davelab6:
I figured this was less important than the mac stuff, but great it was a bitesized fix and is done :)

davelab6:
@Kelvinsong please test :)

Kelvinsong:
It works! Thanks sm!!

PR #2115, opened 2015-03-06
Fix side bearing increment functionality.

frank-trampe:
Use sc->width rather than 0 for the base character width when incrementing the side bearings.

This addresses #1742.


frank-trampe:
@adrientetar, take a look.

PR #2124, opened 2015-03-07
ufo: output OT copyright by default, handle non-ASCII input

adrientetar:
Previously we were always outputting the old PS copyright value which is restricted to ASCII.

r? @frank-trampe

### DIFF
--- a/fontforge/ufo.c
+++ b/fontforge/ufo.c
@@ -367,7 +367,7 @@ xmlNodePtr PythonLibToXML(void *python_persistent, const SplineChar *sc, int has
     xmlNodePtr retval = NULL, dictnode = NULL, keynode = NULL, valnode = NULL;
     // retval = xmlNewNode(NULL, BAD_CAST "lib"); //     "<lib>"
     dictnode = xmlNewNode(NULL, BAD_CAST "dict"); //     "  <dict>"
-    if ( has_hints 
+    if ( has_hints
 #ifndef _NO_PYTHON
          || (python_persistent!=NULL && PyMapping_Check((PyObject *)python_persistent))
 #endif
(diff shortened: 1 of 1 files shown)


fontforge/ufo.c
```diff
@@ -367,7 +367,7 @@ xmlNodePtr PythonLibToXML(void *python_persistent, const SplineChar *sc, int has
     xmlNodePtr retval = NULL, dictnode = NULL, keynode = NULL, valnode = NULL;
     // retval = xmlNewNode(NULL, BAD_CAST "lib"); //     "<lib>"
     dictnode = xmlNewNode(NULL, BAD_CAST "dict"); //     "  <dict>"
-    if ( has_hints 
+    if ( has_hints
```
frank-trampe: Was this necessary? I thought that the whitespace in the following line was sufficient.

fontforge/ufo.c
```diff
@@ -851,7 +851,7 @@ xmlDocPtr PlistInit() {
     xmlDocPtr doc = NULL;
     xmlNodePtr root_node = NULL, dict_node = NULL;
     xmlDtdPtr dtd = NULL;
-    
+
```
frank-trampe: ??

frank-trampe:
There is a lot of puzzling whitespace, but the rest looks okay.

adrientetar:
Oh it's just the editor I used that removes whitespace automatically.

frank-trampe:
@adrientetar, this clutters the change history and makes it harder to review the pull requests. It would be nice to fix this, particularly in the more cluttered of the other pull requests.

adrientetar:
You can add ?w=1 to the URL to ignore whitesp changes. But yea, I see.

PR #2125, opened 2015-03-08
ttf: fix underlinePosition calculation

adrientetar:
- Fix a signs problems in underlinePosition calculation for OpenType fonts. This was noticed here: http://typophile.com/node/103251#comment-555313 but never reported.
- Don't recalculate underlinePosition from the post table if we have a CFF font.

r? @frank-trampe

### DIFF
--- a/fontforge/parsettf.c
+++ b/fontforge/parsettf.c
@@ -641,7 +641,7 @@ static int PickTTFFont(FILE *ttf,char *filename,char **chosenname) {
     names = malloc(cnt*sizeof(char *));
     for ( i=0; i<cnt; ++i ) {
 	names[i] = TTFGetFontName(ttf,offsets[i],0);
-        if ( names[i]==NULL ) 
+        if ( names[i]==NULL )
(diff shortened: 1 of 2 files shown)


frank-trampe:
Aside from the odd whitespace changes, the code changes look good given what you state we are trying to do. Can you reference the appropriate part of the specification on this issue?

adrientetar:
> This is the suggested distance of the top of the underline from the baseline (negative values indicate below baseline). 
The PostScript definition of this FontInfo dictionary key (the y coordinate of the center of the stroke) is not used for historical reasons. **The value of the PostScript key may be calculated by subtracting half the underlineThickness from the value of this field.**

http://www.microsoft.com/typography/otspec/post.htm

frank-trampe:
Excellent. I confirm your math.

I approve of everything except for the whitespace changes.

adrientetar:
Well. They concern mostly empty lines, I'll merge this one for the time being. You can ignore whitespace changes with the `-w` switch of git blame.

PR #2126, opened 2015-03-08
parsettf: read CapHeight and XHeight from the OS/2 table

adrientetar:
r? @frank-trampe


frank-trampe:
Were we misreading the rest of the file previously?

adrientetar:
No, it just ignored the end of the table. There are more fields at the end actually but they aren't needed.

PR #2127, opened 2015-03-08
ufo: output OS/2 fsType array even if it's empty

adrientetar:
An empty bit is actually a legit value in itself.

r? @frank-trampe

### DIFF
--- a/fontforge/ufo.c
+++ b/fontforge/ufo.c
@@ -1156,8 +1156,12 @@ static int UFOOutputFontInfo(const char *basedir, SplineFont *sf, int layer) {
 	    for ( i=fscnt=0; i<16; ++i )
 		if ( sf->pfminfo.fstype&(1<<i) )
 		    fstype[fscnt++] = i;
-	    if ( fscnt!=0 )
-		PListAddIntArray(dictnode,"openTypeOS2Type",fstype,fscnt);
+	    /*
+	     * Note that value 0x0 is represented by an empty bit, so in that case
+	     * we output an empty array, otherwise compilers will fallback to their
+	     * default fsType value.
+	     */
+	    PListAddIntArray(dictnode,"openTypeOS2Type",fstype,fscnt);
 	}
 	if ( sf->pfminfo.typoascent_add )
 	    PListAddInteger(dictnode,"openTypeOS2TypoAscender",sf->ascent+sf->pfminfo.os2_typoascent);


frank-trampe:
This would seem to cause FontForge to denote a meaningful value on export even when there is no meaningful value, no?

adrientetar:
This setting is the default in FontForge so yes. But all compilers have a default, for makeotf it's 0x8 Printable Document which actually is more restrictive than 0x0 Installable Embedding which is the default in FontForge. So we could consider changing the default setting on import or when creating a new font.

frank-trampe:
Clean round-tripping is a fairly high-priority right now. A set flag may be in order, or we could initialize to a negative number. I don't have time to look at the code right now. What do you think?

adrientetar:
We would be better off initializing it to a valid value otherwise we'd have to edit export code for all supported formats to handle it and not just UFO. 
FontForge already uses a default value for numerous fields in the font view (those who are left unset generally have a checkbox labeled "Default") so it wouldn't hurt to do so. I'll have a look.

adrientetar:
@frank-trampe: It seems fsType's default value is -1 and exporters (incl. UFO) check against that when exporting. The TTF exporter has [a default value of 0x8](https://github.com/fontforge/fontforge/blob/e27e285fb052cc52ca5548103494ace48563aecd/fontforge/tottf.c#L3341) like makeotf. Is that ok to merge?

frank-trampe:
I still think that this is not quite right. If we are round-tripping a U. F. O. that lacks that tag, and the user makes no meaningful changes, the output U. F. O. ought to omit the tag. @davelab6?

adrientetar:
So we're not setting a value of -1 when inputting a UFO that does not have this key at all?

adrientetar:
@frank-trampe: Should I add a default setting `sf->pfminfo.fstype = -1` on [that line](https://github.com/adrientetar/fontforge/blob/ufo-fstype/fontforge/ufo.c#L3790)?

[Here is](https://github.com/fontforge/fontforge/blob/8c18b53a635816ae6eed9be4ea2b0653d2daf11e/fontforge/splinefont.h#L1728) the struct decl. and the [SplineFontEmpty() function](https://github.com/fontforge/fontforge/blob/61067f16bcc5536d42e7985dd6b1f392411f1aad/fontforge/splineutil2.c#L3803) &ndash; the latter sets this value to -1 but I don't think it's called when creating a new SplineFont anyway.

frank-trampe:
Yes. I just saw that; I did not quite understand your comment before. This is good to merge.

davelab6:
On 9 March 2015 at 22:48, Adrien Tétar <notifications@github.com> wrote:

> The TTF exporter has a default value of 0x8
> <https://github.com/fontforge/fontforge/blob/e27e285fb052cc52ca5548103494ace48563aecd/fontforge/tottf.c#L3341>
> like makeotf. Is that ok to merge?
>

I would strongly prefer 0x0 to prevent people making DRM restricted fonts
unintentionally :)

PR #2137, opened 2015-03-12
Focus fc-cache better on the Macintosh in order to cut start time.

frank-trampe:
This addresses part of #2055.

I was unable to test this effectively since I have yet to locate the fontconfig cache on Macintosh. I would welcome any advice on this. I have already looked at /usr/local/var/cache/fontconfig, ~/Library/Caches/org.freedesktop.something (which was not present), /Applications/FontForge.app/Contents/Resources/opt/stuff, and /var/caches/stuff (which was not present), and I found nothing.

It is sort of puzzling that there is no fontconfig utility that reports the cache path.

@adrientetar?

### DIFF
--- a/osx/FontForge.app/Contents/MacOS/fcprogress.pl
+++ b/osx/FontForge.app/Contents/MacOS/fcprogress.pl
@@ -13,7 +13,7 @@ my $forcecache = "";
 if ( ! -d "/Applications/FontForge.app/Contents/Resources/opt/local/var/cache/fontconfig" ) {
     $forcecache = "-f";
 } 
-my $fcproc = IO::File->new("$dirname/fc/fc-cache $forcecache  -v|");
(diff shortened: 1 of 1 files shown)


davelab6:
@monkeyiq might know

frank-trampe:
Oops. Never mind. Instead of A-B testing, I was doing B-B testing, and B was broken.

frank-trampe:
All that, and the actual commit was okay. (I forgot to set the executable bit on fcprogress.pl on the test machine.) This works for me. I would still appreciate a double-check.

frank-trampe:
For the curious, the cache (for the bundled fontconfig) is in the directory in the bundle (`/Applications/FontForge.app/Contents/Resources/opt/local/var/cache/fontconfig`) after all.

davelab6:
Looks good! I've asked @monkeyiq to make an app build for me to test with

PR #2145, opened 2015-03-14
fontforge: support styleMap attribute as a subset of OS/2 fsSelection

### LINKED ISSUE #2131: OS/2 fsSelection (style map) support
They are `sf->pfminfo.os2_family_name` and `sf->pfminfo.os2_style_name` in the program. These values are not accessible neither from the UI nor from scripting, they are only roundtripping in OpenType fonts and SFD files for now.

I need these values because they are used for the UFO styleMap (see [here](https://github.com/fontforge/fontforge/blob/c07b08eaa81d807352c455975cc74053bd82d790/fontforge/ufo.c#L1089)) which maps Bold and Italic buttons to select weights e.g on Windows.

adrientetar:
Closes #2131.

r? @frank-trampe

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -2217,6 +2217,8 @@ int SFD_DumpSplineFontMetadata( FILE *sfd, SplineFont *sf )
 
     if ( sf->version!=NULL )
 	fprintf(sfd, "Version: %s\n", sf->version );
+    if ( sf->styleMapFamilyName!=NULL )
+	fprintf(sfd, "StyleMapFamilyName: %s\n", sf->styleMapFamilyName );
     if ( sf->fondname!=NULL )
 	fprintf(sfd, "FONDName: %s\n", sf->fondname );
     if ( sf->defbasefilename!=NULL )
@@ -2274,6 +2276,8 @@ int SFD_DumpSplineFontMetadata( FILE *sfd, SplineFont *sf )
 	SFDDumpBase(sfd,"BaseHoriz:",sf->horiz_base);
     if ( sf->vert_base!=NULL )
 	SFDDumpBase(sfd,"BaseVert:",sf->vert_base);
+    if ( sf->pfminfo.stylemap!=-1 )
+	fprintf(sfd, "StyleMap: 0x%04x\n", sf->pfminfo.stylemap );
     if ( sf->pfminfo.fstype!=-1 )
 	fprintf(sfd, "FSType: %d\n", sf->pfminfo.fstype );
     fprintf(sfd, "OS2Version: %d\n", sf->os2_version );
@@ -2320,7 +2324,9 @@ int SFD_DumpSplineFontMetadata( FILE *sfd, SplineFont *sf )
 	fprintf(sfd, "OS2StrikeYSize: %d\n", sf->pfminfo.os2_strikeysize );
 	fprintf(sfd, "OS2StrikeYPos: %d\n", sf->pfminfo.os2_strikeypos );
     }
+    if ( sf->pfminfo.os2_capheight!=0 )
     fprintf(sfd, "OS2CapHeight: %d\n", sf->pfminfo.os2_capheight );
+    if ( sf->pfminfo.os2_xheight!=0 )
     fprintf(sfd, "OS2XHeight: %d\n", sf->pfminfo.os2_xheight );
     if ( sf->pfminfo.os2_family_class!=0 )
 	fprintf(sfd, "OS2FamilyClass: %d\n", sf->pfminfo.os2_family_class );
@@ -7492,6 +7498,16 @@ bool SFD_GetFontMetaData( FILE *sfd,
 	geteol(sfd,val);
 	sf->version = copy(val);
     }
+    else if ( strmatch(tok,"StyleMapFamilyName:")==0 )
+    {
+    sf->styleMapFamilyName = SFDReadUTF7Str(sfd);
+    }
+    /* Legacy attribute for StyleMapFamilyName. Deprecated. */
+    else if ( strmatch(tok,"OS2FamilyName:")==0 )
+    {
+    char* fname = SFDReadUTF7Str(sfd);
+    if (sf->styleMapFamilyName == NULL) sf->styleMapFamilyName = fname;
+    }
     else if ( strmatch(tok,"FONDName:")==0 )
     {
 	geteol(sfd,val);
@@ -7880,6 +7896,21 @@ bool SFD_GetFontMetaData( FILE *sfd,
 	else
 	    d->last_base->scripts = bs;
 	d->last_base_script = bs;
+    }
+    else if ( strmatch(tok,"StyleMap:")==0 )
+    {
+    gethex(sfd,(uint32 *)&sf->pfminfo.stylemap);
+    }
+    /* Legacy attribute for StyleMap. Deprecated. */
+    else if ( strmatch(tok,"OS2StyleName:")==0 )
+    {
+    char* sname = SFDReadUTF7Str(sfd);
+    if (sf->pfminfo.stylemap == -1) {
+        if (sname == "bold italic") sf->pfminfo.stylemap = 0x21;
+        else if (sname == "bold") sf->pfminfo.stylemap = 0x20;
+        else if (sname == "italic") sf->pfminfo.stylemap = 0x01;
+        else if (sname == "regular") sf->pfminfo.stylemap = 0x40;
+    }
     }
     else if ( strmatch(tok,"FSType:")==0 )
     {
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -2200,12 +2200,6 @@ int SFD_DumpSplineFontMetadata( FILE *sfd, SplineFont *sf )
 	fprintf(sfd, "FullName: %s\n", sf->fullname );
     if ( sf->familyname!=NULL )
 	fprintf(sfd, "FamilyName: %s\n", sf->familyname );
-    if ( sf->pfminfo.os2_family_name ) {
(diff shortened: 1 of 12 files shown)


frank-trampe:
@adrientetar, the Travis build failed.

```
splineutil.c: In function ‘SplineFontFree’:
splineutil.c:6470:67: error: ‘styleMapFamilyName’ undeclared (first use in this function)
splineutil.c:6470:67: note: each undeclared identifier is reported only once for each function it appears in
```

adrientetar:
Updated.

frank-trampe:
@adrientetar, can you explain precisely what you're doing in this pull request?

adrientetar:
I added a styleMap attribute which in turn is a subset of OS/2 fsStyle in order to effectively support styleMapStyleName with possibility for the user to change the setting before export (added a widget for this). I used a bitmask as the repr which corresponds to the OS/2 attribute. This is also used in TTFs if set instead of copying the macStyle attribute (which is corrected to be inline with the styleMap if needed).  
Besides that I implemented fallback mechanisms for various attributes. The styleMapFamilyName fallback is made the same as that of robofab and ufo2fdk.  
I removed OS2FamilyName and OS2StyleName because these were not really OS/2 attributes anyway. And I added support for reading the styleMap out of a TTF.  
I think that's pretty much it. The goal is to standardize styleMapFamilyName, and to support styleMapStyleName properly and with a fallback so you don't need to set it if you use default style mapping attributes. This is important for robofont or ufo2fdk and permits style-linking on Windows for the generated files. Style linking used for the "Bold" or "Italic" buttons or MS Word for instance and style linking is important for grouping fonts properly in certain Windows applications.

adrientetar:
Say, with the current code I have no way to set up `styleMapStyleName` unless picked from a foreign UFO. The current code guarantees round-tripping but it doesn't have those attributes supported "natively".

adrientetar:
Bump.

davelab6:
Frank is mostly offline this week :/

frank-trampe:
I'm looking at this, though.

adrientetar:
No problem, I was just worried that this had been forgotten.

fontforge/ufo.c
```diff
@@ -1086,8 +1103,34 @@ static int UFOOutputFontInfo(const char *basedir, SplineFont *sf, int layer) {
 /* Same keys in both formats */
     PListAddString(dictnode,"familyName",sf->familyname_with_timestamp ? sf->familyname_with_timestamp : sf->familyname);
     PListAddString(dictnode,"styleName",SFGetModifiers(sf));
-    if (sf->pfminfo.os2_family_name != NULL) PListAddString(dictnode,"styleMapFamilyName", sf->pfminfo.os2_family_name);
-    if (sf->pfminfo.os2_style_name != NULL) PListAddString(dictnode,"styleMapStyleName", sf->pfminfo.os2_style_name);
```
frank-trampe: I don't think that we want to add these values automatically if we are exporting something that we imported from U. F. O. without those values.

fontforgeexe/fontinfo.c
```diff
@@ -196,6 +196,15 @@ static GTextInfo ibmfamily[] = {
     { (unichar_t *) N_("Sy Miscellaneous"), NULL, 0, 0, (void *) 0xc0f, NULL, 0, 0, 0, 0, 0, 0, 1, 0, 0, '\0'},
     GTEXTINFO_EMPTY
 };
+static GTextInfo stylemap[] = {
```
frank-trampe: I find the casting sort of confusing. How about putting an annotated set or array of integers before this and storing pointers into it? (I'm not saying that's the best approach.) There at least needs to be a comment describing what is happening.

fontforgeexe/fontinfo.c
```diff
@@ -8433,16 +8447,6 @@ return;
     vgcd[13].gd.u.list = ibmfamily;
     vgcd[13].creator = GListButtonCreate;
```
frank-trampe: Why is this getting dropped?

frank-trampe: Or moved, rather?

fontforgeexe/fontinfo.c
adrientetar: Well I just copied the structure of the surrounding elements.

fontforgeexe/math.c
```diff
@@ -1799,6 +1799,7 @@ static void MKDInit(MathKernDlg *mkd,SplineChar *sc) {
     mkd->dummy_sf.glyphs = mkd->chars;
     mkd->dummy_sf.glyphcnt = mkd->dummy_sf.glyphmax = 4;
     mkd->dummy_sf.pfminfo.fstype = -1;
+    mkd->dummy_sf.pfminfo.stylemap = -1;
```
frank-trampe: Do we perhaps need a function that takes a pointer to a dummy_sf and initializes it as a dummy?

frank-trampe: (Not that that's a defect in this pull request.)

fontforgeexe/fontinfo.c
adrientetar: Because the concerned checkbox is at the end of the window and its positioning is relative to that of the element I added (statically so).

frank-trampe: Does this break the tab order?

fontforgeexe/fontinfo.c
frank-trampe: Oh. You're right. A comment then.

fontforgeexe/fontinfo.c
adrientetar: I don't understand, what do you mean?

frank-trampe: If you tab through the elements in the form, does the cursor move as expected?

fontforgeexe/fontinfo.c
adrientetar: I didn't make this repr so I have nothing particular to say about it. If you want me write something before the stack of arrays, I can do that.

fontforge/ufo.c
adrientetar: Good point, I'll change that.

fontforgeexe/fontinfo.c
adrientetar: I tried to Tab but it really only works with numeric fields, I see no outline and cannot open up list elements when tabbing. 
I'll note that there is existing disorder, for instance OS/2 Version which is the first field of this tab is ordered towards the bottom. Anyways, tab navigation isn't useable (at least on my computer) for this kind of elements.

fontforgeexe/math.c
adrientetar: I don't esp. want to tackle this just now, but you are right that there is duplication with the other function.

fontforgeexe/fontinfo.c
frank-trampe:
```
# Note that we are storing integer data in a pointer field, as others seem to have done before us, hence the casts. They are not to be dereferenced.
```

fontforgeexe/fontinfo.c
frank-trampe: Tabbing does not work for me, either, as it happens, and you are right that the order is already broken anyway. So it's probably not worth fixing in this pull request.

adrientetar:
Updated.

frank-trampe:
Oops. I missed something.

fontforge/sfd.c
```diff
@@ -7465,14 +7465,6 @@ bool SFD_GetFontMetaData( FILE *sfd,
...
 	sf->familyname = copy(val);
     }
-    else if ( strmatch(tok,"OS2FamilyName:")==0 )
-    {
-	if (!sf->pfminfo.os2_family_name) sf->pfminfo.os2_family_name = SFDReadUTF7Str(sfd);
-    }
-    else if ( strmatch(tok,"OS2StyleName:")==0 )
-    {
-	if (!sf->pfminfo.os2_style_name) sf->pfminfo.os2_style_name = SFDReadUTF7Str(sfd);
-    }
```
frank-trampe: In the interest of maximizing backwards compatibility, we ought to process these directives into the new data types with a note that they are deprecated.

fontforge/sfd.c
```diff
@@ -7895,6 +7891,10 @@ bool SFD_GetFontMetaData( FILE *sfd,
 	    d->last_base->scripts = bs;
 	d->last_base_script = bs;
     }
+    else if ( strmatch(tok,"StyleMap:")==0 )
+    {
+    gethex(sfd,(uint32 *)&sf->sfntRevision);
```
frank-trampe: sfntRevision?

adrientetar: I think I noticed it simultanouesly to you while looking at your previous comment, haha. Updated.

frank-trampe:
Oops! I gave you a shell script comment instead of a C comment.

frank-trampe:
Also see my note about the backward compatibility in the s. f. d. parser.

adrientetar:
Updated. I have added backwards-compat reading of SFD attributes.

fontforge/ufo.c
```diff
@@ -1086,8 +1103,38 @@ static int UFOOutputFontInfo(const char *basedir, SplineFont *sf, int layer) {
...
+    {
+        char* preferredFamilyName = fetchTTFAttribute(sf,ttf_preffamilyname);
+        char* preferredSubfamilyName = fetchTTFAttribute(sf,ttf_prefmodifiers);
+        char* styleMapFamily;
+        if (sf->styleMapFamilyName != NULL) {
+            /* Empty styleMapStyleName means we imported a UFO that does not have this field. Bypass the fallback. */
+            if (sf->styleMapFamilyName[0]!='\0')
```
frank-trampe: I would point out that this would fail to round-trip a blank name, but I think that a name cannot be blank.

adrientetar: Yea I think editors don't output an empty field.

frank-trampe:
I'm fine with this.

PR #2160, opened 2015-03-22
wordlistparser: fix glyph pickling

### LINKED ISSUE #2148: UFO: Switching tabs renames and creates glyphs randomly
When editing a UFO where Unicode values are not assigned, like SourceSansPro-Master0.ufo from https://github.com/adobe-fonts/source-sans-pro, switching tabs in the outline window a certain amount of times empties and renames the glyph, creating a random one in its tab.

capture d ecran 2015-03-16 a 07 19 40
becomes this after switching tabs a few times:
capture d ecran 2015-03-16 a 07 19 48
the new glyph is empty and a glyph with its previous name is added at the bottom of the font window.

Using View > Goto, gives the new random empty glyph but double-clicking on the original glyph in the font view opens an outline window with the right glyph name.

adrientetar:
Fix argument mismatch and use NULL-terminated character sequence.

Closes #2148.

r? @frank-trampe

### DIFF
--- a/fontforgeexe/wordlistparser.c
+++ b/fontforgeexe/wordlistparser.c
@@ -3,6 +3,7 @@
 *******************************************************************************
 
     Copyright (C) 2013 Ben Martin
+    Copyright 2014-2015, the FontForge Project Developers.
 
     This file is part of FontForge.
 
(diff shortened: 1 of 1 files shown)


fontforgeexe/wordlistparser.c
```diff
@@ -500,7 +501,7 @@ WordListLine WordlistEscapedInputStringToParsedDataComplex(
 	char glyphname[10];
 	glyphname[0] = ch;
 	glyphname[1] ='\0';
-	SplineChar* sc = SFGetOrMakeChar( sf, ch, 0 );
+	SplineChar* sc = SFGetOrMakeChar( sf, 0, glyphname );
```
frank-trampe: Unicode 0 may exist. Undefined is -1.

adrientetar:
@frank-trampe: any idea how to get the unicodeenc of an unichar? e.g. I paste № (U+2116) in the input field and want to get 2116 from it. Tried various ustring.c functions but no luck.

adrientetar:
What I did here stopped passing incorrect unicodeenc to the function which was creating botchy chars but in fact it's now passing the unicode char as a glyphname to the function which is wrong. It should fetch the unicodeenc from the unicode char and pass it as second argument.

PR #2162, opened 2015-03-23
Fix swapped arguments to SplineCharAntiAlias in EpsGeneratePreview.

frank-trampe:
This addresses a crash from attempting to export a glyph from a typeface supplied by Georg Heeg on the fontforge-devel list on March 6, 2015.


frank-trampe:
@adrientetar, take a look.

PR #2168, opened 2015-03-26
tottf: use SplineFont xheight and capheight values if available

### LINKED ISSUE #2068: OS/2 xheight detection always takes over set value when generating
For test, I set to OS/2 x height correct `ex` for my font. And at this moment my font doesn't contain 'x' character. And `ex` for generated .otf font was incorrect (not from OS/2 x height). Why?

While I add `x` glyph to my font, `ex` detected by this glyph height. 

Why OS/2 x height ignored for `ex`? It's by design?

adrientetar:
Closes #2068.

r? @frank-trampe

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -3528,12 +3528,18 @@ docs are wrong.
     }
 
     if ( os2->version>=2 ) {
-	double xh, caph;
-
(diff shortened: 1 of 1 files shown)


frank-trampe:
It looks good to me.

PR #2172, opened 2015-03-27
sfd,ufo: proper use of strcmp

adrientetar:
r? @frank-trampe


frank-trampe:
Hmm. I guess that we both missed that before somehow.

Those changes look good to me.

Did you mean to put the styleMapFamily initialization into this pull request?

adrientetar:
> Did you mean to put the styleMapFamily initialization into this pull request?

Well yes, I'm just doing little corrections around what I made earlier.

PR #2174, opened 2015-03-29
lookupui: show feature code in the selection dropdown

adrientetar:
It will show e.g. `smcp Small capitals` instead of just `Small capitals` in the selection dropdown. Very useful because one usually look up the feature names by their code, not by a verbose description.

r? @frank-trampe

### DIFF
--- a/fontforgeexe/lookupui.c
+++ b/fontforgeexe/lookupui.c
@@ -1499,7 +1499,9 @@ static GTextInfo *FeatureListFromLookupType(int lookup_type) {
 	cnt = 0;
 	for ( i=0; friendlies[i].tag!=0; ++i ) if ( friendlies[i].masks&mask ) {
 	    if ( k ) {
-		ti[cnt].text = (unichar_t *) copy( friendlies[i].friendlyname );
+		char buf[128];
+		snprintf(buf, sizeof buf, "%s %s", friendlies[i].tagstr, friendlies[i].friendlyname);
+		ti[cnt].text = (unichar_t *) copy( buf );
 		ti[cnt].text_is_1byte = true;
 		ti[cnt].userdata = friendlies[i].tagstr;
 	    }


gdraw/gscrollbar.c
```diff
@@ -520,8 +520,7 @@ static void GScrollBarFit(GScrollBar *gsb) {
     minheight = 2*(gsb->thumbborder+gsb->arrowsize) + GDrawPointsToPixels(gsb->g.base,2);
 
     if ( gsb->g.vert ) {
-	if ( gsb->g.r.width==0 )
-	    gsb->g.r.width = GDrawPointsToPixels(gsb->g.base,_GScrollBar_Width);
+	gsb->g.r.width = GDrawPointsToPixels(gsb->g.base,_GScrollBar_Width);
```
frank-trampe: Dropping the conditional seems a reasonable change to me, but why would somebody have added it in the first place?

adrientetar: Probably in case some other function did another calculation in some cases. Dropping it assures it follows the same rules everywhere.

frank-trampe: I looked at this a bit more. _GScrollBar_Width is a global variable that gets initialized once, so the call to GDrawPointsToPixels would return the same thing every time unless the screen resolution were to change or unless _GScrollBar_Width were used before and after initialization. If you see inconsistent scrollbar widths, there is probably a deeper problem. Can you provide more information about the observed problem?

adrientetar: The commit message says it itself: it's about having scrollbars in the lookup pane with the same width as the other ones in the program.

frank-trampe: Okay. I'll try to figure out why that is happening.

frank-trampe:
I need to get a release out today. I will plan to look at this again on Wednesday.

davelab6:
@frank-trampe any luck?

frank-trampe:
I spent some time narrowing the problem, but the work is not done. I plan to resume next week.

davelab6:
This was submitted in March. Any reason not to merge?

fontforgeexe/lookupui.c
```diff
@@ -1499,7 +1499,9 @@ static GTextInfo *FeatureListFromLookupType(int lookup_type) {
 	cnt = 0;
 	for ( i=0; friendlies[i].tag!=0; ++i ) if ( friendlies[i].masks&mask ) {
 	    if ( k ) {
-		ti[cnt].text = (unichar_t *) copy( friendlies[i].friendlyname );
+        char buf[128];
+        snprintf(buf, sizeof buf, "%s %s", friendlies[i].tagstr, friendlies[i].friendlyname);
+		ti[cnt].text = (unichar_t *) copy( buf );
```
frank-trampe: The indentation is off a bit here.

davelab6:
@frank-trampe @adrientetar anything pending on this? I agree totally this is useful and desirable, so I'd love to see it merged

frank-trampe:
@davelab6, I think that the commit fails to address the root problem and makes it more difficult to solve it in the future. The manifested problem is sufficiently small that it seems acceptable to me to let it slide until my schedule clears or somebody else can dig in. If we do merge it, we need to put a big TODO or FIXME comment above the kludge.

davelab6:
@frank-trampe I don't understand why the root problem must be addressed. I am fine with a big TODO and filing an issue also, and merging this.

adrientetar:
The scrollbars commit is minor and can be dropped.

frank-trampe:
I'll make a new pull request for the feature code commit.

jtanx:
I've updated the branch to just have the feature commit, and split off the other commit into #2809 (the scrollbars branch) which can be updated in the future if/when that gets fixed.

jtanx:
I've also fixed up the indentation.

frank-trampe:
Hmm. That's odd. I thought that I already took care of this a long time ago, but I guess not.

PR #2175, opened 2015-03-31
Fit glyphs in fontview by default, correct fitting

### LINKED ISSUE #2100: turning off fit to em, but ignoring certain large glyphs
if a font includes very large glyphs (giant radical symbols or parentheses for example), turning off fit-to em will make the glyphs too small

adrientetar:
Closes #2100.

r? @frank-trampe


frank-trampe:
Is it wrong to perform the pixel size adjustment in the first case of SplineFontPieceMeal as the existing code does?

adrientetar:
It is, as of this PR, done unconditionally.

frank-trampe:
Ah. I see. In that case, the code looks good to me. Have you tested?

adrientetar:
Yes, it works as intended.

frank-trampe:
Merge it.

PR #2181, opened 2015-04-03
wordlistparser: really fix codepoint lookup

### LINKED ISSUE #2148: UFO: Switching tabs renames and creates glyphs randomly

adrientetar:
#2160 followup. Closes #2148.

r? @frank-trampe

### DIFF
--- a/fontforgeexe/wordlistparser.c
+++ b/fontforgeexe/wordlistparser.c
@@ -498,10 +498,8 @@ WordListLine WordlistEscapedInputStringToParsedDataComplex(
 	    }
 	}
 
-	char glyphname[10];
-	glyphname[0] = ch;
-	glyphname[1] ='\0';
-	SplineChar* sc = SFGetOrMakeChar( sf, -1, glyphname );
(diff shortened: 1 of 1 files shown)


adrientetar:
Well mostly a revert for now... what I did was wrong and problem is still present.

adrientetar:
I have basically broken the display of codepoints in the metrics window with my last PR so I'll merge this one and I can make a followup if needed.

Issue #2184, opened 2015-04-05
Update the OFL to be without reserved font name


andreaslarsen:
Since it's the default for OFL 1.1 shouldn't it be removed from the license text (users wanting reserved font names can add it them themselves)

I'll update https://github.com/fontforge/fontforge/blob/519c5f74a72424125717769fb81a11069f7a3bc5/fontforge/ofl.c if you guys think it's a good idea.

davelab6:
Great idea :)

n7s:
The default? For whom exactly? According to which policy? 

This is what the original OFL.txt example header actually has:  

Copyright (c) <dates>, <Copyright Holder> (<URL|email>),
with Reserved Font Name <Reserved Font Name>.
Copyright (c) <dates>, <additional Copyright Holder> (<URL|email>),
with Reserved Font Name <additional Reserved Font Name>.
Copyright (c) <dates>, <additional Copyright Holder> (<URL|email>

Showing the original header and leaving designers to decide for themselves is probably a better approach.

You should indicate that it's an optional feature and point to the relevant FAQ entry (http://scripts.sil.org/OFL-FAQ_web#cbd32a36 ) but going out of your way to rewrite the original header and hide that feature is really making a decision instead of the authors.

andreaslarsen:
Having the RFN text in the header as default is just as much making a decision as not having it. There are basically two users: 

1. those who know about RFN - they will know how to add it themselves
2. those who don't - they will think it's a mandatory part of the license

In other words:

a. Fontforge with RFN as default => many new users will keep it and has to be told to remove it later on to allow for e.g. simple webfont conversions without the need for renaming

a. Fontforge without RFN as default => advanced users will know how to re add it and new user will just release fonts without RFN => no consequences (at least I haven't heard of a single case of font name hijacking but I might be wrong)

The flaw is obviously in OFL where a license with the same name can be with and without RFN - they should have given one of the licenses a suffix similar to how cc licenses are made. Or even better: no RFN at all - it's a hypothetical scenario anyways.

n7s:
[...]

PR #2185, opened 2015-04-05
Ofl license without rfn

andreaslarsen:
Removed the reserved font name as discussed in #2184


fontforge/ofl.c
```diff
@@ -6,8 +6,7 @@
 #include "ofl.h"
 
 static char *ofl_eng[] = {
-    "Copyright (c) %1$d, %2$s (<URL|email>),",
-    "with Reserved Font Name %s.",
```
adrientetar: You need to remove the corresponding printing code as well.

adrientetar:
Please drop the merge commits as well.

andreaslarsen:
I can fix the merge commits but wouldn't know how to remove the corresponding printing code - sounds like something you can do in a split second so I'm just closing this ;)

adrientetar:
It needs to be refactored to follow [this](https://github.com/adobe-fonts/source-serif-pro/issues/3#issuecomment-82096174) anyway. I'll have a look when I have time.

PR #2192, opened 2015-04-09
sflayout: allow building without native scripting

### LINKED ISSUE #2154: sflayout.c unconditionally calls arrayfree while arrayfree is only if  _NO_FFSCRIPT not defined.
arrayfree() is defined in scripting.c and guard protected by  _NO_FFSCRIPT macro

but sflayout.c unconditionally calls arrayfree without _NO_FFSCRIPT guard. thus result  link error when build without native-scripting support.

adrientetar:
Closes #2154.

r? @frank-trampe

### DIFF
--- a/fontforge/sflayout.c
+++ b/fontforge/sflayout.c
@@ -1187,8 +1187,10 @@ void FontImage(SplineFont *sf,char *filename,Array *arr,int width,int height) {
     li->ps = -1;
     SFMapOfSF(li,sf);
 
+#ifndef _NO_FFSCRIPT
     if ( arr==NULL || arr->argc<2 )
 	arr = freeme = SFDefaultScriptsLines(arr,sf);
+#endif /* _NO_FFSCRIPT */
 
     cnt = arr->argc/2;
     len = 1;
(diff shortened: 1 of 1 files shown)


frank-trampe:
@adrientetar, it is not immediately clear why the call to SFDefaultScriptsLines is unnecessary when native scripting is disabled. Could you explain what you found?

adrientetar:
It's not exactly clear to me either. The Array data type is declared in scripting.h and arrayfree() is unavailable when native scripting is absent so I figured that the function which allocates should be deactivated as well but I'm not sure I understand what's strictly scripting code and what's regular code in there&hellip;

frank-trampe:
I think it more likely that the Array stuff was originally implemented in scripting.h because that is where it was first needed but was then used elsewhere. It probably needs to go into some set of files other than scripting.c and scripting.h. It is probably better for it to go into a new set of files. Or perhaps we could use utils.c, make utils.h, add utils.h to the appropriate build targets, and include it in scripting.h and sflayout.c.

adrientetar:
So then maybe the solution is to take arrayfree() out of ifdefed code?

frank-trampe:
Actually, that's probably fine. I was thinking that the includes for scripting.h would be enclosed in ifdefs in most places, but that seems not to be the case. There might be something cleverly hidden in the build system that drops scripting.c if the build is to happen without scripting, so be sure to test a scripting-free build after making the change.

frank-trampe:
I'm replacing this with #2249.

PR #2193, opened 2015-04-09
ttf,ufo: make version number and id inline with the AFDKO

adrientetar:
The otspec does not require padding the version string but it clears e.g. 1.8 vs. 1.008 ambiguities and makes us inline with the AFDKO. 
As for the unique id, the spec puts no restriction on it, and it's not exported to UFO unless manually set so&hellip; it does not matter as much to me. The datetime that was included is redundant because there is already a field for that in the head table (much like every other thing included here), so dropping it is not a problem.
Round-tripping still unchanged because version number is not recalculated if ttf_version is set.

r? @frank-trampe

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -3793,6 +3793,8 @@ return( copy(str));		/* Either in ASCII (good) or appears to be utf8*/
 return( latin1_2_utf8_copy(str));
 }
 
+char* paddedVersionStr(const char * textVersion, char * buffer);
+
 /* Oh. If the encoding is symbol (platform=3, specific=0) then Windows won't */
 /*  accept the font unless the name table also has entries for (3,0). I'm not */
 /*  sure if this is the case for the CJK encodings (docs don't mention that) */
@@ -3802,6 +3804,9 @@ void DefaultTTFEnglishNames(struct ttflangname *dummy, SplineFont *sf) {
     struct tm *tm;
     char buffer[200];
 
+    char versionStr[6] = "1.000";
+    if (sf->version!=NULL) paddedVersionStr(sf->version, versionStr);
+
     if ( dummy->names[ttf_copyright]==NULL || *dummy->names[ttf_copyright]=='\0' )
 	dummy->names[ttf_copyright] = utf8_verify_copy(sf->copyright);
     if ( dummy->names[ttf_family]==NULL || *dummy->names[ttf_family]=='\0' )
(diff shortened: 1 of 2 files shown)


fontforge/ufo.c
```diff
@@ -1139,9 +1152,11 @@ static int UFOOutputFontInfo(const char *basedir, SplineFont *sf, int layer) {
...
       int versionMajor = -1;
       int versionMinor = -1;
-      if (sf->version != NULL) extractNumericVersion(sf->version, &versionMajor, &versionMinor);
-      if (versionMajor >= 0) PListAddInteger(dictnode,"versionMajor", versionMajor);
-      if (versionMinor >= 0) PListAddInteger(dictnode,"versionMinor", versionMinor);
+      if (sf->version != NULL) {
+          extractNumericVersion(sf->version, &versionMajor, &versionMinor);
+          if (versionMajor >=0) PListAddInteger(dictnode,"versionMajor", versionMajor);
+          if (versionMinor >=0) PListAddInteger(dictnode,"versionMinor", versionMinor);
+      }
```
frank-trampe: I do not understand what changes here. With the previous code, if sf->version is NULL, then versionMajor and versionMinor do not get written anyway.

adrientetar: I can revert that part, I had changed more things initially.

fontforge/ufo.c
```diff
@@ -107,6 +107,16 @@ static void injectNumericVersion(char ** textVersion, int versionMajor, int vers
...
+/* The spec does not really require padding the version str but it clears */
+/* 1.8 vs. 1.008 ambiguities and makes us inline with the AFDKO. */
+char* paddedVersionStr(const char * textVersion, char * buffer) {
+    int major = -1;
+    int minor = -1;
+    if (!extractNumericVersion(textVersion, &major, &minor)) return (char *) textVersion;
+    snprintf(buffer, 6, "%d.%03d", major, minor);
+    return buffer;
+}
+
```
frank-trampe: I cannot confirm or dispute that it is desirable to add zeros like that. So I will call upon @DavidRaymond and @miguelsousa.

fontforge/ufo.c
```diff
@@ -80,23 +80,23 @@ static char *buildname(const char *basedir, const char *sub) {
...
-  if (nextText1[0] == '\0') return;
+  } else return 0;
+  if (nextText1[0] == '\0') return 1;
   tempVersion = strtol(nextText1+1, &nextText2, 10);
   if (tempVersion != -1 && nextText2 != NULL && (nextText2[0] == '\0' || nextText2[0] == ' ' || nextText2[0] == '.')) {
     *versionMinor = tempVersion;
-  } else return;
-  return;
+  } else return 1;
+  return 2;
```
frank-trampe: I think that it is unnecessary and duplicative to add a set of return flags that just mirror the values placed in versionMajor and versionMinor.

fontforge/tottf.c
```diff
@@ -3809,23 +3814,24 @@ void DefaultTTFEnglishNames(struct ttflangname *dummy, SplineFont *sf) {
...
-	tm = localtime(&now);
-	sprintf( buffer, "%s : %s : %d-%d-%d",
-		BDFFoundry?BDFFoundry:TTFFoundry?TTFFoundry:"FontForge 2.0",
-		sf->fullname!=NULL?sf->fullname:sf->fontname,
-		tm->tm_mday, tm->tm_mon+1, tm->tm_year+1900 );
+    int major = -1;
+    int minor = -1;
+	sprintf( buffer, "%s;%s;%s %s",
+        versionStr,
+		BDFFoundry?BDFFoundry:TTFFoundry?TTFFoundry:"UKWN",
```
frank-trampe: I think that the original code works around the fact that FontForge does not come with its default TTFFoundry set to FontForge. If we want to drop the work-around, we need to fix the problem. I'm no TrueType expert; is "FontForge 2.0" a correct default value?

adrientetar: I'm not going to touch ttf codegen after all and I opened up a new PR, #2195.

PR #2195, opened 2015-04-09
ufo: make build-time version number inline with the AFDKO

adrientetar:
#2193 replacement, with less changes.

The otspec says:

~~~
The string must contain a version number of the following form: one or more digits (0-9) of value less than 65,535, followed by a period, followed by one or more digits of value less than 65,535.
~~~

r? @frank-trampe

### DIFF
--- a/fontforge/ufo.c
+++ b/fontforge/ufo.c
@@ -107,6 +107,17 @@ static void injectNumericVersion(char ** textVersion, int versionMajor, int vers
   return;
 }
 
+/* The spec does not really require padding the version str but it clears */
+/* 1.8 vs. 1.008 ambiguities and makes us inline with the AFDKO. */
+char* paddedVersionStr(const char * textVersion, char * buffer) {
+    int major = -1;
+    int minor = -1;
+    extractNumericVersion(textVersion, &major, &minor);
+    if (major < 0 || minor < 0) return (char *) textVersion;
+    snprintf(buffer, 6, "%d.%03d", major, minor);
+    return buffer;
+}
+
 const char * DOS_reserved[12] = {"CON", "PRN", "AUX", "CLOCK$", "NUL", "COM1", "COM2", "COM3", "COM4", "LPT1", "LPT2", "LPT3"};
 const int DOS_reserved_count = 12;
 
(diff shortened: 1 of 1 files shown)


frank-trampe:
This does seem to accomplish the stated objective rather concisely. I'll copy @miguelsousa and @DavidRaymond once more so that they know that the padding question has moved to a new thread.

adrientetar:
The spec allows it per OP and other fonts use this format already.

frank-trampe:
Legality and desirability are distinct qualities. I would like to avoid adding complexity (legal or not) unless this approach is measurably better in some way.

adrientetar:
It's only really for consistency because when this data is absent, the fallback attribute generated by ufo2fdk has padding zeroes.

frank-trampe:
Don't close the issue. It may well be the better way to do this. I just want to be sure that it is.

frank-trampe:
@DavidRaymond, @miguelsousa, is it good to pad the version string like this (3 digits for the minor version)?

miguelsousa:
@frank-trampe that's what the FDK has always done. @readroberts may be able to provide more details.

davelab6:
I think this is a good thing to do. Thanks @adrientetar !

frank-trampe:
@readroberts, what is the rule or system that you use for the version numbers? Do extra digits over 3 in the minor version get trimmed?

frank-trampe:
@readroberts?

readroberts:
I suggest you keep the common tradition of  interpreting at the head table fontRevision field as a 16.16 Fixed value, and round and pad the fractional part to 3 decimal places. 0x00010100L, aka decimal 1.001953125, should be represented as '1.002'

When I first started working on the FDK in 1998, there were several different implementations of the head table fontVersion field. In Both Apple and MS fonts, there are fonts which treated the fontRevision field as two separate short integer values, as well as a 16.16 Fixed value. In both MS and Apple system fonts, some reported  the version in name ID 5 as being padded to 3 decimal places, and some didn't. A discussion on the OpenType list produced a consensus to do what is now common practice. The Open Type spec is now clear that the value is a Fixed value. It is also clear that you do NOT need to zero pad to any number of decimal places - "a period, followed by one or more digits". However, I personally, and quite a few other developers, find this practice useful.

davelab6:
Thank you @readroberts !! I really appreciate you taking the time to participate in FontForge's development :)

frank-trampe:
It seems to me that the proposed changes follow the pattern suggested by @readroberts.

PR #2196, opened 2015-04-10
Unloop ft context

mskala:
Calling FreeTypeFontContext can be an expensive operation.  If it fails consistently, then existing code will call it on *every single glyph*, which can cause a simple rasterization operation to take several hours on a large font.  The only reason to put the call inside the per-glyph loop is to handle the case of no glyphs being worth "outputting"; better to do the operation unconditionally outside the loop, which wastes it harmlessly once in the rare case instead of hundreds of thousands of times in what turned out to be for me a common case.

Patch backported from FontAnvil.


adrientetar:
@mskala Can you squash your commits?

mskala:
What does that mean?

adrientetar:
Concat them into one.

adrientetar:
Thanks!

Issue #2169, opened 2015-03-26
Corruption of 1st field of ttflangnames


adrientetar:
When opening then closing SourceSansPro_0.ufo, stack guards are raised with an invalid free (more likely to be seen when keeping another font open so the program doesn't close). It might be an overrun on the first field of sf->names, I watched the functions that set sf->names and they seem to do their job so it might be something contiguous in memory that makes invalid accesses.

~~~
*** Error in `/usr/local/bin/fontforge': double free or corruption (fasttop): 0x00000000008867c0 ***

Program received signal SIGABRT, Aborted.
0x00007ffff71d2e37 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56	../nptl/sysdeps/unix/sysv/linux/raise.c: Aucun fichier ou dossier de ce type.
(gdb) where
#0  0x00007ffff71d2e37 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff71d4528 in __GI_abort () at abort.c:89
#2  0x00007ffff7213f74 in __libc_message (do_abort=do_abort@entry=1, 
    fmt=fmt@entry=0x7ffff731cf00 "*** Error in `%s': %s: 0x%s ***\n")
    at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007ffff721b586 in malloc_printerr (ptr=<optimized out>, 
    str=0x7ffff731d0c8 "double free or corruption (fasttop)", action=1)
    at malloc.c:4996
#4  _int_free (av=<optimized out>, p=<optimized out>, have_lock=0)
    at malloc.c:3840
#5  0x00007ffff591e2c2 in TTFLangNamesFree (l=0x886840) at splineutil.c:5642
#6  0x00007ffff5921277 in SplineFontFree (sf=0x884840) at splineutil.c:6485
#7  0x00007ffff793e0c2 in FontView_Free (fv=0x15fcd10) at fontview.c:7550
#8  0x00007ffff793caca in FontViewRemove (fv=0x15fcd10) at fontview.c:7156
#9  0x00007ffff793cdfa in fv_e_h (gw=0x19fd050, event=0x7fffffffba40)
    at fontview.c:7237
#10 0x00007ffff681ebfc in _GWidget_Container_eh (gw=0x19fd050, 
    event=0x7fffffffba40) at gcontainer.c:403
[...]

PR #2214, opened 2015-04-23
Fix a crash and a bad access in the U. F. O. code.

frank-trampe:
This fixes a double-free introduced in #2124 (which I then reverted and then partially applied) and a bad access from a long time ago.


frank-trampe:
@adrientetar, take a look.

adrientetar:
Does it fix #2169?

frank-trampe:
I can't test just now. Can you?

davelab6:
@adrientetar were you able to test this?

adrientetar:
No.

Issue #2212, opened 2015-04-23
Fontforge crushes when reading TT gridfit instructions while the SFD's line endings are CRLF.


be5invis:

adrientetar:
@be5invis Can you explain how to reproduce the crash?

be5invis:
just open an ttf with instructions using the windows version of Fontforge, save and reopen, and, it crushes.






发自 Windows 邮件





发件人: Adrien Tétar
发送时间: ‎2015‎年‎4‎月‎24‎日, ‎星期五 ‎0‎:‎33
收件人: fontforge/fontforge
抄送: Invis Belleve





@be5invis Can you explain how to reproduce the crash?

frank-trampe:
@adrientetar, where was the actual crash? That's probably something worth fixing in #2216.

adrientetar:
Well I started saying crash but OP said crushes and I too only get a parse error (parsing gets offset due to not taking cr into account).

frank-trampe:
Good.

frank-trampe:
@be5invis, I just merged a fix from @adrientetar. Would you pull from master and test?

be5invis:
@frank-trampe it works, thanks.
btw please fix #1918 as fast as possible.

PR #2216, opened 2015-04-24
ttfinstrs: handle carriage returns when parsing

adrientetar:
cc #2212.


frank-trampe:
This looks safe to me. Have you tested with a file with each linebreak style?

frank-trampe:
Ping.

adrientetar:
No I have not tested. This will only change behavior for files which have carriage returns.

frank-trampe:
When do you expect to be able to test? I'm cutting the April release soon.

frank-trampe:
I'm going to assume that passing the unit tests is good enough to assure that there are no regressions.

PR #2224, opened 2015-04-26
Use the suggested defaults for RadicalDisplayStyleVerticalGap and Radica...

fred-wang:
RadicalDisplayStyleVerticalGap and RadicalRuleThickness math constants currently default to 0. This patch uses the values suggested in http://www.microsoft.com/typography/otspec/math.htm. It also adds comment for other constants to make explicit when no default is given.

### DIFF
--- a/fontforge/mathconstants.c
+++ b/fontforge/mathconstants.c
@@ -108,6 +108,7 @@ struct MATH *MathTableNew(SplineFont *sf) {
     math->ScriptScriptPercentScaleDown	= 60;
     math->DelimitedSubFormulaMinHeight	= emsize*1.5;
     /* No default given for math->DisplayOperatorMinHeight */
+    /* No default given for math->MathLeading */
     /* No default given for math->AxisHeight */
     sc = SFGetChar(sf,'x',NULL);
     if ( sc!=NULL ) {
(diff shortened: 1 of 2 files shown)


adrientetar:
Thanks!

frank-trampe:
@fred-wang, does the test that you wrote pass for you?

fred-wang:
Yes, it passed for me. However, I noticed C and Python can round up floats differently. For example, on my machine I had to explicitly do -int(10*font.em/18) for the negative value math.RadicalKernAfterDegree.

I guess you can revert the second commit and I'll try to adapt the test again later.

jtanx:
You can probably change the check to see if the value is very close to the expected value, e.g. to within 1e-6 difference. 

E.g. change to:

     math.abs(math.DelimitedSubFormulaMinHeight -  font.em*1.5) < 1e-6

frank-trampe:
The two tests that fail involve no computations, and the values are not close.

```
>>> math.RadicalDisplayStyleVerticalGap != font.uwidth
True
>>> math.RadicalDisplayStyleVerticalGap
0
>>> font.uwidth
50.0
>>> math.RadicalKernBeforeDegree != font.uwidth
True
>>> math.RadicalKernBeforeDegree
277
>>> font.uwidth
50.0
```

frank-trampe:
There is also a duplicate test.

frank-trampe:
Oops. Ignore my reports of the failing values. I was importing an installed (and thus older) version of FontForge. But RadicalDisplayStyleVerticalGap is still tested and set according to different formulae.

fred-wang:
The C code returns only integer values, so I believe it should be sufficient to use "int(...)" in the Python code in order to force the rounding and cast to integer.

Regarding RadicalDisplayStyleVerticalGap, when I tested it math->AccentBaseHeight = 0 so only uwidth was relevant. Indeed, in the c code, the x-height is obtained by measuring the 'x' glyph when one exists. This is not the case in the Python script, since it uses an empty font. A comment should probably be added to explain that.

frank-trampe:
Would `math.RadicalDisplayStyleVerticalGap != font.uwidth + 0.25 * math.AccentBaseHeight` be wrong?

fred-wang:
Yes, I guess this would be better to write it like this in the test (however this would not change the result since math.AccentBaseHeight = 0).

fred-wang:
@frank-trampe @jtanx Using int(...) as I proposed in https://github.com/fontforge/fontforge/pull/2224#issuecomment-97892933 was enough to make the tests pass on travis. See https://github.com/fontforge/fontforge/pull/2241 This also uses a nonzero x-height so that the expression of RadicalDisplayStyleVerticalGap in the test matches the one in the C source.

PR #2225, opened 2015-04-26
Fix bad allocation in ParseComponentTuple

fred-wang:
With the modification in this patch, tests/test1006.py should demonstrate the error:
*** Error in `python': munmap_chunk(): invalid pointer: 0x00007f1ae263a444 ***

With a similar testcase I also obtained
*** Error in `/usr/bin/python': free(): invalid pointer: 0x00007ffff7e5b684 ***

The issue seems to be that PyArg_ParseTuple does not allocate a new string for parts[i].component. This patch make my testcases work as expected.

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -6870,6 +6870,7 @@ return( NULL );
     for ( i=0; i<cnt; ++i ) {
 	PyObject *obj = PySequence_GetItem(tuple,i);
 	int extender=0, start=0, end=0, full=0;
+	char *glyphName;
 	if ( PyType_IsSubtype(&PyFF_GlyphType, Py_TYPE(obj)) ) {
 	    parts[i].component = copy( ((PyFF_Glyph *) obj)->sc->name );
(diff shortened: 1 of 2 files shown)


tshinnic:
Reading the Python docs and checking against other code in `python.c` it does seem correct that use of format 's' does *not* allocate a copy of the string.  A continuing need for the string (like returning it in the `parts[]` array) would require something like `copy()`.

"PyArg_ParseTuple()
s (str) [const char `*`]
A pointer to **an existing string** is stored in the character pointer variable whose address you pass."

As opposed to the usage of a format like 'es':

"PyArg_ParseTuple() will allocate a buffer of the needed size, copy the encoded data into this buffer and adjust `*`buffer to reference the **newly allocated storage**. The caller is responsible for calling PyMem_Free() to free the allocated buffer after use."

Checking many calls in `python.c` with various usages of both formats with `PyArg_ParseTuple()` I see that when the string is needed further `copy()` is called, as done by Fred's fix.  Looks good.

PR #2241, opened 2015-05-01
Add math constants tests. #2224

fred-wang:
@frank-trampe I'm trying again with int(rount(...)) as suggested on http://stackoverflow.com/questions/5997027/python-rounding-error-with-float-numbers

(I'll add more tests with x-height = math.AccentBaseHeight != 0 later)


fred-wang:
OK,  int(rount(...)) was not a good idea. Just doing int(...) as I initially proposed made the test pass.

This second commit also adds tests involving x-height by adding a dummy glyph for 'x'.

frank-trampe:
@fred-wang, I'd like to keep the broken commit out of the history. Would you squash the commits (via `git rebase -i`)?

frank-trampe:
Oh. Actually, could you just make a clean branch from master, apply the change, and push that? I think that your current branch head is wrong.

fred-wang:
Yes, that's what I'll do

PR #2242, opened 2015-05-01
Axis height

fred-wang:
This depends on #2241

@khaledhosny What do you think about the heuristic in this last commit: using the middle of the '+' sign as the default value for axisHeight? (When I tested with Open Type MATH fonts last week, only Libertine had a value significantly different)


khaledhosny:
Should be fine I think, you may want to check other similar symbols as a fallback.

Issue #1827, opened 2014-10-20
Remove outlines and more padding on Tools Pallet 


pathumego:
Graphics
- [x] Create new icons
- [x] Paletteselectedbg.png
- [x] Mouse icon instead of Mse-stuff

Code
- [ ] Support for theme-adjustable palette margins + spacing
- [ ] Support for Paletteselectedbg.png (https://github.com/fontforge/fontforge/issues/1827#issuecomment-97260391)
- [ ] Support for the proposed restyling of Mse and make it "un-clickable"(https://github.com/fontforge/fontforge/issues/1827#issuecomment-98128014) + (https://github.com/fontforge/fontforge/issues/1456)
- [ ] Change the order of the palette icons a bit (https://github.com/fontforge/fontforge/issues/1827#issuecomment-98878311)
- [ ] Use visiblility_off/on.png as default instead of them ugly eyes.

Previously
- [x] <strike>Removing the outlines on the toolbar pallet button would make a pleasant difference. And add 2px padding would be needed.</strike> (see below)

- [x] ~~The eye icon is clipped on layers panel, maybe increase of the width would fix that.~~

andreaslarsen:
What I think needs to be done is not use active icons but have a palleteselectedbg.png that gets put behind any icon when it's selected (also the spire toggle which currently has "up + down")

**Example**

paletteruler.png

![screen shot 2015-04-29 at 01 14 54](https://cloud.githubusercontent.com/assets/3859425/7382274/3a6cc584-ee0d-11e4-9976-01045af65521.png)

palleteselectedbg.png

![screen shot 2015-04-29 at 01 15 01](https://cloud.githubusercontent.com/assets/3859425/7382283/5618da8e-ee0d-11e4-8ee9-4cbcb95ad712.png)

selected and unselected state

![screen shot 2015-04-29 at 01 15 16](https://cloud.githubusercontent.com/assets/3859425/7382290/65497aea-ee0d-11e4-956d-85b788bc80df.png)

andreaslarsen:
Shown together with a couple of more icons


![screen shot 2015-04-29 at 01 19 00](https://cloud.githubusercontent.com/assets/3859425/7382321/c3271c1c-ee0d-11e4-9348-1ae246b38d51.png)

davelab6:
Agree

andreaslarsen:
maybe keep up+down for the spiro toggle though since it's a bit different than the others though

andreaslarsen:
[...]

PR #2245, opened 2015-05-01
New icons - see comment before merging

larsenwork:
I'd be great if someone with c-skills would look into https://github.com/fontforge/fontforge/issues/1827#issuecomment-98128014 before merging


andreaslarsen:
this could be added now if we can live with the slightly cramped mse toolbox 
screen shot 2015-05-25 at 21 30 11

frank-trampe:
Allowing differing spacing values is not the difficult task here. I thought that we needed to support background images for button states.

And I still don't see a reason that FontForge needs to be the piece of software to introduce odd icon sizes to the world. I can't imagine that we're the first people to think of it.

larsenwork:
@frank-trampe Why is uneven an issue (honest question)? What you get is ability to make horizontal and vertical centred lines which is useful in many of these icons.

Icons are traditionally being drawn over an even numbered canvas but that's simply a convention based on scaleability reasons which isn't a concern here so I see no reason not to make them centred (and thus uneven) but I might be missing something?

The BG has nothing to do with the icons per se. It's related to turning the boxy 3D on/off which will complement whatever icons is used. I put it in this pr since it's in the same sprite in my graphics file + if/when someone with code skills want to add the feature the file is already there.

larsenwork:
@frank-trampe ?

frank-trampe:
@andreaslarsen, sorry for the delay. I've been travelling a bit.

I'm not trying to be dogmatic about even icon sizes, it just seems to me that a lot of applications (the vast majority in the world today) get by with even icon sizes, and I would like to avoid rocking the boat in FontForge if there is some way that you can make them work for you.

If not, I ought to be able to work on the icon sizing functionality after I churn through some of the open pull requests.

larsenwork:
@frank-trampe travelling is always nice:) I'll try to make even numbers work - any issues if I opt for bigger sizes (2 or 4px bigger than current)?

frank-trampe:
Actually, now that I think about it, this might be a great time to up the icon sizes considerably. It would still require me to rework the size handling, but screens have been getting a lot larger in the last fifteen years. We could still keep the old theme available for those with smaller screens.

@davelab6, @guiguru, any thoughts?

davelab6:
I don't think you should do more or less than @larsenwork is requesting :)

davelab6:
@larsenwork i think even numbers is ok, just leave a blank row to the right/bottom of what you have. 

The massive improvement of your icons is so big, and the exact centering with an odd numbered icon size is so small a detail, it is a pity to delay shipping the new icons on it

frank-trampe:
@davelab6, I'm adding support for different icon sizes either way. If you don't have an opinion on the ideal size for the new icons, that's fine.

Upon some reflection, I think that four pixels more in each direction is about right.

larsenwork:
@frank-trampe sounds great, if you're adding support for different icon sizes then these uneven would work too, or? If so then I'll clean up this pr so you can merge it when you add support for different icon sizes.

I agree about screen sizes and bigger icons - especially with hdpi displays. What could be cool is if we shipped with maybe 3 icon sizes and let the user could choose in preferences what icon size they want.

guiguru:
since @frank-trampe asked, let me say this:

it is super difficult for me to comment here, since I am not **immersed in** the project. context is _everything_ in design decisions, and I am missing the context.

although, I know one context: has @davelab6 not declared fontforge end-of-life? then I would take the good work already done and not go on paths that lead to more and more rework of the UI.

ps: adding a user preference always smells bad; usually it stinks.

larsenwork:
@guiguru obviously detecting screen dpi and setting the icon size (and UI font size) accordingly would be better but also (I assume) a lot more difficult to code. I don't see how olfactory response is relevant.

Context is trying to make FontForge easier to use for a beginner (@davelab6 call this low hanging fruit) - it's not adding new advanced features which requires too much work and would be easier to do in other apps/projects based on modern frameworks (e.g. web based such as Glyphr)

guiguru:
I was just excusing myself.

context is the _whole_ as-is situation. it would take me at least one super-intense, 8-hour day with you guys (in the same room, nothing remote) to get into the project and know the context. then I can give you the answers you look for. _(no, I am not proposing to do this. just excusing myself.)_

davelab6:
On 24 June 2015 at 05:12, Andreas Larsen <notifications@github.com> wrote:

> @guiguru <https://github.com/guiguru> obviously detecting screen dpi and
> setting the icon size (and UI font size) accordingly would be better but
> also (I assume) a lot more difficult to code.
>
> FF has some of this, as it autodetects the screen width and height (which
the user can hard set in the prefs)

larsenwork:
@davelab6 could that auto detecting be "bound" to icon size:
```
if screen-width >= 1920 ( iconsize: extraLarge)  
if screen-width >= 1366 && screen-width < 1920 ( iconsize: large)
else ( iconsize: default )
```
or something like it

frank-trampe:
@guguru, @davelab6 has declared FontForge end-of-life for him since he prefers another existing editor (Glyphs, I think) and is excited about possible future free typeface design suites. But there are many people for whom proprietary software is not an option and who want to design fonts now. In four or five years, there may be a great free replacement for FontForge, but, until there is, FontForge is far from dead for the people who need to use it on a regular basis.

davelab6:
Glyphr, is web based. I think its the future of libre font drawing.

davelab6:
I spent the last 3 years holding FontForge workshops around the world, and
the 100s of users who paid me to teach them FontForge do not stick around.
They move to Glyphs.

frank-trampe:
You're not using that as your main typeface design suite at this point, though, right?

davelab6:
I think there are a few low hanging fruit left in fontforge - new icons,
new menus, crash fixes - which _could_ stop the mass migration out of
fontforge into proprietary alternatives. But I'm not going to get involved.
I'm going to put my effort into making Glyphr my my design suite asap.

larsenwork:
I have only played around with Glyphr and initial impression is promising but still year(s) away from something I'd use as main/only editor so FontForge is far from end of life to me and by increasing accessibility by making it easier to install and use I think it can be even more inviting to a lot of new users as well. It's a great and very capable program.

larsenwork:
@frank-trampe I never got any wiser on this: https://github.com/fontforge/fontforge/pull/2245#issuecomment-114684906

davelab6:
I'm extremely happy that @larsenwork has the attitude I had a few years ago and am now burned out from :)

larsenwork:
@davelab6 hehe...well I don't know where you stood a few years ago but I have no ambitions (or skills needed) to add features but solely to make the current clusterfuck a bit better presented and organised so it's easier for beginners. 
Closing this because #2396

davelab6:
See I was less modest and wanted that plus ponies like real time collaboration :)

PR #2247, opened 2015-05-04
Add tests for math constants. #2241 #2224

fred-wang:
@frank-trampe here it is.


frank-trampe:
Okay. It looks fine to me. We'll see what Travis thinks.

Since links do not work in pull request and commit names, I am repeating them here. #2224, #2241, #2247.

PR #2248, opened 2015-05-04
Math constants: Use the middle of the '+' sign as the default AxisHeight...

fred-wang:
... value.


fred-wang:
This is a #2242 rebased after #2247

frank-trampe:
It looks good to me.

PR #2249, opened 2015-05-04
Fix scriptless building.

frank-trampe:
Make the array-handling functions available even when building without native scripting.

This addresses #2154 and supersedes #2192.


frank-trampe:
@adrientetar, take a look.

Issue #1918, opened 2014-11-12
Windows: random glyph moved to background on opening


adrientetar:
Sometimes when opening an SFD on Windows, a random glyph gets moved to the background layer. It’s a bit worrying if there’s a memory corruption somewhere, anyhow it’s unclear why I’ve only seen this on Windows so far.

cc @jtanx

sergey-s-betke:
Workaround for python script:

    font = fontforge.open (sourcefile) 
    for glyph in font.glyphs(): 
    	glyph.foreground += glyph.background

frank-trampe:
@adrientetar, can you reproduce this consistently?

adrientetar:
I don't know. It seems rather random, not sure if it's always the same glyphs or not.

frank-trampe:
If you wanted to maximize the odds of reproducing the problem, what would you do? And what are the odds that it would actually happen each time you do that?

Is this a display issue, or is there foreground data in the input file that ends up in the background in the output file?

be5invis:
@frank-trampe an example sfd is [here](https://gist.github.com/be5invis/b1d764250a7b75fa32c6), the glyph uni1DC8's outline went to background.

frank-trampe:
Is the linked file an input or a [corrupted] output?

be5invis:
@frank-trampe input.

be5invis:
@frank-trampe input.

be5invis:
@frank-trampe input.

be5invis:
and @frank-trampe , by removing all `Back` and `Fore` lines in the SFD, the layer-shifting problem is gone.

frank-trampe:
@be5invis, are you able to reproduce this on Linux, are does it only happen on Windows as for Adrien?

be5invis:
@frank-trampe not tested yet. I will try it later using the computer in my home.

frank-trampe:
I can't reproduce on Linux, so this is probably a platform-specific memory corruption problem. @jtanx, can you reproduce the problem on Windows?

jtanx:
I haven't been able to reproduce this bug with a quick test, but since it's a sporadic issue I'll continue to look into it. I really haven't had much time lately though so I don't know how far I'll get with this.

jtanx:
I have been able to reproduce this using `descret-roman.sfd`. It looks like sometimes it writes out the components in different order (e.g. the 'Fore' is still there, but it's being written out after the spline data):

````diff
[...]

PR #2250, opened 2015-05-05
PeekMatch: Don't rewind by calling ungetc more than once (undefined behaviour)

jtanx:
Related issues: #1918

I don't know if this is an issue anywhere else. Someone needs to check all uses of `ungetc` to see if it is called to rewind by more than one character at a time. @frank-trampe? 

Not extensively tested. What I did:

    fontforge -lang=py -c "f=fontforge.open('descret-roman.sfd');f.save('out.sfd')"
    diff descret-roman.sfd out.sfd

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -148,9 +148,9 @@ static int PeekMatch(FILE *stream, const char * target) {
   while (target[pos1] != '\0' && lastread != EOF && lastread == target[pos1]) {
     pos1 ++; lastread = getc(stream);
   }
-  if (lastread != EOF) ungetc(lastread, stream);
-  int pos2 = pos1;
-  while (pos2 > 0) { ungetc(target[--pos2], stream); }
+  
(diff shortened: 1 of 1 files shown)


frank-trampe:
It looks good to me and passes Travis.

I used ungetc rather than fseek originally so as to be able to deal with pipes. That was not strictly necessary, but it would have been nice to see that work.

jtanx:
Yuck, so it turns out there's a really crap bug on Windows with fseek/fsetpos/fgetpos (probably like (this)[www.progtown.com/topic913984-the-strange-behavior-fsetpos.html]).

Essentially, I've found that if it needs to rewind and that point is at/near a newline, it will incorrectly rewind because it assumes CRLF, when the file only has LF newlines. The fix is to open all files in binary mode, since in that case, it doesn't treat CRLF as special. What I've done is [patch Gnulib](https://github.com/jtanx/fontforgebuilds/blob/master/patches/gnulib.patch) to force files to be opened in binary mode, so no need to change code on FontForge's part.

Issue #2230, opened 2015-04-28
Could we remove icons from the menus?


larsenwork:
Can be done by just deleting the .pngs 

I think it looks a lot cleaner without and icons don't really makes sense in menus.

![screen shot 2015-04-28 at 11 54 44](https://cloud.githubusercontent.com/assets/3859425/7368991/1405fdac-edae-11e4-9e39-c482b7a0e49e.png)

andreaslarsen:
To elaborate a bit: 

It only makes sense to use an icon - when it makes sense without the text. This is true for very few of the menu icons e.g. print, save, new, redo, undo etc.
An example of other apps using these icons is google docs. To me it's a bit redundant though because most users (especially users who download fontforge) knows the keyboard shortcuts for these menu items anyway.

The rest of the icons aren't immediately understood or needs the text explanation to be understood. I really can't see what value it adds having an icon next to it. New users won't recognise the icons and old users know where the menu item is and/or use keyboard shortcuts.

pathumego:
@andreaslarsen I agree with this. Most of the icons dose not make sense.

But the I don't know how other users would feel about this.

davelab6:
I believe that icons on every item were a good idea for applications with
toolbars with dozens of icons (like classic Windows 98 era ms office
applications) so that you could learn where a menu item was on the toolbar
by Pavlovian memorisation of the icon.

But for apps without a toolbar, as we are today, then having icons for
everything is the same as no icons:

The icons are most useful for signposting/telegraphing the most commonly
sought menu items, so that users find them faster and click them with less
errors.

However, when all items have icons, this effect is lost.

I'm happy for to say what the most common actions are :)

andreaslarsen:
they can be used for signposting but this was also more relevant years ago - now people know that e.g. the new file is always located at the top of the file menu and/or they already know the keyboard shortcut. I don't think I have a single app installed besides FF that still uses menu icons.

davelab6:
Well it's not part of the Mac hig, not sure about windows

andreaslarsen:
In e.g. Inkscape (gnome/gtk) they are now also disabled by default

jtanx:
I quite like icons in menus. The icons are present on Windows (not entirely sure how this is controlled). I just checked Inkscape 0.91 - they still have icons too.

davelab6:
I think maintaining 2 Mac/win themes could be good to make it more familiar
to users of each

andreaslarsen:
[...]

PR #2251, opened 2015-05-05
Adding windows theme

davelab6:
Taking action on my comment at https://github.com/fontforge/fontforge/issues/2230#issuecomment-99029372 I propose this for @jtanx to consider.

I simply duplicated the tango theme directory to `windows` in its current state; I hope @jtanx can sync the default theme as he wishes as it proceeds


adrientetar:
Why can't we have the same folder for Windows?

davelab6:
So windows can retain icons on menus without code

jtanx:
To be honest, my preference over having icons is not particularly strong enough to warrant having a separate theme just for Windows. We should just reach a consensus, and if that's to remove the icons, then so be it.

davelab6:
Okay cool :)

Issue #2252, opened 2015-05-05
Store hotkeys in prefs


andreaslarsen:
Currently the default file is overwritten every time you update.

If we moved hotkeys to the prefs file instead this wouldn't happen?

Or are there some downside/reason this wasn't done long time ago?

andreaslarsen:
I can quickly fix number 2 and 3, but not number one - but it is a fairly quick thing to fix, no?

andreaslarsen:
This is how number 3 looks when fixed:

![screen shot 2015-05-05 at 14 40 17](https://cloud.githubusercontent.com/assets/3859425/7472638/d268500c-f334-11e4-85c9-2e5b7be2c15f.png)

davelab6:
I think it's good to have separate files but I think the hotkeys file
should work the same as themes.

Does it not?

andreaslarsen:
Maybe I'm using it wrong but I'm just replacing this: /Applications/FontForge.app/Contents/Resources/opt/local/share/fontforge/hotkeys/default

and that file obviously gets replaced when updating FontForge.app

davelab6:
I tested it, by copyiing defaults to ~/.config/fontforge/` and changing a few keys, and FontForge does not pick it up. 

So instead of 

> Store hotkeys in prefs so they don't get overwritten with every update.

I suggest,

Store hotkeys in `.local/fontforge` (just like prefs) so they don't get overwritten with every update.

davelab6:
The other 2 things look awesome, Please do!!!

PR #2253, opened 2015-05-05
ctl->ctrl, shft->shift, ctrl always 2nd last

andreaslarsen:
See #2252 for details, don't merge just yet

### DIFF
--- a/gdraw/gmenu.c
+++ b/gdraw/gmenu.c
@@ -208,9 +208,9 @@ static void _shorttext(int shortcut, int short_mask, unichar_t *buf) {
     unichar_t *pt = buf;
     static int initted = false;
     struct { int mask; char *modifier; } mods[8] = {
-	{ ksm_shift, H_("Shft+") },
+	{ ksm_shift, H_("Shift+") },
 	{ ksm_capslock, H_("CapsLk+") },
-	{ ksm_control, H_("Ctl+") },
+	{ ksm_control, H_("Ctrl+") },
 	{ ksm_meta, H_("Alt+") },
 	{ 0x10, H_("Flag0x10+") },
 	{ 0x20, H_("Flag0x20+") },
--- a/share/default
+++ b/share/default
@@ -32,7 +32,7 @@ CharView.Menu.File.Import...:Shift+Ctrl+I
 CharView.Menu.File.Revert File:Shift+Ctrl+R
 CharView.Menu.File.Revert Glyph:Alt+Ctrl+R
 CharView.Menu.File.Print...:Ctrl+P
-CharView.Menu.File.Execute Script...:Alt+Ctrl+r
+CharView.Menu.File.Execute Script...:Alt+Ctrl+R
 CharView.Menu.File.Quit:Ctrl+Q
 
 CharView.Menu.Edit.Undo:Ctrl+Z
@@ -76,20 +76,20 @@ CharView.Menu.Edit.Unlink Reference:Ctrl+U
 CharView.Menu.Edit.Remove Undoes:No Shortcut
 
 
-CharView.Menu.Point.Tools.Ruler:r
-CharView.Menu.Point.Tools.Pointer:v
-CharView.Menu.Point.Tools.Magnify:a
-CharView.Menu.Point.Tools.Freehand:f
-CharView.Menu.Point.Tools.Scroll:h
-CharView.Menu.Point.Tools.Pen:p
-CharView.Menu.Point.Tools.Knife:k
+CharView.Menu.Point.Tools.Ruler:R
+CharView.Menu.Point.Tools.Pointer:V
+CharView.Menu.Point.Tools.Magnify:A
+CharView.Menu.Point.Tools.Freehand:F
+CharView.Menu.Point.Tools.Scroll:H
+CharView.Menu.Point.Tools.Pen:P
+CharView.Menu.Point.Tools.Knife:K
 
 CharView.Menu.Point.Tools.Curve:2
 CharView.Menu.Point.Tools.Corner:3
 CharView.Menu.Point.Tools.Tangent:4
 CharView.Menu.Point.Tools.HVCurve:5
 
-CharView.Menu.Point.Tools.Toggle Spiro:s
+CharView.Menu.Point.Tools.Toggle Spiro:S
 +CharView.Menu.Point.Tools.G2:1
 +CharView.Menu.Point.Tools.G4:2
 +CharView.Menu.Point.Tools.Corner:3
@@ -213,8 +213,8 @@ CharView.Menu.View.Combinations.Anchor Glyph at Point.Detach:No Shortcut
 CharView.Menu.View.Combinations.Kern Pairs:No Shortcut
 CharView.Menu.View.Combinations.Anchored Pairs:No Shortcut
 CharView.Menu.View.Combinations.Ligatures:No Shortcut
-CharView.Menu.View.Number Points.None:Alt+n
-CharView.Menu.View.Number Points.TrueType:Alt+t
+CharView.Menu.View.Number Points.None:Alt+N
+CharView.Menu.View.Number Points.TrueType:Alt+T
 CharView.Menu.View.Number Points.PostScript:No Shortcut
 CharView.Menu.View.Number Points.SVG:No Shortcut
 CharView.Menu.View.Grid Fit.Show Grid Fit...:Shift+Ctrl+#
@@ -249,10 +249,10 @@ CharView.Menu.View.Show.Vertical Metric Lines:No Shortcut
 CharView.Menu.View.Show.Snap Outlines to Pixel Grid:No Shortcut
 CharView.Menu.View.Zoom out:Alt+Ctrl+-
 CharView.Menu.View.Zoom out:-
-CharView.Menu.View.Zoom out:x
+CharView.Menu.View.Zoom out:X
 CharView.Menu.View.Zoom in:Alt+Ctrl+=
 CharView.Menu.View.Zoom in:=
-CharView.Menu.View.Zoom in:z
+CharView.Menu.View.Zoom in:Z
 CharView.Menu.View.Next Line in Word List:Ctrl+.
 CharView.Menu.View.Previous Line in Word List:Ctrl+,
 
--- a/share/default
+++ b/share/default
@@ -7,29 +7,29 @@ FontView.Popup.Menu.Copy Reference:Ctrl+G
 FontView.Popup.Menu.Copy Width:No Shortcut
 FontView.Popup.Menu.Paste:Ctrl+V
 FontView.Popup.Menu.Clear:No Shortcut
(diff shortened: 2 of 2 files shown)


frank-trampe:
I don't think that it's good to put control after shift. I prefer shift adjacent to the last item since it's the most immediate modifier (shift + p is P and thus easier to remember but control + p is most concisely control + p) and since it tends to make for a more linear keystroke (since one tends to use the control key on the far bottom-left corner of the keyboard). It also feels more natural to have shift as the second-to-last modifier since that's what a lot of applications that I use on a daily basis (like Chrome, Firefox, LibreOffice, OpenOffice) do. Inkscape and the GIMP put shift first, but I don't use those keystrokes as often.

I don't feel that one approach is vastly better, but I slightly prefer the current arrangement, and I don't think that we ought to change this without a strong reason.

frank-trampe:
@guiguru, @davelab6, what do you think?

andreaslarsen:
@frank-trampe it makes the shortcuts a lot more recognisable if they all have the same order and since ctrl/cmd is the most common modifier it makes sense to put it second last which every shortcut heavy app I've checked on my computer does (including entire adobe suite, affinity suite).

I'm not sure I understand your argument about adjacent?

And "feels natural" isn't an argument - it's a statement :)

andreaslarsen:
What I haven't fixed yet is those shortcuts using lowercase letters - e.g. execute script is "Alt+Shift+r" which should be "Alt+Shift+R"

davelab6:
I think we should follow Andreas lead on this

davelab6:
Actually I think always lower case letters and a shift modifier is good, I
think Frank's suggestion is too subtle.

We have some symbol bindings that don't show a shift because on some
keyboards shift is needed to press them and others it is not.

So I think redesigning the key bindings to be more simple and easier to
work across lots of keymaps would be good

andreaslarsen:
hmm...just removed the last odd lowercase letters. 

I haven't seen any application use lowercase and if I look down on my keyboard there are no lowercase letters ;)

frank-trampe:
@andreaslarsen, what I meant about directionality is that it's easier to think about and to execute a keystroke if it moves in a single direction rather than back-and-forth. Control + Alt + Delete is a lot easier to parse than Alt + Control + Delete because one naturally starts his eye and hand at the far left corner of the keyboard and then composes to the right.

@davelab6, I'm not proposing replacing Shift + p (or Shift + P) labels with P labels. What I'm saying is that it makes it easier to think about and to remember certain keystrokes. Perhaps the best example is zooming in with Control + Shift + =, which is easy to remember because it's Control + +.

andreaslarsen:
@frank-trampe if you "just" fix the code so we can tick the last box in #2252 the order will be a non-issue for you and potential users that have rocked fontforge for a long time - you can just batch replace "shift+ctrl" with "ctrl+shift" and be done with it.

New users will benefit from having the shortcuts look more ordered and how they do in other graphics apps.

davelab6:
My preference is to cater to new users, because fontforge base can grow a
lot from where it is today.

I agree that providing workarounds to maintain old configurations is
worthwhile

davelab6:
On 5 May 2015 10:54 pm, "Frank Trampe" <notifications@github.com> wrote:
>
> @andreaslarsen, what I meant about directionality is that it's easier to
think about and to execute a keystroke if it moves in a single direction
rather than back-and-forth. Control + Alt + Delete is a lot easier to parse
than Alt + Control + Delete because one naturally starts his eye and hand
at the far left corner of the keyboard and then composes to the right.

Then read the direction of the icons the other way, char to chord to chord
;)

> @davelab6, I'm not proposing replacing Shift + p (or Shift + P) labels
with P labels. What I'm saying is that it makes it easier to think about
and to remember certain keystrokes. Perhaps the best example is zooming in
with Control + Shift + =, which is easy to remember because it's Control + +

Hmmmmm, not sure, as, I have seen that this gets complicated for non
American keyboards. I think Ctrl + + is a legacy from full PC keyboards
with a num pad

I think our menu key bindings should be non symbols/punctuation as much as
possible, like z+x for zoom

guiguru:
quick question: is FF built on GTK?

because the bigger perspective is that the part that should take care of the configuration and handling of menus (on 4+ different desktop environments) is the GUI toolkit you use.

DIY is inviting a lot of grief

davelab6:
@guiguru No. :) FontForge began in ~2000 when neither GTK nor Qt supported Unicode - before pango - and so George Williams made his own X toolkit. Long story is here :) http://fontforge.github.io/ff-history.html 

This is why at my LGM presentation this year, I repeated a meme that @rrthomas came up with: that FontForge is like a gigantic ancient monumental architectural masterpiece, and conserving it so it doesn't collapse is sensible, but restoring it to a modern habitat isn't worth it. 

There are just a few more 'low hanging fruit' that I think will not be much trouble to implement, like reordering the menus. But big new features like real time collab, multi glyph editing, etc, have been very lonely endeavors.

guiguru:
> No. :) FontForge began in ~2000 when neither GTK nor Qt supported Unicode - before pango - and so George Williams made his own X toolkit.

OK, then the burden is in that DIY toolkit. and I guess that is what this pull request is for.

now about the discussion: **there are laws that govern this**. they are built into the native toolkits of windows, OS-X, GNOME and KDE. each one will be slightly different, but they _are_ the law. there is nothing to choose here for us (no, not even for me ;^).

* for windows, you figure out what the law is, then implement it for the windows version of FF;
* for OS-X, you figure out what the law is, then implement it for the OS-X version of FF;
* for linux, you can fudge a bit: pick either GNOME or KDE (whichever is ‘closer’ to FF) figure out what the law is, then implement it for the linux version of FF.

that’s it

andreaslarsen:
There are no laws in UI/UX. There are study based knowledge, common "standards", OS interface guidelines and personal preferences.

guiguru:
how shortcuts are displayed in menus for a given desktop platform is 99+% nailed down. it is not a discussion.

_(and there is no place for personal preferences in competent UI design)_

davelab6:
@guiguru thanks for your insight. However, FontForge was made by GWW as a bespoke X Windows application, and he did entirely his own thing, sort of Motif like:

fontforge - 2

So FontForge is definitely a vagabond outline in foreign lands, and it has no OS-specific code paths, relying on X Window servers for OSX and Windows to run at all there. Following Reuben's _conservation not restoration_ doctrine, I don't recommend adding such platform codepaths.

To the original discussion with @frank-trampe the Mac OS has the key chord icons in the standard order that @andreaslarsen has proposed:

screen shot 2015-05-06 at 10 52 25 am

I think we should accept @andreaslarsen pull request and move on.

andreaslarsen:
Yep, I agree with you about personal preferences https://github.com/fontforge/fontforge/pull/2253#issuecomment-99231760 but it pops up so often that you can't ignore it that's why I included it on the list. Yesterday I tried to make a case based on research and logic for removing the icons from menus to which one guy replied: "I like icons" :)

It is more or less nailed down (which is why I must insist we use uppercase for letters) but it seems some OS'es (not sure what @frank-trampe use) aren't there yet. 

On MacBook having cmd second last is almost law.

I obviously think we should merge too but my original question was never answered https://github.com/fontforge/fontforge/pull/2253#commitcomment-11045585 :smile:

rrthomas:
I agree with @davelab6: the overriding concern with FF as it is must be not to let it get any more complex, and platform-specific code is a great way to do that. (FWIW I have thought long and hard about how one would rebuild FF from the inside out, but that would require at a minimum a commitment to replace the toolkit /while keeping it working/.)

andreaslarsen:
@rrthomas it would be cool if one could "wrap" FontForge in web views somehow :)

guiguru:
> I think we should accept @andreaslarsen pull request and move on.

yeah, no longer worth the discussion.

davelab6:
Reuben, for desktop stack, i think it would be better to do what Adrien is
doing and use fontforge python only to access it's features from a PyQt ui

rrthomas:
Sure, that's taking a different route, of building a new thing with the old one's guts.

davelab6:
Andreas, the old pyro desktop presented to guadec Birmingham a few years
ago shows it's possible to render x11 using the browser add a compositor.

I really wish gnome had run with this idea for gnome 3.

Exploring this idea a while ago I found
https://github.com/liftoff/GateOne/issues/417 and spoke to the guy offline,
and he has security concerns preventing him from releasing it (doesn't want
to be irresponsible)

davelab6:
The ui sucks anyway so I think this is reasonable :)

andreaslarsen:
@frank-trampe (or someone else) can you confirm that my edits to [gdraw/gmenu.c](https://github.com/fontforge/fontforge/commit/f0b952dbde96d740f9669965d670479135dc1644) are correct, so we can merge this?

frank-trampe:
I probably can't test completely for a few days. Can you?

adrientetar:
> To the original discussion with @frank-trampe the Mac OS has the key chord icons in the standard order that @andreaslarsen has proposed:

That's just Mac OS, on Windows it's Ctrl always first. Ubuntu as well. https://help.ubuntu.com/community/KeyboardShortcuts

andreaslarsen:
@adrientetar I don't use windows so I just googled and the first I found was [this screenshot](http://www.geosetter.de/wp-content/uploads/2010/03/v3346_06_filedate_from_takendate_menu_en.jpg) with "shift+ctrl+" and I actually thought the most common thing in windows was to not show the shortcuts at all?

davelab6:
I think follow mac convensions is better as most designers use or are
familiar with MacOS

andreaslarsen:
@davelab6 true, but if windows (majority of FF users?) standard is the opposite then maybe we should reverse it in the default file and then just substitute "Ctrl+Shift" with ⇧⌘ in gmenu.c for MacIcons (assuming that's a minor task @frank-trampe @adrientetar ?)

adrientetar:
It's doable.

andreaslarsen:
by youable?

adrientetar:
Too busy these days.

davelab6:
@andreaslarsen this is the kind of platform special casing I don't think is worthwhile, but that's just my view. If someone wants to do it, they should! :) From our downloads it does seem like Windows is much much more popular for FontForge users than Mac. 

But you are right, there is already some special casing in the `MacIcons` code, so I think making the hotkeys file use Windows defaults and using the `MacIcons` code path to do the Mac style ordering would be ideal. @frank-trampe what do you think? :)

andreaslarsen:
Changed to windows default so now the `MacIcons` code just need to be updated to replace ⌘⇧ with ⇧⌘ :)

davelab6:
@frank-trampe does this resolve your concerns? :)

frank-trampe:
I still haven't had a chance to test. Hopefully later today, otherwise probably Tuesday.

davelab6:
No worries, thank you Frank!!

frank-trampe:
This tests correctly for me. In the interest of keeping the git history concise, I have squashed this to #2273.

andreaslarsen:
:+1:

Issue #2258, opened 2015-05-06
Using foreground color for the > sign in menus


larsenwork:
I thought I'd found a great way to add border to dropdown menus - but this fucks up the > sign which for some reason is drawn using BorderBrightest, brighter, darker, darkest...

This could be easily fixed by drawing it with foreground color instead

![screen shot 2015-05-06 at 11 29 04](https://cloud.githubusercontent.com/assets/3859425/7490975/0ff38bb6-f3e5-11e4-80fe-e207ffab4469.png)

skef:
This was ultimately addressed as mentioned at https://github.com/fontforge/fontforge/pull/2266#issuecomment-105292661

PR #2266, opened 2015-05-08
An attempt to fix issue #2258

andreaslarsen:
I replaced the `border_*` with `main_foreground` - please check if it's done correctly :)

#2258

### DIFF
--- a/gdraw/gmenu.c
+++ b/gdraw/gmenu.c
@@ -427,19 +427,19 @@ static void GMenuDrawArrow(struct gmenu *m, int ybase, int r2l) {
 	p[1].x = x+1*(as/2);		p[1].y = ybase;
 	p[2].x = p[1].x;		p[2].y = ybase-as;
 
-	GDrawDrawLine(m->w,p[0].x,p[0].y,p[2].x,p[2].y,m->box->border_brighter);
-	GDrawDrawLine(m->w,p[0].x+pt,p[0].y,p[2].x+pt,p[2].y+pt,m->box->border_brighter);
-	GDrawDrawLine(m->w,p[1].x,p[1].y,p[0].x,p[0].y,m->box->border_darkest);
-	GDrawDrawLine(m->w,p[1].x+pt,p[1].y-pt,p[0].x-pt,p[0].y,m->box->border_darkest);
+	GDrawDrawLine(m->w,p[0].x,p[0].y,p[2].x,p[2].y,m->box->main_foreground);
+	GDrawDrawLine(m->w,p[0].x+pt,p[0].y,p[2].x+pt,p[2].y+pt,m->box->main_foreground);
+	GDrawDrawLine(m->w,p[1].x,p[1].y,p[0].x,p[0].y,m->box->main_foreground);
+	GDrawDrawLine(m->w,p[1].x+pt,p[1].y-pt,p[0].x-pt,p[0].y,m->box->main_foreground);
     } else {
 	p[0].x = x;			p[0].y = ybase-as/2;
 	p[1].x = x-1*(as/2);		p[1].y = ybase;
 	p[2].x = p[1].x;		p[2].y = ybase-as;
 
-	GDrawDrawLine(m->w,p[0].x,p[0].y,p[2].x,p[2].y,m->box->border_brighter);
-	GDrawDrawLine(m->w,p[0].x-pt,p[0].y,p[2].x+pt,p[2].y+pt,m->box->border_brighter);
-	GDrawDrawLine(m->w,p[1].x,p[1].y,p[0].x,p[0].y,m->box->border_darkest);
-	GDrawDrawLine(m->w,p[1].x+pt,p[1].y-pt,p[0].x-pt,p[0].y,m->box->border_darkest);
+	GDrawDrawLine(m->w,p[0].x,p[0].y,p[2].x,p[2].y,m->box->main_foreground);
+	GDrawDrawLine(m->w,p[0].x-pt,p[0].y,p[2].x+pt,p[2].y+pt,m->box->main_foreground);
+	GDrawDrawLine(m->w,p[1].x,p[1].y,p[0].x,p[0].y,m->box->main_foreground);
+	GDrawDrawLine(m->w,p[1].x+pt,p[1].y-pt,p[0].x-pt,p[0].y,m->box->main_foreground);
     }
 }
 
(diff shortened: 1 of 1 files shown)


andreaslarsen:
Can @adrientetar, @frank-trampe or someone else confirm this does what I intended?

frank-trampe:
Hmm. I think that the original code was done this way for a decent reason, namely to support the unified three-dimensional appearance of the interface. (The arrows and the menus are presented in fake relief with shadows.) Changing this unconditionally would break that.

If it's a high priority, we could add theming support for a new flag that specifies 3-dimensional arrows or now.

andreaslarsen:
but why can't they just be the same color as the text? I've never seen a menu where the submenu arrow wasn't the same color as the text:)

frank-trampe:
@andreaslarsen, this isn't about paint colors. The arrow (as originally conceived) is a volumetric outset of the menu block rather than a painted label, which is why its coloring is defined in so many different pieces and with the shadowing colors rather than the paint colors.

The three-dimensional look isn't as stylish as it once was, but this was the look of almost all UNIX desktops at one time. Motif still does that, I think.

If we're going to start developing themes that depart from this model, I think that we need to plan carefully how to support that and to implement that support in one changeset.

andreaslarsen:
I don't think we need to be carefull doing something 99% of all OS'es are already doing - we should however be carefull not doing it and I can't see any argument why.

frank-trampe:
Good software is not built by changing part of something without looking at the whole. Nor is it built by breaking existing themes.

andreaslarsen:
Agreed and agreed. 

I think this fits better to the whole as it's what the user would expect it to look like.

And 3D boxy thems will still work (if anyone still use those) they will just have arrows that are styled as the text as they are everywhere else so won't confuse any.

andreaslarsen:
Talked with @davelab6 about this yesterday and the plan is to make [yosemiforge](https://github.com/fontforge/YosemiForge) the new default theme so this definitely needs to be merged.

As written above the point about breaking existing themes is a non issue not worth spending code added theme specific support for.

jtanx:
Unless we don't make it the default on Windows and things break...

BTW: This is how your theme currently looks like on Windows:
winff
winff2

davelab6:
Pathum has an intern who will do the needed theme code changes as Frank
Trampe wants to focus on stability fixes

Jeremy could you list the changes needed in the theme in priority? :)

jtanx:
(Just updated images from previous post to fix pixmap location)
 
1. Fix pixmap include directory - I had:
> Gdraw.GGadget.ImagePath: ~/.FontForge/tango-ct.theme:~/.FontForge/tango-ct.theme/tango:/Applications/FontForge.app/Contents/Resources/opt/local/share/fontforge/pixmaps/:=
   
    I presume that trailing `:=` is important on Windows so that it searches `../share/fontforge/..` for the pixmaps folder
2. No way are menus that gray on Windows (looks like something from Windows 98)
3. Buttons are generally not coloured on Windows (e.g. standard is gray) but I guess this is okay
4. Looking at those screenshots above, there is something wrong with the rounding of corners on buttons
5. Clear Sans, DejaVu sans, monaco etc. don't exist by default on Windows. I can probably fix this with the font loading thing I did a while back, but I'd prefer it if it actually used standard UI fonts (e.g. Segoe UI)

davelab6:
We can do font fall back?

jtanx:
Pango is falling back to 'Sans Not-rotated' which is likely whatever is the standard sans font on the user's desktop. But as you can see in the screenshots, this results in things like text being slightly clipped (e.g. see the menu items - the bottom of the text is clipped) e.g:

ffclip

adrientetar:
Please make something correct on Windows before merging this.

andreaslarsen:
@jtanx as dave said the theme isn't optimised for windows yet... :wink: Is that ugly green color standard on windows nowadays?

The only reason the menu were "OMG! win 98 grey" were for usability reasons (to have enough contrast between against white) - the menus will be white as soon as it's possible to add border around - which is what this pull tries to solve. That is similar to how inkscape and other cross platform apps are doing it.

Buttons are coloured everywhere else - it's just metro that insist one shouldn't be able to tell the difference between buttons and text.

Buttons won't have round corners (the rounding in FF is broken) - online version might be outdated.

My thought was to add Clear Sans to FontForge similar to how the current default menu font is it - I posted an issue about this some time ago.

The online version of yosemiforge is not what I think will be default - it will be more similar to the contrast version.

See my screenshot here to get a better idea. https://www.behance.net/gallery/26418243/FontForge-design-update.

@davelab6 is this intern one I can communicate with somehow?

andreaslarsen:
Can we merge this now so we can test the theme change properly before launching it?

andreaslarsen:
I agree that Segue should be the default for the UI on windows - x11 on mac has a very bad font rendering which is why I tweaked clear sans to look better there. I removed Monaco from the font stack - it was called after Dejavu Sans anyway so didn't really make a difference.

@jtanx what should `400 10pt clear sans` be on win? `400 10pt segue ui`

The "Fontview" font is just Clear Sans with adjusted metrics to center it in the fontview table - not sure why you get the load error though - did you also get it after fixing the pixmaps directory?

pathumego:
CC @lpsandaruwan

frank-trampe:
I've merged this change in modified (flexible) form. In order to render the arrows in paint instead of relief, add the line `Gdraw.GMenu.3DLook: False` to the resources file for the theme.

andreaslarsen:
@frank-trampe sounds great, thank you
@jtanx and @ipsandaruwan - I'll push an update to yosemiforge called fontforge2015 in a minute (the menu is still unfinished) for you to test and create issues over there (it's a bit difficult to get a overview of what needs fixin' in a long thread like this.

jtanx:
@andreaslarsen 
> Is that ugly green color standard on windows nowadays?

No, you can choose any colour you want - mine is set to cycle through a set of predetermined colours though.

larsenwork:
So this is how it currently looks on windows:

@jtanx I'm still having trouble getting the font right. Clear Sans is in the pixmaps folder but FontForge doesn't seem to "pick it up" on windows - I've tried "Clear Sans", "ClearSans" and "Segue UI" all without luck. I've never owned a windows machine so I'm very green here - am I missing something obvious? The updated theme is in the "New Theme" folder in Yosemiforge

win102
win101
win10

jtanx:
Do you have Clear Sans installed as a font? Note also that it's `Segoe UI` not `Segue UI`.

larsenwork:
It's in the pixmaps folder - does it need to be installed by the system too? I'm pretty sure I spelled it right when I tested it. 
I only have windows in my virtual machine so it's probably easier for you to test :)

jtanx:
Well the font's not going to magically load by being in the pixmaps folder. 

Try placing it in the `share/fonts` folder. There should already be some Cantarell fonts there.

larsenwork:
I don't think I suggested magic was involved :confused:  On mac it works just by adding it to the pixmaps folder. I'll the share fonts

larsenwork:
Putting it in share font did the trick - does the win building script automatically pull the fonts from pixmaps or how the the cantorial fonts end up there?

jtanx:
[By this](https://github.com/jtanx/fontforgebuilds/blob/master/ffbuild.sh#L540)

larsenwork:
@jtanx ok, would have been nice if you had shared that information already here (bullet number 5)... https://github.com/fontforge/fontforge/pull/2266#issuecomment-105228062  :wink:

jtanx:
Well, it was never used since it was made, as it ended up not being needed, with the switch to Cairo.

I did mention it in that comment (what I refer to as the 'font loading' feature).

jtanx:
Plus I strongly dislike bundling fonts just to get the UI to work, which was part of the benefit of enabling Cairo support.

larsenwork:
You made it sound like there were some extra thing that needed fixing when apparently the `font loading` feature does it automatically and what I've done is just substituting the previous UI font with Clear Sans. And I can't see how you expect me to deduce from that comment that everything would "magically" work if I just put it in `share/fonts`...

I agree with you about bundling UI fonts is bad but can the current code handle it so FontForge automatically uses system UI font? 
And even if it could I'm pretty sure it would look horrible on mac (small size helvetica + horrible x11 rendering = bad). I don't know about linux/windows.

But if we can make it so it uses UI font on all operating systems and it looks good then I'm all for it. Until then clear sans (or other sans made for UI) will have to do.

jtanx:
> You made it sound like there were some extra thing that needed fixing

Yes, because it is yet another thing that I have to change to get it to work on Windows which is what I'm trying to avoid.

> I agree with you about bundling UI fonts is bad but can the current code handle it so FontForge automatically uses system UI font? 

Usually, you specify 'MS Shell Dlg' and the default font (be it Segoe UI or Meiryo UI or MS Sans Serif) will be used, but I don't know if that works for Pango. As for FF itself, you should be able to specify comma separated lists of fonts to use.

But at the end of the day, I'll just bundle the old resources until this is resolved.

larsenwork:
I wouldn't have a clue as what's easiest for your workflow - just note that the default theme will be updated very soon and that Clear Sans needs to be bundled for that to work.

PR #2268, opened 2015-05-10
fontforge.rb: resync with homebrew

DomT4:
Resync with latest Homebrew/homebrew formulae. Changes:

* Spacing between the `defs` makes it easier to read. Not a big change though.
* Syncs the latest release, bottles, checksum, etc.
* Moves the `head` block into a simple `head` line - You don't need the whole block, if you declared the required dependencies in the normal formula, which is done for both `zeromq` and `czmq`, they'll apply to the whole formula and don't need redeclaring.
* Adds links tracking the Fontconfig issue.
* Fixes the Python issues where when a user has Homebrew's Python and system Python installed and it would hook into the Homebrew Python even though the dependency wasn't declared (`depends_on :python if MacOS.version <= :snow_leopard` = Unless you're using Snow Leopard or below, use system Python) and would cause silent linking problems that way. This was effectively breaking Python scripting a lot of the time. This shouldn't have impacted you guys as presumably you build with a relatively clean Travis system, but it's a useful fix.
* Removes the now redundant `post-install` part because you guys fixed the problem that required that workaround.

### DIFF
--- a/travis-scripts/before_install_osx.sh
+++ b/travis-scripts/before_install_osx.sh
@@ -37,7 +37,8 @@ brew update
 brew config
 
 sed -i -e "s|{TRAVIS_PULL_REQUEST}|${TRAVIS_PULL_REQUEST}|g" ./travis-scripts/fontforge.rb
-sudo cp ./travis-scripts/fontforge.rb /usr/local/Library/Formula/fontforge.rb
+rm /usr/local/Library/Formula/fontforge.rb
+cp ./travis-scripts/fontforge.rb /usr/local/Library/Formula/fontforge.rb
 echo "*****"
 echo "*****"
 echo "***** using homebrew formula fontforge.rb:"
@@ -54,21 +55,19 @@ sudo installer -pkg /Volumes/XQuartz-*/XQuartz.pkg -target /
 popd 
 
 echo "doing an OSX before install step."
-# python may require a little force to install. In October 2014
-# there were issues with easy_install and pip not letting the install
-# step complete.
 set +ev
 brew install python
-brew link --overwrite python
 set -ev
 brew install cairo libspiro fontconfig
 
 #
 # this forces version 4.0.4 and 2.2.0 respectively.
 pushd .
-cd $( brew --prefix )
-git checkout ab7f37834a28b4d6f3c584f08e5b993d8c191653 Library/Formula/zeromq.rb
-git checkout 3ad14e1e3f7d0131b3eee7fb4ce38a65e22a5187 Library/Formula/czmq.rb
+cd $(brew --prefix)
+rm Library/Formula/zeromq.rb
+rm Library/Formula/czmq.rb
+wget https://raw.githubusercontent.com/Homebrew/homebrew/ab7f37834a28b4d6/Library/Formula/zeromq.rb -O Library/Formula/zeromq.rb
+wget https://raw.githubusercontent.com/Homebrew/homebrew/3ad14e1e3f7d0131b/Library/Formula/czmq.rb -O Library/Formula/czmq.rb
 brew install czmq zeromq
 popd
 
--- a/travis-scripts/fontforge.rb
+++ b/travis-scripts/fontforge.rb
@@ -1,12 +1,13 @@
 # based on https://raw.githubusercontent.com/Homebrew/homebrew/master/Library/Formula/fontforge.rb
 #
-# last synced by hand by davelab6 at 2015-03-05
+# last synced by hand by DomT4 at 2015-05-10
 
 class MyDownloadStrategy < GitDownloadStrategy
   # get the PR
   def fetch
     system "rsync -a /Users/travis/build/fontforge/fontforge/. /Library/Caches/Homebrew/fontforge--git"
   end
+
   def reset_args
     ref = case @ref_type
           when :branch then "origin/#@ref"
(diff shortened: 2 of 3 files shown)


travis-scripts/before_install_osx.sh
```diff
@@ -54,21 +55,19 @@ sudo installer -pkg /Volumes/XQuartz-*/XQuartz.pkg -target /
 popd 
 
 echo "doing an OSX before install step."
-# python may require a little force to install. In October 2014
-# there were issues with easy_install and pip not letting the install
```
DomT4: These were fixed a while ago. Doesn't need to be kicked anymore. You may wish to double check, but should be :+1:.

travis-scripts/before_install_osx.sh
```diff
@@ -54,21 +55,19 @@ sudo installer -pkg /Volumes/XQuartz-*/XQuartz.pkg -target /
...
 #
 # this forces version 4.0.4 and 2.2.0 respectively.
 pushd .
-cd $( brew --prefix )
-git checkout ab7f37834a28b4d6f3c584f08e5b993d8c191653 Library/Formula/zeromq.rb
-git checkout 3ad14e1e3f7d0131b3eee7fb4ce38a65e22a5187 Library/Formula/czmq.rb
+cd $(brew --prefix)
+rm Library/Formula/zeromq.rb
+rm Library/Formula/czmq.rb
+wget https://raw.githubusercontent.com/Homebrew/homebrew/ab7f37834a28b4d6/Library/Formula/zeromq.rb -O Library/Formula/zeromq.rb
```
DomT4: Doing it this way instead of exercising a `git` checkout will cause you less pain in the long-term. `git checkout`'s inside the Homebrew directory are *super* prone to breaking things for Homebrew.

travis-scripts/script_osx.sh
```diff
@@ -9,7 +9,7 @@ rm -f /Users/travis/build/fontforge/fontforge/.git/shallow
 set +e
 mkdir -p /tmp/fontforge-source-tree
 echo "brew build starting..."
-brew install --verbose fontforge --HEAD --with-x 
+brew install --verbose fontforge --HEAD --with-x --with-libspiro
```
DomT4: You included `libspiro` in the `brew install` arguments in the `before_install_osx.sh` script but you weren't building FontForge with `libspiro` enabled? Didn't seem deliberate, but will revert if it was.

travis-scripts/before_install_osx.sh
```diff
@@ -37,7 +37,8 @@ brew update
 brew config
 
 sed -i -e "s|{TRAVIS_PULL_REQUEST}|${TRAVIS_PULL_REQUEST}|g" ./travis-scripts/fontforge.rb
-sudo cp ./travis-scripts/fontforge.rb /usr/local/Library/Formula/fontforge.rb
```
DomT4: Using `sudo` here could potentially cause file permission problems for `brew`. If you remove the existing file first, you shouldn't need the `sudo`.

travis-scripts/before_install_osx.sh
DomT4: Side note: You could stop using Homebrew's Python entirely and just use the OS X default one, but the rest of your setup seems to very deliberately use Homebrew's so I left it.

jtanx:
cc @monkeyiq

travis-scripts/script_osx.sh
DomT4: Also changed `--with-x` to `--with-x11` as that's the supported syntax.

travis-scripts/fontforge.rb
```diff
@@ -87,43 +94,32 @@ def install
...
-    # to @executable_path/../Frameworks/Breakpad.framework/Versions/A/Breakpad which
-    # obviously doesn't exist given fontforge and FontForge.app are in different places.
-    # If this isn't fixed within a couple releases, consider dumping everything in libexec.
-    # https://github.com/fontforge/fontforge/issues/2022
-    if build.with? "x11"
-      ln_s "#{share}/fontforge/osx/FontForge.app", prefix
-      system "install_name_tool", "-change", "@executable_path/../Frameworks/Breakpad.framework/Versions/A/Breakpad",
-             "#{bin}/fontforge", "#{share}/fontforge/osx/FontForge.app/Contents/Frameworks/Breakpad.framework/Versions/A/Breakpad"
-    end
-
```
frank-trampe: Why are we dropping this?

travis-scripts/fontforge.rb
```diff
@@ -53,7 +49,6 @@ class Fontforge < Formula
   depends_on "pango"
   depends_on "zeromq"
   depends_on "czmq"
-  depends_on "fontconfig"
```
frank-trampe: Why no fontconfig?

DomT4: It's still there. Moved under the `depends_on :python` line below. It's currently surrounded in bug links, so we moved it for clarity.

travis-scripts/fontforge.rb
DomT4: Someone upstream here fixed the underlying issue in the release before the most recent.

frank-trampe:
This looks fine to me, and I assume that it tests correctly for @DomT4.

DomT4:
Yup, no problems for me. Travis seemed happy as well.

PR #2273, opened 2015-05-13
Use longer hotkey names.

frank-trampe:
This replaces #2253.


frank-trampe:
Note that this may break custom hotkey files, but it is uncommon as far as I know for people to make their own hotkey files anyway.

PR #2274, opened 2015-05-13
Fix read/write of bits USE_TYPO_METRICS and WWS for OS/2 version < 4

fred-wang:
cc'ing @khaledhosny 

Download Asana Math and open it with fontforge:

http://www.ctan.org/tex-archive/fonts/Asana-Math/

ttx on Asana-Math.ttf generates fsSelection value="00000000 01000000" for the OS/2 table, USE_TYPO_METRICS is *not* set.

Open Asana-Math.ttf with fontforge. In the OS/2 menu, you should see that the option "use typo metrics" *is* checked. Generate a new ttf file (you can uncheck and check the box again if you want). A verification with ttx still indicates that USE_TYPO_METRICS is *not* set.

Analysis of the problem: parsettf.c sets use_typo_metrics to true by default but only overrides with the font value when OS/2 version >= 4. Also tottf.c set the bit USE_TYPO_METRICS in OS/2 table to true only if has version >= 4.

This patch makes use_typo_metrics default to false when loading a ttf file, so that it is not set by default when opening a font with OS/2 version < 4. It also tries to force upgrade to OS/2 when use_typo_metrics has been enabled by the user, so that this bit can be set during file generation. Similarly for info.weight_width_slope_only.

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -3333,7 +3333,7 @@ static void setos2(struct os2 *os2,struct alltabs *at, SplineFont *sf,
 	os2->version = 3;
     if ( sf->use_typo_metrics || sf->weight_width_slope_only )
 	os2->version = 4;
-    if ( sf->os2_version!=0 )
+    if ( sf->os2_version > os2->version )
 	os2->version = sf->os2_version;
     if (( format>=ff_ttf && format<=ff_otfdfont) && (at->gi.flags&ttf_flag_symbol))
 	modformat = ff_ttfsym;
--- a/fontforge/parsettf.c
+++ b/fontforge/parsettf.c
@@ -6272,7 +6272,8 @@ SplineFont *_SFReadTTF(FILE *ttf, int flags,enum openflags openflags, char *file
     memset(&info,'\0',sizeof(struct ttfinfo));
     info.onlystrikes = (flags&ttf_onlystrikes)?1:0;
     info.onlyonestrike = (flags&ttf_onlyonestrike)?1:0;
-    info.use_typo_metrics = true;
+    info.use_typo_metrics = false;
+    info.weight_width_slope_only = false;
     info.openflags = openflags;
     info.fd = fd;
     ret = readttf(ttf,&info,filename);


adrientetar:
Thanks! Are you making a math font?

fred-wang:
No, trying to make existing fonts work with Gecko MathML...

fontforge/tottf.c
```diff
@@ -3333,7 +3333,7 @@ static void setos2(struct os2 *os2,struct alltabs *at, SplineFont *sf,
 	os2->version = 3;
     if ( sf->use_typo_metrics || sf->weight_width_slope_only )
 	os2->version = 4;
-    if ( sf->os2_version!=0 )
+    if ( sf->os2_version > os2->version )
```
khaledhosny: I’m not sure why this change was needed, as the line above it will correctly set the version to 4 if use_typo_metrics is set.

khaledhosny:
> ttx on Asana-Math.ttf generates `fsSelection value="00000000 01000000"` for the OS/2 table, USE_TYPO_METRICS is not set.

But the 7th bit is set here.

fred-wang:
> But the 7th bit is set here.
https://www.microsoft.com/typography/otspec/os2.htm#fss
It the "7-th" bit counting from index 0 so actually the 8th bit.

fontforge/tottf.c
fred-wang: because the line just below overrides that :-)

khaledhosny: Right, that is by design the 0 OS/2 version corresponds to “Automatic” in the GUI, where FontForge will chose which fits best the fields being set, but if the user selects a version explicitly it is always used.

khaledhosny: I think the correct fix here is for the font info dialog to warn the user if he is setting a field not supported by the selected version, not to override it in this low level code. Otherwise it makes no sense to even allow setting OS/2 version explicitly (not necessarily a bad thing) and the support for it can just be dropped.

fred-wang: Yes, I guess you are right. However, I don't know for the Python interface if we can do such warning.

fred-wang:
@khaledhosny https://www.microsoft.com/typography/otspec/os2.htm#ver actually says that 0 corresponds to TrueType rev 1.5, so I suspect it is actually not quite correct for fontforge to use it for "automatic"...

khaledhosny:
In SMT I simply [dropped the ability for user to select OS/2 version](https://bitbucket.org/sortsmill/sortsmill-tools/commits/e9b229310d96f54b134bf6c7d6cac85f8078f711) and left only the automatic calculation at build time, I suggest that FontForge do the same.

be5invis:
@khaledhosny So if you drop the ability to select version of OS/2 table, maybe you shouold add an option to make it possible to specify `xAvgCharWidth` if you want to produce monospaced fonts used for CMD. I mean that if your font contains CJK glyphs, `xAvgCharWidth` automatically calculated may make the window too wide since CSRSS (and conemu) calculates window width with this value.

PR #2279, opened 2015-05-19
fontview.c: Fix potential double-free

jtanx:
Preceding code may set `OpenDir = DefaultDir`. This was causing FontForge to crash on 64-bit Windows builds (I finally got these to work again after figuring out what was wrong with gnulib).

Surprisingly this never broke anything on 32-bit Windows builds.


frank-trampe:
This looks good to me.

Issue #1729, opened 2014-09-12
System keyboard layout is ignored on Windows


adrientetar:
FontForge always uses QWERTY apparently.

cc @jtanx

jtanx:
Okay, so I did some testing a while back, and it turns out that it's not an issue from VcxSrv. I was able to get VcxSrv to correctly work with a different keyboard layout (can't remember the settings right now) using another program like gedit, but using the same keyboard layout with FontForge does not work.

I narrowed it down to the call to `XLookpString` in [gxdraw.c](https://github.com/fontforge/fontforge/blob/0d62a06adf6c506ff4b4126d642064dea12f7aa7/gdraw/gxdraw.c#L3087) returning a zero-length response, hence the keyboard not working. I believe it is an encoding issue, but I don't really know that much about Xlib to know what's going on here. I haven't gotten much further than that at this point.

jtanx:
I've been able to work on this some more. So there were a few things that needs fixing:

1. I was compiling libX11 without xlocaledir from environment variable support (`--enable-xlocaledir`)
2. The locales bundled with VcXsrv (locale.alias) were for some reason missing the locale names (e.g. `English_Australia.1252`)
3. **Not all controls use an input context** (e.g. on creation of a text box, should specify `gg_text_xim` in the flags)

That last point is what prevents it from working fully. E.g. in the file open dialogue, you can't type anything with a non-standard keyboard, but in the font view, you can use it just fine because there's an input context.

 A hack would be to check if `((GXWindow) gw)->gic` is NULL ([here](https://github.com/fontforge/fontforge/blob/0d62a06adf6c506ff4b4126d642064dea12f7aa7/gdraw/gxdraw.c#L3087)) and if so, to create an input context then (so that it uses `Xutf8LookupString` instead of `XLookupString`). This works, but then I don't know why an input context isn't created all the time.  E.g. is there any case where an input context is not needed or wanted?

jtanx:
Actually, it looks like at present only the open font dialogue is affected. And in any case you can't open files containing non-ascii characters on Windows right now anyway. So it might actually be okay as it is. Hmm...

jtanx:
[...]

PR #2288, opened 2015-05-25
gxdraw.c: Use Xutf8SetWMProperties where available and remove mingw s…

jtanx:
…pecific window/icon setters/getters.

**Note**: The switch to Xutf8SetWMProperties where applicable is for **all platforms** - so much appreciated testing would be to check that all window titles still display fine.

Regarding the removal of the 'Mingw+WindowsWM hack':
Currently, window titles/icons work as expected without the MINGW specific code (or they will with the next release).

Looking at the original XMing `locale.alias`, it looks like this hack was used because it didn't define Windows specific locales (e.g `English_Australia.1252`) hence why it wasn't working. Using VcXsrv with a `locale.alias` from the standard libX11 distribution, everything works fine. (See also #1729).

So there's no reason to keep this MINGW specific code in place.

cc @frank-trampe

### DIFF
--- a/gdraw/gxdraw.c
+++ b/gdraw/gxdraw.c
@@ -750,162 +750,6 @@ return( XCreateWindow(gdisp->display, gdisp->root,
 	    gdisp->depth, InputOutput, gdisp->visual, wmask, &attrs));
 }
 
-
-#if defined(__MINGW32__)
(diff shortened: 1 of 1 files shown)


frank-trampe:
It tests fine for me on Linux.

PR #2289, opened 2015-05-25
Use git.mk for breakpad

jtanx:
This should prevent the `.gitignore` files in breakpad from showing as a modified item in git.

cc @monkeyiq @frank-trampe


frank-trampe:
Great!

PR #2291, opened 2015-05-25
Items in File->Call Script work only first time.

michal-n:
My scripts from Call Script menu only work first time I use any. Subsequent attempts end up with message:
```
  File "script.py", line 2, in <module>
    import fontforge
SystemError: _PyImport_FixupExtension: module fontforge not loaded
```
I need to restart FontForge to avoid it.

Looks like it stores references from `Py_InitModule3()` for reuse. But they are "borrowed references", and in my setup (Ubuntu 14.04, Python 2.7.6) python forgets them after `Py_Finalize()`.

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -18158,7 +18158,8 @@ static PyObject* CreatePyModule( module_definition *mdef ) {
     PyObject *module;
 
     if ( mdef->runtime.module != NULL )
-	return mdef->runtime.module;
+	if ( mdef->runtime.module == PyImport_AddModule(mdef->module_name) )
+            return mdef->runtime.module;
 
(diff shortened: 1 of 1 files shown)


frank-trampe:
@michal-n, do you mean that a script listed in the Script Menu (just beneath the Execute Script entry) runs once and then ceases to work?

michal-n:
Yes.

frank-trampe:
This seems to be not the right way to fix the problem. PyImport_AddModule returns a new reference, which gets dropped.

I have prepared another possible solution in #2293. Try it and tell me what you think.

michal-n:
> PyImport_AddModule returns a new reference

Whoops...

> I have prepared another possible solution in #2293

Works well and looks much better.

PR #2292, opened 2015-05-25
Make menu arrow relief optional.

frank-trampe:
Add the ability to paint rather than to chisel menu arrows using the Gdraw.GMenu.3DLook value in the theme resources file.

This supports #2266.


frank-trampe:
This tests correctly for me.

I would prefer to develop a single theme value that affects all paint/relief choices in the application, but I don't have time for that right now, and @andreaslarsen has asked that we merge this change now so as to facilitate his work.

I would hope that, in the future, we could support a top-level value Gdraw.3DLook and drop this option.

PR #2293, opened 2015-05-25
Clear dangling module references when finalizing Python.

frank-trampe:
The embedded Python was holding references to modules across Python sessions, which was bad.

This fixes that and supersedes #2291.


frank-trampe:
@adrientetar, take a look.

@michal-n, test.

michal-n:
+1

Issue #2290, opened 2015-05-25
Windows: Cannot open/save files with paths/folders containing non-standard characters


jtanx:
On Windows, functions like fopen/opendir/stat all only accept filenames in the current active codepage, which excludes opening files with characters outside of that codepage (e.g. Chinese and Cyrillic characters for a Windows/English setup).

Some aspects (like directory scanning) can be easily solved by using GLib and passing UTF-8 encoded filenames on Windows. Other aspects are less easily solved, as most function calls involving files are spread out, and most assume `u2def` (Unicode to default character encoding) is sufficient when it is not (on Windows - 99.9% of the time the default charset is 8859-1).

---

To be fixed:
- [x] fopen/open (in Gnulib)
- [x] <strike>freopen</strike> (not used)
- [x] opendir/readdir
- [x] rmdir (Unused; removed GFileRmDir)
- [x] unlink
- [x] access
- [x] mkdir
- [ ] getenv
- [x] `u2def` -->`u2fsys` / `def2u` --> `fsys2u` as necessary
- [x] stat/lstat
- [x] rename
- [x] remove (present in gtkui; left unchanged; unused elsewhere)
- [x] chmod (unused)
- [x] chdir 
- [x] creat? (unused)
- [ ] fsys functions
  - [ ] GFileGetHomeDocumentsDir - trailing slash
- [ ] Font loading code
- [ ]  endswithi is broken
- [ ] Check all instances of tolower

Anything else from https://developer.gnome.org/glib/stable/glib-File-Utilities.html

be5invis:
Same problem on Chinses Windows. You should use "wide" version of WINAPIs (which accepts UTF-16 encoded strings) and perform a "widen" transformation if the internal encoding of strings is UTF-8.

jtanx:
Yes, that is what GLib does - on Windows, it expects paths to be UTF-8 encoded, and internally it will do the conversion to UTF-16/call the Unicode/wide character version of the functions.

jtanx:
Hmm... I was contemplating switching to using `g_fopen` from GLib, but in testing, we can't do that because we dynamically link to GLib, so it has its own CRT, meaning the FILE* won't work in our code. 

A solution that would work is using gnulib for fopen/open, then patching gnulib's implementation to accept UTF-8 strings on input...

PR #2299, opened 2015-05-29
Multiple fixes

jtanx:
Some work towards fixing #2290. Also fixes some other bugs along the way.

General direction: Use GLib for all actions that require interaction with the filesystem (still a long long way to go here) where possible.
Rationale: GLib properly handles unicode filenames (on Windows you give it a UTF-8 filename and it does the rest). It's way too much effort to have a separate codepath just for Windows.

-------

This PR includes:
* Fixing compiler warnings about collab code (use of `g_signal_new`). Requires testing by someone who uses collab.
* Ignore more files from git history (ltargz.m4, uthash)
* Removal of `mkdir_p` (use `g_mkdir_with_parents`) and `smprintf` (use `xasprintf`)
* Replace usage of opendir with GLib equivalents (this is still ongoing)
    * The file browser can now show Unicode filenames properly. You probably still can't open them though.
* Fix bug in the autosave recover function on Windows. (Previously if you selected one of the 'Forget' options, it wouldn't forget the file because of an open file handle preventing deletion)
* Fixing bug in saving hotkeys on Windows (for much the same reason as the autosave bug)
* Getting rid of fork/exec in autotrace and using `g_spawn_sync` instead. This removes the Windows specific code path.
* Porting the MetaFont import functionality to Windows. You need a LaTeX distribution installed (e.g. MikTeX) for this to work. Probably also fixed this process for files with spaces in their names.

cc @frank-trampe


larsenwork:
I'm no windows and/or colab user but just wanted to say :+1: on the work

jtanx:
Thanks.

I've briefly tested the collab and autotrace code on Linux, and I don't think there's any (introduced) bugs caused by this. I did fix one segfault bug in the collab code that was unrelated to these changes.

P.S: I notice the collab server doesn't close by itself when FontForge closes... I don't think this is caused by me though.

frank-trampe:
I expect to have sometime to look at this on Wednesday.

jtanx:
I have switched over all remaining uses of opendir to the GLib equivalents (except in acorn2sfd.c which appears to be a separate program in itself).

I also removed the header `fileutil.h`, which was redundant, given that all the function prototypes in it already existed in `gfile.h`.

fontforge/python.c
```diff
@@ -7612,7 +7606,7 @@ return( NULL );
     free(filename);
 
     /* Check if the file exists and is readable */
-    if ( access(locfilename,R_OK)!=0 ) {
+    if ( GFileReadable(locfilename) ) {
```
jtanx: Oops, should be !GFileReadable

jtanx:
Hmm, just going through, why was 5360843effed654834d437fcc5c812d5937d23a8 done again? The functions in `usprintf` are still unused.

adrientetar:
Too often I see reverted commits without explanatory comment. IMO you should always explain why the revert in the log.

Unicode/ucharmap.c
```diff
@@ -640,4 +640,4 @@ char *utf82def_copy(const char *ufrom) {
     ret = u2def_copy(u2from);
     free(u2from);
     return( ret );
-}
+}
```
frank-trampe: Why are we dropping the trailing linebreak?

Unicode/ustring.c
```diff
@@ -858,6 +862,95 @@ void utf8_strncpy(register char *to, const char *from, int len) {
...
+ * @return The output buffer (NULL on error). Must be freed by the caller.
+ */
+char *utf82fsys_copy(const char *ufrom) {
+#ifdef __MINGW32__
+    return copy(ufrom);
+#else
+    return utf82def_copy(ufrom);
+#endif
+}
+
```
frank-trampe: Did you test these functions? What's the easiest way to do that?

fontforge/autosave.c
```diff
@@ -25,136 +25,121 @@
...
+static char *GetAutoDirName() {
+    char *userdir = getFontForgeUserDir(Config);
+    char *autodir = NULL;
+    
+    if (userdir != NULL) {
+        int ret = asprintf(&autodir, "%s/autosave", userdir);
+        free(userdir);
+        if (ret == -1) {
+            return NULL;
+        } else if (!GFileExists(autodir) && GFileMkDir(autodir) == -1) {
```
frank-trampe: A comment explaining that we are creating the directory if non-existent might be in order.

fontforge/autosave.c
```diff
@@ -25,136 +25,121 @@
...
-	fprintf( stderr, "Recovering from %s... ", buffer);
-	if ( (sf = SFRecoverFile(buffer,inquire,&inquire_state)) ) {
-	    any=true;
-	    if ( sf->fv==NULL )		/* Doesn't work, cli arguments not parsed yet */
-		FontViewCreate(sf,false);
-	    fprintf( stderr, " Done\n" );
-	}
+    if (recoverdir == NULL) {
+        return false;
+    } else if ((dir = g_dir_open(recoverdir, 0, NULL)) == NULL) {
```
frank-trampe: Hmm. Now I need to get out the GLib documentation.

frank-trampe: What's the intended improvement here?

Unicode/ucharmap.c
jtanx: Oops, that wasn't intentional.

fontforge/autosave.c
jtanx: OK

Unicode/ustring.c
jtanx: Yes; tested by essentially anything involving file operations (e.g. opening/closing files) and seeing if they still work. Moreover testing to see if it can open files containing e.g. Cyrillic characters (which it now can).

On anything other than Windows (MinGW, not CygWin), it boils down to calling exactly the same function, so the only potential effect is on Windows. 

Although one thing I've been contemplating is if it makes sense to just use the `2def/def2` functions, and force UTF-8 encoding for Windows.

fontforge/autosave.c
jtanx: On Windows, opendir won't work if the path contains nonstandard characters (characters outside of their current codepage). GLib ensures that the paths are UTF-8 encoded (on Windows) so this will work irregardless of what current codepage the user has.

fontforge/autotrace.c
```diff
@@ -25,26 +25,22 @@
  * ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 #include "fontforgevw.h"
```
frank-trampe: This is a big diff. Would you outline the changes?

fontforge/autowidth.c
```diff
@@ -1086,7 +1086,7 @@ int AW_ReadKernPairFile(char *fn,WidthInfo *wi) {
     unichar_t buffer[300];
     struct kernsets ks;
 
-    filename = utf82def_copy(fn);
+    filename = utf82fsys_copy(fn);
```
frank-trampe: There are only about 8 comments in all of `ucharmap.c`. What is the difference between these two functions?

fontforge/python.c
```diff
@@ -7580,14 +7579,9 @@ Py_RETURN( self );
 static PyObject *PyFFGlyph_autoTrace(PyObject *self, PyObject *UNUSED(args)) {
     SplineChar *sc = ((PyFF_Glyph *) self)->sc;
     int layer = ((PyFF_Glyph *) self)->layer;
-    char **at_args;
-
-    at_args = AutoTraceArgs(false);
-    if ( at_args==(char **) -1 ) {
-	PyErr_Format(PyExc_EnvironmentError, "Bad autotrace args" );
-return(NULL);
```
frank-trampe: What was wrong here?

fontforge/sfd.c
```diff
@@ -2983,26 +2983,24 @@ return( err );
...
-    unlink(filename);		/* Just in case it's a normal file, it shouldn't be, but just in case... */
-    dir = opendir(filename);
+    g_unlink(filename);		/* Just in case it's a normal file, it shouldn't be, but just in case... */
+    dir = g_dir_open(filename, 0, NULL);
     if ( dir==NULL )
 return;
     buffer = malloc(strlen(filename)+1+NAME_MAX+1);
-    while ( (ent = readdir(dir))!=NULL ) {
-	if ( strcmp(ent->d_name,".")==0 || strcmp(ent->d_name,"..")==0 )
-    continue;
```
frank-trampe: Was this logic necessary even before?

fontforge/sfd.c
```diff
@@ -9155,47 +9149,49 @@ static int ask_about_file(FILE *asfd,int *state,char *filename) {
...
-      case 3:
-	*state = 2;
-	/* Fall through */
-      case 4:
-	unlink(filename);
-return( false );
-      default:
-      break;
+    ret = ff_ask(_("Recover old edit"),(const char **) buts,0,3,_("You appear to have an old editing session on %s.\nWould you like to recover it?"), pt);
+    switch (ret) {
```
frank-trampe: We need comments on the various states if we're selectively using fclose.

fontforge/autotrace.c
```diff
@@ -573,259 +473,231 @@ return;
...
+ * Finds the autotrace/potrace executable.
+ * 
+ * @param [out] path The buffer to store the location, or NULL if not required.
+ * @param [in] bufsiz The size of the output buffer.
+ * @return nonzero if potrace was found, zero if autotrace was found, negative
+ *         if none was found. Note previously, FontForge determined itself if
+ *         the resulting executable was potrace or autotrace, irrespective of
+ *         setting AUTOTRACE or POTRACE. (i.e. AUTOTRACE=potrace was possible).
+ */
+int FindAutoTrace(char *path, size_t bufsiz) {
```
frank-trampe: This seems a backward way of getting a string. Could we just return a dynamically allocated string or a null pointer?

gutils/fsys.c
```diff
@@ -912,9 +847,9 @@ return NULL;
 #if defined(__MINGW32__)
 	/* Allow for preferences to be saved locally in a 'portable' configuration. */ 
 	if (getenv("FF_PORTABLE") != NULL) {
-		buf = smprintf("%s/preferences/", getShareDir());
+		buf = xasprintf("%s/preferences", getShareDir());
```
frank-trampe: What is the improvement?

gutils/giofile.c
```diff
@@ -34,6 +34,8 @@
 #include <unistd.h>
 #include <dirent.h>
 #include <errno.h>
+#include <glib.h>
```
frank-trampe: I'm taking a break; I'll pick up here later.

fontforge/autowidth.c
jtanx: As mentioned already before - **all** places involving filenames **must** be UTF-8 encoded on Windows.

fontforge/sfd.c
jtanx: `g_dir_read_name` discards the `"."` and `".."` entries.

fontforge/autotrace.c
jtanx: The issue then is how do you know it was potrace or not?

gutils/fsys.c
jtanx: Not having to implement yet another function that does the same thing?

fontforge/autotrace.c
jtanx: It's in the comments I made earlier, but briefly:
* Get rid of the Windows specific code path by switching to using GLib
* Port the MetaFont importing function to Windows

fontforge/sfd.c
jtanx: Ok. It will call `fclose` when it fails.

fontforge/autotrace.c
frank-trampe: Take an argument specifying which one you're trying to find.

jtanx: Oh right, now I remember, most of the time, the actual path to autotrace is not needed, just knowing that it exists.

fontforge/python.c
jtanx: Personal preference I mostly, but it seemed to be breaking encapsulation - outside of `python.c`, `_SCAutoTrace` and `AutoTraceArgs` are only ever called in `autotrace.c`. Plus by calling `_SCAutoTrace` directly, it misses [other checks](https://github.com/jtanx/fontforge/blob/unidir/fontforge/autotrace.c#L487) that are made in `SCAutoTrace`.

So I guess right it's a tradeoff between less informative error messages and more error checking.

frank-trampe:
@jtanx, I'm not sure about switching every file call to use GLib. It seem ridiculous to me that all of the standard library file handling functions would be broken on Windows. Surely there are other pieces of software that work correctly on Windows using the standard calls (perhaps via some changes to the compilation process)?

jtanx:
@frank-trampe if you want to support opening files with Unicode filenames, GLib is the easiest way. On Windows, all the standard C library functions assume what is called the multi-byte version of functions.

What this means is that they expect filenames (and all other strings for that matter) to be in what is known as the 'current active codepage' - but this is *independent* of how filenames are represented, so you can always have the case that a particular filename that cannot be represented by the current active codepage. 

If you look at the Microsoft documentation for say [fopen](https://msdn.microsoft.com/en-us/library/yeby3zcb.aspx), you'll notice that there is a 'wide-character' version that expects the filename to be encoded in UTF-16  - being Unicode, it works with *all* filenames.

It is for this very reason that internally, GLib **must** mandate that all strings on Windows for filenames are UTF-8 encoded, and internally they do the re-encode to UTF-16 and then call the wide character version of the functions. 

For the greatest portability, I think FontForge should be looking into using more of GLib and not less.

jtanx:
@frank-trampe I plan on splitting this PR to make things easier to merge.

PR #2304, opened 2015-05-30
Fix point info dialog for interpolated points

michal-n:
Point info dialog is unusable for interpolated points.

Looks like pointers to some gadgets are left uninitialized, and so the routine that shuffles them for various point types actually can't complete its work. This fixes it.

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -18158,8 +18158,7 @@ static PyObject* CreatePyModule( module_definition *mdef ) {
     PyObject *module;
 
     if ( mdef->runtime.module != NULL )
-	if ( mdef->runtime.module == PyImport_AddModule(mdef->module_name) )
-            return mdef->runtime.module;
+	return mdef->runtime.module;
 
     if ( mdef->types != NULL && FinalizePythonTypes( mdef->types ) < 0 )
 	return NULL;
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -18158,7 +18158,8 @@ static PyObject* CreatePyModule( module_definition *mdef ) {
     PyObject *module;
 
     if ( mdef->runtime.module != NULL )
-	return mdef->runtime.module;
+	if ( mdef->runtime.module == PyImport_AddModule(mdef->module_name) )
+            return mdef->runtime.module;
 
     if ( mdef->types != NULL && FinalizePythonTypes( mdef->types ) < 0 )
 	return NULL;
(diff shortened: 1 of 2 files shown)


frank-trampe:
@michal-n, I think that this branch has more commits than you had intended. Try pulling from master and making a clean branch.

I also don't see what your added code accomplishes that the memcpy command at line 2728 misses. If the added code does indeed make a difference (and if I'm interpreting the code correctly), there is probably a memory corruption problem somewhere between the memcpy line and your fix lines. If you give me more specific steps to reproduce, I can take a look.

frank-trampe:
Oops. Never mind that second part. ret is the returned value from the widget's constructor, which is why it changes in that region.

This seems like a hack to me as does the existing code's copy of an incomplete GCreateData tree (which probably happened so as to avoid needing to follow the branches of the tree). Let me look at options for dealing with this more completely.

michal-n:
OK

davelab6:
@michal-n to confirm, did you make a clean branch? 

@frank-trampe anything blocking this? I propose to merge in 7 days if no further comments

frank-trampe:
@davelab6, it's very bad practice to merge a self-reverting stack of commits.

davelab6:
Looks like only fe25350 should be merged then; @michal-n can you rebase?

davelab6:
Oh, its a tiny change. @frank-trampe would you like to apply the diff?

frank-trampe:
@davelab6, I've resumed examining the original problem in search of a comprehensive fix.

frank-trampe:
@michal-n, test #2541.

michal-n:
I'd gladly do so, if I only could compile it:

```
bootstrap: running: libtoolize --quiet
bootstrap: running: git clone --depth 365 'git://git.sv.gnu.org/gnulib' 'gnulib'
Cloning into 'gnulib'...
fatal: unable to connect to git.sv.gnu.org:
git.sv.gnu.org[0: 208.118.235.72]: errno=Connection timed out
```

Do I miss something?

frank-trampe:
It pings for me now; try again.

michal-n:
Still no luck.

frank-trampe:
I just tested, and I'm able to clone the gnulib repository. Try pinging the server and see what happens.

michal-n:
100% ping loss. But I found out I can clone over HTTP succesfully. As for interpolated points, looks like problem's fixed.

PR #2319, opened 2015-06-03
Using e.g. "Fontname" instead of the generic "Postscript name"

larsenwork:
Bumped into this alert today and figured it wasn't precise enough (especially for a beginner who are the ones most likely to make the mistake.)

I've changed it to Fontname, Family Name and Name For Humans as that's what the labels are called.

screen shot 2015-06-03 at 18 52 58

### DIFF
--- a/fontforgeexe/fontinfo.c
+++ b/fontforgeexe/fontinfo.c
@@ -2657,7 +2657,7 @@ static int CheckNames(struct gfi_data *d) {
     buts[0] = _("_OK"); buts[1] = _("_Cancel"); buts[2]=NULL;
 
     if ( u_strlen(ufont)>63 ) {
-	ff_post_error(_("Bad Font Name"),_("A PostScript name should be ASCII\nand must not contain (){}[]<>%%/ or space\nand must be shorter than 63 characters"));
+	ff_post_error(_("Bad Font Name"),_("A Fontname should be ASCII\nand must not contain (){}[]<>%%/ or space\nand must be shorter than 63 characters"));
 return( false );
     }
 
@@ -2688,7 +2688,7 @@ return( false );
     }
     while ( *ufamily ) {
 	if ( *ufamily<' ' || *ufamily>=0x7f ) {
-	    ff_post_error(_("Bad Font Family Name"),_("A PostScript name should be ASCII\nand must not contain (){}[]<>%%/ or space"));
+	    ff_post_error(_("Bad Font Family Name"),_("A Family Name should be ASCII\nand must not contain (){}[]<>%%/ or space"));
 return( false );
 	}
 	++ufamily;
@@ -2705,7 +2705,7 @@ return( false );
 		*pt=='(' || *pt=='[' || *pt=='{' || *pt=='<' ||
 		*pt==')' || *pt==']' || *pt=='}' || *pt=='>' ||
 		*pt=='%' || *pt=='/' ) {
-	    ff_post_error(_("Bad Font Name"),_("A PostScript name should be ASCII\nand must not contain (){}[]<>%%/ or space"));
+	    ff_post_error(_("Bad Font Name"),_("A Fontname should be ASCII\nand must not contain (){}[]<>%%/ or space"));
 return( false );
 	}
     }
@@ -3316,8 +3316,8 @@ static int GFI_AddOFL(GGadget *g, GEvent *e) {
 	ff_post_notice(_("Using the OFL for your open fonts"),_(
 	    "The OFL is a community-approved software license designed for libre/open font projects. \n"
 	    "Fonts under the OFL can be used, studied, copied, modified, embedded, merged and redistributed while giving authors enough control and artistic integrity. For more details including an FAQ see http://scripts.sil.org/OFL. \n\n"
-	    "This font metadata will help users, designers and distribution channels to know who you are, how to contact you and what rights you are granting. \n" 
-        "When releasing modified versions, remember to add your additional notice, including any extra Reserved Font Name(s). \n\n" 
+	    "This font metadata will help users, designers and distribution channels to know who you are, how to contact you and what rights you are granting. \n"
+        "When releasing modified versions, remember to add your additional notice, including any extra Reserved Font Name(s). \n\n"
 	    "Have fun designing open fonts!" ));
     }
 return( true );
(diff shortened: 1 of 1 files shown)


jtanx:
Hmm, well we are in the  "PS names" tab.

adrientetar:
> Bumped into this alert today and figured it wasn't precise enough (especially for a beginner who are the ones most likely to make the mistake.)

Really? This PR is only making it less precise, as far as I can see. Also, I don’t see anything wrong with the current message.

fontforgeexe/fontinfo.c
```diff
@@ -2657,7 +2657,7 @@ static int CheckNames(struct gfi_data *d) {
     buts[0] = _("_OK"); buts[1] = _("_Cancel"); buts[2]=NULL;
 
     if ( u_strlen(ufont)>63 ) {
-	ff_post_error(_("Bad Font Name"),_("A PostScript name should be ASCII\nand must not contain (){}[]<>%%/ or space\nand must be shorter than 63 characters"));
+	ff_post_error(_("Bad Font Name"),_("A Fontname should be ASCII\nand must not contain (){}[]<>%%/ or space\nand must be shorter than 63 characters"));
```
adrientetar: Either "A font name" or just "FontName" (referencing the attribute).

Anyway, I don’t think this statement is true outside of Postscript.

Also I note that the current message says Postscript name because this rule applies to all Postscript names. Only the window title ought to display the individual attribute (see how the window titles differs across the different messages).

larsenwork: I hadn't noticed the title - the error should always say the same as the title. Anything else is bad UX. (And the title still isn't 100% = label name).

I'm no code expert but I'm fairly sure it only responds to "Bad Font Name" so there's no reason for not using the precise "Fontname" 

(the "A" should be removed, yes)

adrientetar: > I'm no code expert but I'm fairly sure it only responds to "Bad Font Name" so there's no reason for not using the precise "Fontname"

Yes, but the person who types an invalid character there probably does not know that this applies to all fields in the window so this is extra information.

larsenwork:
> Really? This PR is only making it less precise, as far as I can see. Also, I don’t see anything wrong with the current message.

The current message doesn't tell the user what label needs fixing if that's not wrong then I don't know...(yes the title is there but it is also wrong (Font Name vs Fontname) and the interpretation of an error message should never depend on the title anyway) 

PS: I don't know in french but starting a reply with "really" is juvenile and rude in english so I'd recommend you'd refrain from doing so if you want to contribute to a positive tone in discussions.

larsenwork:
@jtanx yep, but the error message alone doesn't tell the user if it's Fontname, Family Name, Name For Humans that is wrong

fontforgeexe/fontinfo.c
larsenwork: > Yes, but the person who types an invalid character there probably does not know that this applies to all fields in the window so this is extra information.

An error message is no place for "extra information" it should tell the user what he/she did wrong as short and precise as possible. 
If you absolutely want to educate your users in error messages then you should make it clear that you are doing so e.g. 
```
Fontname should be ASCII\nand must not contain (){}[]<>%%/ or space\nand must be shorter than 63 characters.\nThis is true for all Postscript names.
```

larsenwork:
What we could (and probably should) do is to label it "Postscript name" instead of "Fontname" so we have:

* Postscript name
* Name For Humans (on mac this is called "Full name" - is that more correct?)
* Family name

I'll update this pr to reflect this but I don't know how to change the labels.

larsenwork:
Name For Humans could also be "Human name"

larsenwork:
> Hmm, well we are in the "PS names" tab.

Yep, your point being? My point is that when we tell the user he/she made an error and we know what error that is, then there's no reason not to tell the user that in the shortest and most precise way possible.

Which is what I'm trying to do here - constructive suggestions welcomed.

jtanx:
My point being that "PS Names" --> "PostScript names", hence the same warning applying to all fields in that tab. It's an accurate error message.

larsenwork:
@jtanx have you read this: https://github.com/fontforge/fontforge/pull/2319#discussion_r31754384

pathumego:
@jtanx it is an accurate message. But I agree with @larsenwork on this change.  I faced this confusion when I started working with FF first. For a beginner it is confusing since there is no field called `PostScript names` and for someone who is not good with English and/or not familiar with terminology  assuming PS Names is `PostScript names` could be confusing. Current message is accurate, but dose not tell what field has the problem.

jtanx:
Not before posting that reply, which was simply replying to your "yes, and?" question.

larsenwork:
@jtanx ok, that's why I asked :) any comments on that comment?

larsenwork:
An alternate solution could be to change `Fontname` to `Postscript Name` like so:

screen shot 2015-06-05 at 10 20 20

pathumego:
@larsenwork This is better :) Im not sure about `Human Name` is there a better term for this? and shouldn't `Weight` be `Style Name` too?

larsenwork:
Mac uses "Full name" but I don't think it's better.
Weight: Makes sense - I only "fixed" the names with "name" in them

frank-trampe:
@larsenwork, I disagree with your characterization of the comment from @adrientetar as appearing rude to a native English speaker. When I am not sure that I am hearing something correctly, I often lead into my request for confirmation/clarification with the same expression (Really?).

frank-trampe:
I find the pull request as written to be inconsistent. The font name and the family name are both PostScript names. Perhaps we could start each message by stating the rules for a PostScript name and then indicating that a particular value (font name or family name or ...) currently fails to comply.

Or we could do nothing. I don't think that this is a big issue, and I don't see a solution that increases specificity without also increasing verbosity.

larsenwork:
@frank-trampe my updated suggestion was based on what FontBook on mac calls them but I actually think we to rename and rearrange a few things to make it more correct and less confusing for beginners - I found it confusing when I started out too as did @pathumego and I'm fairly sure we aren't the only two although that's just an assumption. 

I think we will do it as part of the menu makeover we're doing anyways. So I'm closing this as we can't agree on an easy fix and we'll probably return with a more improved and thorough solution as the current leaves a lot of room for improvement:)

"Really?" Depends on how you follow it up but let's not go there. I was just tired being shot down without constructive and valid arguments and I've already apologised so I see no reason to stir that pot any further.

PR #2330, opened 2015-06-05
updated FF app, debug, install and uninstall icons

larsenwork:
@jtanx should be what you need and if not then let me know. I hope you read my apology in the previous thread.


frank-trampe:
The 22x22 icon is missing.

larsenwork:
@frank-trampe for the .ico files and/or just the png? 
According to [stackoverflow](http://stackoverflow.com/questions/3236115/which-icon-sizes-should-my-windows-applications-icon-include) the stack should be: 16, 24, 32, 40, 48, 64, 96, 128 and 256. 
In which case I would need to add 24, 40, 48 and 96 too?

frank-trampe:
I don't know about the ico file right now. But the png-format images for 22x22, 24x24, and 48x48 are missing. We don't need any icons that were not present before.

larsenwork:
@frank-trampe Ok, I'll just leave the ico for now until @jtanx does the :+1: or tells me what sizes needs adding (they can be made from the svg already in this pull but I would like to do it for pixel perfecting at smaller sizes:) Png's will be up in a minute.

larsenwork:
@frank-trampe note that I haven't run any png optimisation as I'm quite sure it broke the pixmaps when I tried it on them but can I do it to these?

frank-trampe:
What sort of optimization?

frank-trampe:
So it seems that you moved/removed the svg sources (`fontforge/desktop/icons/src/icon-16x16-apps-fontforge.svg` and the like) also. Try building the target `distdir` until it completes successfully.

larsenwork:
Yep, I moved those - I thought they were only there as a "design resource" and not used in build. I'll add and ping you when Travis builds without problems. 
Png: I normally use imageoptim which uses a combination of PNGOUT, Pngcrush, AdvPNG and extended OptiPNG to reduce the file size for when I "ship" pngs for app and/or web use.

larsenwork:
@frank-trampe "All is well" 
One additional question: I can see you've made a top level folder name "debian-old" would it make sense for me to do:
```
Old
    Debian
    Icons
        Old
        Older
    Pixmaps
        Pango
        2012
    etc...
```
so we have one top level folder for all old stuff?

jtanx:
@larsenwork thanks, I really appreciate the work.

wrt `ff.ico`, it's missing the 48x48 size (this is probably the size that gets used the most). To save on space, the 64x64 and 128x128 sizes can also be switched to the PNG variant, since Windows XP doesn't use those resolutions anyway.

larsenwork:
> @larsenwork thanks, I really appreciate the work.

cheers, so just 48 or 24, 40, 48 and 96 (https://github.com/fontforge/fontforge/pull/2330#issuecomment-109357858)

larsenwork:
@jtanx done, Travis just need to confirm

larsenwork:
@jtanx I think I forgot to add 96x96 but unless it's critical the rest should be good to go.

jtanx:
Ok thanks; really I've never seen 24x24 and 96x96 used in practice (in .icos) anyway.

I'll leave final merging to @frank-trampe, since I think other releases depend on the other files you've changed.

larsenwork:
:+1:

frank-trampe:
I see that this pull request changes from one icon set to the other, but what is the reason for the change, or what are we trying to accomplish?

We'll need to squash commits anyway, so I would suggest branching from this branch, rebasing, squashing the commits, giving the commit a description explaining the change and the reason for the change, and then making a new pull request with that information (and the squashed commits). Or I can do this once I have the information necessary to write the description.

larsenwork:
@frank-trampe sorry it got a bit messy because it took a couple of commits for me to get all the sizes right. I'm no git jedi (yet) so thanks for offering to do the squashing.
Description: 
> Updated FontForge app, debug, install and uninstall icons on win + linux to fit the new icon used on website, GitHub and mac.

PR #2341, opened 2015-06-08
Windows: Fix clipboard clearing. Fixes #2286.

### LINKED ISSUE #2286: Windows: Clicking on the text box clears the clipboard for external inputs
If you copy from one program into FontForge, it will paste correctly as long as you don't explicitly click on the text box. As soon as you click on the text box, the clipboard is cleared for unknown reasons. This bug is also present in the 2012 version.

jtanx:
I've made this apply to both MinGW and CygWin. Although primary clipboard
pasting works on CygWin, it too suffers from the fact that clicking on
the text field will clear the clipboard, which more often than not is
not the desired action.

### DIFF
--- a/gdraw/gxdraw.c
+++ b/gdraw/gxdraw.c
@@ -3805,6 +3805,11 @@ static void GXDrawGrabSelection(GWindow w,enum selnames sel) {
 	if ( gd->selinfo[sel].owner->eh!=NULL )
 	    (gd->selinfo[sel].owner->eh)((GWindow) gd->selinfo[sel].owner, &e);
     }
+    //Only one clipboard exists on Windows. Selectively set the selection owner
+    //as otherwise the Windows clipboard will be cleared.
+#ifdef _WIN32
+    if (sel == sn_clipboard)
+#endif
     XSetSelectionOwner(gd->display,gd->selinfo[sel].sel_atom,gw->w,gd->last_event_time);
     GXDrawClearSelData(gd,sel);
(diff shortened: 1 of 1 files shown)


frank-trampe:
I don't entirely understand the nature of the original problem. Is the clipboard implementation on Windows flawed, or were we using it incorrectly?

larsenwork:
@frank-trampe http://designwithfontforge.com/en-US/Configuring_FontForge.html#windows

jtanx:
So as mentioned in #2286, there is no direct equivalent to X11 clipboards on Windows. Whereas in X11, you can have any number of clipboards, such a concept does not exist on Windows. 

What happens is the X11 server (eg XWin, XMing, VcXsrv) translates the X11 clipboard into an equivalent on Windows. 

What XWin does is it translates the primary clipboard into the Windows clipboard. VcXsrv seemingly clears the clipboard when the primary clipboard is set, which may be an implementation issue.

**But** the real issue is that because as soon as you click into the text field, the clipboard is set, it clears any copied input that the user wants to paste into the field.

This isn't an issue on Linux because it just sets the primary clipboard which is independent of the 'explicit' clipboard, but it is a major issue on Windows.

Since the whole concept of a 'primary' clipboard doesn't exist on Windows anyway (it's not expected that highlighting text in one application and middle-clicking into another will do a paste), this PR alters the behaviour to only expose 'explicit' copy events, which do work well with VcXsrv.

frank-trampe:
That makes sense.

Issue #2296, opened 2015-05-27
Error when trying to brew from source on mac


larsenwork:
See the terminal output here: 
https://gist.github.com/andreaslarsen/ee78d2f28880ec26c405

I tried to follow this guide: 
https://github.com/fontforge/fontforge/blob/master/INSTALL-git.md

Installing dependencies without problems. But building source or release gives me the same build error

```
BuildError: Failed executing: ./bootstrap 
```

andreaslarsen:
Running a backtrace gives me this:
```
Choose an action: backtrace
/usr/local/Library/Homebrew/debrew.rb:11:in `raise'
/usr/local/Library/Homebrew/formula.rb:873:in `block in system'
/usr/local/Library/Homebrew/formula.rb:830:in `open'
/usr/local/Library/Homebrew/formula.rb:830:in `system'
/usr/local/Library/Formula/fontforge.rb:85:in `install'
/usr/local/Library/Homebrew/debrew.rb:22:in `block in install'
/usr/local/Library/Homebrew/debrew.rb:97:in `debrew'
/usr/local/Library/Homebrew/debrew.rb:22:in `install'
/usr/local/Library/Homebrew/build.rb:129:in `block in install'
/usr/local/Library/Homebrew/formula.rb:521:in `block in brew'
/usr/local/Library/Homebrew/formula.rb:911:in `block in stage'
/usr/local/Library/Homebrew/resource.rb:91:in `block in unpack'
/usr/local/Library/Homebrew/extend/fileutils.rb:17:in `mktemp'
/usr/local/Library/Homebrew/resource.rb:88:in `unpack'
/usr/local/Library/Homebrew/resource.rb:81:in `stage'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/forwardable.rb:171:in `stage'
/usr/local/Library/Homebrew/formula.rb:909:in `stage'
/usr/local/Library/Homebrew/formula.rb:517:in `brew'
/usr/local/Library/Homebrew/build.rb:107:in `install'
/usr/local/Library/Homebrew/build.rb:179:in `<main>'
```

jtanx:
May be related: #2019

andreaslarsen:
Ok, maybe @DomT4 can shed some light :)

DomT4:
[...]

PR #2358, opened 2015-06-12
INSTALL-git.md: Update homebrew section

davelab6:
https://github.com/fontforge/fontforge/issues/2296

### DIFF
--- a/INSTALL-git.md
+++ b/INSTALL-git.md
@@ -129,7 +129,7 @@ export PATH=${PATH}:/usr/local/opt/gettext/bin
 
 ## On Mac OS X with Homebrew
 
-Homebrew no longer packages FontForge with a UI.
+Homebrew no longer packages FontForge with a UI, but the release GUI app bundles are available from Cask. 
 You can install FontForge from the current github source code using a package manager, such as MacPorts or Fink. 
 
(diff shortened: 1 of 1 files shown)


larsenwork:
Sounds a little confusing telling people they can't use Homebrew and then moves on to use brew commands - could maybe be clarified a bit:)

davelab6:
They can use homebrew for the python, but not the ui

INSTALL-git.md
```diff
@@ -154,3 +126,41 @@ Then edit then your shell profile (eg, `~/.bash_profile` or `~/.zprofile`) and a
 ```
 export PATH=${PATH}:/usr/local/opt/gettext/bin
 ```
+
+## On Mac OS X with Homebrew
+
+Homebrew no longer packages FontForge with a UI, but the release GUI app bundles are available from Cask. 
+You can install FontForge from the current github source code using a package manager, such as MacPorts or Fink. 
+
+Here is how to install FontForge without a UI:
```
larsenwork: Maybe make this

##Install FontForge without UI using HomeBrew

davelab6: Make the PR :)

PR #2367, opened 2015-06-14
Set freetypecontext to NULL in `AnchorD_FreeChar`. Closes #2366.

### LINKED ISSUE #2366: [PATCH] Unallocated memory access in anchors dialog
Now that AnchorD_FreeChar() is called not only on dialog destruction, we should initialize a->freetypecontext to NULL after freeing it. The crash can be reproduced by going to anchor control dialog, and then clicking a different mark glyph.
```
diff --git a/fontforgeexe/anchorsaway.c b/fontforgeexe/anchorsaway.c
index 71fec0d..e9dd047 100644
--- a/fontforgeexe/anchorsaway.c
+++ b/fontforgeexe/anchorsaway.c
@@ -117,8 +117,10 @@ static void AnchorD_FreeChar(AnchorDlg *a) {
     for ( i=0; i<a->cnt; ++i )
        BDFCharFree(a->apmatch[i].bdfc);
     free(a->apmatch); a->apmatch = NULL;
-    if ( a->freetypecontext!=NULL )
+    if ( a->freetypecontext!=NULL ) {
        FreeTypeFreeContext(a->freetypecontext);
+        a->freetypecontext = NULL;
+    }
 }
 
 static void AnchorD_FreeAll(AnchorDlg *a) {
```

jtanx:
Fixes a use-after-free bug reported by @tuffnatty.

cc @frank-trampe

### DIFF
--- a/fontforgeexe/anchorsaway.c
+++ b/fontforgeexe/anchorsaway.c
@@ -117,8 +117,10 @@ static void AnchorD_FreeChar(AnchorDlg *a) {
     for ( i=0; i<a->cnt; ++i )
 	BDFCharFree(a->apmatch[i].bdfc);
     free(a->apmatch); a->apmatch = NULL;
-    if ( a->freetypecontext!=NULL )
-	FreeTypeFreeContext(a->freetypecontext);
+    if ( a->freetypecontext!=NULL ) {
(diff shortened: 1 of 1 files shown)


frank-trampe:
Good.

PR #2371, opened 2015-06-15
More static code analysis fixes

jtanx:
I was mostly looking at the 'incorrect expression' issues.

### DIFF
--- a/fontforge/scripting.c
+++ b/fontforge/scripting.c
@@ -633,7 +633,7 @@ static void bStrSplit(Context *c) {
 	    c->return_val.u.aval->argc = cnt;
 	    c->return_val.u.aval->vals = malloc(cnt*sizeof(Val));
 	} else {
-	    if ( *pt!='\0' ) {
+	    if ( *pt!='\0' && cnt < max ) {
 		c->return_val.u.aval->vals[cnt].type = v_str;
 		c->return_val.u.aval->vals[cnt].u.sval = copy(pt);
 	    }
--- a/fontforge/scripting.c
+++ b/fontforge/scripting.c
@@ -175,6 +175,7 @@ return;
 	    arrayfree( dica->entries[i].val.u.aval );
     }
     free( dica->entries );
+    dica->entries = NULL;
 }
 
 static int DicaLookup(struct dictionary *dica,char *name,Val *val) {
@@ -218,8 +219,10 @@ static void calldatafree(Context *c) {
             free( c->a.vals[i].u.sval );
             c->a.vals[i].u.sval = NULL;
         }
-	if ( c->a.vals[i].type == v_arrfree || (c->a.vals[i].type == v_arr && c->dontfree[i]!=c->a.vals[i].u.aval ))
+	if ( c->a.vals[i].type == v_arrfree || (c->a.vals[i].type == v_arr && c->dontfree[i]!=c->a.vals[i].u.aval )) {
 	    arrayfree( c->a.vals[i].u.aval );
+	    c->a.vals[i].u.aval = NULL;
+    }
 	c->a.vals[i].type = v_void;
     }
     DictionaryFree(&c->locals);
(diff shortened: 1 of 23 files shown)


fontforge/scripting.c
```diff
@@ -3278,9 +3281,8 @@ static void bLoadTableFromFile(Context *c) {
 	sf->ttf_tab_saved = tab;
     } else
 	free(tab->data);
-    tab->len = len;
```
adrientetar: It seems the `len` variable is unused after your patch, could be cleaned up completely.

jtanx: It's used to `malloc` the space required and for the `fread` length. Since `fread` may read less bytes than what is specified, `tab->len` is set to the result of `fread`

jtanx: Unless you mean to just replace `len` with `statb.st_size`?

adrientetar: Yea.

jtanx: Done

frank-trampe:
I'm travelling today. I will plan to look at this tonight or tomorrow night.

jtanx:
Build is failing for unrelated reasons - it seems that pango is being compiled [without xft support](https://travis-ci.org/fontforge/fontforge/jobs/67294819#L2785) on the Mac build. @DomT4 any idea what's going on?

DomT4:
@jtanx #2380

fontforge/sfd.c
```diff
@@ -4855,8 +4857,7 @@ void SFDGetKerns( FILE *sfd, SplineChar *sc, char* ttok ) {
 		while ( (ch=nlgetc(sfd))==' ' );
 		ungetc(ch,sfd);
 		if ( ch=='{' ) {
-		    kp->adjust=chunkalloc(sizeof(DeviceTable));
-		    SFDReadDeviceTable(sfd,kp->adjust);
+		    kp->adjust = SFDReadDeviceTable(sfd, NULL);
```
frank-trampe: I dislike the way SFDReadDeviceTable works. Allocating storage because an argument is NULL is not expected behavior.

fontforge/sfd.c
```diff
@@ -5881,8 +5881,10 @@ return( 0 );
 	if ( ch!='\n' && ch!='\r' )
 	    getint(sfd,&vwidth);
     }
-    if ( enc<0 ||xmax<xmin || ymax<ymin )
-return( 0 );
+    if ( enc<0 ||xmax<xmin || ymax<ymin ) {
+        BDFCharFree(bfc);
```
frank-trampe: Why are we only freeing bfc here and not in the other abortive cases above?

frank-trampe: Did you mean to move the allocator as below?

frank-trampe:
This looks fine aside from the things upon which I have commented.

fontforge/sfd.c
jtanx: Okay, but it was already being done like this in the past. If you have a separate allocation as before, there's a possibility that you will have allocated memory but `SFDReadDeviceTable` will return NULL due to some error.

fontforge/sfd.c
jtanx: Nice find

fontforge/sfd.c
```diff
@@ -5848,8 +5848,6 @@ static int SFDGetBitmapChar(FILE *sfd,BDFFont *bdf) {
     EncMap *map;
     int ch;
 
-    bfc = chunkalloc(sizeof(BDFChar));
-    memset( bfc,'\0',sizeof( BDFChar ));
```
jtanx: This memset isn't needed because chunkalloc boils down to calloc which zeroes bfc

fontforge/sfd.c
frank-trampe: I wasn't blaming you for the function syntax. I missed the upside of letting the function return the storage.

frank-trampe:
Okay.

Issue #2369, opened 2015-06-15
Bad SVG font export


puzrin:
How to reproduce:
1. open ttf font from https://github.com/google/material-design-icons/tree/master/iconfont
2. create svg font
3. -> svg font is broken

Problems:
1. svg font can't be opened in FontForge (error about bad format)
2. `glyph-name` and `unicode` attrs **swapped** **AND** their content **improperly encoded**:
   - `unicode` should be defined like ~ `&#xf000`
   - `glyph-name` (actually ligature) should not encode english letters

Example from FF (bad output):

```
<glyph glyph-name="uniE046" unicode="snoo&#x7a;e" 
d="M192 277v43h128v-38l-77 -90h77v-43h-128v39l77 89h-77zM256 85q62 0 105.5 44t43.5 106t-43.5 105.5t-105.5 43.5t-105.5 -43.5t-43.5 -105.5t43.5 -106t105.5 -44zM256 427q80 0 136 -56.5t56 -135.5t-56 -135.5t-136 -56.5t-136 56.5t-56 135.5t56 135.5t136 56.5z
M469 390l-27 -33l-98 83l27 32zM168 440l-98 -82l-27 32l98 82z" />
```

PS. Ubuntu 14.04 LTS and latest FontForge from PPA.

jtanx:
@puzrin, from the [specs](http://www.w3.org/TR/SVG/fonts.html#GlyphElementGlyphNameAttribute), the glyph-name and unicode parameters look correct to me. E.g. `uniE046` is the glyph name assigned by FontForge because that glyph is in the private use area and has no special name. Typing e.g. 'snooze' will be replaced by that glyph. 

The encoding of English characters and FontForge failing to open the generated font are issues though.

puzrin:
@jtanx there are 2 problems:

1. Saved SVG can not be opened back. That means, something is completely wrong.
2. opened TTF already has assigned codes, and i don't  see reasons to change/miss those on SVG save.

Actually, `unicode` should contain character code, and `glyph-name` is not mandatory but can contain ligature.

jtanx:
@puzrin, yes I already said that failing to open the SVG is an issue.

> Actually, unicode should contain character code, 

The [specs say](http://www.w3.org/TR/SVG/fonts.html#GlyphElementGlyphNameAttribute)
> One or more Unicode characters indicating the sequence of Unicode characters which corresponds to this glyph
[...]

PR #2372, opened 2015-06-15
Attempt to fix some issues identified in #2369

jtanx:
#2369

### DIFF
--- a/fontforge/svg.c
+++ b/fontforge/svg.c
@@ -755,7 +755,7 @@ static void svg_scdump(FILE *file, SplineChar *sc,int defwid, int encuni, int vs
 	c = LigCnt(sc->parent,best,univals,sizeof(univals)/sizeof(univals[0]));
 	fputs("unicode=\"",file);
 	for ( i=0; i<c; ++i )
-	    if ( univals[i]>='A' && univals[i]<'z' )
+	    if ( univals[i]>='A' && univals[i]<='z' )
 		putc(univals[i],file);
 	    else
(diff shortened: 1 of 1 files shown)


fontforge/svg.c
```diff
@@ -755,7 +755,7 @@ static void svg_scdump(FILE *file, SplineChar *sc,int defwid, int encuni, int vs
 	c = LigCnt(sc->parent,best,univals,sizeof(univals)/sizeof(univals[0]));
 	fputs("unicode=\"",file);
 	for ( i=0; i<c; ++i )
-	    if ( univals[i]>='A' && univals[i]<'z' )
+	    if ( univals[i]>='A' && univals[i]<='z' )
```
frank-trampe: That's funny!

Was this silently and mysteriously breaking something for you?

fontforge/svg.c
```diff
@@ -3356,7 +3358,7 @@ return( NULL );
 		xmlStrcmp(kids->name,(const xmlChar *) "missing-glyph")==0 )
 	    ++cnt;
     }
-    if ( sf->descent==0 ) {
+    if ( !has_font_face ) {
```
frank-trampe: That does seem much more reasonable.

frank-trampe:
I like it.

fontforge/svg.c
jtanx: Not particularly, but it did end up encoding strings like `snoo&#x7a;e` instead of `snooze`

PR #2373, opened 2015-06-15
Fix and better document the native scripting Chr() function.

mskala:
It looks like this function never actually worked as documented.  Change backported from FontAnvil.

### DIFF
--- a/doc/html/scripting-alpha.html
+++ b/doc/html/scripting-alpha.html
@@ -494,11 +494,14 @@ a[4][0] = "Nested array";
 	    <A NAME="Chr">C</A>hr(int)<BR>
 	    Chr(array)
 	  <DD>
-	    Takes an integer [0,255] and returns a single character string containing
-	    that code point. Internally FontForge interprets strings as if they were
-	    in utf8 (well really, FontForge almost always just uses ASCII-US internally).
-	    If passed an array, it should be an array of integers and the result is the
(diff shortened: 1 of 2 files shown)


frank-trampe:
Let me do some mining just to be sure.

frank-trampe:
I know. I'll get to it soon.

JoesCat:
Hi @frank-trampe ?

Hi mskala, I did a diff between fontanvil/scripting.c and fontforge/scripting.c and noted an interesting improvement you've made with builtins[] which dramatically reduces code using min/max args. Wold you be interested in creating a back-port pull request based on it? I'd like to add another pull request built on it for further compiled executable size reduction.

frank-trampe:
Sorry about that. I've been tied up (figuratively, mostly) and out of town, a lot. I repeat my pledge to look at this soon, probably early next week.

JoesCat:
Ok, can work-around this...

...Code compression was planned at docall:(found->func)(&sub); and minion routines...

JoesCat:
Hi @mskala, never-mind my request for a backport of the lookup table...this needs more thought, I think a different plan is needed to include some of the more complex argument types....

davelab6:
@frank-trampe any update? I think we can trust @mskala on this and merge, so I propose to merge after 7 days from now if no further comments

frank-trampe:
It looks fine to me.

PR #2374, opened 2015-06-15
Fix more collab code warnings and ignore more files from git history

jtanx:
This has been split off from #2299.

cc @frank-trampe

### DIFF
--- a/fontforgeexe/collabclientui.c
+++ b/fontforgeexe/collabclientui.c
@@ -433,8 +433,8 @@ static void collab_sessioncallbacks_init (CollabSessionCallbacks *self) {}
 static void
 collab_sessioncallbacks_class_init (CollabSessionCallbacksClass *klass)
 {
-    GType argtypes[] = { G_TYPE_POINTER, NULL };
-    
+    GType argtypes[] = {G_TYPE_POINTER};
+
     g_signal_newv ("joining",
 		   (collab_sessioncallbacks_get_type ()),
 		   G_SIGNAL_RUN_LAST | G_SIGNAL_NO_RECURSE | G_SIGNAL_NO_HOOKS,
@@ -471,30 +471,34 @@ gpointer getSessionCallbacksObject()
 void collabclient_notifySessionJoining( cloneclient_t *cc, FontViewBase* fv )
 {
 #ifdef BUILD_COLLAB
-    
-    printf("AAA collabclient_notifySessionJoining() fv:%p\n", fv);
+
+    printf("AAA collabclient_notifySessionJoining() cc: %p, fv:%p\n", cc, fv);
     g_signal_emit_by_name( getSessionCallbacksObject(), "joining", fv );
-    
+
 #endif    
 }
 void collabclient_notifySessionLeaving( cloneclient_t *cc, FontViewBase* fv )
 {
 #ifdef BUILD_COLLAB
-    
-    cc->sessionIsClosing = 1;
-    g_signal_emit_by_name( getSessionCallbacksObject(), "leaving", fv );
+
+    if (cc) {
+        cc->sessionIsClosing = 1;
+        g_signal_emit_by_name( getSessionCallbacksObject(), "leaving", fv );
+    } else {
+        LogError(_("collabclient_notifySessionLeaving: cc was NULL"));
+    }
 
 #endif
 }
 void collabclient_addSessionJoiningCallback( collabclient_notification_cb func )
 {
     void* data = 0;
-    g_signal_connect( getSessionCallbacksObject(), "joining", (GCallback)func, data );
+    g_signal_connect( getSessionCallbacksObject(), "joining", G_CALLBACK(func), data );
 }
 void collabclient_addSessionLeavingCallback( collabclient_notification_cb func )
 {
     void* data = 0;
-    g_signal_connect( getSessionCallbacksObject(), "leaving", (GCallback)func, data );
+    g_signal_connect( getSessionCallbacksObject(), "leaving", G_CALLBACK(func), data );
 }
 
 
--- a/fontforgeexe/collabclientui.h
+++ b/fontforgeexe/collabclientui.h
@@ -165,7 +165,11 @@ extern int Collab_getLastChangedPos( void );
 extern int Collab_getLastChangedCodePoint( void );
 
 
-typedef void (*collabclient_notification_cb)( FontViewBase* fv );
+/**
+ * See https://developer.gnome.org/gobject/stable/signal.html
+ * for implied parameters in GLib callback functions.
+ */
+typedef void (*collabclient_notification_cb)( gpointer instance, FontViewBase* fv, gpointer user_data );
 extern void collabclient_addSessionJoiningCallback( collabclient_notification_cb func );
 extern void collabclient_addSessionLeavingCallback( collabclient_notification_cb func );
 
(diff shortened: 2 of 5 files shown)


fontforgeexe/collabclientui.c
```diff
@@ -433,8 +433,8 @@ static void collab_sessioncallbacks_init (CollabSessionCallbacks *self) {}
 static void
 collab_sessioncallbacks_class_init (CollabSessionCallbacksClass *klass)
 {
-    GType argtypes[] = { G_TYPE_POINTER, NULL };
-    
+    GType argtypes[] = {G_TYPE_POINTER};
```
frank-trampe: What was the warning here?

fontforgeexe/collabclientui.c
```diff
@@ -471,30 +471,31 @@ gpointer getSessionCallbacksObject()
 void collabclient_notifySessionJoining( cloneclient_t *cc, FontViewBase* fv )
 {
 #ifdef BUILD_COLLAB
-    
-    printf("AAA collabclient_notifySessionJoining() fv:%p\n", fv);
-    g_signal_emit_by_name( getSessionCallbacksObject(), "joining", fv );
-    
+    if (cc) {
```
frank-trampe: Is there ever a case in which it would be okay for cc to be NULL in this context? If not, we ought to have a menacing error message in the else block.

fontforgeexe/collabclientui.h
```diff
@@ -165,7 +165,11 @@ extern int Collab_getLastChangedPos( void );
 extern int Collab_getLastChangedCodePoint( void );
 
 
-typedef void (*collabclient_notification_cb)( FontViewBase* fv );
+/**
+ * See https://developer.gnome.org/gobject/stable/signal.html
+ * for implied parameters in GLib callback functions.
+ */
```
frank-trampe: Did this change, or were we just doing it wrong? And would this possibly be responsible for that mysterious crash that I can't quite remember?

fontforgeexe/collabclientui.c
jtanx: `GType` is not a pointer type, so using `NULL` in this case doesn't make sense. Usually lists in GLib need to be NULL terminated, but in this case it doesn't need to be because the size parameter is specified in `g_signal_newv`.

fontforgeexe/collabclientui.h
jtanx: It was just being done wrong. Since I don't think many use the collab feature, it's never been fully highlighted as an issue I guess.

fontforgeexe/collabclientui.c
jtanx: I don't know if it's okay or not, but thinking about it, the NULL condition is probably only needed in the `notifySessionLeaving` function since `cc` isn't even (currently) being used here. I guess I added it more for consistency.

frank-trampe: Were we getting crashes in the other block, though?

jtanx: Yes

frank-trampe: What was the occasion? Was this just happening when there was no network connection?

jtanx: IIRC it occurred when disconnecting from the server. I'll see if I can reproduce the exact steps

jtanx: * `Collaborate->Start session`
* `Collaborate->Connect to session...` (Choose the only option)
* In the original window: `Collaborate->Close local server`
* In the original window: `Collaborate->Disconnect`

frank-trampe: It seems to me that we ought not to be calling this function during the disconnection process, no? I'm away from my real computer right now, so I cannot do much digging.

jtanx: I think it's meant to be called because it's signalling that we're leaving the collab session. But then again I don't know much about what the collab code is meant to do.

frank-trampe: There is a function called collabclient_notifySessionLeaving that might be more appropriate for signalling that. I'll plan to dig into this a bit more when I'm back at my desk.

frank-trampe: @jtanx, would you provide a backtrace of the crash? I want to see which function is calling this one and whether it ought to be calling the other function instead.

jtanx:
`````
0x00007ffff7a7c872 in collabclient_notifySessionLeaving (cc=0x0, fv=0x86e4c0)
    at collabclientui.c:484
484	    cc->sessionIsClosing = 1;
(gdb) bt
#0  0x00007ffff7a7c872 in collabclient_notifySessionLeaving (cc=0x0, 
    fv=0x86e4c0) at collabclientui.c:484
#1  0x00007ffff7a7c951 in collabclient_sessionDisconnect (fv=0x86e4c0)
    at collabclientui.c:507
#2  0x00007ffff794a0f4 in FVMenuCollabDisconnect (gw=0x8ac970, 
    UNUSED_mi=0x8e6700, UNUSED_e=0x7fffffffbdc0) at fontview.c:5730
#3  0x00007ffff686b9d3 in gmenu_mouse (m=0x971c90, event=0x7fffffffbdc0)
    at gmenu.c:943
#4  0x00007ffff686d487 in gmenu_eh (w=0x9338a0, ge=0x7fffffffbdc0)
    at gmenu.c:1378
#5  0x00007ffff68b9164 in dispatchEvent (gdisp=0x69acc0, event=0x7fffffffbf70)
    at gxdraw.c:3296
#6  0x00007ffff68b96fc in GXDrawEventLoop (gd=0x69acc0) at gxdraw.c:3404
#7  0x00007ffff683fa7b in GDrawEventLoop (gdisp=0x69acc0) at gdraw.c:791
#8  0x00007ffff7a55873 in fontforge_main (argc=1, argv=0x7fffffffe2a8)
    at startui.c:1382
#9  0x00000000004007dd in main (argc=1, argv=0x7fffffffe2a8) at main.c:39
`````

Comment (frank-trampe) [inline fontforgeexe/collabclientui.c]: Then it isn't crashing here.

Comment (frank-trampe) [inline fontforgeexe/collabclientui.c]: I think that we ought to keep the check for safety (since it's always a shame for somebody to lose his work) and to flag a non-fatal error if cc is NULL here.

Comment (frank-trampe) [inline fontforgeexe/collabclientui.c]: Ping (@jtanx).

File: fontforgeexe/collabclientui.c [hunk 4]
```diff
@@ -471,30 +471,32 @@ gpointer getSessionCallbacksObject()
 void collabclient_notifySessionJoining( cloneclient_t *cc, FontViewBase* fv )
 {
 #ifdef BUILD_COLLAB
-    
-    printf("AAA collabclient_notifySessionJoining() fv:%p\n", fv);
+
+    printf("AAA collabclient_notifySessionJoining() cc: %p, fv:%p\n", cc, fv);
```
jtanx: @frank-trampe, since `cc` wasn't even used in `collabclient_notifySessionJoining`, I removed the check on `cc` but added it to this debug statement

frank-trampe: @jtanx, I think that I meant to say that we ought to have an error message in the event that cc is null in the other function. Otherwise we might silently miss an error there.

jtanx: Ah, right. I've added in a warning message now (it pops up in that `Warnings` box)

frank-trampe: Excellent.

frank-trampe:
Looks good.

PR #2375, opened 2015-06-15
Windows Mingw+CygWin patches

jtanx:
This was split from #2299.

* Fix autorecover 'forget about it' and 'forget to all' bug on Windows (MinGW) @frank-trampe I added more comments and slightly reworked it to make it more obvious that the file pointer may be closed by `ask_about_file`.
* Fix hotkey saving on Windows (MinGW)
* Pull in most of the patches from the [CygWin build](http://www.geocities.jp/meir000/fontforge/fontforge-20150121-cygwin.patch.txt).

-----
On Windows, you can only unlink a file that has no open file handles.
So do `fclose` before `unlink`.

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -9142,63 +9142,81 @@ return( NULL );
 return( sf );
 }
 
-static int ask_about_file(FILE *asfd,int *state,char *filename) {
+/**
+ * Asks the user whether or not to recover, skip or delete an autosaved file.
+ * If requested by the user, this function will attempt to delete the file.
+ * @param [in] filename The path to the autosaved file.
+ * @param [in,out] state The current state.
+ *                 state&1: Recover all. state&2: Forget all.
+ * @param [out] asfd Location to store the file pointer to the autosaved file.
+ * @return true iff the file is to be recovered. If true, asfd will hold the
+ *         corresponding file pointer, which must be closed by the caller. If
+ *         false, asfd will hold NULL.
+ */
+static int ask_about_file(char *filename, int *state, FILE **asfd) {
     int ret;
     char *buts[6];
     char buffer[800], *pt;
 
-    if ( *state&1 )
-return( true );
-    else if ( *state&2 ) {
-	unlink(filename);
-return( false );
+    if ((*asfd = fopen(filename, "r")) == NULL) {
+        return false;
+    } else if (*state&1) { //Recover all
+        return true;
+    } else if (*state&2) { //Forget all
+        fclose(*asfd);
+        *asfd = NULL;
+        unlink(filename);
+        return false;
     }
 
-    fgets(buffer,sizeof(buffer),asfd);
-    rewind(asfd);
-    if ( strncmp(buffer,"Base: ",6)!=0 )
-	strcpy(buffer+6, "<New File>");
+    fgets(buffer,sizeof(buffer),*asfd);
+    rewind(*asfd);
+    if (strncmp(buffer,"Base: ",6) != 0) {
+        strcpy(buffer+6, "<New File>");
+    }
     pt = buffer+6;
-    if ( strlen(buffer+6)>70 ) {
-	pt = strrchr(buffer+6,'/');
-	if ( pt==NULL )
-	    pt = buffer+6;
+    if (strlen(buffer+6) > 70) {
+        pt = strrchr(buffer+6,'/');
+        if (pt == NULL)
+            pt = buffer+6;
     }
 
     buts[0] = _("Yes"); buts[1] = _("Yes to _All");
     buts[2] = _("_Skip for now");
     buts[3] = _("Forget _to All"); buts[4] = _("_Forget about it");
     buts[5] = NULL;
-    ret = ff_ask(_("Recover old edit"),(const char **) buts,0,3,_("You appear to have an old editing session on %s.\nWould you like to recover it?"), pt );
-    switch ( ret ) {
-      case 0:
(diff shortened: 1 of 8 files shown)


jtanx:
Apparently the locale hack is also needed for CygWin (taken from the CygWin [patch](http://www.geocities.jp/meir000/fontforge/))

jtanx:
I've added in a commit from the CygWin patch. It seems that the activation of the hand tool by the spacebar has been broken since #6 was done.

When you press the spacebar, the hand tool should activate, but it shouldn't immediately act as if it was 'clicked'. E.g. it shouldn't drag the window around just by pressing the spacebar.

frank-trampe:
Would you add comments on the meanings of the various states?

jtanx:
@frank-trampe  - Do you mean in the switch-case statement?

frank-trampe:
I was looking at the if/then/else statement before that, but comments in the switch case statement would be handy, too.

jtanx:
Sure thing

fontforge/sfd.c
```diff
@@ -9142,63 +9142,81 @@ return( NULL );
...
+ * If requested by the user, this function will attempt to delete the file.
+ * @param [in] filename The path to the autosaved file.
+ * @param [in,out] state The current state.
+ *                 state&1: Recover all. state&2: Forget all.
+ * @param [out] asfd Location to store the file pointer to the autosaved file.
+ * @return true iff the file is to be recovered. If true, asfd will hold the
+ *         corresponding file pointer, which must be closed by the caller. If
+ *         false, asfd will hold NULL.
+ */
+static int ask_about_file(char *filename, int *state, FILE **asfd) {
```
frank-trampe: Do we really need to shuffle the names? If you're trying to avoid using asfd incorrectly, I think that renaming it is a better option.

fontforge/sfd.c
```diff
@@ -9142,63 +9142,81 @@ return( NULL );
...
-return( false );
-      default:
-      break;
+    ret = ff_ask(_("Recover old edit"),(const char **) buts,0,3,_("You appear to have an old editing session on %s.\nWould you like to recover it?"), pt);
+    switch (ret) {
+        case 1: //Recover all
+            *state = 1;
+            break;
+        case 2: //Skip one
+            fclose(*asfd);
```
frank-trampe: I don't like conditionally closing a file that isn't opened in this function. It seems like the sort of thing that would cause confusion later. Was there a particularly compelling reason to do this?

frank-trampe: Oh. I didn't see that we were now opening it in this function. I guess this is okay.

fontforgeexe/charview.c
```diff
@@ -272,7 +272,7 @@ static struct resed charview2_re[] = {
 };
 
 /* return 1 if anything changed */
-static int update_spacebar_hand_tool(CharView *cv) {
+static void update_spacebar_hand_tool(CharView *cv) {
```
frank-trampe: Why are we dropping the return values?

fontforgeexe/charview.c
```diff
@@ -5246,18 +5243,15 @@ static void CVMouseMove(CharView *cv, GEvent *event ) {
     GEvent fake;
     int stop_motion = false;
     int has_spiro = hasspiro();
-    int spacebar_changed;
```
frank-trampe: Why are we dropping this functionality?

gdraw/hotkeys.c
```diff
@@ -329,6 +329,10 @@ void hotkeysSave() {
     // Atomic rename of new over the old.
     //
     char* newpath = getHotkeyFilename(0);
+#ifdef __MINGW32__
+    //Atomic rename doesn't exist on Windows.
+    unlink(newpath);
+#endif
```
frank-trampe: How did you notice this?

frank-trampe: Ah. Just saw the commit comment.

inc/dynamic.h
```diff
@@ -39,10 +39,6 @@ extern const void *gwwv_NSAddImage(char *name,uint32_t options);
 #  elif defined(__MINGW32__)
 #  else
 #   include <dlfcn.h>
-#   ifdef __CygWin
```
frank-trampe: I have no idea what's happening here, but Cygwin isn't really my business anyway.

frank-trampe:
Okay. Almost everything looks good here. I have two open questions in the change comments.

fontforge/sfd.c
jtanx: Mostly it was just about ordering input parameters before output parameters.

fontforgeexe/charview.c
jtanx: :shrug: I'm pulling in the [Cygwin patch](http://www.geocities.jp/meir000/fontforge/). I know it correctly implements the behaviour that #6 didn't correctly re-implement (as mentioned earlier), so I guess this information isn't needed.

fontforgeexe/charview.c
jtanx: [As before](https://github.com/fontforge/fontforge/pull/2375#discussion_r33462163)

fontforge/sfd.c
frank-trampe: I don't prefer one philosophy to another, but a lot of the standard library functions put the output parameters first, and I wouldn't go to the trouble to change away from that.

frank-trampe:
I'm happy.

PR #2380, opened 2015-06-18
Travis: Homebrew fixes

DomT4:
Re: https://github.com/fontforge/fontforge/pull/2371#issuecomment-113025801

The PR to remove X11 support from GTK+ in `Homebrew/homebrew` was merged yesterday afternoon, but because your build scripts do `--with-x11` and the local Fontforge formula here wasn't resynced essentially Fontforge was expecting to find `pangoxft` which no longer exists.

I'm not quite sure what the packaging intention is with [the Homebrew app build script](https://github.com/fontforge/fontforge/blob/master/travis-scripts/create-osx-app-bundle-homebrew.sh). It seems to lean on Homebrew but at the same time uses MacPorts paths. Homebrew hasn't been correctly producing working `FontForge.app` forever though, so that would imply you're packaging it up in a different way rather than using a Homebrew-generated `FontForge.app`?

If I can grab some clarity into the packaging process I'm happy to help rewrite it to use older versions of Homebrew formulae, or to switch it to another package manager, or even build the required elements from source directly, but I'm not quite sure step-by-step what `create-osx-app-bundle-homebrew.sh` is intended to do.

This PR at least will stop your Travis builds dying on you during `configure`.

### DIFF
--- a/travis-scripts/fontforge.rb
+++ b/travis-scripts/fontforge.rb
@@ -1,6 +1,6 @@
 # based on https://raw.githubusercontent.com/Homebrew/homebrew/master/Library/Formula/fontforge.rb
 #
-# last synced by hand by DomT4 at 2015-05-10
+# last synced by hand by DomT4 at 2015-06-18
 
 class MyDownloadStrategy < GitDownloadStrategy
   # get the PR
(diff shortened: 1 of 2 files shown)


frank-trampe:
@DomT4, we currently satisfy all dependencies for the Mac .app via MacPorts. It would have been nice to be able to consolidate to HomeBrew, but we seem to be moving farther away from that rather than closer.

DomT4:
Thanks Frank. The X11 in GTK+ ship has sailed, but if there's anything else we can tackle to make life easier for you feel free to shout. Always happy to review an idea or request.

bagong:
With all due respect to MacPorts, Homebrew is an overwhelming success, and in the areas I have been observing it, has left MacPorts behind both in quality and update speed... Looking for an 10.6 compatible universal build of Qt4.8? You'll find it on Homebrew, along with the latest version building in the latest development environment, usually not more that a weak after the vanilla release. It would be lovely (for me :)) if the FF community could decide to switch full-heartedly to Homebrew.

jtanx:
With all the talk about the OS X builds not working, I butchered the Travis build script so that apparently all dependencies (that I can see) are satisfied via homebrew (https://github.com/jtanx/fontforge/commit/1768cf96df6ca5abcab5ed29b4f7868279841e30).

I got stuck at the packaging stage (create-osx-app.sh) and I must say that is some really arcane stuff - packaging on OS X is horrendous as compared to Windows!

Granted, I don't have a Mac to test anything on, so I have no idea if it works. @DomT4, it seems that homebrew Cairo accepts the `--with-x11` flag, but Pango does not - if I use a [custom formula](https://github.com/jtanx/fontforge/blob/1768cf96df6ca5abcab5ed29b4f7868279841e30/travis-scripts/pango.rb) that removes the hard-coded `--without-xft` then I can get FontForge to compile with X11 support again. Something that I still haven't figured out is why if I pass [this](https://github.com/jtanx/fontforge/commit/1768cf96df6ca5abcab5ed29b4f7868279841e30#diff-e2abd974a91922bf60a2713ccbbeb430R46) variable to the `depends_on` for Cairo/Pango (copied from the R formula), it doesn't work and installs the bottle(??), whereas if I hardcode the "with-x11" value, it compiles it from source as it should (I use [this line](https://github.com/jtanx/fontforge/blob/1768cf96df6ca5abcab5ed29b4f7868279841e30/travis-scripts/install_osx.sh#L11) to install dependencies)...

PR #2383, opened 2015-06-19
Fix configure.ac m4 syntax in homebrew PYTHON setup

### LINKED ISSUE #1836: y: command not found
I keep noticing this crop up in the ``` ./configure ``` readout, and I've been wondering if it's an actual bug that needs squashing or just a bit of nothing.

It's a fun little error that only displays for a few seconds before it gets drowned out by the rest of the ``` ./configure ``` process.

This is the error on the ``` ./configure ``` readout:

```
found python pkg_config information: /blah/blah/blah/blah/lib/etc
./configure: line 17616: y: command not found
```

That line inside the generated configure script is:

```
if test "y$HOMEBREW_BREW_FILE" != "y"; then
   echo "detected a homebrew build environment"
   PYLIBDIR=$(${HOMEBREW_PREFIX}/bin/python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
   PYPKGCONFIG="$( cd "$PYLIBDIR/../../pkgconfig" && pwd )"
   export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:$PYPKGCONFIG"
   echo "found python pkg_config information: $PYPKGCONFIG"
   if  "y$PYTHON" = y  ; then
     export PYTHON="${HOMEBREW_PREFIX}/bin/python"
   fi
fi
```

The line is the: ``` if  "y$PYTHON" = y  ; then ``` line.

Any ideas what's going on here? My curiosity has gotten the better of me and I had to ask.

apjanke:
Fixes #1836 

This fixes a problem with the m4 syntax in the homebrew-specific section of `configure.ac`. The impact of this error was that `PYTHON` was never getting set because the test was failing due to trying to run the bogus command `y` instead of doing a string comparison.

Here's the difference I see in my `./configure` output on OS X 10.9 under a Homebrew environment.

```
$ diff configure-old.out configure.out
158,164c158,162
< ./configure: line 18119: y: command not found
< checking for a Python interpreter with version >= 2.7... python
< checking for python... /usr/local/bin/python
< checking for python version... 2.7
< checking for python platform... darwin
< checking for python script directory... ${prefix}/lib/python2.7/site-packages
< checking for python extension module directory... ${exec_prefix}/lib/python2.7/site-packages
---
> checking whether /usr/local/bin/python version is >= 2.7... yes
> checking for /usr/local/bin/python version... 2.7
> checking for /usr/local/bin/python platform... darwin
> checking for /usr/local/bin/python script directory... ${prefix}/lib/python2.7/site-packages
> checking for /usr/local/bin/python extension module directory... ${exec_prefix}/lib/python2.7/site-packages
```

With this patch applied, FontForge builds and installs for me with `brew install fontforge`. (It did before too; just checking for regression.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -220,7 +220,7 @@ if test "y$HOMEBREW_BREW_FILE" != "y"; then
    PYPKGCONFIG="$( cd "$PYLIBDIR/../../pkgconfig" && pwd )"
    export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:$PYPKGCONFIG"
    echo "found python pkg_config information: $PYPKGCONFIG"
-   if [ "y$PYTHON" = y ] ; then 
+   if test "y$PYTHON" = "y"; then 
      export PYTHON="${HOMEBREW_PREFIX}/bin/python"
    fi
 fi


apjanke:
The Travis build failed on OS X due to a missing pango dependency. I don't think this is a regression introduced by this PR, because I looked at a couple of other PRs and their OS X Travis builds are failing due to the same error. I believe #2380 explains and fixes this.

```
checking for PANGO... no
configure: error: Package requirements (pango >= 1.10 pangoxft) were not met:
No package 'pangoxft' found
Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.
Alternatively, you may set the environment variables PANGO_CFLAGS
and PANGO_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
```

frank-trampe:
Looks good.

Issue #2384, opened 2015-06-19
Closing expand stroke dialog with esc -> crash


larsenwork:
I haven't figured out how to do a proper crash rapport on Mac but you might be able to reproduce. 

**Always**
Open Outline Window -> open expand stroke dialog -> press esc -> crash

**Sometimes**
Open Outline Window -> open expand stroke dialog -> press "cancel" button -> crash

jtanx:
Duplicate(?) of #2029

larsenwork:
@jtanx duplicate yes :oops: looks like you've already fixed? :smile:

larsenwork:
closing as duplicate of #2029 - haven't tested it yet since we are having mac building difficulties :)

Issue #2029, opened 2015-01-03
Expand stroke bug


m93a:
How to reproduce:
- select an element
- click on Element → Expand Stroke
- set Main Stroke Width and Minor Stroke Width to zero
- hit Apply

What should happen:
- an error message should appear
- the Main and Minor Stroke Width values should be reverted to default

What happens:
- the app crashes
- all work is lost

larsenwork:
Expand stroke bug 2

How to reproduce: 
* Element -> Expand Stroke
* Cancel (either press escape or the cancel button)

What happens:
* Crash

Gist of crash report: https://gist.github.com/larsenwork/73c58e60f44046f40a40

jtanx:
The 'zero stroke width' segfault:

~~~~~
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff58e16cc in HideStrokePointsCircle (c=0x7ffffffc1760)
    at splinestroke.c:809
809		    if ( op->me.x==p->me.x && op->me.y==p->me.y )
(gdb) bt
#0  0x00007ffff58e16cc in HideStrokePointsCircle (c=0x7ffffffc1760)
    at splinestroke.c:809
#1  0x00007ffff58e2151 in FindStrokePointsCircle (ss=0x9fbcd0, 
    c=0x7ffffffc1760) at splinestroke.c:945
#2  0x00007ffff58f1c5e in SplineSet_Stroke (ss=0x9f9b10, c=0x7ffffffc1760, 
    order2=0) at splinestroke.c:3596
#3  0x00007ffff58f1d77 in SplineSets_Stroke (ss=0x9f9b10, c=0x7ffffffc1760, 
    order2=0) at splinestroke.c:3622
#4  0x00007ffff58f2264 in SplineSetStroke (ss=0x9f9b10, si=0x7ffffffc1950, 
    order2=0) at splinestroke.c:3685
#5  0x00007ffff78c4f29 in CVStrokeIt (_cv=0x9085e0, si=0x7ffffffc1950, 
    justapply=0) at cvstroke.c:84
#6  0x00007ffff78c5be2 in _Stroke_OK (sd=0x7ffff7d85960 <strokedlg>, isapply=0)
    at cvstroke.c:258
#7  0x00007ffff78c5ca0 in Stroke_OK (g=0x9e6f70, e=0x7ffffffc2ad0)
    at cvstroke.c:268
[...]

PR #2385, opened 2015-06-19
Fix a few expand stroke bugs

jtanx:
This addresses a couple of issues outlined in #2029 and #2384. 

* Fix crash when pressing 'esc' in the expand stroke dialogue
* Fix crash when trying to expand stroke with a stroke width of 0

@larsenwork I couldn't reproduce the crash that occurred sometimes when pressing the cancel button.

cc? @frank-trampe

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -3933,7 +3933,7 @@ static void CVCharUp(CharView *cv, GEvent *event ) {
 	    TRACE("was on charselector\n");
 	    GWidgetIndicateFocusGadget( cv->hsb );
 	}
-	else
+	else if ( cv->charselector != NULL )
 	{
 	    TRACE("was on NOT charselector\n");
 	    GWidgetIndicateFocusGadget( cv->charselector );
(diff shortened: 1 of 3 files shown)


larsenwork:
@jtanx "cancel" crash happens only when I press it very quickly after dialog appeared (muscle memory from other programs: cmd+shift+e = export) so I'd guess it's a memory issue somehow. But it's a non-frequent minor issue.

fontforgeexe/charview.c
```diff
@@ -3933,7 +3933,7 @@ static void CVCharUp(CharView *cv, GEvent *event ) {
 	    TRACE("was on charselector\n");
 	    GWidgetIndicateFocusGadget( cv->hsb );
 	}
-	else
+	else if ( cv->charselector != NULL )
```
frank-trampe: So there was a crash when cv->charselector was NULL? Why would cv->charselector be NULL in this context?

jtanx: @frank-trampe it's explained in the [commit message](https://github.com/jtanx/fontforge/commit/074fe7586fc50dd3b04288fd2ab11dc02e2f9464).

frank-trampe:
Looks good.

PR #2387, opened 2015-06-20
Fix progress bar display on Windows.

jtanx:
The font instance was not being set on the root window. This fixes a bug in how the progress window was being displayed since Cairo support was enabled.

Before:
prog-old

After:
prog-fix


frank-trampe:
Looks good.

PR #2349, opened 2015-06-09
parsettfaat: allocate memory as needed

adrientetar:
Closes #2348.

PR #2392, opened 2015-06-25
Revert "parsettfaat: allocate memory as needed"

adrientetar:
Reverts fontforge/fontforge#2349.

Not the right fix.


frank-trampe:
Looks good.

Issue #2143, opened 2015-03-14
Enable space in hotkeys (so it can be used for hot preview)


davelab6:
When holding `space` to pan around, it would be nice to activate the fill while held, similar to how it is currently activated with ``` ` ``` on a US keyboard, which is often much harder to access on european keyboards - holding down ``` ` + space ``` gives the effect of what I would like with only space - and this in addition to the regular fill preview. This is a simple way to get better International keyboard support.

However, `space` isn't possible to use in the hotkeys file.

jtanx:
@davelab6, slightly off-topic but #2375 changes the behaviour slightly of the spacebar pan activation. In particular, it reverts it back to 2012-era behaviour of only panning when you click and drag vs. immediately dragging when the spacebar is activated.

I presumed this was because #6 never implemented the previous behaviour correctly, but is this actually intended behaviour?

davelab6:
I think it's not very important; the way it works in 2012/now is like
Photoshop and how I changed out was to be like gimp is, which I prefer. But
if the preview mode is to be triggered with space bar as well as hand pan
tool, then reverting to the PS style is ok for me

anthrotype:
:+1: for activating the fill while holding the space bar

pathumego:
+1

anthrotype:
I checked out @jtanx's patch #2393 and it works great!
I was also wondering, would it be possible to allow users to disable the default spacebar behaviour whereby the hand tool starts moving the canvas as soon as the spacebar key is pressed? I guess many other users who come from other font editors, or regularly switch between them, are more used to a  spacebar *plus* mouse-click hand tool...
Thanks anyway.

davelab6:
> would it be possible to allow users to disable the default spacebar behaviour whereby the hand tool starts moving the canvas as soon as the spacebar key is pressed?

#2375 does change the behavior this way

davelab6:
Just tested this in https://github.com/fontforge/fontforge/releases/download/20160403/FontForge-Mac-20160403.zip and its awesome!!!

PR #2393, opened 2015-06-25
Allow the spacebar to enable preview mode.

jtanx:
Related issues: #2143.

A new preference, 'ShowFillWithSpace' allows this behaviour to be toggled by
the user. It is enabled by default.

I will need to update #2375 if this gets merged first or vice versa.

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -79,6 +79,7 @@ int DrawOpenPathsWithHighlight = 1;
 #define default_cv_height 540
 int cv_width = default_cv_width;
 int cv_height = default_cv_height;
+int cv_show_fill_with_space = 1;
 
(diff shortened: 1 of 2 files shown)


frank-trampe:
I thought that there was a different keystroke for this already.

jtanx:
Yes, the backtick (`), but @davelab6 suggested that it would be good for this to also happen when the spacebar is pressed, as described in #2143.

davelab6:
The backtick is often not a key on european keyboards, and was done to be compatible with FontLab; but now Glyphs is the most popular editor and sensibly uses the space key for this, so we should update to use both.

jtanx:
I have rebased the commit.

frank-trampe:
Looks good.

PR #2405, opened 2015-07-03
OSX shell fixes

glyphobet:
This fixes a few annoying problems when trying to launch FontForge on Mac OS X from the command line a la:

     $ /Applications/FontForge.app/Contents/MacOS/FontForge ~/path/to/Font.sfd

Previously, this would open the font *and* a new, "Untitled" font window. This changes the behavior so that FontForge only creates a new font when no font to open was specified.

Previously, this also only worked if the path to the file to open was an absolute path. If you passed a relative path, FontForge would pop up a dialog saying "The requested file, Font.sfd, does not exist" and would not open the font. That was because of the `cd ~` in this script. Now the script only changes to the `~` directory if it's invoked via normal Mac OS X methods (double clicking on the application, an SFD file, or launching via Spotlight, for example). If the application is invoked via the command line, it assumes the user knows what they are doing and won't change the working directory, which also means you can pass relative paths on the command line and FontForge will open them no problem. So the following now works:

    $ cd ~/path/to/
    $ /Applications/FontForge.app/Contents/MacOS/FontForge Font.sfd

Tested on Mac OS X 10.9 Mavericks.

These changes mean that you can now put an alias in your `.bashrc` to invoke FontForge from the command line:

    alias FontForge='/Applications/FontForge.app/Contents/MacOS/FontForge'

### DIFF
--- a/osx/FontForge.app/Contents/MacOS/FontForge
+++ b/osx/FontForge.app/Contents/MacOS/FontForge
@@ -124,4 +124,9 @@ cd ~/ || true
 
 export PATH="$PATH:$bundle_bin"
 echo $WRAPPER $bundle_bin/fontforge > /tmp/oo
-$WRAPPER $bundle_bin/fontforge -new "$@" &
+
+if [ $# -eq 0 ] ; then
+    ( exec $WRAPPER $bundle_bin/fontforge -new )
+else
+    ( exec $WRAPPER $bundle_bin/fontforge "$@" )
+fi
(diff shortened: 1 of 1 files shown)


glyphobet:
Updated the pull request to fix the relative vs. absolute path issues.

frank-trampe:
@glyphobet, the ampersand may be unnecessary now that we're using exec. Would you test?

glyphobet:
@frank-trampe already tested with and without the `&`. It does appear to be necessary despite the presence of `exec`.

frank-trampe:
As in the dock icon keeps bouncing?

glyphobet:
@frank-trampe yes, exactly. Without the `&`, the dock icon appears to bounce indefinitely. With the `&`, the script launches FontForge inside XQuartz and then quits, which seems to terminate the FontForge.app, and the dock icon disappears.

glyphobet:
Any chance of this getting merged anytime soon? It would be great to see it included in the next OS X build.

frank-trampe:
Sorry about that. There wasn't a blocking issue; I just forgot about it.

Issue #2401, opened 2015-07-01
Random files are generated, and become unresponsive


enathu:
When working on a font, FontForge keeps storing new files with the extension of *.sfd-[number] in the same directory - Here the [number] is a variable but serial numbers. After some point, FontForge becomes unresponsive. Usually I need to end task using Task Manager, and restart FontForge to work again.

I am using Windows 8.1, and 2015-06-12 FF Release. Here is the screenshot of how it looks with the files created by FF. ![fontforge](https://cloud.githubusercontent.com/assets/62147/8455333/bd368afc-1ffb-11e5-9b8c-8bb9323b79a5.PNG)

Is it anything to do with AutoSave? or any fix available to stop this file generation?

Thank you.

jtanx:
Strange. Do you know if this happened with any of the older versions of FontForge?

enathu:
In older versions, I experienced crashes repeatedly. But haven't seen any file generation like this. I only saw this file generation with 2015-06-12 FF version.

davelab6:
The Revisions to Retain preference is here, set it to 0:

![screen shot 2015-07-01 at 9 52 08 am](https://cloud.githubusercontent.com/assets/261579/8456053/e0061220-1fd6-11e5-84dd-d3da6b4e2685.png)

enathu:
Hi @davelab6 I have set it to 0 and tried with a new file, and it keeps generating the files. So, I went to the preferences and checked, but it showed 32. So, again I set RevisionsToRetain to 0 and the files started to generate. And the 0 now became 32 in the preferences. The cycle continues. Any idea, please?

jtanx:
Do you have any non-English/non-ASCII characters in your install path?

jtanx:
Also the preferences box is a bit strange - how are you saving the preferences? If you click on the 'X' button it will not save preferences - you have to press 'enter' to save

davelab6:
When you change the setting in the UI, quit FF and check the prefs file, and see if it is `0` there

enathu:
@jtanx No, I haven't got any non-English/non-ASCII characters in path.

Oh, that's strange. I really thought that 'Preferences' works like the Glyphs app, 'Font Info' dialog box, and I did save by clicking on the 'X'. So, now I used the Enter key to save it. So, it isn't saving any files in the there. Thank you for the enlightenment.

Do you think, does this file generation directly related to the unresponsiveness of FF? I'll keep you posted on this as I am going to work on FF, and see.

[...]

PR #2407, opened 2015-07-05
Update prefs

davelab6:
Hack to fix https://github.com/fontforge/fontforge/issues/2401#issuecomment-118649636

### DIFF
--- a/share/prefs
+++ b/share/prefs
@@ -12,3 +12,4 @@ SnapOutlines: 0
 SnapDistance: 2
 MarkExtrema: 1
 OutlineThickness: 3
+RevisionsToRetain: 0


adrientetar:
I do not want to see this enabled by default, personally;

davelab6:
It is already enabled by default, has been for a couple of years; this PR turns it off.

adrientetar:
Sorry, I meant set to 0.

frank-trampe:
Is this discussion over? I don't follow.

davelab6:
This is a work around because the feature causes unprompted crashes on
windows. I doubt a fix will be soon. I'm happy to see it reenabled after
crashes are fixed :)

Issue #2390, opened 2015-06-24
Crash when using Spiro mode


jtanx:
If you add enough points in Spiro mode it will eventually crash. The warnings box says `Spiros did not converge`. I think it's a use-after-free issue but I don't know the exact origin of where this is ocurring.

~~~~~
==20620== Invalid read of size 8
==20620==    at 0x4EE77FA: CVInfoDrawText (charview.c:4086)
==20620==    by 0x4EE80CF: CVInfoDraw (charview.c:4147)
==20620==    by 0x4F22273: CVMouseDownSpiroPoint (cvaddpoints.c:408)
==20620==    by 0x4F223C0: CVMouseDownPoint (cvaddpoints.c:454)
==20620==    by 0x4EEB72F: CVMouseDown (charview.c:5024)
==20620==    by 0x4EEEBE2: v_e_h (charview.c:5735)
==20620==    by 0x6135232: _GWidget_Container_eh (gcontainer.c:403)
==20620==    by 0x61B3163: dispatchEvent (gxdraw.c:3296)
==20620==    by 0x61B36FB: GXDrawEventLoop (gxdraw.c:3404)
==20620==    by 0x6139A7A: GDrawEventLoop (gdraw.c:791)
==20620==    by 0x50F3872: fontforge_main (startui.c:1382)
==20620==    by 0x4007DC: main (main.c:39)
==20620==  Address 0xfdcfbf0 is 192 bytes inside a block of size 240 free'd
~~~~

I had a backtrace for this but I lost it and it seems to be somewhat difficult to reproduce.

--------

This one may be a separate crash (after deleting some points, `spline->to` is NULL). This one seems more reproducible - e.g. make some points, delete some in-between then try to add another point.
~~~~
#0  0x00007ffff78c4ed3 in CVClearSel (cv=0x905870) at cvpointer.c:162
#1  0x00007ffff78c792f in CVMouseDownPointer (cv=0x905870, fs=0x7fffffffbb30, 
    event=0x7fffffffbdc0) at cvpointer.c:664
#2  0x00007ffff784d5e7 in CVMouseDown (cv=0x905870, event=0x7fffffffbdc0)
    at charview.c:4983
#3  0x00007ffff7850be3 in v_e_h (gw=0x964c70, event=0x7fffffffbdc0)
    at charview.c:5735
#4  0x00007ffff683b233 in _GWidget_Container_eh (gw=0x964c70, 
[...]

PR #2415, opened 2015-07-09
A few Spiro fixes

jtanx:
Related issues: #2390 

I think quite a few Spiro related crashes were the result of the fairly recent change made by a692c52.

* Fix crash/weird behaviour when deleting Spiro points (use after free issue; a692c52)
* Fix crash when deleting points and when the Spiro curve 'failed to converge' (e.g. where you see straight lines instead of Spiro curves)
* Fix the point info dialog (when you right-click on a control point and go to `Get Info...`) Ok/Cancel functions. (Before `Ok` did nothing for Spiro points, and I believe that `Cancel` didn't actually cancel/revert state for both Spiro and normal points)
* Don't display the Spiro end point when it 'fails to converge' e.g.
zeropt

cc @frank-trampe


fontforge/spiro.c
```diff
@@ -231,7 +262,7 @@ void SSRegenerateFromSpiros(SplineSet *spl) {
 	SplinePoint *sp, *first, *last;
 	if ( (last=first=SplinePointCreate(spl->spiros[0].x,spl->spiros[0].y))==NULL )
 	    return;
-	for ( i=1; i<spl->spiro_cnt; ++i ) {
+	for ( i=1; spl->spiros[i].ty != SPIRO_END; ++i ) {
```
jtanx: This addresses the last point I made in the PR comments

fontforge/splineutil.c
```diff
@@ -1494,7 +1493,6 @@ static SplinePointList *SplinePointListSplitSpiros(SplineChar *sc,SplinePointLis
 	    memset(temp+(i-start),0,sizeof(spiro_cp));
 	    temp[i-start].ty = SPIRO_END;
 	    cur = SpiroCP2SplineSet( temp );
-	    free( temp ); temp = NULL;
```
frank-trampe: What keeps the reference to temp? Does it get freed somewhere else?

jtanx: `SpiroCP2SplineSet` sets `cur->spiros=temp` (i.e. `temp` is used in the return value)

frank-trampe:
Looks good.

PR #2420, opened 2015-07-11
Use CFF over PFB as intermediary for FT Rasterisation. Fixes #685

### LINKED ISSUE #685: FreeType rasterization doesn't work, and has been made mandatory
This is a priority issue for me because it's blocking the Tsukurimashou 0.8 release, planned for tomorrow.  Encountered in the latest git master version of FontForge, under 64-bit Linux.

Same problem I reported two years ago to the mailing list (http://sourceforge.net/mailarchive/forum.php?thread_name=20111125.100615.444195158.wl%40gnu.org&forum_name=fontforge-devel ):  when FontForge uses FreeType to generate bitmaps, it sometimes produces complete garbage.  The bitmaps, if saved to BDF, are mostly empty, with just a few containing large irregular black areas.  If I generate a bitmap strike in the GUI and then try to look at it in the bitmap window, it sometimes segfaults, suggesting in-memory data structures contain garbage.

The file I posted for that bug report is no longer online, but I've posted a fresh one at https://github.com/fontforge/debugfonts/blob/master/jieub-bt-ps.sfd

In the GUI, there is an option during bitmap generation to use FreeType or not; but native scripting doesn't offer such a choice and will always attempt to use FreeType.

Werner Lemberg found and fixed an unrelated bug in FreeType in response to my report (see the mailing list thread) - apparently FreeType couldn't *read* BDF files with large indices.  But FontForge does not use FreeType for BDF files; that bug is unrelated to mine and only happened to also be noticeable from the test case I provided.  He disclaimed responsibility for my actual bug, which is specific to the case where FontForge generates a bitmap using FreeType.

In 2011, a workaround for me was to build FontForge without FreeType.  But pull #157 made that impossible.

At this point it looks to me like my best option may be to dig into the FontForge source and modify the hardcoded "use FreeType" setting to "no,"  because I need this to work *now*.  But in the longer term, we should figure out what's actually wrong with the FreeType rasterization and fix it, or convince Werner that it is his business to do so.

jtanx:
Seems to work.

@mskala it would be great if you could have a quick look to see if I've missed anything.

cc @frank-trampe

### DIFF
--- a/fontforge/freetype.c
+++ b/fontforge/freetype.c
@@ -212,7 +212,7 @@ return( NULL );
 	    /* like the metricsview, which may expect the full glyph set */
 	    /* available, and will crash if that is not the case */
 	    /* So if anything needs autohinting, do it now */
-	    if ( (ff==ff_pfb || ff==ff_pfa || ff==ff_otf || ff==ff_otfcid ) &&
+	    if ( (ff==ff_pfb || ff==ff_pfa || ff==ff_otf || ff==ff_otfcid || ff==ff_cff) &&
 		    autohint_before_generate ) {
 		extern int preserve_hint_undoes;	/* users don't expect that rasterizing the glyph will cause an undo */
 		int phu = preserve_hint_undoes;		/* if metrics view & char view are open, and a change is made in */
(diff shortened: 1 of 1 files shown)


mskala:
Yes, it fixes the problem for me.  Thanks!

PR #2426, opened 2015-07-14
Replace `rm -rf` with more portable code. Fixes #2424.

### LINKED ISSUE #2424: Font Generation: When overwriting UFO, save fails
When generating the font (File> Generate Fonts), UFO generation works fine. But, let's say I made some changes in the font and want to generate the latest UFO by overwriting the previously generated UFO in the same path. Obviously, FF says the file already exists and asks whether to replace it or not.

ff-replace

When selecting the Replace option. FF shows a dialog box with 'Save Failed'

ff-ufo-save-fail

I have tried with other font formats, but others are working fine as intended. Only the UFO generation has this issue. 

I am using latest FF on Win 8.1 & I also tried with Win7 and FF having the same issue.

Is there any fix for this? 

Thank you.

jtanx:
Using `rm -rf` is not portable, especially given that it doesn't exist on Windows, unless you are running through CygWin.

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -3043,8 +3043,7 @@ return;
 	    else
 		sprintf( markerfile,"%s/" FONT_PROPS, buffer );
 	    if ( !GFileExists(markerfile)) {
-		sprintf( markerfile, "rm -rf %s", buffer );
-		system( buffer );
+		GFileRemove(buffer, false);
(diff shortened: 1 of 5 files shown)


frank-trampe:
I find it amusing that we were doing that in the first place.

I like the way you've encapsulated the outside GLib calls.

PR #2445, opened 2015-08-06
configure.ac: intelligently check for <execinfo.h>

blueness:
Currently, it is assume that only MINGW32 and CYGWIN on windows do
not provide backtrace.  However, other libc's in the Linux world
do not provide it either.  uClibc only provides it if configured
to do so, and musl doesn't provide it at all.  As a result, the
build fails on those systems.

This patch checks for <execinfo.h> in configure.ac and skip its
inclusion in fontforge/cvundoes.c if it is not available.

Signed-off-by: Anthony G. Basile <blueness@gentoo.org>


JoesCat:
Thanks!

Issue #2454, opened 2015-08-16
Update Python documentation for undefined math constants


fred-wang:
(cc'ing @khaledhosny)

The documentation for the Python interface mentions at several places "The special value of -32768 (0x8000)": https://github.com/fontforge/fontforge/blob/master/doc/html/python.html

Actually, these special values correspond to TEX_UNDEF which is 0x7fff, so the doc should be updated
https://github.com/fontforge/fontforge/blob/176a172222df093902e25ad4792fa3e2afa69825/fontforge/splinefont.h#L1534

(it may also be good to expose a python constant "TEX_UNDEF")

JoesCat:
safe to close issue?

fred-wang:
I think so.

PR #2459, opened 2015-08-21
Python API: expose a constant fontforge.unspecifiedMathValue to descr…

fred-wang:
…ibe unspecificied math values and fix the documentation. #2454 

(cc'ing @khaledhosny)

### DIFF
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -1858,7 +1858,7 @@ pen = None;				# Finalize the pen. This tells FontForge
     <TR>
       <TD><CODE>italicCorrection</CODE></TD>
       <TD colspan=2>The glyph's italic correction field. Used by both TeX and MATH.
-	The special value of -32768 (0x8000) means the value is unspecified (An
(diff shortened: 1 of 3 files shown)


frank-trampe:
I don't quite understand. Were we inconsistently using 0x8000 in some places and 0x7FFF in others? And is TEX_UNDEF constant across platforms such that an undefined value saved to disk on one machine (either in an S. F. D. or a U. F. O.) would read as an undefined value on any other machine?

fred-wang:
AFAIK: we only use TEX_UNDEF internally for all these math values and this tells fontforge not to save the value (for example if the italic correction is not set for a glyph then it will just not appear in the MathItalicsCorrectonInfo table http://www.microsoft.com/typography/otspec/math.htm). So the actual value does not matter, but it is defined as 0x7FFF in the C code while the HTML documentation says it is 0x8000.

frank-trampe:
After grepping for TEX_UNDEF, I have a better understanding of what you're fixing. It looks good to me.

PR #2466, opened 2015-08-27
Windows: Add `tmpfile` to bootstrap.conf.

### LINKED ISSUE #2465: Font Generation Fails Completely
I have updated the FontForge with [latest build release for Windows](https://github.com/fontforge/fontforge/releases/download/20150824/FontForge-2015-08-24-Windows.exe), and with this new release when I trying to generate font (all formats - ttf, otf and other) it shows the message box "Save Failed" -

save-failed

Once I click 'OK' on the "Save Failed" message and cancel the Font Generation dialog box, it shows this error in a message box. 

error-1

I've also, tried with different and brand new source files to get generate the font, and got the same error with subtle number changes in the message box.

error-2

I have inspected the destination folder, where I tried to generate the font, I've noticed that FontForge created a file with the font extension each time I was trying to generate a font, but all the files are in the size of 1KB - and when I try to open the file, it shows "The requested file fontfile.ttf is not a valid font file." as an error message. 

The latest build completely fails in generating font.

Any fix for this? I look forward to hearing from you. Thank you.

jtanx:
`tmpfile` generates temporary files in the root (e.g. C:\) directory, which
normally will fail for non-administrators. The gnulib implementation fixes
this by not trying to generate temporary files in the root directory but
it a more sensible location.

I'm still unclear as to how this managed to work in the past.

Fixes #2465.  cc @frank-trampe

### DIFF
--- a/bootstrap.conf
+++ b/bootstrap.conf
@@ -51,6 +51,7 @@ gnulib_modules='
 	strcase
 	strtok_r
 	strndup
+	tmpfile
 	xalloc
 	xvasprintf
 '


frank-trampe:
It looks good to me.

PR #2467, opened 2015-08-27
Support loading encondings from a GlyphOrderAndAliasDB file

felipesanches:
Also did some cleanup and refactoring on the functions related to this portion of the code.

Issue #1502, opened 2014-07-11
Run entire codebase through GNU Indent?


davelab6:
@adrientetar suggest we should run the entire C codebase through GNU Indent, because he says right now it's annoying to work with existing code.

I note that there's been some protest against that as it would lose git blame but perhaps the tradeoff is worthwhile :)

Adrien noted that one can ignore whitespace change in blame with the `-w` switch, but GNU Indent would also change the coding style a bit (i.e., get rid of returns with parenthesis) and GitHub's blame view - eg https://github.com/googlefonts/fontbakery/blame/1ffef8f0c5ec72e5114450a209a7f4eaa79cb37e/CONTRIBUTORS.txt - doesn't use that switch. 

Thoughts?

Issue #830, opened 2013-10-07
define a codingstyle


lynxis:
The project should define a codingstyle. It doesn't matter which one, but it must easy to understand and improve readability of the code. At the moment there are lot of different coding styles in the code.

To start a discussion about the needs of this project, I suggest the linux [kernel coding style](https://www.kernel.org/doc/Documentation/CodingStyle)

skef:
An email on February 19 2021 had this conclusion:

> On 2/19/21 6:33 AM, Frank Trampe wrote:
> > That works for me.
> >
> > On Fri, Feb 19, 2021, 04:21 T J <jtanx@outlook.com> wrote:
> >
> > If we can agree on one, then sure. I mean all my added files follow the same code style (e.g. ggdkdraw.c), I even committed the astyle command I ran in one of the commit messages:
> >
> > `astyle --style=java --pad-oper --unpad-paren --pad-header --indent=spaces --add-brackets --align-pointer=name`
> >
> > It tends to follow the general style of FF code, except no mixed tabs/spaces and no undented returns.

So, basically, mostly keep doing what has been done except where it comes to indentation, which should be done with spaces and without the shallow return statements. 

I've tried to codify this in a brief code style wiki here: https://github.com/fontforge/fontforge/wiki/Code-Style . Feel free to modify, extend, or delete as needed. Closing this.

PR #2468, opened 2015-08-27
Overall code cleanup

felipesanches:
Warming up the work day... Decided to review a few functions while cleaning up their code-style to make them more legible.

No rush to merge this as I may still add a few more cleanup commits here. Anyway, It should only be merged after pull request #2467 because it was done on a branch of that branch :-)

### DIFF
--- a/fontforge/encoding.c
+++ b/fontforge/encoding.c
@@ -238,7 +238,7 @@ static int TryEscape(Encoding *enc, const char *escape_sequence) {
     return enc->has_2byte;
 }
 
-Encoding *_FindOrMakeEncoding(const char *name,int make_it) {
+Encoding *_FindOrMakeEncoding(const char *name, int make_it) {
     Encoding *enc;
     char buffer[20];
     const char *iconv_name;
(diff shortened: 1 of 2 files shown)


cicku:
+1

JoesCat:
Hi Felipe,
A re-sorting of code somewhat white-washes all the git blame code, which from my point of view isn't really helping improve FontForge, and unfortunately, different programmers have different tastes in coding styles, so it can be quite possible for someone to come in later and "cleanup" code according to their tastes in style. For myself, I tend to use the old-code style as markers to point-out code that needs further analysis, so when you "clean" code, it changes the style of the code and makes it look like it was possibly improved, but, unfortunately, resorting, isn't fixing or improving...
If you really intend to improve code, I would highly recommend, looking at each routine, and asking the question what would each statement do, and have you handled all conceivable possibilities of failures?
For suggestions, I'll start with the first file in this list here:
"Fixing indentation / code-style of the ParseEncodingFile function..."
very first line....ask the question...what happens if this function fails?
at the moment, the routine returns void regardless of if it works or does not, so... ..maybe better to let the caller-routine know that the routine can fail...therefore, change the function from....
void RemoveMultiples(Encoding *item) {
to
int RemoveMultiples(Encoding *item) {
(for int, I prefer 0=pass, -1=fail or less for other possible errors, and -1 works well if you use ! to test for opposite of a return value, examples are various IO routines that tend to use 0==pass).

next question, next line down... if (test)
note that, NULL can be zero, but not all operating systems describe NULL as zero.... did the calling function convert a failure to zero, or did it return NULL? If you are going to test for if (test), then I recommend that you also set the value to zero as well to ensure it really is zero...

...then keep going through each line.... for example, a more fortified file error check is to do:
if (ch<0) {
instead of
if (ch==EOF) {
because errors can also happen too, such as file locked, file lost (eg: SDcard), etc...

...last but not least, we assume functions work, but let's assume the possibility that functions fail at any point in the routine....please do a housecleaning, memory cleanup as you back-out of routines. tricky ones are things like realloc where you lose prior pointers if the realloc fails....

long story short, please review the routines you propose to cleanup, and apply code-hardening whenever and wherever possible. If not, then please just leave the routines as-is unless you find bugs that need fixing. If you want an example, look at gutils/gimagereadrgb.c where there are very few spots left over for possible segfaults or various file errors.

I hope this makes sense, and I hope it convinces you to go beyond simply "just" formatting changes.
Thanks in advance (if you do improve code).

mskala:
If I may say so, I have a differing opinion.

George's wacky code layout costs developer time and increases the risk of future bugs whenever anyone needs to touch the code in question.  It should be regarded as a bug in itself.  Code in a wacky layout is broken even if it compiles into machine language that works.  Spending new time imitating the wacky layout in new or newly touched code is counterproductive; writing new code in a more reasonable style, fixing old code as it's touched, and even going out and pre-emptively fixing broken code that superficially appears okay because it compiles into working software, are all worthwhile activities.

The git blame issue doesn't convince me because when someone edits code and git blame says so, then git blame is doing exactly what it's supposed to.  Leaving code broken just because you don't want to have a record of the fact that you fixed it, doesn't seem like a good trade-off.  Saying that one wants git blame to record other edits but not indentation edits is saying that indentation edits aren't important, and I simply don't agree.

When I contribute to FontForge I attempt to imitate the layout of the existing code because that's the project's policy and it's not up to me to change it, but it's not what I think the policy should be, nor the policy I set in other projects where it _is_ my call.  And it does reduce the possibility for code I write elsewhere to be contributed to FontForge, and that also becomes part of the cost to the project of maintaining this policy.

George isn't coming back.  Is the "preserve George's formatting" policy really worth what it costs?

jtanx:
I agree that formatting the code in a saner way needs to be done, but I would prefer for it to be done all in one go, possibly using some sort of formatting tool (and sticking to that standard once it's done).

@felipesanches are there any logical changes to the code in this PR, or is it solely formatting related?

JoesCat:
@mskala, I think you misunderstood what I meant. I'm not suggesting to maintain the code style, I'm asking, if you look at the code to improve it, please go ahead and improve it, but don't just simply do a reformat, since it gives a false sense of improvement. Take a look at the last pull-request I've added, the coding style isn't following George's, and the various gutil/gimage*.* files I've looked at don't follow George's odd return/exit indentations either. Other coders also apply their own style too, like Ben and others did.
If we're going to hard-code a style policy, we could let the make system just reformat every single routine on it's own...but...in my opinion, it makes old code (that can use improvement) very hard to identify....I prefer to see code formatting applied when logical improvements are made.

mskala:
It's certainly better to do it in an automated way all at once rather than a little bit at a time.  One of the first things I did in FontAnvil was run the whole works through GNU indent.  Unfortunately, the output of that isn't perfect - it didn't do everything exactly the way I wanted - so I'm still making small changes by hand whenever convenient (usually, but not always, along with other changes).

JoesCat:
@mskala - different strokes for different folks  ;-)
If we are to work together, we need to acomidate the different coding styles and requests.

no compassion, no peace,
know compassion, know peace.

For myself, for example,, I prefer to see as much code in a screenshot as I can instead of scrolling through 3 screenshots to see 5 lines of code, but I tolerate the fact other developers prefer 1 line of code per line.
If you want to run each routine you've improved through gnu formatting, then go ahead, but please do that if you've bullet-proofed the routine as well as you could, and please avoid doing such a thing if all you've done is simply changed 1 line in a 100-line function (that could take a better look-at to ensure it's bullet-proof). If you've looked at the entire routine, and it looks bullet-proof with that 1 line change, then go ahead and gnu format it, but if you haven't looked at the rest of the routine, then just leave the code you haven't looked at, as-is ...but, that's just my request.

felipesanches:
Except for a few reasonable comments above, the overall rationale behind this policy seems non-sense to me. This almost feels like Stockholm Syndrome.

And you're also probably losing code contributors because of that.

jtanx:
This has been a recurring issue; see #830 and #1502. 

I'll just reiiterate this, but I'm all for reformatting the code. I'd like to see it done all in one hit so that we can clearly delineate actual logic changes to the code from just reformatting.

> If we are to work together, we need to acomidate the different coding styles and requests.

If you we actually set a coding style to follow, then I would expect that all future pull requests/contributions would follow that same code style, and not people just doing their own thing.

felipesanches:
I did not use an automated tool. I did actually read all the code I changed.

I believe that software freedom requires the ability to read and understand the code. So, code can be classified in 2 aspects: brokeness and readability. And we have 4 combinations:

* working and readable <-- That´s the goal!
* working and messy <-- this is bad because it discourages improvement (it is always scary to modify something we don't understand with ease at a glympse)
* broken and readable <-- that's what you seem to be afraid of. But this is actually what gives you better chances of actually seeing the bugs.
* broken and messy <-- this is dangerous because the brokeness can be hidden forever.

Please, seriously consider giving up this strategy of using code uglyness as a marker to know which code has been reviewed and which hasnt been. Anything else you can use for that purpose will certainly be better than this ill-conceived practice.

jtanx:
@felipesanches I'm getting a bit confused at who you're responding to.

> I did not use an automated tool. I did actually read all the code I changed.

Were there any logic changes as part of that change? If there were, then it makes it difficult for anyone else to review what those logic changes were (which is why I would like to see the formatting changes separate/done in one go).

felipesanches:
Someone asked but I did not answer yet: yes, there are logic refactoring commits in this pull request as well.

There are places where it used to do:

if (logic_expression) /* do_something */;
if (oposite_of_the_same_logic_expression) /* do_something_else */;

This was changed to if/else

There was also junk like:

if (ch=='a' || ch=='b' || ch=='c' || ch=='d'){
if (ch=='a') do_something;
if (ch=='b') do_something;
/*and then nothing is done for ch=='c' or for ch=='d' This is crazy weird !!! */
}

There was also adoption of switch/case statements (instead of a series of if/else if/else statements) for the sake of clarity of code-intent.

felipesanches:
@jtanx I am replying to the ideas expressed here by other people. I'm not sure exactly who said what. I would have to review the whole conversation to map that. I am simply replying to the ideas. :-)

mskala:
>There are places where it used to do:
>
>if (logic_expression) /* do_something /;
>if (oposite_of_the_same_logic_expression) / do_something_else */;
>
>This was changed to if/else

Be careful!  If the first /* do_something */ has the effect of changing the value of the logic expression, then the original code will execute both if bodies, and changing it to if/else will execute only one.  This is the sort of thing I would expect to see done on purpose in the code when handling things like initialization of static variables, without any comment warning the reader about what's going on.  It's arguably bad style, but that has never been a barrier to finding such things in the code base.

Also worth knowing that if the if/else construction really is equivalent, then the compiler's optimizer is likely to compile it into an if/else anyway, so changing the source code only serves the (important, I agree) purpose of making it more understandable to human readers.

jtanx:
...And that's why I would like to see logic changes separate to indent/formatting changes.

So far, I see [one case](https://github.com/fontforge/fontforge/pull/2468/files#diff-35a548f59116f2cb22bfb1c8cb562244L346) of the `if(cond)` `if(!cond)` refactoring which seems okay.

> /*and then nothing is done for ch=='c' or for ch=='d' This is crazy weird !!! */

The only place I can find something similar to that is [line 494](https://github.com/fontforge/fontforge/pull/2468/files#diff-dbea8a90f6a82c8f5c9b4970183d95c6L494) which does have an action for `ch=='c'` or `ch=='d'` - to return `pt_unknown`. Although I do agree that a switch/case statement for this part makes it clearer.

JoesCat:
@felipe,
Please take a day or two to digest what was written before doing name calls 
and finger pointing. Describing a comment or two as giving you Stockholm 
Syndrome is a bit over the top and suggests that any little comment that 
doesn't agree with you, or your direction, you'll find offensive.

I'll write it again...please take a day or two to digest what was written.

On September 2, 2015 05:21:10 AM Felipe Corrêa da Silva Sanches wrote:
> Except for a few reasonable comments above, the overall rationale behind
> this policy seems non-sense to me. This almost feels like Stockholm
> Syndrome.
> 
> And you're also probably losing code contributors because of that.

With the last statement, there are code developers that disagree, but then 
again, looking at 2012, and comparing now, there are a lot of contributors 
that have helped. I hope you will too, but read what I wrote 
earlier...please see if you can bulletproof the code you modify. There is a 
lot of code existing right now that assummes all is well, and then users 
see the seg faults and fontforge vanishing for no apparent reason. If you 
make the routines you modify "bullet-proof", then there is less code that 
is prone to die unexpectedly.

I'll write it again...please take a day or two to digest what was written.

felipesanches:
Please excuse-me if I sounded aggressive in my previous comments. That was definitely not my intention. I did refer to Stockholm Syndrome (using "This almost feels like" and not "This is") only because it was the closest analogy I was able to conceive for describing a situation in which people start to sympathize with something that is actually harming them. The intent was not to point fingers or to acually imply any sort of sickness. I'm here trying to help the project and I'm trying to convince people that the current policy is problematic and should definitely be changed to something better.

Once again, sorry if I sounded aggressive. I'm here in peace :-)

davelab6:
@felipesanches well, @jtanx said 

> I would like to see logic changes separate to indent/formatting changes

So I suggest to do that and rebase this and open a 2nd PR, or close this and forget about it and spend the time on trufont

PR #2500, opened 2015-09-24
WIP making Source Code Sans build cleanly

mycroftxxx:


adrientetar:
cc @khaledhosny (you have touched featurefile.c in sortsmill).

The limit of 31 chars in glyph names in explicitly stated in the feature file specification. If Adobe itself uses longer names then maybe it could use an amendment (cc @readroberts about that).

mycroftxxx:
There's the spec, and there's reality… and reality always wins.  The spec
also shows that size/parameters always has 4 arguments, but in fact the
Source fonts only use two.

khaledhosny:
If Adobe

khaledhosny:
If Adobe’s implementation differs from the spec, then it should be reported so that the spec is updated. Personally I’d go for supporting arbitrary length of glyph names if possible.

khaledhosny:
As for UFO support, I’d just kill FontForge’s C implementation, and write a glue layer for robafab (extending FontForge’s Python interface as necessary), and added a way to extend FontForge’s file format support using Python hooks (similar to the exiting support for importing and exporting images). I did some incomplete work for the former point (though targeting Sorts Mill Tools) and I can push it somewhere if anyone wants to have a look, as I might not resume work on it anytime soon.

readroberts:
About the 31 character limit in PostScript glyph names: although the spec clearly states the limit of 31 characters, I increased the limit in the AFDKO tools to 63 a year or to ago for practical reasons. Use of the rules to name complex ligatures leads to longer glyph names, and I had reasonable use cases that would exceed this limit. A number of the limits stated in the spec are not in fact enforced any the rasterizes I have available for testing.  I changed this limit in the AFDKO tools in order to test if we could in practice get away with longer glyph names. Although there are few such cases, I have not heard of any reports of difficulties.  I agree that the spec should be changed.

adrientetar:
@mycroftxxx Do tell when you are done.

@frank-trampe Can you have a look?

mycroftxxx:
Actually, I've already gotten it to build.  Beyond what I've posted, there
are three remaining issues:

* two argument size/parameters
* use of cvParameters
* UFO glyph format 2

I have some changes for the last that I don't think are quite prime time
yet, but I might get to finishing them tomorrow.

(What actually got me started on this path was tweaking something in
Fanwood for a particular application, and discovering that both the Adobe
tools and FontForge barfed on it—but I got past that fairly quickly.)

frank-trampe:
I've been a bit behind on pull requests lately due to heavy work-related travel. People stop meeting and returning calls pretty soon due to Christmas, so I expect to be back in the FontForge groove in a little over a week.

fontforge/ufo.c
```diff
@@ -2364,8 +2364,8 @@ static SplineChar *_UFOLoadGlyph(SplineFont *sf, xmlDocPtr doc, char *glifname,
     glyph = xmlDocGetRootElement(doc);
     format = xmlGetProp(glyph,(xmlChar *) "format");
     if ( xmlStrcmp(glyph->name,(const xmlChar *) "glyph")!=0 ||
-	    (format!=NULL && xmlStrcmp(format,(xmlChar *) "1")!=0)) {
-		LogError(_("Expected glyph file with format==1"));
+	    (format!=NULL && xmlStrcmp(format,(xmlChar *) "1")!=0 && xmlStrcmp(format,(xmlChar *) "2")!=0)) {
```
adrientetar: Why && and not || here? It's either one format or the other, no?

khaledhosny: It is checking that format **is not** one of the accepted versions.

adrientetar: Right, thanks!

adrientetar:
> I just looked at Source Serif Pro. It fails to comply with any present U. F. O. specification.

Yes, it's a known problem that iirc we hit last year already. Basically Adobe produces ufo2 files with some ufo3 functionality notably layers. 
I suspect they do this so to be able to use their UFOs in robofont which doesn't support ufo3 at the moment while having this extra bit of functionality (which they may make use of in their UFO tools). Again, that's just iirc.

Also, I agree that we shouldn't complexify our code to accept files that are out-of-spec.

mycroftxxx:
On Sat, Dec 26, 2015 at 8:16 PM, Adrien Tétar <notifications@github.com>
wrote:

> Also, I agree that we shouldn't complexify our code to accept files that
> are out-of-spec.
>
> The problem with that logic is that the code isn't any more complex, at
all.  In fact, I significantly reduced the complexity in one area because
the existing code was needlessly overcomplicated.

davelab6:
@mycroftxxx are you planning to do any more on this PR? :)

davelab6:
@frank-trampe anything blocking this from being merged? s

frank-trampe:
I've been able to review/confirm the smaller changes, but I still need to chug through the big commit.

davelab6:
@mycroftxxx I'm sorry this hasn't been merged for over a year. Did you test it? If it works for you, I think it should be merged.

frank-trampe:
I feel bad about that too, but it's a big changeset, and I can't merge it until I've reviewed it. The year of merging everything without review was a disaster, and we're not doing that again. I considered cherry-picking the commits from here that I had already reviewed, but then we'd have a merge conflict from this pull request.

frank-trampe:
Actually, I already merged five of the eight commits in #2655, but it didn't cause a conflict.

frank-trampe:
There are also some open questions/issues in this thread, which are in turn complicated by the fact that GitHub has broken the associations between the code and the pull request comments. Let me try to reconstruct.

frank-trampe:
Agh. The line comments don't even show up on the code now. I'm going to make a new pull request with just the unmerged commits and start over.

davelab6:
Thank you Frank!!

ctrlcctrlv:
> I'm going to make a new pull request with just the unmerged commits and start over.

I don't know if this ever happened or not, but as many years have passed it's clear that this PR is unlikely to ever be merged, so I'm closing it in favor of @frank-trampe's plan, executed or not.

Issue #2510, opened 2015-10-01
Catalan translation


rferranp:
Hi, I'm Rafael Ferran i Peralta, author of catalan translation of FontForge wich is about 50% complete.
Now I can continue working with this translation. Please, say me what are the next steps I must do in order to collaborate (I can't find any instructions about how translators can help, I'm sorry). Thanks in advance.

Rafa

JoesCat:
Hi Rafael,
The easiest step for us would be for you to "fork" a copy of Fontforge into your github profile, then make a copy on your computer, add the copy there, upload the updated copy to your github FontForge fork, and then github will notice the update, and add a "do you want to do a pull request?". You would click that, and that would create a pull request on the main Fontforge pull request page.

If you have trouble doing that, then zip the catalan updated file and send it as email to one of us so that we update the file. This file is too big to post on sourceforge as a zipped attachment.

While you are doing that, can you also look at file desktop/fontforge.appdata.xml and add a catalan translation? You can see the file here: https://github.com/fontforge/fontforge/blob/master/desktop/fontforge.appdata.xml and you can use a plain text editor to add the extra translation for lang=ca
If you have trouble using github and pull requests, then put the edited file in a zip file, and email that file too. Since this second file is small, you can send it to the email list on sourceforge, preferably the fontforge developer email list so you don't create large traffic on the user list.

rferranp:
Thank you very much, Joe.
I'll try to do that what you say.

PR #2512, opened 2015-10-02
Scripting minions layer

JoesCat:
moving in the direction of eliminating exit(x) from scripting and use return pass/fail so that the GUI doesn't vanish if error encountered in scripting.


JoesCat:
Found Rafael's patch and posted it on the end here ( see issue #2510 ).

Issue #926, opened 2013-11-07
WOFF generated by FF have wrong 'head' checksum


coolwanglu:
Open <https://dl.dropboxusercontent.com/u/475186/ff_bug/demo.html> with different browsers:

OS: Win 8.1

`IE 11`
![capture1](https://f.cloud.github.com/assets/734614/1491333/4d25e9de-47ac-11e3-9bc9-355bf3cbe245.PNG)

`Firefox 25`
![capture2](https://f.cloud.github.com/assets/734614/1491348/84590206-47ac-11e3-9eb9-f25c227ee4a1.PNG)

`Chrome 30`
![capture3](https://f.cloud.github.com/assets/734614/1491359/a40c029c-47ac-11e3-9efc-94607572cc6c.PNG)

[f7.woff](https://dl.dropboxusercontent.com/u/475186/ff_bug/f7.woff) and [f9.woff](https://dl.dropboxusercontent.com/u/475186/ff_bug/f9.woff) are respectively used in the two lines, which are generated from [f7.pfa](https://dl.dropboxusercontent.com/u/475186/ff_bug/f7.pfa) and [f9.pfa](https://dl.dropboxusercontent.com/u/475186/ff_bug/f9.pfa) with FontForge respectively. Note that both fonts contain only a subset of characters.

It can be observed that `f7.woff` is accepted by all three browsers, while `f9.woff` works only on IE. Firefox shows the following message regarding `f9.woff`

![capture4](https://f.cloud.github.com/assets/734614/1491381/06073ea8-47ad-11e3-9196-77b83a7ba6ce.PNG)

I heard that Chrome is using the same validator as Firefox, but I'm not sure.

I wonder if it is a bug of Fontforge, or bugs of the validator used in Firefox and chrome, or maybe something wrong in `f9.pfa` ?

jimmywarting:
+1

fred-wang:
Comparing the table directory of the ttf output generated by FontForge and the one constructed by calcHeadChecksum in validator.py, the only difference seems to be the offset of tables:

    fontforge sfnt output

    tag | checkSum | offset |length    
    CFF , 1370159785, 2116, 3754, 
    FFTM, 1992083500, 5904, 28, 
    GDEF, 2555936, 5872, 30, 
    OS/2, 1502503344, 288, 96, 
    cmap, 909254778, 1256, 826, 
    head, 95901027, 188, 54, 
    hhea, 88605330, 244, 36, 
[...]

PR #2520, opened 2015-10-12
Recalculate checkSumAdjustment when generating a WOFF file. #926

fred-wang:


fred-wang:
I verified locally that the test passes with "make check" but travis seems to fail before (when downloading test fonts)...

jtanx:
It looks like the path to the test font has changed from ftp://ftp.naist.jp/pub/NetBSD/packages/distfiles/adobe-cidfonts/ to ftp://ftp.naist.jp/NetBSD/packages/distfiles/adobe-cidfonts/

fred-wang:
@jtanx thanks, that seems to work now. However, for some reason many tests are skipped.

It's possible that instead of recalculating the checkSumAdjustment, we just want to write the font tables in their offset order instead of the table directory order so that checkSumAdjustment may not need to be changed (otherwise the spec says we should drop DSIG tables). I'll do more experiments on this later.

fred-wang:
Closing per my previous comment.

PR #2524, opened 2015-10-13
WOFF: Compress & write the font tables in the same order as the origi…

fred-wang:
…nal sfnt font so that the checkSumAdjustment does not change when the font is decompressed again. Also add a Python test to verify that the WOFF files generated by fontforge are valid. #926

### DIFF
--- a/fontforge/woff.c
+++ b/fontforge/woff.c
@@ -362,6 +362,23 @@ return( NULL );
 return( sf );
 }	
 
+typedef struct {
+    int index;
+    int offset;
+} tableOrderRec;
+
+static int
+compareOffsets(const void * lhs, const void * rhs)
+{
+    const tableOrderRec * a = (const tableOrderRec *) lhs;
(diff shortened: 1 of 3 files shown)


fred-wang:
The relevant quotation from http://www.w3.org/TR/WOFF/:

> Font tables in WOFF files MUST be stored in the same order as the well-formed input font. The table order is implied by offset values in the table directory; sorting table directory entries into ascending offset value order produces a list of entries in an order equivalent to that of the font tables.

frank-trampe:
It looks good.

fred-wang:
Thanks! I guess #926 can also be closed, unless there are reasons why checkSumAdjustment can be invalid other than wrong table ordering.

jtanx:
So, after looking at #2762, I've found that the added test doesn't work on Python 3 - which all FF unit tests should conform to. Travis wasn't picking this up, because `test926.py` wasn't added to `EXTRA_DIST` in `Makefile.am` so the test was being skipped.

PR #2532, opened 2015-10-15
Fix crash when generating FON without 'X'.

bungeman:
The conditional is fixed by doing the same thing currently done a few lines up with ' '. This allows generation of bitmap FON files without an 'X'. Currently generating a FON file without an 'X' glyph mysteriously crashes.


frank-trampe:
This looks good to me.

PR #2544, opened 2015-10-25
Check for math.h and then check for libm and halt if not found.

JoesCat:
It's practically a given expectation math.h and libm exists,
but check for both of them anyways. Some operating systems don't
have libm, therefore check for function instead.
Do a search for library first, but do a check if not found since
some older distros have trouble using search for user-installed
libraries.


JoesCat:
Not quite answered in 2533, but seems ok with travis checker.

PR #2553, opened 2015-10-31
Fix building error with Python 3.5.

RavuAlHemio:
PyModuleDef has lost m_reload and gained m_slots in Python 3.5. Set the correct member to NULL according to the detected version. The Python changes are described in PEP489.


JoesCat:
references: "[Python-Dev] Changing PyModuleDef.m_reload to m_slots"
https://mail.python.org/pipermail/python-dev/2015-April/139186.html

PR #2557, opened 2015-11-04
Add INSTALL instruction for Alpine Linux

SydOps:


SydOps:
the problem with `git://` behind proxy can be fixed by the way in http://www.emilsit.net/blog/archives/how-to-use-the-git-protocol-through-a-http-connect-proxy/

But i prefer to do it via `https://`

JoesCat:
Would you mind checking if there is any issues with libspiro HEAD or libuninameslist HEAD.
I expect no complications, but worth knowing if you have issues in alpine linux.

You'll need to install them first before doing fontforge's ./configure (should be able to autodetect both without user needing to add configure flags).

SydOps:
@JoesCat 

Not sure how to check. I make it in a docker image.
```
$ docker images |grep fontforge
fontforge                    latest              290cb202ba26        4 hours ago         788.1 MB

$ docker run -ti --rm fontforge sh
/fontforge # fontforge
 Based on sources from 02:55 UTC  4-Nov-2015-D.
 Based on source from git with hash: 1ff2e1d83f7b14b1d7f5cd099388bb3f94c46de0
Internal Error: Your version of iconv does not support the "Mac Roman" encoding.
If this causes problems, reconfigure --without-iconv.
>>> exit()
/fontforge # apk search spiro
/fontforge # apk search libspiro
/fontforge # apk search libuni
libunique3-doc-3.0.2-r0
libunique-1.1.6-r5
libunique-dev-1.1.6-r5
libunique3-3.0.2-r0
libunique3-3.0.2-r0
libunique3-dev-3.0.2-r0
libunique-1.1.6-r5
libunique-doc-1.1.6-r5
/fontforge #
```

So I don't find package `libspiro` or `libuninameslist` 

Can i run some tests to provide if the compiler `fontforge` works as expect in the docker container?

SydOps:
The `Dockerfile` for fontforge is at 

https://github.com/BWITS/Docker-builder/blob/master/fontforge/alpine/Dockerfile

JoesCat:
Both libraries are optional, but also "nice to have".
If you can run FontForge, then to check if libuninameslist is running, you can hover your mouse over a character, such as "B" and you should see a pop-up bubble with additional info. If the pop-up bubble is fairly basic, such as a hex digit, unicode value and a simple statement like "B" Latin Character, then FontForge did not link with libuninameslist. If the description is much more detailed, then FontForge linked with libuninameslist. Some characters have descriptions that can be several lines in detail, and libuninameslist is more or less updated when unicode.org updates their charts (currently at version 8 right now). It's a nice to have for people sometimes fix or modify font(s), but seems to be a wish-list must-have for type experts that do fonts at a professional level.

libspiro can be checked by selecting a single character from the chart so that it is editable in a separate window. You should see a button that looks like a pencil dragging over a sideways "S". If you click it, it should toggle 5 other buttons in that menu if libspiro is available. You'll notice lines are drawn differently if you have it toggled on or off. FontForge functions without it, but some type experts make use of it to help improve their curves. Fedora and several distros are using 0131, Debian is currently using the latest tag since it answers problems with older CPUs (but I created an error that caused problems with the BSDs - now fixed in HEAD). I recommend trying HEAD since it's eventually going to be the next version, however, if you want to use a release tag and not HEAD, try ver 0.5 (unless you run into same problem hit by BSD or 0.3 (functionally same as 0.5).

If you are able to view ./configure (on the command line), it will indicate if found or not.

jtanx:
@SydOps could you please squash your "update INSTALL" commits? Thanks!

SydOps:
@JoesCat 

Thanks for so detail information. 

I think I didn't install both libraries. Alpine is lightweight linux, I may need compile them from source code. 

Second, I didn't set GUI in alpine, not sure how to test the fonts without it. Seems it is not good idea to install fontforge in Alpine linux. 

Anyway, my purpose is to used fontforge with another project (https://github.com/coolwanglu/pdf2htmlEX) in alpine. 

@jtanx 

I am not good in English, could you please explain what need be done with `squash`?

jtanx:
Squash in git means to combine those commits into one, so instead of having three "update INSTALL" commits, you have one. See here for an example: http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git and http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

In this case, it would be something like:

      git rebase -i 11b3dc6

In the prompt you would choose:

      pick 65dbc98
      squash 1ff2e1d
      squash  bbbb507

You will have to force-push back to github as you have essentially rewritten your git history:

      git push -f origin master

Note that normally you don't want to force push (somewhat dangerous), but in this case it is ok.

SydOps:
@jtanx 

I have updated with `squash`. 

That's really good function in git, I never use it before. Thanks to let me know that.

JoesCat:
Hi @SydOps, if your target is https://github.com/coolwanglu/pdf2htmlEX then the libraries are not necessary since they are meant for creating or editing new fonts, while what you describe seems more like converting from one format to another.
I believe Coolwanglu maintains a fork of FontForge here https://github.com/coolwanglu/fontforge to support the pdf2html you mention, and there is also another fork that prefers running without GUI and emphasising scripts located as a subset here: https://github.com/mskala/Tsukurimashou/tree/master/fontanvil but I'm not sure if you can compile it seperately.

For FontForge, you would need to turn-off several defaults to run without GUI such as:
/configure --without-cairo --without-x
...but I agree, FontForge is a very heavy application for such a lightweight distro such as Alpine. The smallest size you could get is to compile it, strip it, run it through upx, and distribute just the compiled binary if you are trying to keep the size as small as you can.

SydOps:
Thanks, @JoesCat . Seems it is not suitable to raise the pull request to add alpine compile in `INSTALL`

I will close it directly.

Regarding `git://` or `https://` in `bootstrap`, if you think it is a good idea, then I will raise another pull request for that.

JoesCat:
On November 4, 2015 09:43:46 PM Bill W wrote:
> Regarding `git://` or `https://` in `bootstrap`, if you think it is a
> good idea, then I will raise another pull request for that.

You found something that appears worth fixing.
It is also likely worth notifying upstream to bootstrap too (you may have 
noted the reference to gvvaughan/bootstrap, issue#1)

I would recommend making a pullrequest to modify bootstrap.conf so that 
bootstrap can be updated without having to carry added mods. The upstream 
bootstrap appears to have mods already.

SydOps:
@JoesCat 

Done with #2561

mskala:
I'm not sure FontForge can build and run without GUI libraries, even if it can be used headlessly.  FontAnvil is a much smaller program, with fewer dependencies, and may be more suitable for headless use.

PR #2560, opened 2015-11-05
Resurrect acorn2sfd

mjwoodcock:
Hi,

Please resurrect the acorn2sfd tool.  I have added it to Makefile.am, but I'm not an automake expert so it may need tidying up.

You have to compile it yourself by running:
cd fontforgeexe
make acorn2sfd

which is how it worked in older versions of fontforge.

I fixed a couple of crashes as well.

Thanks


JoesCat:
In fontforgeexe/Makefile.am, please add acorn2sfd.c so that when anyone runs 'make dist' or 'make distcheck' it gets included (make dist is currently broken):
```c
EXTRA_DIST = fontimage.pe fontlint.pe sfddiff.pe	\
	fontimage.1 fontlint.1 sfddiff.1		\
	acorn2sfd.c darwinsetup.in macobjective.m
```

Please add your instructions on how to build the file in acorn2sfd.c.
It is also useful to understand what or why the file is included, my guess is due to historic significance as one of the first ARM machines, but you may have a better reason to show why.
Please comment file so that it won't be deleted by someone looking to cleanup cruft.

mjwoodcock:
Thanks for the feedback.

I have made the changes you have suggested.

The OS that uses these fonts is still in active development (http://www.riscosopen.org) so the interest is not purely historic, although it is a niche interest.

make dist works for me, but make distcheck still fails, but it fails in some code that I haven't touched:
libtool: compile:  gcc -std=gnu99 -std=gnu99 -g -c -I../inc -I../../inc -I../lib -I../../lib -DSHAREDIR=\"/home/jw/fontforge/fontforge-2.0.20140101/_inst/share/fontforge\" -DDOCDIR=\"/home/jw/fontforge/fontforge-2.0.20140101/_inst/share/doc/fontforge\" -I../../uthash/src -DFF_UTHASH_GLIF_NAMES=1 -DPLUGINDIR=\"/home/jw/fontforge/fontforge-2.0.20140101/_inst/lib/fontforge/plugins\" -Wl,--no-undefined -pthread -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -pthread -I/usr/include/pango-1.0 -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/harfbuzz -I/usr/include/freetype2 -pthread -I/usr/include/pango-1.0 -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng12 -I/usr/include/freetype2 -I/usr/include/libxml2 -I/usr/include/python2.7 -I/usr/include/x86_64-linux-gnu/python2.7 -pthread ../../fontforge/splinerefigure.c -o splinerefigure.o >/dev/null 2>&1
  CCLD     libfontforge.la
.libs/libfontforge_la-spiro.o: In function `SSRegenerateFromSpiros':
/home/jw/fontforge/fontforge-2.0.20140101/_build/fontforge/../../fontforge/spiro.c:274: undefined reference to `_SpiroCP2SplineSet'
collect2: error: ld returned 1 exit status

JoesCat:
Interesting coincidence. 2 Niche operating systems within (practically) 2 days of one another.
issue #2557 is related to an OS which seems to target very light platforms (like routers).

mskala:
Are there some fonts in acorn format that can be distributed under FontForge's license (that is, GPL3 or compatible with GPL3) and could be used for testing this code?

mjwoodcock:
> Are there some fonts in acorn format that can be distributed under FontForge's license (that is, GPL3 or compatible with GPL3) and could be used for testing this code?

Not that I have found so I have converted some GNU free fonts to Acorn format.  I think this is legal as long as I include the license and point to .zip file containing the original .ttf files somewhere in a readme.

Would that be acceptable to you?

mskala:
That sounds good.

It seems to me that ideally, FontForge should simply be able to read Acorn fonts the same as other formats, instead of needing a separate utility to convert them.  I'd like to try implementing such a feature in FontAnvil, and if it works, maybe it could then be merged into FontForge as well.  It seems like it shouldn't be very difficult because acorn2sfd already links with FontForge code and loads the font into FontForge in-memory format before saving it as SFD; it seems like it should be easy to just skip the save/load steps and put the code of acorn2sfd directly into the part of FontForge that reads other file formats.  I can't really attempt such a thing without fonts to test it on, though.

mjwoodcock:
That sounds like a good idea.

For your own testing you can find a whole bunch of fonts here: http://arcade.demon.co.uk/filepages/file39.htm.  Ignore any BBC fonts, they are something different entirely.  Anything with -zip in their filename can be extracted using the standard unzip.    For all others, you will need to extract them with nspark from here: https://github.com/mjwoodcock/nspark.  Compiling and running it should be simple enough, but let me know if you get stuck.  Or I can extract them all and send them to you in .zip format.

I have a script I use to do some standard fixups on the fonts I have converted from Acorn format which I can send to you.

Should I put the GNU fonts into git?  Or should I send you a copy?

mskala:
Thanks.  I think checking them into a repository somewhere would be good; that way they can be available to others working on acorn2sfd and fontforge.  My own availability is unpredictable.

Issue #2574, opened 2015-11-16
Fontforge produces invalid XML


tmccombs:
For a certain font (LibelSuit-Regular.otf), when it is converted to an SVG, the output XML contains an XML declaration of version="1.0", but then also contains the following line:

`<hkern u1="&#x17;" u2="j" k="19" />`

but `&#x17;` is illegal in XML 1.0. Either the XML declaration should specify XML 1.1, in which the character is valid, or not output the illegal character (for example, use `u1="U+17"` instead).

tmccombs:
See http://stackoverflow.com/a/28152666/2543666 for a discussion of which characters are legal in XML 1.0 and 1.1

PR #2576, opened 2015-11-17
Better xml

tmccombs:
This fixes a couple of bugs where fontforge outputs invalid XML for SVG fonts.

### DIFF
--- a/fontforge/svg.c
+++ b/fontforge/svg.c
@@ -91,12 +91,16 @@ static int svg_outfontheader(FILE *file, SplineFont *sf,int layer) {
     }
     fprintf( file, "</metadata>\n" );
     fprintf( file, "<defs>\n" );
-    fprintf( file, "<font id=\"%s\" horiz-adv-x=\"%d\" ", sf->fontname, defwid );
+    fprintf( file, "<font id=\"");
+    latin1ToUtf8Out(file, sf->fontname);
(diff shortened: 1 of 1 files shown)


mskala:
Using XML 1.1 may not be a good thing.  Many popular parsers can't handle it.

tmccombs:
In that case fontforge needs to make sure the output is valid for XML 1.0. Some characters are illegal in XML 1.0, even if escaped using XML entities. (such as &#x17;), but fontforge includes them. See #2574.

mskala:
I agree, that's a bug that should be fixed.  From my reading of the SVG Font specification at http://www.w3.org/TR/SVG/fonts.html#HKernElementU1Attribute , and the discussion it links to, it looks to me like the easiest way to handle the specific example you raise in #2574 might be to write a single-character instance of the "range" syntax, like "U+0017", instead of using an escape.  I don't know if readers of SVG fonts can read that, but it _is_ in the standard, and at least it won't break XML 1.0 parsers in general.

tmccombs:
that works for me. I'll revert the commit XML 1.1 so the other commit can be merged.

jtanx:
Please rebase your branch to remove the reverted commits.

tmccombs:
done. I also made sure fontname is escaped.

PR #2577, opened 2015-11-18
fix StrSplit, and test for it

mskala:
A recent fix for a buffer overflow in StrSplit was an incomplete solution:  it fixed the case that was breaking (running past the optional element count limitation), but broke a case that was formerly working (unlimited element count).  This change should fix it in both cases, and updates the test to check them both.


JoesCat:
Excellent!

Issue #2490, opened 2015-09-15
Creation time / modification time not propagated from source file [Debian]


n8willis:
Debian's reproducible-builds project has opened a bug - https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=774148 - citing the fact that when FontForge is used to build TTFs, it inserts the current (i.e., build) time as the creation time and modification time of the generated font, as opposed to the modification time of the font source.

This causes TTFs generated with FontForge to fail the reproducible-build test, which Debian is moving toward for all packages.

There's a patch included in the downstream bug report. It checks the creation time and mod time of the font source, and uses those as the values written to the TTF.  I have not reviewed the patch, only read through it.

JoesCat:
I'm digging through some rather big patches for native scripting plus libspiro at the moment, but following your lead here... ...I think we need to visit debian font bugs since the bug you list seems to reference two patches, plus imply the need to add scripting 'creation time's in python plus native scripting.
https://bugs.debian.org/cgi-bin/pkgreport.cgi?maint=pkg-fonts-devel%40lists.alioth.debian.org

n8willis:
I suppose at this point, one thing we could do is see if there is anyone who *objects* to the change (proposed) itself.  Because it still constitutes a change, justifiable or otherwise....

JoesCat:
On September 17, 2015 09:06:04 AM n8willis wrote:
> I suppose at this point, one thing we could do is see if there is anyone
> who *objects* to the change (proposed) itself.  Because it still
> constitutes a change, justifiable or otherwise....

...appears justifiable, plus looking at the debian page, seems there's a lot 
of broken fonts, maybe some are waiting for such a fix as you've indicated.

My main thought for this patch you found is (like any other), please credit 
the original author if you can for what they made. They made the effort, so 
would be nice to credit them for it. Ideally, they would post a patch 
themselves, but that isn't always the case, or sometimes they are too busy, 
or moved-on....I don't know.

...Also following the bug you mentioned, I've noted that there seems to be 
a couple other bugs that could/might be followed-up if you (or anyone else) 
are interested in following-up.

...sometimes bugs like this leads into other things that could be fixed, so 
it's a good starting point for people interested in starting somewhere but 
not knowing what to do.

...at this moment, I'm a bit loaded with a rather large native scripting 
patch (sort of imports mskala's fontanvil builtins table idea, plus I'm 
adding a big libspiro patch together but haven't quite settled on some 
details.

[...]

PR #2609, opened 2016-01-17
propagate creation and modification times to TTF files.   Fix: #2490

mapreri:
This enable TrueType Font to be built reproducibly.
See https://reproducible-builds.org/ for more about reproducible builds.

See #2490 and https://bugs.debian.org/774148 for more info.


davelab6:
@mapreri wow this is awesome!! I hope @frank-trampe can take a look at this as it seems like a 1 minute review :)

adrientetar:
Thanks

davelab6:
Thanks @adrientetar !

PR #2613, opened 2016-01-26
do not crash on eof (invalid input data)

pgajdos:
See https://bugzilla.suse.com/show_bug.cgi?id=963023 for testcase.


JoesCat:
Thanks for the patch!

PR #2666, opened 2016-04-03
Fix broken glyph view. Fixes #2665.

### LINKED ISSUE #2665: i can't see glyphs in the font view any more
um…the glyphs in the fontview suddenly got really really faint. I tried reinstalling fontforge & everything.
selection_001

I'm using the April 2 version from github, Ubuntu 15.10.

jtanx:
This is a partial revert of 92086911ed0be6f3c466643389d777646015b738. cc @frank-trampe


frank-trampe:
I was a bit wary of that change. But it passed Travis.

PR #2667, opened 2016-04-03
misc updates

JoesCat:
Update AUTHORS file a little further up to date.
cleaned some trailing spaces.
Need to comment EncodingFree, might as well remove duplicate too.
Add more info to configure for x.org and Potrace (Autotrace 0.31-1 needs patch before enabling check).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,6 +2,7 @@ George Williams
 10 years and approximately 600,000 lines of code.
 
 Abel Cheung
+Adam Goode
 Adolfo Jayme Barrientos
 Adrien Tétar
 Aktado <aktado@users.sourceforge.jp>
@@ -9,6 +10,7 @@ Alex Raftis
 Anand
 Andreas Larsen
 Andrew Janke
+Andrey Panov
 Anthony G. Basile
 Arthaey Angosii
 Alexey Kryukov - nowakowskittfinstr.c
@@ -23,12 +25,17 @@ BensongChen, bungeman
 Ben Wagner
 Bill Wang
 catharanthus
+Charles M. Hannum
 Christopher Meng
 Claudio Beccari
 Craig Lennox
 Dave Crossland (dave@lab6.com)
 Deron Meranda
 Dave Levine
+David Lemon
+David Opstad
+David Shaal
+Divay Prakash
 Dominyk Tiller
 Dong Yang
 Dražen Lučanin
@@ -40,7 +47,7 @@ Frank Trampe
 Frédéric Wang
 Georg A. Duffner
 Giuseppe Bilotta
-Giuseppe Ghibo
+Giuseppe Ghibò
 Greg Ford
 Grzegorz
 Harald Harders
@@ -51,6 +58,7 @@ James Cloos
 James Crippen
 James Woodcock
 Jason Pagura
+Jérémy Bobbio
 Jeremy Tan
 Jim Killock
 Jiwon Choi
@@ -68,15 +76,19 @@ Julien Nabet <serval2412>
 Just Fill Bugs <mozbugbox>
 Kartik Mistry
 KANOU Hiroki
-Kauzuyuki
+Kazuki Ohta (Kauzuyuki, Kazuyuki)
 Kelvinsong
 Kemie Guaida - Tango PixMaps
+Kengo Ichiki
 Kevin Fenzi
 Khaled Hosny
 Koki Takahashi
 Lasse Fister
+László Károly
 Lu Wang <coolwanglu@gmail.com>
 Luiz Matheus <lzmths>
+Markus Schwarzenberg
+Marius Larsen
 Mark Oteiza
 Martin Hosken
 Mathias Wollin
@@ -88,11 +100,15 @@ Max Rabkin - excepthook.py
 Maxim Iorsh
 Michal Mazurek
 Michal Nowakowski - nowakowskittfinstr.c
+midzer <midzer@gmail.com>
+Nicolas Kaiser
 Nicolas Spalinger
 Ondřej Hošek
 Parag Nemade
+Petr Gajdos
 Phil Krylov
 Phil Lindsay
+Pierre Hanser
 Primoz PETERLIN
 Qmpel
 R.L. Horn
@@ -103,6 +119,7 @@ Reuben Thomas
 Richard Hughes
 Rogier van Dalen
 Ryusei Yamaguchi
+Saeed Rasooli
 Scott Pakin - Packaging
 Scott Romack
 Silvan Toledo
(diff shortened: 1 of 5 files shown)


AUTHORS
```diff
@@ -23,12 +25,17 @@ BensongChen, bungeman
...
 Bill Wang
 catharanthus
+Charles M. Hannum
 Christopher Meng
 Claudio Beccari
 Craig Lennox
 Dave Crossland (dave@lab6.com)
 Deron Meranda
 Dave Levine
+David Lemon
```
frank-trampe: Did we get code from David Lemon?

fontforge/encoding.c
```diff
@@ -498,12 +498,19 @@ static char *getPfaEditEncodings(void) {
     return encfile;
 }
 
-static void EncodingFree(Encoding *item) {
+void EncodingFree(Encoding *item) {
+/* TODO: Fix error if repeatedly freed more than 2x */
```
frank-trampe: I don't think that we need to do that in this function. The caller ought to set the pointer of a freed object to NULL.

fontforge/splineutil.c
```diff
@@ -39,6 +39,9 @@
 #ifdef FF_UTHASH_GLIF_NAMES
 # include "glif_name_hash.h"
 #endif
+#ifndef _ENCODING_H
```
frank-trampe: When would this happen?

fontforge/splineutil.c
```diff
@@ -6541,7 +6529,7 @@ return;
         }
       }
       free(sf->layers); sf->layers = NULL;
-    }
```
frank-trampe: I really don't like these whitespace changes. They just make blame harder to use.

frank-trampe: But at least they're in a separate commit.

JoesCat:
> Did we get code from David Lemon?
David Lemon provided "technical" information (of benefit), which can lead a programmer in the right direction. George thought it was worth giving credit in a commit, I think the benefitial... it's one of shade of grey decisions since it's not black nor white.... If you want to leave him out, it's your call, look here: 2003-10-23 16:42:43

> I don't think that we need to do that in this function. The caller ought to set the pointer of a freed object to NULL.
The TODO is more to do with why is FontForge falling-over here? can we fix it so it doesn't break?
Replacing a memory location with NULL is a memory leak, and not a clean exit, maybe considered okay with some programmers...I prefer making things (eventually) robust.
If you want a code example to test it out, I'll FWD it to your github specified gmail.

> When would this happen?
I wasn't too sure whether to do an include "encoding.h" or simply use an external just for that 1 routine. I thought it more important to keep the "TODO" and remove the duplicate routine before it got forgotten. The extern was quick-n-easy for the moment, and I somewhat think there are more duplicate stuff to cleanup later...ie splineutil.c seems to know what structure is passed even though encoding.h isn't referenced.

> I really don't like these whitespace changes. They just make blame harder to use.
Kwrite seems to catch these dangling trailing spaces that should have been left-out by the original programmers to begin with...there's no IQ present with those blanklines, so I let it go...

Link for bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=819864

frank-trampe:
@JoesCat, I'm not saying to set the pointer to NULL without freeing the indicated object. I'm saying that the calling code ought to set the pointer to NULL after freeing the object. It makes it much easier to track double frees.

frank-trampe:
Oh. I see. splineutil.c wasn't including encoding.h.

This solution still leaves us with duplicate function declarations, which could cause silent (buried in warnings) problems if the implemented function changes. What happens if we just include encoding.h?

JoesCat:
> What happens if we just include encoding.h?

Sorry about the long delay... tried it looks okay to use encoding.h

I had done the ifndef with some other test-code and copied it over thinking about the fact splineutils was using encoding structures, yet never referring to encoding.h ...needing some further investigating...I'll have to fix this on the weekend, no time right now, or if you want to go ahead and edit/merge, please go ahead - Thanks in advance.

JoesCat:
problem not double free outside the routine, but actually something inside the EncodingFree() that seems to have a problem.

JoesCat:
I'll have to leave the EncodingFree() for a little later.
I noticed the extra free which I think is what you mentioned, fixed now.

configure.ac
```diff
@@ -317,6 +317,20 @@ AM_CONDITIONAL([MAINTAINER_TOOLS],[test x"${i_do_have_maintainer_tools}" = xyes]
 AC_SUBST([MAINTAINER_TOOLS],["${i_do_have_maintainer_tools}"])
 
 #--------------------------------------------------------------------------
+# Check for autotrace and potrace (don't do anything with this yet)
```
jtanx: What's the point of adding this check? None of the autotrace code is dependent on a compile time check to see if autotrace is present or not. autotrace/potrace could be installed/removed after fontforge was compiled, and this would make no difference to fontforge - as long as it's in PATH or the AUTOTRACE variable is set when the autotracing function is used.

jtanx: Ping @JoesCat

jtanx: I'm still not sure I see at as being useful. Did said user also compile from source? Otherwise this is a pretty pointless check, since they will never see that it indicated yes/no to autotrace/potrace.

I think this is an indicator that better documentation or other indicators within FontForge itself are necessary, but not to the build system itself. Doing so implies that it's a build time dependency, which it's clearly not. I can certainly see this itself becoming a point of confusion because now people can be wondering if they need to install autotrace/potrace for the build to work, or if it will influence the build in any way.

Furthermore, this particular check can do no more - it doesn't (and shouldn't) influence compilation in any particular way.

JoesCat: @jtanx, this patch was a quick solution to what I saw as a problem, and it is not all that important if "said-user" saw the ./configure results or not.
To put more perspective to this...
From the last release tag until now, Fontforge has had over 100,000 downloads.
If only 1% happened to see the results of ./configure, that would be 1000 users who saw autotrace/potrace in the end-results.
If 1% of those users happen to be on the mailing list, that is 10 potential users who could also help answer the very same basic question you had to answer yourself.

We have X hours in the day, and the question(s) you don't need to answer, means more time available for hunting bugs and development instead of answering questions.
I think 1% is a bit pessimistic, therefore some numbers higher than 1% and 1% means more potential eyeballs able to help.
At the moment, I think the patch helps for now (better solution than nothing).
I agree with you documentation is a better solution, and when you write it, I'll be happy to help review/merge it.

jtanx: @JoesCat, as indicated before, my primary concern is that this is actually more misleading than it's meant to help. See my comment wrt. it being confused as actually affecting the build during compile time. Certainly when I first saw the check, my first thought was *what part of the build now relies on autotrace/potrace*?

I'm in favour of removing this check altogether as it just doesn't belong in the build system.

PR #2671, opened 2016-04-03
Fixup mkdir usage on Windows (Using Gnulib's mkdir now.)

jtanx:
Since Gnulib now handles the mkdir abstraction, the #ifdef'ed code is no longer necessary. cc @frank-trampe

### DIFF
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -793,11 +793,7 @@ static void ffensuredir( const char* basedir, const char* dirname, mode_t mode )
 
     snprintf(buffer,buffersz,"%s/%s", basedir, dirname );
     // ignore errors, this is just to help the user aftre all.
-#if !defined(__MINGW32__)
     mkdir( buffer, mode );
-#else
-    mkdir( buffer );
-#endif
 }
 
 static void ensureDotFontForgeIsSetup() {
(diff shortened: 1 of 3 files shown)


frank-trampe:
@jtanx, I'm still not completely thrilled with gnulib, and I don't want to do anything that makes it harder to move away from it in the future unless it provides some major benefit.

jtanx:
@frank-trampe hmm okay. But gnulib is already currently providing a number of portability related fixes for Windows and if I were to ever release another Windows build then I would need to apply this patch anyway.

frank-trampe:
@jtanx, how were your Windows builds working before?

jtanx:
I've been patching in GLib changes (e.g. https://github.com/jtanx/fontforge/commits/august-release) to get unicode filename support for a few releases now, which avoided this issue because it was using GLib's `g_mkdir`.

On current master, Gnulib's handling of `mkdir` seems to be a side effect of having it handle `tmpfile` - necessary to work around a permissions issue with `tmpfile` on Windows (e33edd67c3385667a92e590398ddb7c9391d7b35). 

In any case, I would like current master to be compiling on Windows without any patches, so this a step in that direction.

frank-trampe:
So you're saying that the present code breaks the Windows build by using an incompatible function prototype?

jtanx:
Yes - Windows provides `mkdir(filename)` but Gnulib replaces this with `mkdir(filename, mode)`

PR #2690, opened 2016-04-11
Update Ukrainian translation

yurchor:
This patch includes bug fixes for some typos in the previous version of the translation and the translation of two new strings. It is tested for validity with msgfmt -vc.

Thanks for merging this patch into the repo.


JoesCat:
Thanks Yuri!

PR #2694, opened 2016-04-12
cppcheck: fix some reports (part 5)

serval2412:
[contrib/cidmap/src/mapcns1.c:98]: (error) Uninitialized variable: j
[contrib/cidmap/src/mapgb1.c:89]: (error) Uninitialized variable: j
[contrib/cidmap/src/mapjapan1.c:93]: (error) Uninitialized variable: j
[contrib/cidmap/src/mapjapan2.c:88]: (error) Uninitialized variable: j
[contrib/cidmap/src/mapkorean.c:89]: (error) Uninitialized variable: j
[contrib/fonttools/showttf.c:5744]: (error) Uninitialized variable: i
fontforge/fvcomposite.c:2643]: (error) Uninitialized variable: ret
[fontforge/splinechar.c:1478]: (error) Uninitialized variable: start1
[fontforge/splinechar.c:1485]: (error) Uninitialized variable: start2

[fontforge/encoding.c:1205]: (style) Expression is always false because 'else if' condition matches previous condition at line 1203.
[fontforge/print.c:2635]: (style) Expression is always false because 'else if' condition matches previous condition at line 2633.
(+ Czech language is "cs", Czech Republic is "cz", see https://en.wikipedia.org/wiki/Locale, POSIX example)
[fontforgeexe/charview.c:5834]: (style) Expression is always false because 'else if' condition matches previous condition at line 5832.
[fontforgeexe/cvimportdlg.c:470]: (style) Expression is always false because 'else if' condition matches previous condition at line 466.
[fontforgeexe/cvstroke.c:585]: (style) Expression is always false because 'else if' condition matches previous condition at line 572.
[fontforgeexe/fontinfo.c:4144]: (style) Expression is always false because 'else if' condition matches previous condition at line 4141.
[gtkui/gwwvask.c:322]: (style) Expression is always false because 'else if' condition matches previous condition at line 319.

### DIFF
--- a/contrib/cidmap/src/mapcns1.c
+++ b/contrib/cidmap/src/mapcns1.c
@@ -37,7 +37,7 @@ return( *buffer=='\0' );
 }
 
 static int getnth(char *buffer, int col, int *mults) {
-    int i,j, val=0, best;
+    int i,j=0, val=0, best;
     char *end;
     int vals[10];
     int pua, nonbmp;
--- a/contrib/cidmap/src/mapgb1.c
+++ b/contrib/cidmap/src/mapgb1.c
@@ -36,7 +36,7 @@ return( *buffer=='\0' );
 }
 
 static int getnth(char *buffer, int col,int *mults) {
-    int i,j, val=0, best;
+    int i,j=0, val=0, best;
     char *end;
     int vals[10];
 
--- a/contrib/cidmap/src/mapjapan1.c
+++ b/contrib/cidmap/src/mapjapan1.c
@@ -39,7 +39,7 @@ return( 5 );
 }
 
 static int getnth(char *buffer, int col, int *mults) {
-    int i,j, val=0, best;
+    int i,j=0, val=0, best;
     char *end;
     int vals[10];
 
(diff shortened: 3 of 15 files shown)


JoesCat:
Hi Julien,
Can you check this, I believe there is an additional error higher-up in print.c at line 2179...

references [fontforge/print.c:2635] and http://en.wikipedia.org/wiki/List_of_ISO_639-1_codes and _simplelatnlangs[] where SLY=Slovene=sv (correct), and Czech=cs={but not CSY}

serval2412:
Hi JoesCat,

I'm not sure because _simplelatnlangs patterns seem to have to match with struct langsamples
If not or if both should be changed, in this case according to https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes, Slovene should be 'SLV' (not 'SLK' which is Slovak), Poland should 'POL' (not 'PLK'), Serbian should be 'SRP' not 'SRB'... In brief, it seems everything should be reviewed here :-(
I think a specific patch for these should be done (perhaps by an expert of l10n)

JoesCat:
Thanks for the fixes!

Adding yet another reference codepage: https://www.microsoft.com/resources/msdn/goglobal/default.mspx

serval2412:
I don't use Microsoft websites when I'm looking for standards ! :-)

PR #2697, opened 2016-04-16
Fix spelling: elipse->ellipse

jtanx:


frank-trampe:
@jtanx, I think that we ought to refrain from changing the names in the code. It doesn't give us a lot of upside, and it can make merges to/from other FontForge branches more difficult.

PR #2700, opened 2016-04-17
Update Russian translation

prokoudine:
Updated ru.po to match some of the code/feature changes over the past 3 years.


frank-trampe:
I notice that a lot of relative paths change. Is there a particular reason for the change?

prokoudine:
Apparently at some point in the past the 'FontForge.pot' target was moved to the Makefile inside the 'po' directory. This might be it.

JoesCat:
Hi Frank,
there may appear to be large changes, but in actuality there would be few translation additions for this... here's a bit of history... last translation update for ru.po was 2012
Directory fontforge was split into fontforge and fontforgeexe Oct 27,2013
There was also talk of how FontForge's po files were very much out of shape, so I ran a bunch of them thru approx Nov 27,2013. This in reality accumillated new pointers, but no "intelligence" and "translation" of the sort Alexander is able to provide here.
Fontforge.po was moved Dec 16,2013 and some extra info added to help make it easier to update or create new po files too.
I expect you may see a large shift if you update any/all po files older than Dec 16,2013...or before for some extra margin... 20140101

I would expect the "real" improvements now with Alexander's update here, since the last "translation" update is really more like 2012 and not Dec2013.

Good question Frank, hopefully the above history helps answer it.

JoesCat:
Thanks Alexander - much appreciated!
Thanks Frank for checking details

PR #2702, opened 2016-04-19
Parsing of GlyphOrderAndAliasDB encoding

felipesanches:
This piece of code implements parsing of GlyphOrderAndAliasDB encoding.
It is equivalent to #2467 but it includes only the set of changes that are not merely cosmetic (such as code cleanup).

### DIFF
--- a/fontforge/encoding.c
+++ b/fontforge/encoding.c
@@ -576,6 +576,69 @@ return( NULL );
 return( item );
 }
 
+/* Parse the GlyphOrderAndAliasDB file format */
+static Encoding *ParseGlyphOrderAndAliasDB(FILE *file) {
+    char* names[1024];
+    char buffer[256];
+    int32 encs[1024];
+    Encoding* item = NULL;
+    size_t i, any;
+    int max, enc;
+
+    for (i=0; i < sizeof(names)/sizeof(names[0]); ++i) {
+        encs[i] = -1;
+        names[i] = NULL;
+    }
+
+    max = -1; any = 0;
+    i = 0;
+    while (fgets(buffer, sizeof(buffer), file) != NULL) {
+        max = i;
+
+        int tab, tab2;
+        for (tab = 0; tab < sizeof(buffer) && buffer[tab] != '\t'; tab++);
+        if (tab < sizeof(buffer))
+            buffer[tab]='\0';
+
+        for (tab2 = tab+1; tab2 < sizeof(buffer) && buffer[tab2] != '\t'; tab2++);
+        if (tab2 < sizeof(buffer))
+            buffer[tab2]='\0';
+
+        if (strcmp(buffer, ".notdef") == 0) {
+            encs[i] = -1;
+        } else if ((enc = UniFromName(buffer, ui_none, &custom)) != -1) {
+            encs[i] = enc;
+
+            /* Used not to do this, but there are several legal names */
+            /*  for some slots and people get unhappy (rightly) if we */
+            /*  use the wrong one */
+            names[i] = copy(&buffer[tab+1]);
+            any = 1;
+        } else {
+            names[i] = copy(&buffer[tab+1]);
+            any = 1;
+        }
+        i++;
+    }
+
+    if (max != -1) {
+        if (++max < 256) max = 256;
+        item = calloc(1,sizeof(Encoding));
+        char* buf = strdup(_("Please name this encoding"));
+        item->enc_name = ff_ask_string(buf, "GlyphOrderAndAliasDB", buf);
+        item->char_cnt = max;
+        item->unicode = malloc(max*sizeof(int32));
+        memcpy(item->unicode, encs, max*sizeof(int32));
+
+        if (any) {
+            item->psnames = calloc(max, sizeof(char *));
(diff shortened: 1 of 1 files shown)


JoesCat:
Thanks Felipe for redoing this.

Some suggestions to look at:
```c
  for (i = sizeof(names)/sizeof(names[0])-1; i >= 0; --i) {
       encs[i] = -1;
       names[i] = NULL;
  }
```
The above looks crazy, but the start point gets computed once vs the end point getting repeatedly computed (for inefficient compilers, but hopefully today's compilers are smarter enough to compute the end-point just once). Also, some processors have branch instructions like 'test' and 'jge', making the end-point a little quicker too). No worries, about your code, but just see it as a speed improvement (especially if dealing with interpretive languages like python).

The next suggestion I suggest as worth revisiting:
```c
  while (fgets(buffer, sizeof(buffer), file) != NULL) {
```
Anywhere you have a file read or write, don't use !=NULL, but preferably check for '<0' because there are other errors that cause problems, for example, file-lock, out of file space, corrupt/unreadable input, etc, etc, etc....
This I recommend revisiting.

...otherwise....comments aside...

Thanks for your help/improvements.
Joe

jtanx:
Hmm. That comment on `fgets` doesn't make sense - [documentation](http://linux.die.net/man/3/fgets) explicitly states it returns either the pointer passed to it (`buffer` in this case), or NULL on error. So checking against NULL is correct.

JoesCat:
On April 19, 2016 08:07:53 AM Jeremy Tan wrote:
> Hmm. That comment on `fgets` doesn't make sense -
> [documentation](http://linux.die.net/man/3/fgets) explicitly states it
> returns either the pointer passed to it (`buffer` in this case), or
> NULL on error. So checking against NULL is correct.

My err, was thinking of fgetc() and fread(), fwrite().
Here's fgets: http://www.drpaulcarter.com/cs/common-c-errors.php#4.4

fontforge/encoding.c
```diff
@@ -576,6 +576,69 @@ return( NULL );
...
+
+    max = -1; any = 0;
+    i = 0;
+    while (fgets(buffer, sizeof(buffer), file) != NULL) {
+        max = i;
+
+        int tab, tab2;
+        for (tab = 0; tab < sizeof(buffer) && buffer[tab] != '\t'; tab++);
+        if (tab < sizeof(buffer))
+            buffer[tab]='\0';
```
frank-trampe: If the input item is 256 long, it seems that we would feed an unterminated string into UniFromName.

fontforge/encoding.c
```diff
@@ -576,6 +576,69 @@ return( NULL );
 return( item );
 }
 
+/* Parse the GlyphOrderAndAliasDB file format */
+static Encoding *ParseGlyphOrderAndAliasDB(FILE *file) {
+    char* names[1024];
+    char buffer[256];
```
frank-trampe: I am not familiar with the file format. Is this limit of 255 characters in the specification?

fontforge/encoding.c
```diff
@@ -576,6 +576,69 @@ return( NULL );
...
+
+        if (strcmp(buffer, ".notdef") == 0) {
+            encs[i] = -1;
+        } else if ((enc = UniFromName(buffer, ui_none, &custom)) != -1) {
+            encs[i] = enc;
+
+            /* Used not to do this, but there are several legal names */
+            /*  for some slots and people get unhappy (rightly) if we */
+            /*  use the wrong one */
+            names[i] = copy(&buffer[tab+1]);
```
frank-trampe: I don't see over-run protection for names.

fontforge/encoding.c
```diff
@@ -576,6 +576,69 @@ return( NULL );
...
+            any = 1;
+        }
+        i++;
+    }
+
+    if (max != -1) {
+        if (++max < 256) max = 256;
+        item = calloc(1,sizeof(Encoding));
+        char* buf = strdup(_("Please name this encoding"));
+        item->enc_name = ff_ask_string(buf, "GlyphOrderAndAliasDB", buf);
```
frank-trampe: I don't think that ff_ask_string frees the title and message strings.

fontforge/encoding.c
frank-trampe: Wait. Never mind that.

frank-trampe: But it wouldn't kill us to have some comments.

JoesCat:
Seems I merged this too early.
@felipesanches any fixes? answers?

felipesanches:
I'll review these things right now. Thanks for pointing out.

frank-trampe:
Ping, @felipesanches.

fontforge/encoding.c
```diff
@@ -576,6 +576,69 @@ return( NULL );
...
+            any = 1;
+        } else {
+            names[i] = copy(&buffer[tab+1]);
+            any = 1;
+        }
+        i++;
+    }
+
+    if (max != -1) {
+        if (++max < 256) max = 256;
```
frank-trampe: Why do we pad the encoding?

Issue #2710, opened 2016-04-27
Quit with custom status don't work any longer


jimmywarting:
Seems like something has broken
Before I could exit with a custom exit status

Mac OS

Based on sources from 10:51 PDT 13-Sep-2015-D.
```
$ fontforge --lang ff -c 'Quit(2)'
$ echo $?
$ 2
```

But now with the newer version it always returns 0
Based on sources from 16:16 BST  7-Apr-2016-D.
```
$ fontforge --lang ff -c 'Quit(2)'
$ echo $?
$ 0
```

frank-trampe:
Just to clarify, this isn't the .app bundle, right?

jimmywarting:
erm, no... I'm scripting (using terminal)

JoesCat:
I've been trying to move the exit() out of sub/sub/sub functions, and I must have missed a copy-n-paste.
I think it may be sub instead of c-> but will have to check (search for quit in fontforge/scripting.c).

I'll look after this on the weekend.

jtanx:
I'd recommend adding this (calling Quit() with non-zero return value) to the unit tests.

JoesCat:
On April 29, 2016 07:48:52 AM Jeremy Tan wrote:
> I'd recommend adding this (calling Quit() with non-zero return value) to
> the unit tests.

I'm going to try it on my old system first incase something changed.

PR #2716, opened 2016-04-30
Check for error on return from native scripting, fixes issue #2710

JoesCat:
Missed this earlier as part of a larger effort to move exit() out of
lower sub function levels (eventually to exit() in main() ).

Fixes issue "Quit with custom status don't work any longer" #2710


frank-trampe:
It tests correctly for me.

PR #2720, opened 2016-05-07
Update to use new toolbar icons

monkeyiq:
Now uses a 'selected' image to show the button background, allows most flexibility because one might decide to change the colors of the image itself rather than just stamp a background behind the active button.

I've also tested against 300% sized buttons to see that size doesn't matter. You should make sure that all buttons are the same size, but if you want 64x64 icons then that should work fine.


davelab6:
@monkeyiq wow awesome!!!! :D 

(However, it would be great to leave PRs open for a day or two so that @frank-trampe and @adrientetar and @khaledhosny and others can review :)

davelab6:
cc @larsenwork @pathumego :D

fontforgeexe/cvpalettes.c
```diff
@@ -819,100 +817,281 @@ static void CVPolyStar(CharView *cv) {
...
+
+static int getSmallIconsHeight()
+{
+    return GIcon_smallpointer.u.image->height;
+}
+
+static int getToolbarWidth( CharView *cv ) 
+{
+    int cache = 0;
+    if( !cache ) {
```
jtanx: This looks like `cache` should be static.

fontforgeexe/cvpalettes.c
```diff
@@ -819,100 +817,281 @@ static void CVPolyStar(CharView *cv) {
...
+	    cache = MAX( cache, buttons[0][i][0]->u.image->width + buttons[0][i][1]->u.image->width );
+	}
+    }
+    return cache;
+}
+
+static int getToolbarHeight( CharView *cv ) 
+{
+    int cache = 0;
+    if( !cache ) {
```
jtanx: Same here.

jtanx: Except I don't see the point of caching this value.

davelab6:
thanks for the review @jtanx ! :)

fontforgeexe/cvpalettes.c
```diff
@@ -819,100 +817,281 @@ static void CVPolyStar(CharView *cv) {
...
+static int getSmallIconsHeight()
+{
+    return GIcon_smallpointer.u.image->height;
+}
+
+static int getToolbarWidth( CharView *cv ) 
+{
+    int cache = 0;
+    if( !cache ) {
+	GImage* (*buttons)[14][2] = (CVInSpiro(cv) ? spirobuttons : normbuttons);
```
frank-trampe: I don't understand this declaration. What does it mean? Is it a 14 by 2 array of `GImage**`?

fontforgeexe/cvpalettes.c
```diff
@@ -819,100 +817,281 @@ static void CVPolyStar(CharView *cv) {
...
+				      int selected,             // is this button the active tool
+				      void* udata );            // user data
+
+
+/**
+ * Visit every button in the toolbar calling the function 'v' with a
+ * collection of interesting state data.
+ */
+static void visitButtons( CharView* cv, visitButtonsVisitor v, void* udata ) 
+{
```
frank-trampe: It would be very helpful to have comments here explaining what happens. The logic is not particularly obvious.

fontforgeexe/cvpalettes.c
```diff
@@ -819,100 +817,281 @@ static void CVPolyStar(CharView *cv) {
...
-	{
-	  // These are from charview.c.
-	  extern int cvbutton3d; // Default 1.
-	  extern Color cvbutton3dedgelightcol; // Default 0xe0e0e0.
-	  extern Color cvbutton3dedgedarkcol; // Default 0x707070.
-	  if (cvbutton3d) {
-	    GDrawDrawLine(pixmap,j*27,i*27,j*27+25,i*27,norm?cvbutton3dedgelightcol:cvbutton3dedgedarkcol);
-	    GDrawDrawLine(pixmap,j*27,i*27,j*27,i*27+25,norm?cvbutton3dedgelightcol:cvbutton3dedgedarkcol);
-	    GDrawDrawLine(pixmap,j*27,i*27+25,j*27+25,i*27+25,norm?cvbutton3dedgedarkcol:cvbutton3dedgelightcol);
-	    GDrawDrawLine(pixmap,j*27+25,i*27,j*27+25,i*27+25,norm?cvbutton3dedgedarkcol:cvbutton3dedgelightcol);
```
frank-trampe: What happened to this logic? How are we drawing the relief now?

fontforgeexe/cvpalettes.c
davelab6: @monkeyiq what do you think?

fontforgeexe/cvpalettes.c
davelab6: @monkeyiq what do you think?

fontforgeexe/cvpalettes.c
davelab6: @monkeyiq can you explain this please? :)

fontforgeexe/cvpalettes.c
davelab6: @monkeyiq can you add comments for this please?

fontforgeexe/cvpalettes.c
davelab6: I believe we are not drawing the relief, and assuming that it is drawn into the PNGs used for the toolbar. @monkeyiq please explain :)

fontforgeexe/cvpalettes.c
monkeyiq: it does look like I was heading for a static variable here. Not sure if all the conditions are met to make it static or not though.

fontforgeexe/cvpalettes.c
monkeyiq: either way, just trim the if( !cache ) condition and leave it as non static.

fontforgeexe/cvpalettes.c
monkeyiq: this is likely a block of code that was doing something relating to the dimensions already, but placed into a visitor.

fontforgeexe/cvpalettes.c
monkeyiq: every button has two images, one for clicked, one for non clicked. If you want the relief then the images contain them. If you want to use the color to show selected tool and avoid the clutter of the boxes then you are free to do that too.

fontforgeexe/cvpalettes.c
monkeyiq: it is a definition like the one that was there to declare the toolbar (spirobuttons and normbuttons). I've just used buttons to allow sniffing around without complicating the code with inSpiro() checks throughout the rest of the function.

fontforgeexe/cvpalettes.c
```diff
@@ -819,100 +817,281 @@ static void CVPolyStar(CharView *cv) {
     ps_pointcnt = temp;
 }
 
+/* Note: If you change this ordering, change enum cvtools */
```
frank-trampe: @monkeyiq, why did we pull these declarations out of the ToolsExpose function?

Issue #2726, opened 2016-05-19
segfault when generating truetype/opentype fonts


cbaakman:
On fontforge 14:57 GMT 31-jul-2012 version,

for a particular fontforge file, a segfault always occurs when I try to generate truetype fonts or other formats. SVG fonts still work however.

The SFD file contains glyphs for latin as well as non-latin characters and references are also used. Furthermore, it has a horizontal kern-by-classes table.

If there's a mistake in my font setup, I'd like to know what it is so that I can correct it, rather than getting a segfault.

If you (the developer) need the SFD file, I could send it.

JoesCat:
There are over 9800 patches and updates added to FontForge of 2012 and the current HEAD, some of them minor, some fairly major.
Please try a more recent version first.

cbaakman:
OK sorry about that! I thought the apt-get version was the latest stable.

Yesterday (19th of may 2016, evening) I built the most recent master branch version on ubuntu 14.
It has the same segfault however.

JoesCat:
I'm way overloaded with other things that need doing and hope someone else has the time to look at this, ...however, best to send a copy. If it needs to be private, zip it, and you can find my email in the git logs, or my profile page. If it can be posted for anyone, you'll potentially have more eyes looking at it.

JoesCat:
Hi @cbaakman,
Okay, I tried:
File->Generate Fonts.
NOTE: I do have checked-on = "Validate before saving"
I then selected ttf and then pressed "Generate" (ttf file)

Results:
"The font contains errors
Self intersecting
Wrong direction
Missing Points at Extrema
Would you like to review the errors or save the font anyways?

Yes - Review (results):
I noted 40 errors that require fixing.
All errors show "Missing Points at Extrema"

I tried this again with "Validate before saving" turned-off this time.
and it creates segfault.

Right now, the easiest/quickest fix for me would be to simply refuse saving without having all the problems with this font fixed first. This means you (and everyone else) would be forced to fix their errors before being able to export a bad ttf file. Before I add such a fix, I'd rather understand and go through this code in more detail. So at this point it's considered won't fix yet, plus a bad font file that needs fixing first.

From your point of view, turn on the check "Validate before saving", and please fix all errors listed until you have zero errors listed. After you have fixed all the errors listed, and if this font file still creates a segfault, I'd like to check again, but before this point in time, it's too early for me to do anything with this yet.

cbaakman:
[...]

PR #2727, opened 2016-05-24
Prevent utf8_ildb from testing a NULL non-existant string

JoesCat:
Partial fix for issue #2726. More fixes ARE required to tottf.c to make it check for and recover from errors.


frank-trampe:
So are we waiting for more commits?

frank-trampe:
Ping.

JoesCat:
This "symptom" of the problem was easy to find. The real problem causing the issue is buried elsewhere and requires a bit more chasing. In summary - this issue is still unresolved for now.

Issue #2643, opened 2016-03-12
license/authors clarifying


henrich:
Hi,
 I'm one of fontforge package maintainer in Debian.
 We're still staying old fontforge version because update package is rejected due to lack of license clarifying. Could anyone help those files license/authors information, please?
 
> 
> Files: fontforge/ofl.*
> Copyright: ?
> License: OFL-1.1
> 
> Files: fontforge-debug-lldb.sh
>        fontforge/acorn2sfd.c
>        fontforge/autowidth.h
>        fontforge/bezctx_ff.c
>        fontforge/bezctx_ff.h
>        fontforge/crctab.c
>        fontforge/encoding.h
>        fontforge/flaglist.c
>        fontforge/flaglist.h
>        fontforge/is_LIGATURE.h
>        fontforge/langfreq.c
>        fontforge/lookups.h
>        fontforge/othersubrs.c
>        fontforge/pua.c
>        fontforge/strlist.c
>        fontforge/strlist.h
>        fontforge/threaddummy.c
>        fontforge/usermenu.h
>        fontforge/zapfnomen.c
>        gdraw/gkeysym.c
>        gdraw/xkeysyms_unicode.c
>        gdraw/genkeysym.c
>        gdraw/gxcdrawP.h
>        gutils/divisors.c
>        tests/link_test.c
>        inc/dynamic.h
>        inc/utype.h
>        inc/chardata.h
>        inc/carbon.h
>        pyhook/*
>        win/make-package.sh
>        win/expand.sh
>        win/fetch.sh
>        gtkui/gwwvprogress.h
>        gtkui/gwwv.h
>        gtkui/gwwvask.h
>        gtkui/gwwvmenubar.h
>        gtkui/callbacks.h
>        gtkui/gwwvmenubar.c
>        gtkui/stubs.c
>        po/toengb.c
>        osx/create-osx-app-bundle.sh
>        osx/FontForge.app/Contents/MacOS/fcprogress.pl
>        plugins/gb12345.c
>        fonttools/showttf.c
>        fonttools/stripttc.c
>        fonttools/findtable.c
>        fonttools/pfadecrypt.c
>        Unicode/ArabicForms.c
>        Unicode/alphabet.c
>        Unicode/backtrns.c
>        Unicode/cjk.c
>        Unicode/combiners.h
>        Unicode/unialt.c
>        Unicode/utype.c
>        cidmapsrc/*.c
>        nodejs/collabwebview/server.js
[...]

Issue #2732, opened 2016-06-05
Debian Packaging: Sane options for configure for building a distribution package


copyninja:
Hi,

I'm updating the Debian package of fontforge and I'm now wondering what are the options I need to enable for configure phase. The old version from 2012 had these options.

``` make
DEB_CONFIGURE_EXTRA_FLAGS = --prefix=/usr \
              --libdir=\$${prefix}/lib \
              --mandir=\$${prefix}/share/man \
              --with-regular-link \
              --enable-devicetables \
              --enable-type3 \
          --with-freetype-src=$(CURDIR)/freetype \
          --with-freetype-bytecode
```

Some of these options might not be even present today. So can any one suggest me a default set of options I need to use, considering we are shipping a general package to all font developers.

JoesCat:
Hi @copyninja,
Please note that going from 20120731-b1 to 20140101 (and today), FontForge has had a very large modification to the make build system. (1) The easier to recognize change is going from {./configure make} used in 20120731, to {./bootstrap, ./configure, make} used in the latest releases. (2) The less noticed, but I think also, fairly important change is having had the dynamic linking mechanism stripped-out (this isn't all too important to source built distros, but is important to binary distributed distros).

(1) In terms of the ./bootstrap system, I would like to eventually see it get reduced to autoconf -i, automake, but we would need to put more time into the make structure, and most of our time has been concentrated on bug fixing and improvements in the FontForge source code. If you can successfully get FontForge built using autoconf automake, please aim for that first. If you cannot, then do use bootstrap.

(2) In terms of building a general package for all font developers, and if you have a binary package distribution system, since dynamic linking isn't available as in the older 2012, you will need to increase the dependencies to keep font designers happy and minimize issues/complaints/false-bugs being mentioned. I would suggest to run ./configure --help and look at the --without- options and force several of them into --with- so that these extras are checked for and added on purpose.
Ones that come to mind right away are --with-giflib --with-libpng, --with-libjpeg (to allow designers to create webpage bitmaps). There have been numerous bugs and issues with libspiro, and a few with libuninameslist, so I think both of these are desired features for font developers --with-libspiro --with-libuninameslist. If you are making a graphical FontForge, a substantial amount of fixing was added for cairo, so you may want to add --with-cairo. When you are in graphical mode, and press the 'F1' key for help, you will open a web-browser help, so you may need to add an additional reference for the html help documentation directory.

[...]

Issue #1747, opened 2014-09-24
Fragment joining failure in stroke expander + bad instruction arguments


Pecita:
Dear developers, 

The font that I build with FontForge is corrupt. 
It is perfectly usable when installed but is rejected by ots when embedded in a web page. 
I do not know how to fix this problem.

The material related with this bug is here: http://pecita.eu/bugCFF/

*http://pecita.eu/bugCFF/version5.xhtml*
"downloadable font: CFF : failed to parse table (font-family: "Pecita" style:normal weight:normal stretch:normal src index:0)
source: http://pecita.eu/bugCFF/Pecita.otf police.css
downloadable font: rejected by sanitizer (font-family: "Pecita" style:normal weight:normal stretch:normal src index:0)
source: http://pecita.eu/bugCFF/Pecita.otf"

(Debian testing)
fontforge -version
Copyright (c) 2000-2012 by George Williams.
 Executable based on sources from 14:57 GMT 31-Jul-2012-ML.
 Library based on sources from 14:57 GMT 31-Jul-2012.
fontforge 20120731
libfontforge 20120731-ML

Regards,
Philippe Cochy

davelab6:
@frank-trampe did you work on those 2 issues?

frank-trampe:
I did work on the fragment joining problem. I can't remember what happened.

Pecita:
Steps to reproduce:
1) local copy of http://pecita.eu/bugCFF/ (compile.eu+PecitaSource.sfd)
2) chmod +x compile.eu then execute

Result:
./compile.pe
Copyright (c) 2000-2014 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Based on sources from 21:13 UTC 28-Feb-2015-ML-D.
 Based on source from git with hash: a875b4d41cc48aded120aa060ae520cc0113a7a7
Some fragments did not join in thornSome fragments did not join in afii10024[...]Validation error

TY, Frank. Please do not release as long as this bug is not fixed.

davelab6:
@Pecita are you saying this worked okay in the last 2012 release?

Pecita:
Yes!

2015-07-13 7:18 GMT+02:00 Dave Crossland <notifications@github.com>:

Pecita:
[...]

PR #2735, opened 2016-06-25
Copyrights concerning Issue 2643 and Issue 2732

JoesCat:
Work in progress to "help" resolve issue #2643 so that issue #2732 can proceed.
Basically a lot of reading, plus implying/adding authors who didn't add their names to the credits, and hopefully acknowledged as "okay" by the authors themselves (preferred they did it, but, well...).

### DIFF
--- a/fontforge/bezctx_ff.c
+++ b/fontforge/bezctx_ff.c
@@ -1,5 +1,29 @@
-/* This file written by George Williams to provide a gateway to fontforge */
-/* it it a modification of Raph's bezctx_ps.c */
+/*
+Copyright: 2007 Raph Levien
+License: GPL-2+
+    bezctx_ps.c - ppedit - A pattern plate editor for Spiro splines.
+Copyright: 2007 George Williams
+    bezctx_ff.c - Provide a gateway into FontForge with added modification.
+Contributions: Yuri Chornoivan, Jonathan Hanna, Joe Da Silva, Reuben Thomas,
+    Frank Trampe
+
+This program is free software; you can redistribute it and/or
+modify it under the terms of the GNU General Public License
+as published by the Free Software Foundation; either version 2
+of the License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+02110-1301, USA.
+
+*/
+
 #include <basics.h>
 #include <stdio.h>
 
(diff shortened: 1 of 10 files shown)


McSinyx:
I think line 12 in `fontforge/ofl.c` is missing a comma at the end.

JoesCat:
Thanks @McSinyx, you're absolutely right.
I looked at it earlier, and the warning from github said something to the effect of ""Powering down the server NOW" or something similar, so I though it was a glitch due to maybe some sort of server patching. My error for not having compiled before committing this copy-n-paste...after all, it's only comments, what could possibly go wrong here?
;-P
Thanks for the catch.

fontforge/bezctx_ff.c
```diff
@@ -1,28 +1,52 @@
-/* This file written by George Williams to provide a gateway to fontforge */
-/* it it a modification of Raph's bezctx_ps.c */
+/*
+Copyright: 2007 Raph Levien
+License: GPL-2+
+    bezctx_ps.c - ppedit - A pattern plate editor for Spiro splines.
+Copyright: 2007 George Williams
+    bezctx_ff.c - Provide a gateway into FontForge with added modification.
+Contributions: Yuri Chornoivan, Jonathan Hanna, Joe Da Silva, Reuben Thomas,
+    Franke-Trampe
```
frank-trampe: Name is wrong.

fontforge/bezctx_ff.c
```diff
@@ -1,28 +1,52 @@
...
+02110-1301, USA.
+
+*/
 #include <basics.h>
 #include <stdio.h>
 
 #ifndef _NO_LIBSPIRO
 #include "bezctx_ff.h"
-#include "fontforgevw.h"	/* For LogError, else splinefont.h */
+#include "fontforgevw.h" /* For LogError, else splinefont.h */
```
frank-trampe: Could we just fix the header and not mess with the whitespace and minor verbiage? I don't think it's worth the changelog/blame clutter.

fontforge/bezctx_ff.c
JoesCat: On July 19, 2016 02:42:12 PM Frank Trampe wrote:
> Name is wrong.

Sorry about that, should have double checked - my error.

fontforge/bezctx_ff.c
JoesCat: On July 19, 2016 02:42:56 PM Frank Trampe wrote:
> Could we just fix the header and not mess with the whitespace and minor
> verbiage? I don't think it's worth the changelog/blame clutter.

Was thinking about the older machines like MIPs etc that really can't get 
much beyond 80chars and seem to struggle with more modern structures (as 
per several of the edits I had to throw into libspiro to move it closer to 
C89).

I'll have to put is_LIGATURES.* in a seperate thread since it's going to 
have to be more than a plain header fix.

JoesCat:
will continue further...later...

copyninja:
@JoesCat Thanks for this. Btw I've uploaded fontforge to the Debian Experimental for now. We just need to make sure everything builds fine before uploading it to unstable. We are also waiting for the resolution  of #1747 before we can upload to unstable.

Issue #2025, opened 2015-01-02
Stroke expansion regressions


joshuakraemer:
Until now, I have used Fontforge version 2012-07-31.  I tried version 2014-11-26 now and found two regressions in the stroke expansion tool.  The attached images show the correct expansion result from the old version on the left, and the faulty result from the new version on the right.  The green line shows the original stroke in the background.  The strokes have been expanded with the "round" setting for line cap and line join.  The problems are the following:

(1) At some line joins, the expanded stroke is cut off:

![3](https://cloud.githubusercontent.com/assets/10373011/5597013/4f72147a-929b-11e4-8264-178a9a8549be.png)

(2) At some line joins, the expanded stroke is distorted:

![5](https://cloud.githubusercontent.com/assets/10373011/5597012/4f7214b6-929b-11e4-8e80-4e87fabddc56.png)
![w](https://cloud.githubusercontent.com/assets/10373011/5597014/4f768b72-929b-11e4-80c5-48126b69c1a0.png)

dimitre:
Second this issue, same thing happening here

dimitre:
hello @joshuakraemer did you find any way of avoiding this issue? I'm keeping an 2012 verion installed in an old system.

frank-trampe:
@dimitre, is there any chance that you could bisect the problem? I'm stretched rather thin right now, and this is a lot easier for me to fix if I know which commit caused it.

dimitre:
I'll seek information on how to do it and try it @frank-trampe

joshuakraemer:
@ <https://github.com/dimitre>dimitre <https://github.com/dimitre>,
unfortunately not. I'm keeping an old version, too.

jtanx:
Would you be able to provide a sample to test this on?

dimitre:
Great @jtanx I've just generated a sample to test here:
The last two lines are just useful if you are in OS X so it opens the generated file. You will recognize the regression because this generates a bump in the center of W letter. 

```python
# -*- coding: UTF-8 -*-
'''
cd ~/Desktop/WORK_DMTR/syntype_python/
python regression.py

'''

import fontforge
import random
font = fontforge.font()
font.encoding = "UnicodeBmp"
font.familyname = 'test' + str(random.randint(0,99999))

caracter = 'W'
font.createMappedChar(caracter).unicode
pen = font[caracter].glyphPen()
contorno = fontforge.contour()
aresta = 100.0
[...]

PR #2739, opened 2016-07-09
Attempted fix of #2025 (bug in expand stroke)

jtanx:
4f883fd broke expand stroke (see #2025) while attempting to fix an infinite loop found in regression testing (test1003.py, circular stroke as applied to backslash character in StrokeTests.sfd).

This attempts to provide a better fix for the infinite loop condition while not causing the effects of #2025.

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -708,7 +708,7 @@ return;		/* Essentially colinear */ /* Won't be perfect because control points l
 	    p->needs_point_left = p->needs_point_right = false;
 	    p->left_hidden = bends_left;
 	    p->right_hidden = !bends_left;
-	    if ( rot.x<=fabs(diff_angle.x) ) {
+	    if ( rot.x<=diff_angle.x || (diff_angle.x <= -1 && rot.x <= -0.999999) ) { /* close enough */
 		p->right = done.right;
 		p->left = done.left;
 		p->needs_point_left = p->needs_point_right = true;


fontforge/splinestroke.c
```diff
@@ -708,7 +708,7 @@ return;		/* Essentially colinear */ /* Won't be perfect because control points l
 	    p->needs_point_left = p->needs_point_right = false;
 	    p->left_hidden = bends_left;
 	    p->right_hidden = !bends_left;
-	    if ( rot.x<=fabs(diff_angle.x) ) {
+	    if ( rot.x<=diff_angle.x || (diff_angle.x <= -1 && rot.x <= -0.999999) ) { /* close enough */
```
frank-trampe: I have not read all of the code carefully, but this seems like an asymmetric fix to me. In particular, it seems odd to me that the negative case have a wider threshold than the positive case.

dimitre: maybe the command is doing two different things, expanding the path correctly and avoiding the previous issue. 
I think it worths merging because the expand stroke defect is present in all versions since 2012.

davelab6:
This is a very serious problem and 'close enough' sounds good to me. I urge the merging of this :)

fontforge/splinestroke.c
frank-trampe: Pinging @jtanx.

frank-trampe:
We can merge it, but I don't want to lose the opportunity to get a good fix before @jtanx forgets the results of his digging.

fontforge/splinestroke.c
jtanx: Looking at the code, it appears the ranges are [-1, 1] for `diff_angle.x` and [-1, 1] for `rot.x`. Now, `rot.x` will always [start in the range of [0, 1]](https://github.com/jtanx/fontforge/blob/e7ff15b015c72cef235fe4c55449c54f59125559/fontforge/splinestroke.c#L684). In the subsequent loops, `rot.x` is gradually decreased until it is less than `diff_angle.x`, which is the termination condition.

However, for what I presume is due to rounding errors, `diff_angle.x` can never reach -1. Instead, it will approach close to -1 before increasing again (it oscillates in the range of (-1,1)), hence causing the infinite loop.

In this case, I think an asymmetric fix (instead of say, `fabs(diff_angle.x-rot.x) < 0.0001`) is OK, since the opposite case of `diff_angle.x=1` won't cause the same issue.

jtanx: But what I do find strange this same code [doesn't cause an infinite loop](https://github.com/fontforge/fontforge/issues/2025#issuecomment-231816669) in the 2012 version.

PR #2742, opened 2016-07-16
Write device name as empty string in FON file header, issue #2741

### LINKED ISSUE #2741: FON file is displayed in Windows with multiple bogus face names
I created a bitmap font called Mungre.  Each glyph has several sizes.  I exported it as a FON file.

I installed the FON file in Windows 10.  The installation process worked as normal.

I tried to select my font in the standard Windows font selection dialog box (in notepad, not that it matters).  Each different point size of the font appeared with a different face name.  These face names were "Q", "R", "S", etc.

I've made a fix for the problem and hope to make a pull request shortly.

mungre:
### Motivation and Context
Closes #2741

### Types of changes

### Description
The FON file header is constructed by copying data from temporary FNT
files.  The code was previously copying five bytes too many which
left the device name non-empty.  The device name ran into the face
name, causing Windows to interpret whatever junk came next as the
face name.

### Final checklist

### DIFF
--- a/fontforge/winfonts.c
+++ b/fontforge/winfonts.c
@@ -1035,11 +1035,17 @@ return( false );
     for(res = first_res, i = 0; i < num_files; i++, res++) {
         lputshort(fon,res);
 
+        /* The first 0x6D bytes of the FNT file and the FONTDIRENTRY match */
         rewind(fntarray[i]);
-        fread(buf, 0x72, 1, fntarray[i]);
+        fread(buf, 0x6D, 1, fntarray[i]);
         fnt_header = (struct _fnt_header *)buf;
         fnt_header->fi.dfBitsOffset = 0;	/* I can ignore endianness here. all is 0 */
-        fwrite(buf, 0x72, 1, fon);
+        fwrite(buf, 0x6D, 1, fon);
+
+        /* FONTDIRENTRY.dfReserved */
+        lputlong(fon,0);
+        /* FONTDIRENTRY.szDeviceName */
+        fputc(0x00, fon);
 
 	fseek(fntarray[i], 0x69, SEEK_SET);
 	off = lgetlong(fntarray[i]);
--- a/fontforge/winfonts.c
+++ b/fontforge/winfonts.c
@@ -729,52 +729,6 @@ typedef struct
      */
 } NE_TYPEINFO;
 
-typedef struct
-{
-    INT16 dfType;
-    INT16 dfPoints;
-    INT16 dfVertRes;
-    INT16 dfHorizRes;
-    INT16 dfAscent;
-    INT16 dfInternalLeading;
-    INT16 dfExternalLeading;
-    CHAR  dfItalic;
-    CHAR  dfUnderline;
-    CHAR  dfStrikeOut;
-    INT16 dfWeight;
-    BYTE  dfCharSet;
-    INT16 dfPixWidth;
-    INT16 dfPixHeight;
-    CHAR  dfPitchAndFamily;
-    INT16 dfAvgWidth;
-    INT16 dfMaxWidth;
-    CHAR  dfFirstChar;
-    CHAR  dfLastChar;
-    CHAR  dfDefaultChar;
-    CHAR  dfBreakChar;
-    INT16 dfWidthBytes;
-    LONG  dfDevice;
-    LONG  dfFace;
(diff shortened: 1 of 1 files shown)


mungre:
The Travis CI Linux build succeeds but the OS X build fails with "Error: Empty installation".

This error comes from the Homebrew package manager and indicates that it can't find any software to install.  However, the actual building of the software seems to succeed.

I compared the most recent successful OS X log to this failing log.  One obvious difference is that the destination prefix has changed from "/usr/local/Cellar/fontforge/HEAD" to "/usr/local/Cellar/fontforge/HEAD_1".  Though this might be a red herring.

Anyway, I'm nearly certain this failure has nothing to do with my commit, but I don't have a Mac so I can't easily investigate it.

davelab6:
@mungre thanks for this! I have a mac and so does @pathumego and @frank-trampe and @monkeyiq but Pathum and I are currently unable to build from sources on OS X. I'm hoping @frank-trampe or @monkeyiq can supply build instructions over on the other PR

frank-trampe:
@mungre, could you point out the section of the FON specification that pertains to this? I want to verify that we are doing this properly and have it on record on GitHub.

mungre:
@frank-trampe  Of course.

A FON file is a codeless 16-bit Windows executable file.  The font directory is stored in a resource of type RT_FONTDIR.

This consists of a [FONTGROUPHDR][1] which contains a sequence of [DIRENTRY][2] plus [FONTDIRENTRY][3].

The FONTDIRENTRY structure strongly resembles the [FNT file format][4].

By inspection, the fields from dfVersion through dfFace are the same.  By adding up, these common fields occupy 0x6D bytes.

After these 0x6D bytes the FONTDIRENTRY has the dfReserved, szDeviceName and szFaceName fields.

The equivalent bytes in the FNT file start with the dfBitsPointer and dfBitsOffset fields.

[1]: https://msdn.microsoft.com/en-us/library/windows/desktop/ms648015(v=vs.85).aspx
[2]: https://msdn.microsoft.com/en-us/library/windows/desktop/ms648013(v=vs.85).aspx
[3]: https://msdn.microsoft.com/en-us/library/windows/desktop/ms648014(v=vs.85).aspx
[4]: https://support.microsoft.com/en-us/kb/65123

jtanx:
Hmm... now that I look at the structures, I'm wondering if packing is an issue - e.g. if `FONTINFO16` should be declared with `#pragma pack(1)`.

fontforge/winfonts.c
```diff
@@ -1035,11 +1035,17 @@ return( false );
     for(res = first_res, i = 0; i < num_files; i++, res++) {
         lputshort(fon,res);
 
+        /* The first 0x6D bytes of the FNT file and the FONTDIRENTRY match */
         rewind(fntarray[i]);
-        fread(buf, 0x72, 1, fntarray[i]);
+        fread(buf, 0x6D, 1, fntarray[i]);
         fnt_header = (struct _fnt_header *)buf;
         fnt_header->fi.dfBitsOffset = 0;	/* I can ignore endianness here. all is 0 */
```
jtanx: Wouldn't this become redundant since we're only writing out until `dfFace`?

jtanx:
Well that's strange... Inspecting the code, that `_fnt_header` struct looks like it was specifically defined just so that [this](https://github.com/mungre/fontforge/blob/b1f62938d83896133e1baa1873f51fd90db43543/fontforge/winfonts.c#L1042) line could be done. So I guess *if* `fnt_header` was manipulated (within 0x6d bytes), then struct packing would come into play, but because it isn't, then it doesn't matter. Nevermind...

mungre:
@jtanx You're right, that line is now redundant, and fixing the structure packing would have made my original problem go away.

However, zeroing a field in the source structure because it coincides with a different field that needs to be zero in the destination structure is obscure, so I think the intent of the new code is clearer.

I'll update the pull request (or create a new one, I'm new to this) with the redundant two lines removed.  I guess I should also remove the _fnt_header and FONTINFO16 definitions.  Anyway, this won't happen until tomorrow.

mungre:
The Travis builds have failed due to a (presumably temporary) lack of connectivity to archive.org.

jtanx:
I finally got the travis builds working again.

LGTM. Thanks!

PR #2749, opened 2016-07-21
Zero pfminfo structure to fix FNT file metrics, closes #2748

### LINKED ISSUE #2748: FON file is created with random external leading
Create a bitmap font.  If it has several different sizes the problem will be more obvious.

Export the font as a FON file.  Compare the external leading for each different point size.  They will be set to 0 or 1 (typically), apparently at random.

Export the font again as a second FON file.  The external leading for each point size will again be random, and won't match the first FON file.

I hope to make a pull request with a fix shortly.

mungre:
### Motivation and Context

Closes #2748

### Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

### Description

The _FntFontDump function uses the SFDefaultOS2Info function to populate a pfminfo structure with font metrics that are written to the FNT file.

The SFDefaultOS2Info function requires the pfminfo structure to be zeroed otherwise it can produce bogus values for a few metrics including linegap, which is used to calculate the external leading.

This fix zeroes the structure.  It also fixes the problem for FON files because they are made of FNT files.

With one exception, every other call to SFDefaultOS2Info either uses the pfminfo embedded in the SplineFont structure, or zeroes its local structure first.  The exception is in scripting.c, so I fixed that one too.  I don't know that this one was causing any observable problem.

### Final checklist

- [X] My code follows the code style of this project found [here](https://github.com/fontforge/fontforge/blob/master/CONTRIBUTING.md#coding-style).

### DIFF
--- a/fontforge/scripting.c
+++ b/fontforge/scripting.c
@@ -9346,6 +9346,7 @@ static void handlename(Context *c,Val *val) {
 		val->u.aval = malloc(sizeof(Array));
 		val->u.aval->argc = 10;
 		val->u.aval->vals = malloc((10+1)*sizeof(Val));
+		memset(&pfminfo,'\0',sizeof(pfminfo));
 		SFDefaultOS2Info(&pfminfo,sf,sf->fontname);
 		for ( cnt=0; cnt<10; ++cnt ) {
 		    val->u.aval->vals[cnt].type = v_int;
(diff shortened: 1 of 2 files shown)


adrientetar:
Can you fix SFDefaultOS2Info instead of doing a memset ahead of it?

mungre:
@adrientetar SFDefaultOS2Info supports the initialization of a pfminfo structure which has already been partially initialized.  This is indicated by the hheadset and vheadset members.

If every stack-allocated pfminfo was zeroed (as I said above) then we could do the zeroing in SFDefaultOS2Info.  I double-checked using grep and found that github search had missed both calls to SFDefaultOS2Info in fontforgeexe/fontinfo.c.  One of these uses a stack-allocated pfminfo that is preinitialized to SplineFont.pfminfo with some additional changes.

So we can't distinguish between a stack-allocated pfminfo structure that needs to be zeroed, and one that has already been partially initialized.

PR #2754, opened 2016-07-24
cppcheck: fix Redundant assignment

serval2412:
[fontforgeexe/charinfo.c:4711]: (warning) Redundant assignment of 'vargcd[i][4].gd.popup_msg' to itself.
+ 1 similar case


jtanx:
Hmm, I'm not sure if this is correct. How do you know that the popup messages for those elements should be the same as a previous element's popup message?

PR #2756, opened 2016-07-25
splinestroke.c: Fix buffer overrun in LineJoin

jtanx:
Related issue: #1747


frank-trampe:
@jtanx, could you explain what was going wrong and what the new code does?

jtanx:
@franke-trampe, the issue occurs in the case of the buffer (`c->all`) becoming full (`c->cur == c->max`). In every other case, this condition is checked before attempting to write to the buffer, but in this case it wasn't.

As such, it was possible to overrun the buffer, which causes the program to crash (you can test this on the sample provided in #1747). 

The fix is to therefore check if the buffer is full, and if so, to expand the buffer as necessary (I've copied the style in which it was done similar to other places in the code).

frank-trampe:
That makes sense.

PR #2758, opened 2016-07-27
splinesave.c: Ensure that hintmask is properly set for referenced chars

jtanx:
Note: No need to check for autohinting (i.e. running `SplineCharAutoHint`)
Because autohinting would have already been run on the base char if
needed (in `SplineChar2PS2`), which would have autohinted ref chars.

Related issue #1747


frank-trampe:
I haven't fully tested, but it looks okay.

Issue #2494, opened 2015-09-19
Ligatures


CarMir:
Good morning.
I am creating a new font from an ancient Italian language. I have done all of the glyphs designs and now I have to move on.
In this type of font there is little kerning involved but has some amount of ligature to be done.
My first problem is that this language is written from right to left, and I need to know how to set it up to write the font in this way.
Second problem, how do I build the ligatures, if this language is written from right to left, in which order the letters go placed, do they follow in  the order they are typed?
Third problem, I followed the tutorial for ligatures “compound”, I got into the glyph info windows, and I can not manage to mark the glyph as a ligature, and I am stuck.
Could you please tell me how I mark this gliphs for ligature?
Thank you.

pathumego:
cc @davelab6

khaledhosny:
Whether a script is written from right-to-left or not can not be controlled from OpenType font files, if the characters are given proper BiDi properties in Unicode then conferment layout engines should do the right thing. If we are talking about [Old Italic](https://en.wikipedia.org/wiki/Old_Italic_script) then this is not the case, unfortunately, as Unicode encodes it as a left-to-right script, so you will have to enclose your text with U+202E/U+202C to force it to be set right-to-left.

As for ligatures, your rules should be in the logical (typing) order, not the visual (text display) one.

PR #2762, opened 2016-08-02
makeutype is_Ligatures.c

JoesCat:
Allow makeutype to build an up-to-date is_Ligature.c file, and add several native script functions to allow read access to internal ligature and fraction tables.


Unicode/makeutype.c
```diff
@@ -177,15 +187,53 @@ static void FigureAlternates(long index, char *apt, int normative) {
...
-static void processAssignment(long index,char *pt) {
+static int processAssignment(long index,char *pt,long *flg) {
     static long first=-1;
     long i;
 
-    if ( index>0x11ffff )
-return;
+    if ( index<0 ) return( -1 );
+    if ( index>0x11ffff ) return( 0 );
     ++pt;							/* move past semicolon */
```
frank-trampe: Where does pt originate? Are we sure that it is of non-zero length?

Unicode/makeutype.c
```diff
@@ -128,11 +129,20 @@ unsigned long flags2[MAXC];
 unichar_t alts[MAXC][MAXA+1];
 unsigned long assignedcodepoints[0x120000/32];	/* 32 characters represented per each long value */
 
+int frc, frm, lgc, lgm, vfc, vfm;		/* identify all ligatures and fractions */
+#define LG_MAX 550
+#define FR_MAX 100
+#define VF_MAX 50
+unsigned long ligature[LG_MAX];
+unsigned long fraction[FR_MAX];
+unsigned long vulgfrac[VF_MAX];
```
frank-trampe: Gah. More global variables. I suppose there's no sense in sessionizing just some of them, though.

Unicode/makeutype.c
```diff
@@ -177,15 +187,53 @@ static void FigureAlternates(long index, char *apt, int normative) {
...
+	if ( strstr(pt,"LIGATURE") ) {
+	    /* This codepoint index is a ligature */
+	    /* fprintf( stderr, "ligature[%d]=U+%X, = %s\n", lgm, index, pt ); */
+	    if ( lgm >= LG_MAX ) {
+		fprintf( stderr, LgFrcTooMany, "ligatures", "ligature", lgm, index );
+		return( 5 );
+	    }
+	    ligature[lgm] = index;
+	    if ( index < MAXC ) {
+		*flg |= _LIG_OR_FRAC; lgc++;
```
frank-trampe: Could we get some comments explaining what lgc does?

Unicode/makeutype.c
```diff
@@ -630,6 +683,168 @@ static void dumparabicdata(FILE *header) {
     fclose( data );
 }
 
+static void buildtables(FILE *data,long unsigned int *dt,int s,int m,char *t) {
+    int i,j;
+
+    fprintf( data, "const uint16 %s16[] = {\n", t );
+    for ( i=j=0; i<s; i=i+j ) {
+	fprintf( data, "  0x%04x", dt[i] );
+	for ( j=1; j<8 && i+j<s; ++j ) {
```
frank-trampe: Why `j<8`?

frank-trampe: Oh, wait, I see, j isn't a second dimension, it's just the position on the line.

Unicode/makeutype.c
```diff
@@ -630,6 +683,168 @@ static void dumparabicdata(FILE *header) {
...
+	if ( i+j>=s )
+	    fprintf(data, "};\n\n" );
+	else
+	    fprintf(data, ",\n" );
+    }
+    if ( s<m ) {
+	fprintf( data, "const uint32 %s32[] = {\n", t );
+	for ( i=s; i<m; i=i+j ) {
+	    fprintf( data, "  0x%08x", dt[i] );
+	    for ( j=1; j<5 && i+j<m; ++j ) {
```
frank-trampe: Could we do 4 per line instead? It seems more computer-like.

Unicode/makeutype.c
```diff
@@ -630,6 +683,168 @@ static void dumparabicdata(FILE *header) {
     fclose( data );
 }
 
+static void buildtables(FILE *data,long unsigned int *dt,int s,int m,char *t) {
```
frank-trampe: Could we get a brief description of what the function does and the meanings of the parameters?

Unicode/makeutype.c
```diff
@@ -630,6 +683,168 @@ static void dumparabicdata(FILE *header) {
...
+    }
+}
+
+static void dumpbsearch(FILE *data,int s) {
+    fprintf( data, "static int compare_codepoints%d(const void *uCode1, const void *uCode2) {\n", s );
+    fprintf( data, "    const uint%d *cp1 = (const uint%d *)(uCode1);\n", s, s );
+    fprintf( data, "    const uint%d *cp2 = (const uint%d *)(uCode2);\n", s, s );
+    fprintf( data, "    return( (*cp1 < *cp2) ? -1 : ((*cp1 == *cp2) ? 0 : 1) );\n}\n\n" );
+}
+
```
frank-trampe: Breaking here; I'll be back later.

Unicode/makeutype.c
JoesCat: hehehe ;-) your turn to go off-topic ;-P
see: https://github.com/fontforge/fontforge/pull/2735
"Could we just fix the header and not mess with the whitespace and minor verbiage? I don't think it's worth the changelog/blame clutter."

I've already had to rebase patches a few times for throwing in extras, so, I'll just leave that pt alone for now.  ;-P

A lot of code assumes things work fine and doesn't deal with "what if?" scenarios, so...as you note here...it's too easy to go fix one-more thing and throw it into a patch while you're there now.

This entire pull request is sort-of a subset of adding authors and copyrights to satisfy #2735, but this patch focuses on creating is_Ligatures.c as a substitute for the python code that had to be removed in the past. Would prefer to throw-in python script commands to keep this all together as a mini-howto of what needs attention, but this single patch is pretty big already without it.

Unicode/makeutype.c
JoesCat: makeutype.c is an administrator type of utility meant to be run very seldom, and not as part of general 'make all'.
This patch is already pretty big, and messing with that goes off-topic with the patch.

While in Rome, do like the Romans do.

Unicode/makeutype.c
JoesCat: ...was thinking about MAXC at the time, but linguistically probably makes more sense to call it lg16 as it's better related to uint16 I suppose. If no reply, will change to lg16, ...and comment to point-out as within uint16 range since MAXC was smaller in the past.

Unicode/makeutype.c
JoesCat: 8 was going beyond 79 char width. 4 looked too narrow.... will make it 4.

Unicode/makeutype.c
JoesCat: and break the coding style?
ok....

Unicode/makeutype.c
```diff
@@ -630,6 +683,178 @@ static void dumparabicdata(FILE *header) {
...
+/* Treat both (uint16, uint32) tables as if it is only one table. Error=-1. */
+    fprintf( data, "int32 %s_get_U(int n) {\n", n );
+    fprintf( data, "    if ( n<0 || n>=%d )\n\treturn( -1 );\n    ", m );
+    if ( s<m )
+	fprintf( data, "if ( n<%d )\n\t", s );
+    fprintf( data, "return( (int32)(%s16[n]) );\n", t );
+    if ( s<m )
+    fprintf( data, "    else\n\treturn( (int32)(%s32[n-%d]) );\n", t, s );
+    fprintf( data, "}\n\n" );
+}
```
frank-trampe: So `m` is table size. What are `t`, `n`, and `s` in this context?

Unicode/utype.c
```diff
@@ -41001,7 +41001,7 @@ const uint32 ____utype[] = { 0,
   0x180c2010, 0x00080000, 0x00004000, 0x00004000, 0x00004000, 0x00004000, 0x00000000, 0x00000000,
   0x10000000, 0x00000000, 0x10200100, 0x00000000, 0x00000000, 0x00020000, 0x00000000, 0x10000000,
   0x00004000, 0x00004000, 0x10000400, 0x10000400, 0x10010000, 0x10200101, 0x00000000, 0x00000000,
-  0x10000000, 0x10000400, 0x10200100, 0x00000000, 0x10000000, 0x10000000, 0x10000000, 0x00080000,
+  0x10000000, 0x10000400, 0x10200100, 0x00000000, 0x30000000, 0x30000000, 0x30000000, 0x00080000,
```
frank-trampe: What is the purpose/meaning of the table changes?

fontforge/python.c
```diff
@@ -810,6 +810,114 @@ static PyObject *PyFF_UnicodeNamesListVersion(PyObject *UNUSED(self), PyObject *
     return( ret );
 }
 
+/* Ligature & Fraction information based on current Unicode (builtin) chart. */
+/* Unicode chart seems to distinguish vulgar fractions from other fractions. */
+/* These routines test value with internal table. Returns true/false values. */
```
frank-trampe: I think that it would be worthwhile to specify that the argument to each of these new functions is a numeric codepoint. It is not obvious.

Unicode/is_Ligature.c
```diff
@@ -0,0 +1,212 @@
...
+	if ( p16 ) n = p16 - ____fraction16;
+    } else {
+	p32 = (uint32 *)(bsearch(&uCode, ____fraction32, 4, \
+				sizeof(uint32), compare_codepoints32));
+	if ( p32 ) n = p32 - ____fraction32 + 25;
+    }
+    return( n );
+}
+
+/* Codepoints whose names match the regular expression (yes==0) */
```
frank-trampe: What's this about names?

Unicode/makeutype.c
frank-trampe: This is not whitespace, and you're already adjusting this file. Surely adding `if (pt[0] == '\0') return -1` would not ruin your day? I'm happy to do it if necessary, but it's much more efficient if you do it.

inc/utype.h
```diff
@@ -145,6 +147,35 @@ extern struct arabicforms {
     unsigned int required_lig_with_alef: 1;
 } ArabicForms[256];	/* for chars 0x600-0x6ff, subtract 0x600 to use array */
```
frank-trampe: We ought to put a comment noting that these things come from is_Ligature.c.

frank-trampe:
And thanks for making tests!

Unicode/makeutype.c
JoesCat: 'm' is the entire maximum length of the table *dt which will got from reading unicode list and then be named 't' when building the tables and functions in is_Ligatures.c. Looking at a byte count, it made less waste to have an array of 500+ ligatures, 20+ vulgars, and about 50 odd fraction definitions (approx 600 x 4bytes = 2400bytes) instead of assigning an individual 1bit for each in the lookup table (approx 1bit x MAXC = 1/8byte x 65536 = 8k). doing this for ligs, vulgs and frac would make it 8k x 3 = 24k used in flag space. Since I was already building a program, it seemed worth going the extra step and compressing everything in {0..65535} into uint16 (so we split the table 0...s), and anything left afterwards goes into {65536....->) which is uint32 (and is the remainder of the table from s+1...m. so instead of using 2400bytes, now is closer to 1200bytes to define about 600 ligs/vuls/fracs.
'n' is used to name the various functions using the tables.
These routines are reusable 3x to build ligs, vuls, fracs., so testing against one verifies the other two will build similarly the same.

Unicode/utype.c
JoesCat: I heard complaints of how some changes made the fontforge-fork slower in some respects.
As there are approx only 600 values of interest, that means there are approx 65000 values that aren't of interest. Dedicating 1bit as a flag (8kbytes) allows you to avoid doing a bsearch on the 65000 other values that are not ligatures, vulgars, or fractions as you already know they are, or are not a code of interest.

fontforge/python.c
JoesCat: Everything is all in the same Pull request.
There are the bare-bone functions in unicode library. Then there are working script functions in native scripting and also in python scripting. Then there are actual test scripts in the testsuite as working examples. And then there is documentation in scripting and python html help files.

Unicode/is_Ligature.c
JoesCat: If you search unicode list for fractions, these would be the fractions that do not include 'VULGAR' in the definition.

inc/utype.h
JoesCat: At this point in time, I think it is best to leave that working the way it works now.
If someone is willing to explain 0600..06ff a little better, we can work at making that section more customized, but at this point it seemed to make sense to have a base to cover 'Vulgars' as originally done in the older IS_LIGATURES.c, and while searching, also expand into the other odd fractions and ligatures too.

JoesCat:
I was thinking of adding more stuff, but the only thing I got working ATM is vulgar alt expansions. The unicode list appears inconsistent for getting ligatures and other fractions ATM (needs more thought, and ver9.0 is a larger table, maybe improved too), plus you announced a planned 2week window on the next update, so no point on making it more complex at the moment.
There might be some interest for this PR from https://github.com/fontforge/fontforge/issues/2494, but it's only a partial solution ATM

Unicode/makeutype.c
JoesCat: Looking at the structure of unialt.c shows that fontforge is unable to cope with codepoint==0, or has difficulty and needing workaround exceptions. Older bitmaps have references to char 0, and as you point-out, it won't kill us to make it 1 instead of zero, but why limit ourselves? These functions here will work fine if the value is zero. Youll note that there is another routine in ustring that also handles zero too. unialt could probably use a rewrite and I believe we can add zero capability into it too. Where 0 is a problem is with higher level fonts, so it makes sense to test for zero upthere...but down here should be okay to look for zero without having to limit ourselves from the get-go.

frank-trampe: That's not what I meant. What I meant was that there is a potential segfault here if argument pt is a zero-length string because it gets incremented past the first value before it gets used. All we need is `if (pt[0] == '\0') return -1`, and I'm hoping that you can just put it into this patch.

Unicode/utype.c
frank-trampe: So this table is a bit-per-field mask, and we're interested in more codepoints than before.

Unicode/is_Ligature.c
frank-trampe: So codepoints whose names include FRACTION but not VULGAR?

fontforge/python.c
frank-trampe: ?? So you're going to add a commit to this pull request with the additional comment?

inc/utype.h
frank-trampe: I don't understand what you're saying. These functions do come from is_Ligature.c as of this pull request, right? If not, then the need for a comment here is even greater due to the non-obviousness of the definition sources.

Unicode/makeutype.c
JoesCat: ok - will add test for zero on weekend.

Unicode/utype.c
JoesCat: On August 25, 2016 08:14:33 AM Frank Trampe wrote:
> > @@ -41001,7 +41001,7 @@ const uint32 ____utype[] = { 0,
> So this table is a bit-per-field mask, and we're interested in more
> codepoints than before.

yes.
For this PR, we're interested in avoiding a bsearch if we already know 
there is no such character {0..MAXC-1}
There are very few values above MAXC that are lig|vulg|frac, so we can 
probably tolerate a simple min<x<max before bsearching those higher values.

inc/utype.h has 32bit_field allotted for flag definitions. I shared 1 bit for 
{lig&vulg&frac} and there is still an unused flag bit that could be used for 
something else 0x4000000. After that, we'd have to build a 40bit flag or 
larger for more bits.

fontforge/python.c
JoesCat: reworded val to codepoint.

Unicode/is_Ligature.c
JoesCat: Yes. there are 19 vulgar fractions (and they have consistency in the unicode charts too).
There are additional 29 fractions (and their definitions are also not easy to extract if I was going to attempt expanding, alternates, substitution based on the info provided in the unicode list).

inc/utype.h
JoesCat: I'm puzzled how ArabicForms is involved here. Maybe something in the future, but at this point, this PR has no connection.

frank-trampe: I'm not talking about ArabicForms. I'm talking about the following function declarations. I think that, since we are using local declarations, it is worthwhile adding a comment noting where they are defined. It's not a big deal, but it's easy to do when making other adjustments.

JoesCat: added.

Unicode/is_Ligature.c
frank-trampe: So can we update the comment to reflect this?

Unicode/makeutype.c
frank-trampe: Great. Can we get this (or a shorter version of it) in a comment?

inc/utype.h
frank-trampe: And why did we abolish is_LIGATURE.h?

Unicode/is_Ligature.c
JoesCat: We could.
I'm not sure what you are looking for.
What do you want the comment to say?

inc/utype.h
JoesCat: is_LIGATURE.h is not needed as it is already covered by inc/utype.h, and is_LIGATURE.h would have to include inc/utype.h within it.

is_LIGATURE.c currently exists because I did not include it in utype.c even though technically speaking, from a makeutype.c point of view, it could have been appended to utype.c

is_LIGATURE.c makes more sense keeping separate from utype.c because (1) it keeps utype.c from growing further (2) adds some sense of clarification between utype.c and is_LIGATURES.c (3) the original code was written by an author that was pro-GPL, and therefore my best guess was to also call this GPL for patch "Moved is_LIGATURES into Unicode library instead of fontforge library" to satisfy the request to add copyright notices (4) even though there is only about 3 lines of original code left from the old is_LIGATURE.c to the new is_LIGATURE.c, there are also additional routines I would like to add under the GPL banner, and keeping it seperate, adds clarity (utype.c==BSD+3, is_LIGATURE.c==GPL).

tests/test1009.py
```diff
@@ -0,0 +1,45 @@
+# test some ligatures and fractions close to U0000. Avoid high values such as
+# UFxxx since the Unicode chart may add new unicode values later.
+
+import sys, fontforge
+
+print "Get Table Array Totals."
```
jtanx: Pretty sure all tests should conform to Python 3 standards.

jtanx: Hmm why was the test skipped on Travis? Anyway adding brackets to the print calls should be enough for python 3 compatibility. `test1009.py` also needs to be added to `Makefile.am` `EXTRA_DIST` in the test folder.

tests/test129.pe
```diff
@@ -0,0 +1,49 @@
+# test some ligatures and fractions close to U0000. Avoid high values such as
```
jtanx: File needs to be added to `EXTRA_DIST` in `Makefile.am`

tests/test1009.py
jtanx: Ok yep. The reason why it wasn't failing on Travis when it should have (which compiles FontForge with Python3 scripting) is because `test1009.py` wasn't in `Makefile.am`.

JoesCat:
Good catch with the python3 @jtanx 
I was tempted to add test926.py as well to the Makefile.am
We'll need to add one PR, update, add the other PR, or put the 926 on another line so we won't have a merge conflict....or add the 926 here since I'm already here...(even though it's not related).

jtanx:
I had a brief look at 926, but it looks like it's harder to get that working on Python 3 due to bytes not being implicitly converted to strings. I'd definitely leave that one for another PR.

inc/utype.h
frank-trampe: What I was trying to ask was why we were merging the header files but not the code? It makes tracing function definitions slightly harder, and it seems to me that the old header was already included in all of the necessary places.

I didn't realize how much of the new file you authored. Is it divisible enough for you to mark off your part and dual-license that part?

Unicode/is_Ligature.c
frank-trampe: So instead of saying that we are looking for things that match "the" regular expression yes == 0, which I did not understand even after looking at the functions, we would say either that we are looking for non-vulgar fractions or that we are looking for things whose names include FRACTION but not VULGAR.

fontforge/python.c
frank-trampe: That makes it much clearer.

Unicode/makeutype.c
frank-trampe: We probably ought to rename parameter n to something like nam so as to avoid confusion with the other n.

Are these comments right?
```
// m is the maximum size of the table.
// s is the size of the lower partition of the table.
// t is the first part of the data type name.
// nam is the function name base.
```

jtanx:
ping @joescat to rebase because #2811 was merged

Unicode/makeutype.c
JoesCat: added notes

inc/utype.h
JoesCat: I don't think we needed to split makeutype.c into BSD/GPL since we treat the info as "text and data strung-together to create something else", but to keep it simple, I'll switch is_LIGATURE.c to BSD license (treat it as new file replacing old - file while keeping same name).

JoesCat:
had to pull-out test1009.py to avoid merging conflict in tests/Makefile.am due to #2811 - will add it later after rebasing (later).

frank-trampe:
I'm done here. @jtanx?

jtanx:
@JoesCat are you going to rebase to add back test1009 to the Makefile? Otherwise lgtm

JoesCat:
Gets sort of messy trying to do a PR and rebase at the same time, so test1009.py would sit somewhere after the rebase. I'll add it afterwards.with another isLigature.c PR.

JoesCat:
Weird - seems like we both hit merge at the same time - hehehe
....last comment 7min ago, merged 6min ago, branch del 5min ago. Sort of explains github's weird reaction when I clicked buttons.

finally done - time to move-on.

Issue #2769, opened 2016-08-08
Missing translation ID for "ShowFillWithSpace" in Preferences


jrbastien:
- [ ] Screenshot
- [x] The FontForge version and the operating system you're using
  FontForge 17:10 Ubuntu 16.04
- [x] The behavior you expect to see, and the actual behavior
  Expected to see an ID for ShowFillWithSpace in fr.po, The ID for the tooltip is also missing
- [x] Steps to reproduce the behavior
  Search fr.po for ShowFillWithSpace, nothing is found.
- [ ] \(optional) Possible solution/fix/workaround

jtanx:
Hmm I'm not too familiar with gettext. I think you need to regenerate the .po file? @JoesCat maybe?

JoesCat:
On August 7, 2016 11:38:57 PM Jeremy Tan wrote:
> Hmm I'm not too familiar with gettext. I think you need to regenerate
> the .po file? @JoesCat maybe?

gettext on FontForge is a thin version missing several heavy duty things 
that wouldn't work in some operating systems, but for what you are 
interested in there's two steps involved. Generate, and merge.

Some operating systems have a gettext that could use updating (as warned in 
po/Makefile.am) so your mileage may vary. If you have access to a recent 
gettext, or one that is known not to have problems with utf8 text, use 
that. If you are not certain, you may want to run gettext on a linux 
machine, or from a live distro that runs from CDrom, DVDrom, or USBstick.

If you look in po/Makefile.am, there is two generate instructions.
Take a look at the number of options thrown in, and either use the same or 
fewer. If you want to use the same number of options, then go into the po 
directory and run:
make FontForge.pot

this will generate a new po file based on the existing Fontforge file 
structure. Then, you will need to merge the existing po with this new file. 
Using fr.po as an example, I would suggest...

mv fr.po fr-old.po
msgmerge -N fr-old.po FontForge.pot > fr.po

this new fr.po is the one you update.
You can compare with the old:
diff fr-old.po fr.po
or delete the old one:
rm fr-old.po

there is also a menu po too....it's been a while since I looked at po, but 
looking at po/Makefile.am, you would go to the po directory and run:
make FontForge-MenuShortCuts.pot

I'll see about updateing po/Makefile.am to include an update command.

Even though there are tools for modifying the po files, for myself, I've 
used a plain text editor to modify text, such as kwrite. gnome has gedit. 
Windows, you'll probably need to run an enhanced Notepad since these files 
[...]

PR #2772, opened 2016-08-09
Update po file so it can be edited. Resolves issue #2769

JoesCat:
This updates a ??.po file to meet the current FontForge build, by
allowing a user to run make and specify which po file to upgrade.
For this example, fr.po is specified:
make FontForge-Update-po 1=fr


JoesCat:
Closes #2769

Issue #1458, opened 2014-06-25
Moving a curve point may generate an infinite stack recursion.


frank-trampe:
Here is a slice of the stack trace.

```
    at splineorder2.c:1630
#1200 0x00007ffff5ba15f5 in SplinePointListTransformExtended (base=0x97c7b0, 
    transform=0x7fffff8d5990, tpt=<optimized out>, tpmask=<optimized out>)
    at splineutil.c:1851
#1201 0x00007ffff799b44a in CVMoveSelection (cv=0xe27340, dx=1762, 
    dy=4168.0000915527344, input_state=0) at cvpointer.c:1270
#1202 0x00007ffff799ce1f in CVMouseMovePointer (cv=0xe27340, 
    event=0x7fffff8d6150) at cvpointer.c:1513
#1203 0x00007ffff795ef83 in CVMouseMove (cv=0xe27340, event=0x7fffff8d6150)
    at charview.c:5312
#1204 0x00007ffff7961a3f in v_e_h (gw=0x973f40, event=0x7fffff8d6150)
    at charview.c:5652
#1205 0x00007ffff69b5390 in _GWidget_Container_eh (gw=0x973f40, 
    event=0x7fffff8d6150) at gcontainer.c:403
#1206 0x00007ffff6a0065d in dispatchEvent (gdisp=<optimized out>, 
    event=0x7fffff8d64c0) at gxdraw.c:3451
#1207 0x00007ffff6a01929 in GXDrawProcessOneEvent (gdisp=0x8213e0)
    at gxdraw.c:3483
#1208 0x00007ffff69b7c8f in RunError () at gdrawerror.c:82
#1209 0x00007ffff69b8155 in GDrawIError (fmt=<optimized out>)
    at gdrawerror.c:178
#1210 0x00007ffff7a8cd84 in UI_IError (format=<optimized out>) at uiutil.c:423
#1211 0x00007ffff5b59713 in SplineRefigure2 (spline=0x97cda0)
    at splineorder2.c:1270
#1212 0x00007ffff5b5a003 in SplineRefigureFixup (spline=0x97cda0)
    at splineorder2.c:1630
#1213 0x00007ffff5ba15f5 in SplinePointListTransformExtended (base=0x97c7b0, 
    transform=0x7fffff8d7ed0, tpt=<optimized out>, tpmask=<optimized out>)
    at splineutil.c:1851
[...]

Issue #1402, opened 2014-06-08
GDraw segfault with current FontForge


adrientetar:
So I switched to another program, then clicked back to bring FontForge on the foreground and it crashed.

```
Program received signal SIGSEGV, Segmentation fault.
GDrawIsVisible (w=w@entry=0xffcc772200000000) at gdraw.c:148
148     while ( w!=NULL && ( w->is_visible || w->is_pixmap ))
```

The backtrace is very long and loops through a few functions I excerpted below:

```
(gdb) bt
#0  GDrawIsVisible (w=w@entry=0xffcc772200000000) at gdraw.c:148
#1  0x00007ffff69b6762 in GDrawRequestExpose (w=0xffcc772200000000, rect=0x0, 
    doclear=0) at gdraw.c:741
#2  0x00007ffff79a3b89 in CVMouseMoveRuler (cv=0x43ccf90, event=0x7ffffffc29f0)
    at cvruler.c:654
#3  0x00007ffff7963d9c in CVMouseMove (cv=cv@entry=0x43ccf90, 
    event=event@entry=0x7ffffffc29f0) at charview.c:5338
#4  0x00007ffff7965257 in v_e_h (gw=0x3fbf460, event=0x7ffffffc29f0)
    at charview.c:5651
#5  0x00007ffff69b4ad8 in _GWidget_Container_eh (gw=0x3fbf460, 
    event=0x7ffffffc29f0) at gcontainer.c:403
#6  0x00007ffff69fcc51 in dispatchEvent (gdisp=gdisp@entry=0x785660, 
    event=event@entry=0x7ffffffc2bb0) at gxdraw.c:3442
#7  0x00007ffff69fe5d3 in GXDrawProcessPendingEvents (gdisp=0x785660)
    at gxdraw.c:3512
[...]
```

Complete gist [here](https://gist.github.com/adrientetar/0c0caf4b8d7a38e260ee) (there's more than 700 stack frames in here it's scary).

Print:

```
(gdb) p w
$1 = (GWindow) 0xffcc772200000000
(gdb) p w->is_visible
Cannot access memory at address 0xffcc772200000000
```

Build is from May 31th.

PR #2773, opened 2016-08-11
cvruler.c: Reenable the expose call for the ruler.

jtanx:
Without this call, the ruler information won't be updated as the ruler
is used. I don't see the need to call GDrawProcessPendingEvents,
and indeed, it works without this call. Calling GDrawRequestExpose
will not cause an infinite stack recursion.

Similarly, it's not necessary to reexpose the whole CharView window,
just the ruler's window (`cv->ruler_w`)

Related to PR #1464 (Issues #1402 and #1458).

Before:
input
After:
after

cc @frank-trampe


frank-trampe:
Looks good.

Issue #2028, opened 2015-01-03
Provide a way to silence the output of fontforge when used for a script.


Earnestly:
I currently use a simple script to build my fonts as part of a `Makefile`.  The problem with this is that a lot of useless information is output with no way to prevent it outside of redirecting `2> /dev/null` which I would like to avoid as errors should display if there are any.

For example this is what I see:
```sh
teapot earnest %i ~fonts/fncp6 make test     
/usr/bin/fontforge -script mk/generate-opentype.pe src/FNCP6-Bold.sfd
Copyright (c) 2000-2014 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Based on sources from 02:27 GMT 26-Dec-2014-ML-TtfDb-D.
 Based on source from git with hash: 036cd6eb23d333ea56cdebf54016927f49464bf5
/usr/bin/fontforge -script mk/generate-opentype.pe src/FNCP6-Regular.sfd
Copyright (c) 2000-2014 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Based on sources from 02:27 GMT 26-Dec-2014-ML-TtfDb-D.
 Based on source from git with hash: 036cd6eb23d333ea56cdebf54016927f49464bf5
/usr/bin/fontforge -script mk/validate-opentype.pe FNCP6-Bold.otf FNCP6-Regular.otf
Copyright (c) 2000-2014 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Based on sources from 02:27 GMT 26-Dec-2014-ML-TtfDb-D.
 Based on source from git with hash: 036cd6eb23d333ea56cdebf54016927f49464bf5
```
Would it be possible to add a `--silent` or `--quiet` flag to prevent this copyright and version information from being printed?

jtanx:
[...]

PR #2781, opened 2016-08-22
Fix a few crashes relating to empty/invalid lookups.

frank-trampe:
This branch still requires testing. Wait for Frank to merge.

### Important

### Motivation and Context

This fixes a crash reported by @pmjdebruijn when exporting to OpenType after subsetting a typeface.

### Types of changes

### Description

### Final checklist

### DIFF
--- a/fontforge/tottfgpos.c
+++ b/fontforge/tottfgpos.c
@@ -559,17 +559,26 @@ static SplineChar **SFOrderedGlyphs(SplineChar **glyphs) {
 return( NULL );
     for ( cnt=0; glyphs[cnt]!=NULL; ++cnt);
     qsort(glyphs,cnt,sizeof(SplineChar *),sc_ttf_order);
-    if ( glyphs[0]->ttf_glyph==-1 ) {
+	// We want to eliminate invalid glyphs.
+    if ( cnt > 0 && glyphs[0]->ttf_glyph < 0 ) {
 	/* Not sure if this can happen, but it's easy to fix */
-	for ( k=0; k<cnt && glyphs[k]->ttf_glyph==-1; ++k);
+	// We count the invalid glyphs.
+	for ( k=0; k<cnt && glyphs[k]->ttf_glyph < 0; ++k);
+	// Then we move the valid glyphs down to the base of the range.
 	for ( i=0; i<=cnt-k; ++i )
 	    glyphs[i] = glyphs[i+k];
+	cnt -= k;
+	fprintf(stderr, "Eliminated %d invalid glyphs.\n", k);
+	// And we null-terminate.
+	glyphs[i] = NULL;
     }
     for ( i=0; i<cnt-1; ++i )
         if (glyphs[i]->ttf_glyph==glyphs[i+1]->ttf_glyph) {
+						fprintf(stderr, "Duplicate glyph.\n");
             memmove(glyphs+i, glyphs+i+1, (cnt-i)*sizeof(SplineChar *));
             --cnt;
         }
+	glyphs[i] = NULL;
 return( glyphs );
 }
 
@@ -632,6 +641,7 @@ static SplineChar **TTFGlyphsFromNames(SplineFont *sf,char *names) {
 				output[j] = glyphs[i];
 				j++;
 			}
+		fprintf(stderr, "Using %d of %d glyphs.\n", j, i);
 		free(glyphs);
 		return output;
 }
@@ -642,6 +652,8 @@ static SplineChar **OrderedGlyphsFromNames(SplineFont *sf,char *names) {
 
     if ( glyphs==NULL || glyphs[0]==NULL )
 return( glyphs );
+
+#if 0
 
     for ( i=0; glyphs[i] != NULL && glyphs[i+1]!=NULL; ++i ) for ( j=i+1; glyphs[j]!=NULL; ++j ) {
 	if ( glyphs[i]->ttf_glyph > glyphs[j]->ttf_glyph ) {
@@ -658,7 +670,12 @@ return( glyphs );
 	    }
 	}
     }
+
 return( glyphs );
+#endif // 0
+
+// We are going to try using SFOrderedGlyphs here.
+	return SFOrderedGlyphs(glyphs);
 }
 
 static void gposvrmaskeddump(FILE *gpos,int vf1,int mask,int offset) {
@@ -874,7 +891,7 @@ static void dumpGPOSsimplepos(FILE *gpos,SplineFont *sf,struct lookup_subtable *
 	for ( pst=glyphs[cnt]->possub; pst!=NULL; pst=pst->next ) {
 	    if ( pst->subtable==sub && pst->type==pst_position ) {
 		if ( first==NULL ) first = pst;
-		else if ( same ) {
+		else if ( same && first && pst ) {
 		    if ( first->u.pos.xoff!=pst->u.pos.xoff ||
 			    first->u.pos.yoff!=pst->u.pos.yoff ||
 			    first->u.pos.h_adv_off!=pst->u.pos.h_adv_off ||
@@ -906,7 +923,7 @@ static void dumpGPOSsimplepos(FILE *gpos,SplineFont *sf,struct lookup_subtable *
     coverage_pos = ftell(gpos);
     putshort(gpos,0);		/* offset to coverage table */
     putshort(gpos,bits);
-    if ( same ) {
+    if ( same && first ) {
 	if ( bits&1 ) putshort(gpos,first->u.pos.xoff);
 	if ( bits&2 ) putshort(gpos,first->u.pos.yoff);
 	if ( bits&4 ) putshort(gpos,first->u.pos.h_adv_off);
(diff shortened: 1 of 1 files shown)


frank-trampe:
@pmjdebruijn, does this fix your problem? Is the result valid as far as you can tell?

frank-trampe:
I've run a few typefaces through this and then through ttx without any problems, but I would appreciate testing by a few other people. @mskala, @davelab6, @cbaakman?

frank-trampe:
@jtanx, take a look.

jtanx:
Will do - probably tomorrow (for me). Although is there any particular test case?

frank-trampe:
I'll forward the e-mail from @pmjdebruijn.

fontforge/tottfgpos.c
```diff
@@ -484,13 +484,23 @@ static void dumpcoveragetable(FILE *gpos,SplineChar **glyphs) {
...
-	last = glyphs[i]->ttf_glyph;
-    }
+	// We will not emit glyphs with -1 identifiers.
+	// We count the valid glyphs and the ranges.
+	int glyph_cnt = 0;
+	for (i=0; glyphs[i]!=NULL; i++) {
+		if (i > 0 && glyphs[i]->ttf_glyph <= glyphs[i-1]->ttf_glyph)
+			IError("Glyphs must be ordered when creating coverage table");
+		if (glyphs[i]->ttf_glyph < 0) {
+			fprintf(stderr, "-1 glyph index in dumpcoveragetable.\n");
```
jtanx: I think we should either be using the `TRACE` macro (if only for debugging) or `LogError`, which will either dump warnings into the warning window (GUI) or into the console (if scripting). See also #2028

jtanx: We should also probably clean up any other `fprintf`/`printf` calls to stdout/stderr. Plus I don't know if we should also make these strings translatable.

frank-trampe: This is all true. I suppose LogError is appropriate here, even if it's not an error. I'll make the change when I add more comments and squash.

frank-trampe: As for translations, agh. I guess we need to do that too.

frank-trampe: Writing good software is too much work. I guess that's why so few people do it.

fontforge/tottfgpos.c
```diff
@@ -502,24 +512,35 @@ static void dumpcoveragetable(FILE *gpos,SplineChar **glyphs) {
...
+	// start is the index in the glyph array of the starting glyph. last is the ttf_glyph of the ending glyph.
+	r = 0; // r keeps count of the emitted ranges.
+	// We follow the chain of glyphs, ending and emitting a range whenever there is a discontinuity.
+
+	for (i=0; glyphs[i]!=NULL; i++) {
+		if (i > 0 && glyphs[i]->ttf_glyph <= glyphs[i-1]->ttf_glyph)
+			IError("Glyphs must be ordered when creating coverage table");
+		if (glyphs[i]->ttf_glyph < 0) {
+			fprintf(stderr, "-1 glyph index in dumpcoveragetable.\n");
+		} else {
```
jtanx: Aren't we basically doubling up on the error messages?

fontforge/tottfgpos.c
```diff
@@ -591,14 +616,32 @@ return( calloc(1,sizeof(SplineChar *)) );
 return( glyphs );
 }
 
+static SplineChar **TTFGlyphsFromNames(SplineFont *sf,char *names) {
```
frank-trampe: And I definitely need to add a comment explaining what this function does. It is a wrapper for SFGlyphsFromNames that forwards only glyphs with valid TrueType encodings. Which, when we're exporting to TrueType, is all we need and all we can do safely.

fontforge/tottfgpos.c
frank-trampe: Below, yes. I copied and pasted and left that in.

fontforge/tottfgpos.c
```diff
@@ -591,14 +616,32 @@ return( calloc(1,sizeof(SplineChar *)) );
...
+    SplineChar **glyphs = SFGlyphsFromNames(sf,names);
+		if (glyphs == NULL) IError("Glyph-finding error.");
+		int i, j;
+		// Count the valid glyphs.
+		for (i=0; glyphs[i] != NULL; i++)
+			if (glyphs[i]->ttf_glyph >= 0) j++;
+		SplineChar **output = calloc(j+1,sizeof(SplineChar *));
+		if (output == NULL) IError("Memory error.");
+		for (i=0; glyphs[i] != NULL; i++)
```
jtanx: Waiiit - surely `j` should be reset to 0 again here?

frank-trampe: Oooh. Yes. I'm not sure how that slipped through my tests.

fontforge/tottfgpos.c
```diff
@@ -591,14 +616,32 @@ return( calloc(1,sizeof(SplineChar *)) );
 return( glyphs );
 }
 
+static SplineChar **TTFGlyphsFromNames(SplineFont *sf,char *names) {
+    SplineChar **glyphs = SFGlyphsFromNames(sf,names);
+		if (glyphs == NULL) IError("Glyph-finding error.");
+		int i, j;
```
jtanx: Where is j initialised?

fontforge/tottfgpos.c
frank-trampe: I never initialize it, in fact.

fontforge/tottfgpos.c
frank-trampe: Nowhere! I probably need to fix that.

frank-trampe:
@jtanx, I fixed a few things but decided not to mess with translations for now. The error is rare, and the error message isn't really useful unless somebody knows what he's doing.

frank-trampe:
Anyway, if it looks good, I'll squash to another pull request.

jtanx:
@frank-trampe wrt translations, I meant just using the `_()` macro, like `LogError(_("Glyphs must be ordered when creating coverage table"));`

You can also just squash and force push to this current branch instead of making a new pull request. Or if you don't mind there being only *one* commit, you can select the 'Squash and merge' option on GitHub to let GitHub do the squashing for you (note I wouldn't use this option if you have/want distinct commits).

frank-trampe:
@jtanx, I suppose we can go underscore on this. It makes me feel all the more guilty about not making some passing attempt at updating the translation files.

I don't like rewriting the history of a branch, which is why I always make a new pull request. I wish there were some way to switch the branch of a pull request.

jtanx:
Eh, I don't see the problem of rewriting history on branches - in the end you're going to delete this branch anyway once the changes get merged. I think it's cleaner having it all in just one PR.

fontforge/tottfgpos.c
```diff
@@ -502,24 +512,34 @@ static void dumpcoveragetable(FILE *gpos,SplineChar **glyphs) {
...
-		    putshort(gpos,glyphs[start]->ttf_glyph);	/* start glyph ID */
-		    putshort(gpos,last);			/* end glyph ID */
-		    putshort(gpos,start);			/* coverage index of start glyph */
-		    ++r;
+	// start is the index in the glyph array of the starting glyph. last is the ttf_glyph of the ending glyph.
+	r = 0; // r keeps count of the emitted ranges.
+	// We follow the chain of glyphs, ending and emitting a range whenever there is a discontinuity.
+	for (i=0; glyphs[i]!=NULL; i++) {
+		if (i > 0 && glyphs[i]->ttf_glyph <= glyphs[i-1]->ttf_glyph)
+			// LogError(_("Glyphs must be ordered when creating coverage table"));
```
jtanx: Ah, good thing I compiled with `-Wall`:

~~~~
tottfgpos.c: In function 'dumpcoveragetable':
tottfgpos.c:519:6: warning: suggest explicit braces to avoid ambiguous 'else' [-Wparentheses]
   if (i > 0 && glyphs[i]->ttf_glyph <= glyphs[i-1]->ttf_glyph)
~~~~

Because you commented out the line, this if statement is actually falling though to the next if statement e.g.

if (A)
    if (B)

frank-trampe: Ooh. Indeed!

frank-trampe:
Now that we seem to have resolved the outstanding issues, I did some testing, and it seems to work.

jtanx:
I realised that I had built without python scripting support so I translated the script into pe. Did you get this output?
https://gist.github.com/jtanx/97e5b835e4487ada36832e8c21233536

frank-trampe:
I tested only normal round-tripping just now. I remember getting those messages when testing with Pascal's file, but I seem to remember being unbothered by them for some reason. Let me try to figure out why.

frank-trampe:
I reran, and I get vastly fewer of those errors than you do. Only 16, and all before the -1 errors.

jtanx:
Huh. I get completely different results between the python and pe versions. The python version *seems* to generate the wrong output. I don't know if it's related to this pr or not.

frank-trampe:
I don't quite understand what you mean. Pascal's script, since it is written in Python, only works if you compile FontForge with Python support, right?

frank-trampe:
Remember, this case caused a crash before, so we cannot compare the output from the before and after test cases. I did do before/after tests on TrueType output from Charter Nova with a run through ttx, and the non-timestamp, non-checksum data were identical.

jtanx:
Yeah - I recompiled with python support and ran his script. The OTF output seems to be garbage.

frank-trampe:
Well that's odd. I didn't check the output this time, and it is indeed broken, but, on some previous iteration of this branch, it was valid.

jtanx:
Anyway that seems like another issue. I'm okay to merge this as is.

jtanx:
Oh it worked before? Hmm strange

frank-trampe:
I had this working before. I'm going to fiddle with it a little bit more.

frank-trampe:
Yeah. And when I look at the changes since then, the only substantive ones fixed very obvious problems or did not change anything. I wonder whether the uninitialized `j` was somehow making it work before.

jtanx:
I just checked out bccbc6aeefb56f2420e46efa580840979cf5bff4 and it was even worse - a 0 length OTF.

jtanx:
Ah I found the issue. The problem was with the script itself. Each selection command was clearing the previous selection because the `more` flag was not specified, like 

    font.selection.select(("unicode","more",None),0x0021)

Although I don't know why the `None` is there.

frank-trampe:
The current code on this branch still drops a few things from Charter Nova relative to current master. I will look a bit more later.

frank-trampe:
Oh. But it does emit an openable output from subset.py.

frank-trampe:
Charter Nova now outputs identical from this as from master.

@jtanx, would you test one more time?

jtanx:
On testing it seems to work. It no longer outputs the `-1 glyph index in dumpcoveragetable.` message with the original sample.

frank-trampe:
That was a long road!

frank-trampe:
Closing via #2807.

Issue #2668, opened 2016-04-03
Brew installed FontForge doesn't open with UI


weiweihuanghuang:
OS X 10.11.4

```
 Based on sources from 11:56 CEST  3-Apr-2016-D.
 Based on source from git with hash: 1cb7b2e2ee653d32a0b2ff8244c3cea8d55fc716
```

Installed with `brew install -v --debug fontforge --HEAD  --with-giflib --with-x11 --with-libspiro`

And no UI opens? I just get

```
Copyright (c) 2000-2014 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Based on sources from 11:56 CEST  3-Apr-2016-D.
 Based on source from git with hash: 1cb7b2e2ee653d32a0b2ff8244c3cea8d55fc716
>>> 

```

jtanx:
This is a known issue - IIRC X11 support was removed from the brew version of Cairo which means that the GUI cannot be built with homebrew. I believe you have to use MacPorts(?) to build a version with the GUI. But since I don't use Mac, I could be totally off. @frank-trampe might be able to clarify.

weiweihuanghuang:
Thanks, I will use the packaged app version for UI, and the brew version for scripting.

frank-trampe:
Something like that. I made a fuss somewhere on GitHub and got confirmation that the problem on Homebrew would not be fixed. The additional problem is that our Macintosh .app bundle stopped working sometime last year. I tried posting a new one yesterday, and I'd be thrilled to have somebody test [it] (https://github.com/fontforge/fontforge/releases/tag/20160403) and report back.

bagong:
What would proper testing involve? I downloaded it and started it. It comes up for me (el-capitan and XQuartz 2.9.7 rc2). The first time I started it, it claimed that Untitled1.sfd was missing, from then on it came up fine.

I opened the font musica.ttf. Seemed to go well.

Two issues I noticed:

- in the default Window the glyphs display in a very light grey
- I was scrolling up and down with a trackpad-drag, and apparently overdid it. The program got into a state where scrolling never stopped until ff crashed.

Best
.r.

frank-trampe:
Your first issue is #2665, which we just fixed. The second one is one that we seem to have unsuccessfully fixed many times. I'll take another look.

[...]

Issue #2381, opened 2015-06-18
Dialogs show up "too centered" on a multi monitor setup 


Pomax:
Similar to https://github.com/fontforge/fontforge/issues/2256, but actually much easier to fix:  On a multi-monitor setup Fontforge will place dialog windows in the center of "the total screen width" rather than centered on the monitor fontforge itself is open in. On odd numbered monitor setups, that's a little funny, but on a dual setup, it puts the dialog half on screen 1 and half on screen 2, and that's really problematic =)

larsenwork:
So how's it easy to fix? :)

Pomax:
Don't confuse "easier" with "easy", they are absolutely different things. #2256 is hard because the only proper solution is to figure out of the positioning conflicts with other applications. Here, the problem is internal to FontFont and the solution requires not using the total display resolution to figure out where to place dialogs, but asking windows which screen it's on, and then using that screen's dimensions for dialog positioning.

jtanx:
See that's the problem, it's not a native app at all. As has been mentioned multiple times before, it uses X11 directly and rolls its own GUI toolkit. You're more than welcome to fix this if (a) you can fix the [X11 server](http://sf.net/p/vcxsrv) to work with multi-monitor setups, or (b) you can rewrite FF to use a standard GUI toolkit.

Pomax:
Not having been part of this project or having ever look at this issue tracker before yesterday, that is not something I can reasonably be expected to already know =)

I have no knowledge whatsoever of X11, but it sounds like this is a bug in X11 then, which should probably be filed upstream if that hasn't happened yet, so that rolling over to the next X11 update automatically fixes it for Fontforge.

That said, the "you're welcome to fix it yourself" comment is pretty low. You know damn well I don't have the skills to do so, which is why I filed an issue, not a PR. If this is a bug you know exists in X11, then I assume that because X11 is important to Fontforge, you've already filed this over on the X11 bug tracker, and this issue doesn't need to be solved by anyone working on the project, X11 will fix this. Just drop a link to that bug in this ticket and mark it as closed because eventually X11 will fix their bug. No need to be mean.

larsenwork:
@Pomax no-one is mean. I think the confusion comes from your 
> Similar to #2256, but actually much easier to fix
which frankly only makes sense to say if you have looked in the code and have an idea how to fix it :wink: 

[...]

Issue #2156, opened 2015-03-20
AltGr support


AlexandruCobzas:
Please provide support for this key. AltGr is the RightAlt key and it's used on many European keyboard layouts. It should be treated differently from the LeftAlt. AltGr is a fast and convenient way to input various symbols, practically doubling (combined with Shift) the total number of symbols available. Good for testing kerning etc.

Thank you for your effort. FF is getting better and better on Windows!
Alex

jtanx:
See also: #2065

jtanx:
@AlexandruCobzas, please try this [test build](http://sourceforge.net/projects/fontforgebuilds/files/i686/Portable/FontForge-mingw-w64-i686-d17f76-r1.7z). To test, modify `fontforge.bat` - where you see 

~~~~~
start /B "" "%FF%\bin\VcXsrv\vcxsrv.exe" :%FF_XPORT% -multiwindow -clipboard -silent-dup-error
~~~~~

Replace with:

~~~~~
start /B "" "%FF%\bin\VcXsrv\vcxsrv.exe" :%FF_XPORT% -multiwindow -clipboard -silent-dup-error -xkbmodel pc105 -xkblayout fr,us -xkboptions grp:ctrl_shift_toggle
~~~~~

This presumes a (French?) French layout keyboard - if it is not, then try changing `fr` to your country code (e.g. `de` for German layout and `ca` for Canada - if `ca`, specify `-xkbvariant fr` for Canadian French layout etc.). It also allows switching to a standard US keyboard layout by pressing Ctrl+Shift. 

Note that in the "Open Font Dialog" you can only type ASCII characters, but everywhere else you should be able to use your desired keyboard layout.

ghost:
I uses Czech keyboard and all English characters I write with AltGr on Win10. But probably last version where is functional `AltGr+Key` in inputboxes is 20201107. I tried newer versions 2023-01-01 and 2025-10-09-Windows-x64 but `AltGr+Key` isn't functional.

Issue #2065, opened 2015-02-04
can't change keyboard layout in windows


sergey-s-betke:
I have read #1908 and #1729. And I can't change keyboard layout:

    start /B "" "%FF%\bin\VcXsrv\vcxsrv.exe" :%FF_XPORT% -multiwindow -clipboard -silent-dup-error -xkbmodel pc105 -xkblayout en,ru -xkbvariant winkeys -xkboptions grp:ctrl_shift_toggle

I can input just latin and digits.

If I test next command line:

    start /B "" "%FF%\bin\VcXsrv\vcxsrv.exe" :%FF_XPORT% -multiwindow -clipboard -silent-dup-error -xkblayout ru

I can't input any chars (and latin, and digits). 

Please, help me with correct command line for support `ru` layout.

jtanx:
I've just had a quick check, and it loks like VcXsrv isn't handling the keymap change:

~~~~~
Warning: Locale not supported by X, falling back to 'C' locale.
(II) AIGLX: enabled GLX_SGI_make_current_read
(II) AIGLX: enabled GLX_MESA_copy_sub_buffer
(II) AIGLX: enabled GLX_SGI_swap_control and GLX_MESA_swap_control
(II) AIGLX: enabled GLX_SGIX_pbuffer
(II) AIGLX: enabled GLX_ARB_multisample and GLX_SGIS_multisample
(II) 60 pixel formats reported by wglGetPixelFormatAttribivARB
(II) AIGLX: Set GLX version to 1.4
(II) GLX: Initialized Win32 native WGL GL provider for screen 0
(EE) Error loading keymap C:\Users\\AppData\Local\Temp\server-0.xkm
(EE) XKB: Failed to load keymap. Loading default keymap instead.
winBlockHandler - pthread_mutex_unlock()
winClipboardThreadProc - DISPLAY=127.0.0.1:0.0
~~~~~

sergey-s-betke:
For a workaround, may be, I must to copy some files to %temp%? Or not?

jtanx:
Hmm... I don't know. I tried with a full install of VcXsrv and it didn't work either. I wouldn't know where to get that file either.

sergey-s-betke:
may be previous versions of VcXsrv? 

Other question: how I can encode Unicode string for "manual" storing them in to .sfd file? As a workaround...

jtanx:
[...]

PR #2782, opened 2016-08-23
Add experimental GDK GUI backend

### LINKED ISSUE #1729: System keyboard layout is ignored on Windows

### LINKED ISSUE #2096: Mniature pop-up windows
miniature pop-up windows on taskbar show the same image.
i expect to see a thumbnail of each opened window of FF like in other applications. 

when taken the curser over them.
on windows 7 latest stable fontforge 

img_20150302_142914

### LINKED ISSUE #2256: Windows open off-screen on multi-monitor setup (Windows)
When opening a font, that window is moved to what FontForge believes is the upper-right corner of the screen. Apparently, it does this by looking at the total dimensions of all connected monitors, not the dimensions of any single monitor. Consequently, the window may be positioned in an area that is not actually visible. 

Example/to recreate: I run a three-monitor setup with a large QFHD monitor in the center and a smaller FHD monitor at the right, bottom aligned with the bottom of the center monitor. The font window is opened in the non-imaged space above the right-hand FHD monitor.

Workaround: while FontForge's windows do not respond to the usual Alt+Space, m shortcut to move windows, they do respond to Win+left arrow and Win+right arrow keyboard shortcuts will move them to visible areas.

OS: Windows 8.1 Pro x64
Version: 09:31 AWST 1-May-2015

### LINKED ISSUE #2461: Pulldown menus can become detached from main menu (Windows)
When opening a pull-down menu (for example the "File" menu) and then dragging the program window will detach the pull-down menu and leave it floating on the screen:
menu_detach
Expected behavior would be that the pull-down menu is automatically being closed if the program window is being dragged.

Related to this issue is the undesired behavior of FontForge of not closing an open pull-down menu when another application window is being activated (by mouse-click or Alt+TAB).

OS: MS Windows 7 Pro x64
FontForge version: FontForge-mingw-w64-i686-99da6f-r2

### LINKED ISSUE #2462: Menus shown behind taskbar (Windows)
If the FontForge window is placed at an unfortunate screen location, menus will open behind the taskbar, thus making the bottom menu entry/entries inaccessible until the FontForge window is being moved and the menu reopened:
taskbar
The pull-down menus should either always being displayed on top of the taskbar, or alternatively the space occupied by the taskbar needs to be taken into account when deciding whether pull-down menus should open downwards or upwards.

Related to issue #2256:
When deciding whether pull-down menus should open upwards or downwards, FontForge bases this decision incorrectly on the dimension of the virtual-screen bounding-box. This can lead to situations where a pull-down menu is being drawn off-screen:
menu_cut
The situation as displayed in this screenshot can for example occur with the following multi-monitor setup when FontForge is placed close to the bottom of monitor 1:
mm

OS: MS Windows 7 Pro x64
FontForge version: FontForge-mingw-w64-i686-99da6f-r2

### LINKED ISSUE #2590: How to change window's title name from VcXsrv X ??
I want to change window's title name from VcXsrv, but I couldn't find window title function, where is it.
Could you answer Where is it??

jtanx:
Note: Not to be merged as of yet.

---

This adds an experimental GDK backend to GDraw. It is meant to be more or less a drop-in replacement for the X11 backend. This backend is enabled by passing `--enable-gdk` at the configure stage. This will add a dependency on `libgdk-3.0` or `libgdk-2.0` (provided by `libgtk-3-dev` and `libgtk2.0-dev` or equivalent respectively) - it will prefer `libgdk-3.0` to `libgdk-2.0`, but either should work. If you wish, you may force it to explicitly build with either GDK2 or GDK3 by passing `--enable-gdk=gdk2` or `--enable-gdk=gdk3`.
### Motivation and Context

By using the GDK backend, it removes the hard dependency on X11. This is beneficial on Windows and OS X as it removes the need to run an X server (VcXsrv or XQuartz). Note that most of the benefits of this backend would be on Windows.

Side-effects of using GDK:
- Better multi-monitor support - windows should no longer open halfway between monitors  - fixes #2256 and #2381
- Menus open in front of the taskbar - fixes #2462. Also fixes #2461 as a side-effect of not being able to move the window without first dismissing the menu.
- Proper centering of windows on screen (by taking into account the size of window decorations) and not opening windows under the cursor (not really a side-effect since I specifically coded this in)
- Native support for the user's current keyboard, including IME support (IME support on Windows only) - there's no longer a need to mess with VcXsrv config files to get the right keyboard. Fixes #1729, #2065, and probably #2156.
- Fixes #2096 (preview windows when hovering over taskbar in Windows)
- Window titles (on Windows) are displayed properly - fixes #2590
- RGBA cursors can now be used (if someone wants to make them)
- The palette windows can now be docked/undocked within a single session on OS X (due to removal of reparenting code)
- Technically supports running natively on Wayland by virtue of GDK supporting it - but this is untested - and server side decorations would have to be provided, which I don't think happens as of now.
- Apparently better performance under OS X, especially when scrolling (XQuartz performance related issues)

Technically the GUI can now also be built on Homebrew using this backend (#2668), although I don't have enough experience with the OS X packaging. 
#### Caveats

Expect this backend to crash a bit more than on X11. I have been testing quite a lot, but it's difficult to emulate the behaviour of that backend exactly. There may be differences in behaviour that I may have overlooked which causes odd things to happen.
### Types of changes

Changes to FontForge outside of adding this backend have been kept as minimal as possible. However, some changes did have to be made:
- Removal of XOR mode drawing. It's simply not supported by Cairo, which is the only way to draw things when using GDK. Luckily this only affects a few things - the cursor blinking, the ruler position markers and the guidelines in the charview. I've made changes to all three of these to make sure that they still work (either by doing expose calls - or in the case of the cursor, using 'difference' mode). There may be more efficient ways to do this.
  - Update 2016/09/12: I've made rulers get drawn to an offscreen pixmap which then gets drawn to the screen. When the marker position changes, I copy part of that pixmap back over the screen where the old marker was before drawing the new marker. Performance is now roughly equivalent to what it was with just XOR mode drawing.
- Removal of all code that used the reparenting feature - which was only the palette toolbars. GDK's support of this was flakey - especially on OS X. Even XQuartz couldn't do it right (try undocking then docking the palettes on a current version of FF). Anyway, this was fixed by simply destroying then re-creating the palettes whenever it's docked/undocked, which I don't think is a big deal.
- I implemented the `DrawArc` function in Cairo for the X11 backend.
- Popups (tooltips) no longer show if FF doesn't have focus 
- Undocked palette windows now behave properly especially on Windows - they remain above the charview even if the charview has focus.
### Known remaining issues
- I still don't understand the start routine for OS X. To be honest, `startui.c` is in particularly bad shape. I remember #1277. What I do know is that it has a separate event loop to Windows and Linux, which seems odd.
- I don't know if resolution is detected correctly on OS X (particularly on HiDPI displays). GDK has a function ([gdk_screen_get_resolution](https://developer.gnome.org/gdk3/stable/GdkScreen.html#gdk-screen-get-resolution)) to get it, but on my test rig (_cough_) it reports 72dpi, even though it should be rendered at 96dpi. Anyway, I added a property called "ScreenResolution", so this value can be manually set (no messing around with "ScreenWidth..." settings).
- The clipboard should work on all three platforms. Note however, that copy/pasting to/from external applications on OS X is limited to text only.
- Drag and drop to/from external applications is not supported. Although I'm led to believe that this didn't actually work with the current version anyway. 
- IME support doesn't exist on OS X/Linux. Unlikely to change because this support lies in the GTK layer. 
- The palette windows now lack a black border around them when docked - they relied on the window manager to do this (as reparented toplevel windows), but this no longer happens because they're implemented as child windows now.
- Minimum size windows (GDK constraint vs FF asking for ridiculous min widths)
- Modal focus is sometimes broken - FF allows for some really weird window combinations.
- Repeated mouse events when the mouse isn't moving is a bug(difference?) of VcXsrv on Windows - can reproduce with FF running on Linux X-forwarded to Windows


mskala:
I'd be happier to see a move away from GDraw than a patch to make it easier to continue using GDraw, but this is impressive work and if it makes the package play better with desktop environments, it's a good thing.

jtanx:
IMO, the only option of moving away from GDraw would be to move to Qt. Now I've had time to see it, I hate GTK (in particular GTK3). In any case, that would be a major undertaking, far greater than this patch would ever be.

jtanx:
Here's a GDK2 build for Windows: https://ci.appveyor.com/project/fontforge/fontforge/build/1.0.111/artifacts (I set it to GDK2 because I normally build GDK3 and I want to ensure it still builds on GDK2)

Some noted differences between GDK2 and GDK3:
* GDK2 and GDK3 seem to have different drawing models. On GDK3, invalidating a child window (which can happen by calling gdk_window_invalidate_region, or by resizing or raising the window) will cause the parent window to also be invalidated, which results in flicker. The same doesn't happen on GDK2
* GDK3 appears to have better drawing performance, at least on Windows
* The API instability on GDK3 is insane. Functions deprecated left right and centre. Not looking forward to GDK 3.22, although apparently that's meant to be the one with a stable API.
*GDK3 can apply motion compression whereas the same cannot be done easily on GDK2 (this was the point of `GDrawSkipMouseEvents`, but GDK doesn't really provide much in the way of queue manipulation). This means that on complex glyphs, using GDK2 can appear to lag behind.

Known issues remaining:
* Ctrl clicking + dragging in the charview lags, especially on GDK2. Yet to investigate, but the hand tool does some pretty stupid things like invalidate the whole window if the rulers are enabled.
* Sometimes modals break things - such as right clicking in font view -> expand stroke. The way FontForge handles modals is... interesting.

jtanx:
Uh. So I noticed a slow-down when you zoomed in too far on a glyph - see [here](https://github.com/fontforge/fontforge/projects/1#card-98098). I had fixed a bug where a path that was indicated as to be drawn with a dashed line was not actually drawn as such (see [here](https://github.com/fontforge/fontforge/pull/2782/files#diff-9148cd615a33f3feaca272dc1bb6f7a2R176). 

So old FF was drawing a solid line while I was drawing a dashed line. It turns out if you zoom in too far on a dashed line, Cairo runs slow.

stroke

So.. does anyone know what the dashed vs solid line is meant to mean?

pathumego:
@jtanx I can arrange testing on Win and Mac on a daily workflow if you require.

davelab6:
@jtanx this is awesome work!!

jtanx:
gnulib is broken again.

davelab6:
@jtanx wrote, 

> So... does anyone know what the dashed vs solid line is meant to mean?

That's odd. I thought it was meant to be solid, not dashed.

davelab6:
@pathumego did you check out https://ci.appveyor.com/project/fontforge/fontforge/build/1.0.111/artifacts ? :)

gdraw/ggdkcdraw.c
```diff
@@ -51,7 +51,16 @@ static void _GGDKDraw_CheckAutoPaint(GGDKWindow gw) {
             _GGDKDraw_CleanupAutoPaint(gw->display);
             gw->display->dirty_window = gw;
         }
+
+        // Shut up! I know what I'm doing.
+#ifdef GGDKDRAW_GDK_3_22
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
```
khaledhosny: You can simply use `G_GNUC_BEGIN_IGNORE_DEPRECATIONS`/`G_GNUC_END_IGNORE_DEPRECATIONS` from glib.

jtanx: Thanks.

khaledhosny:
> I don't know if resolution is detected correctly on OS X (particularly on HiDPI displays). GDK has a function (gdk_screen_get_resolution) to get it, but on my test rig (cough) it reports 72dpi, even though it should be rendered at 96dpi. Anyway, I added a property called "ScreenResolution", so this value can be manually set (no messing around with "ScreenWidth..." settings).

Is this for hi-dpi auto detection? You may want to use `gdk_monitor_get_scale_factor ()` for this instead.

khaledhosny:
I just tried this on Linux and it is working on hi-dpi screen without any manual settings:
image
(The glyphs in the fontview are blurry, but that is probably they did not take the scale factor into account when rasterized, and Cairo just scaled the bitmaps.)

khaledhosny:
Any updates here? Since this keeps the default X backend essentially unchanged, I think it should be merged and further work on GDK backend can be continued (potentially allowing others to contribute).

jtanx:
Sorry about the delay, yeah I'm fairly happy to merge it (any objections, @frank-trampe?); it more or less works. I got busier, and then I saw the breaking changes coming in GDK4. That coupled with debating what version of GDK to support, or if this would have been possible/better done at the GTK level, and I kind of got put off by making any further changes.

I'll see if I can clean it up a little before merging though.

khaledhosny:
I think GTK3/GDK3 will be around for some time and the API is now stabilized, so that is a win. GTK3 was first released in 2011 and GTK2 is still a thing today.

Doing this work with GTK instead of GDK is something to explore indeed, but I think it would be easier to do on top of this work.

davelab6:
Agree with Khaled on all points

On Mar 10, 2018 10:10 AM, "Khaled Hosny" <notifications@github.com> wrote:

I think GTK3/GDK3 will be around for some time and the API is now
stabilized, so that is a win. GTK3 was first released in 2011 and GTK2 is
still a thing today.

Doing this work with GTK instead of GDK is something to explore indeed, but
I think it would be easier to do on top of this work.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/fontforge/fontforge/pull/2782#issuecomment-372015557>,
or mute the thread

jtanx:
Thanks for testing! Yeah sure, feel free to create issues for anything you find.

MaddTheSane:
> GDK has a function (gdk_screen_get_resolution) to get [the desktop DPI], but on my test rig (cough) it reports 72dpi, even though it should be rendered at 96dpi.

For historical reasons, Mac OS X uses 72 DPI for the base 1x resolution.

PR #2783, opened 2016-08-23
Travis OSX: Fix path to formula

jtanx:
Homebrew loves breaking changes


DomT4:
To be fair, there was about 3 months notice on this particular breaking change 😕.

jtanx:
Ah, fair enough I guess. I don't follow changes to Homebrew closely so the only time I notice is when it breaks the build on Travis.

DomT4:
We're hoping the repo split actually helps with this sort of thing, because people can watch/star https://github.com/Homebrew/brew & not get hundreds upon hundreds of notifications for formula changes they probably don't care about. `brew` is just the core code now.

PR #2785, opened 2016-08-23
Fix a few things in the GlyphOrderAndAliasDB handler.

frank-trampe:
Tie up a few loose ends from #2702.

### Important

### Motivation and Context
It fixes a potential over-run, a memory leak, and an invalid data problem.

### Types of changes

### Description

### Final checklist

### DIFF
--- a/fontforge/encoding.c
+++ b/fontforge/encoding.c
@@ -592,15 +592,15 @@ static Encoding *ParseGlyphOrderAndAliasDB(FILE *file) {
 
     max = -1; any = 0;
     i = 0;
-    while (fgets(buffer, sizeof(buffer), file) != NULL) {
+    while (fgets(buffer, sizeof(buffer), file) != NULL && i < 1024) {
         max = i;
 
         int tab, tab2;
-        for (tab = 0; tab < sizeof(buffer) && buffer[tab] != '\t'; tab++);
+        for (tab = 0; tab < sizeof(buffer) && buffer[tab] != '\t' && buffer[tab] != '\0'; tab++);
         if (tab < sizeof(buffer))
             buffer[tab]='\0';
 
-        for (tab2 = tab+1; tab2 < sizeof(buffer) && buffer[tab2] != '\t'; tab2++);
+        for (tab2 = tab+1; tab2 < sizeof(buffer) && buffer[tab2] != '\t' && buffer[tab] != '\0'; tab2++);
         if (tab2 < sizeof(buffer))
             buffer[tab2]='\0';
 
@@ -622,20 +622,29 @@ static Encoding *ParseGlyphOrderAndAliasDB(FILE *file) {
     }
 
     if (max != -1) {
+        // If we have mappings, we make an encoding.
+        // We pad the map in accordance with existing code in FindOrMakeEncoding and elsewhere.
+        // Nobody knows why.
         if (++max < 256) max = 256;
-        item = calloc(1,sizeof(Encoding));
-        char* buf = strdup(_("Please name this encoding"));
-        item->enc_name = ff_ask_string(buf, "GlyphOrderAndAliasDB", buf);
-        item->char_cnt = max;
-        item->unicode = malloc(max*sizeof(int32));
-        memcpy(item->unicode, encs, max*sizeof(int32));
-
-        if (any) {
-            item->psnames = calloc(max, sizeof(char *));
-            memcpy(item->psnames, names, max*sizeof(char *));
+        char *tmp_name = ff_ask_string(_("Encoding name."), "GlyphOrderAndAliasDB", _("Please name this encoding."));
+        if (tmp_name != NULL) {
+          if (tmp_name[0] == '\0') {
+						// Encodings must be named.
+						free(tmp_name);
+            tmp_name = NULL;
+					} else {
+            item = calloc(1,sizeof(Encoding));
+            item->enc_name = tmp_name;
+            item->char_cnt = max;
+            item->unicode = malloc(max*sizeof(int32));
+            memcpy(item->unicode, encs, max*sizeof(int32));
+            if (any) {
+                item->psnames = calloc(max, sizeof(char *));
+                memcpy(item->psnames, names, max*sizeof(char *));
+            }
+          }
         }
     }
-
     return item;
 }
 


frank-trampe:
@jtanx, take a look.

jtanx:
This is about the function in general, but:
* Having a limit of 1024 seems... low. I'm not familiar with this format, but I found [this](https://github.com/i-tu/Hasklig/blob/master/GlyphOrderAndAliasDB) on the internet, and I can see 1587 entries. If you look at [ParseConsortiumEncodingFile](https://github.com/fontforge/fontforge/blob/b92f8364475ef4dd272202bc7081339ea288a49d/fontforge/encoding.c#L547), its limit is `0x10000` or 65536.
* The code surrounding extracting the two tab separated values seems fragile. But anyway, looking at that file I found, I don't understand how there can be a second tab? I see one tab separating the first and second columns, so `tab2` would just run to the end of the line (and include CR LF or LF in the string)
* Looking at how `AddEncoding` does it, it looks like `i` may be calculated wrong - `max` corresponds to `char_cnt` while `i` corresponds to the number of values below 256 which is contiguously `-1`:

    ````c
for ( i=0; i<256 && i<max; ++i )
	if ( enc_to_uni(i)!=-1 )
    break;

    if ( i<256 && i<max )
	enc->has_1byte = true;
    if ( max<256 )
	enc->only_1byte = true;
    else
enc->has_2byte = true;
    ````

fontforge/encoding.c
```diff
@@ -622,20 +622,29 @@ static Encoding *ParseGlyphOrderAndAliasDB(FILE *file) {
...
-        char* buf = strdup(_("Please name this encoding"));
-        item->enc_name = ff_ask_string(buf, "GlyphOrderAndAliasDB", buf);
-        item->char_cnt = max;
-        item->unicode = malloc(max*sizeof(int32));
-        memcpy(item->unicode, encs, max*sizeof(int32));
-
-        if (any) {
-            item->psnames = calloc(max, sizeof(char *));
-            memcpy(item->psnames, names, max*sizeof(char *));
+        char *tmp_name = ff_ask_string(_("Encoding name."), "GlyphOrderAndAliasDB", _("Please name this encoding."));
```
jtanx: I see elsewhere in `encoding.c` that there's a string: `Please name this encoding` (without the period at the end) - I think we should match it with this so that it doesn't mess with the translation strings/ so that we don't have to translate another string

frank-trampe: Oops. I probably did that when I was retyping it.

jtanx:
Ah, well I see [this](https://forums.adobe.com/thread/395771) which indicates that it can either be two or three columns of data.

frank-trampe:
Hmm. That's a good point about the map size. But I'm not thrilled about the prospect of statically allocating 65536 records.

jtanx:
This would be my version - https://gist.github.com/jtanx/1d81deb5344e33c7e461594ad5b593b6. Not well tested at all though.

frank-trampe:
Ooh. Thanks. I was just starting on my own dynamic memory adventure. I'll look at yours now.

jtanx:
Be warned - I do use GLib. I'm also not entirely sure how to interpret the third column, or if I'm setting the `has_1byte` flag correctly.

frank-trampe:
* I really want to limit GPL creep into the core codebase where possible, but I suppose we could reimplement GArray in ten minutes or less if we needed.
* At line 19, isn't split already null-terminated before the next tab?
* Having assignments and conditionals together without delineating parentheses makes me dizzy.
* I agree with the `has_1byte` logic. As far as I can tell, it indicates whether there is meaningful data in the first 256 entries and thus that there is a valid single-byte index. I wish I had the time to find out why somebody would want such a thing.
* It's a waste to do this much fishing and digging and to leave no comments.

jtanx:
* FWIW, GLib is LGPL, so it's not like it's forcing the entire source to be GPL too. I feel like re-implementing a dynamically sized array is just a waste of time.
* I'll go through and comment it a bit I guess. But the split logic does make sense. We want to split it into two, possibly three columns, depending on what's present. At line 19, what's at `split` would be column two and three - e.g.

    For line: "Value1\tValue2\tValue3"
    buffer="Value1"
    split="Value2\tValue3"

    Running strchr again, we get
    split2="\tValue3"
    etc.

frank-trampe:
* I wasn't saying that we need to implement smart arrays now. I'm saying that if it ever becomes an issue, it's easy to fix.
* In may too-far-past-bed-time state, I was thinking that strstrip was so crude that it null-terminated at the first whitespace after non-whitespace (such that, after line 16, strip would be null-terminated after Value2). But it does wait until the end, so the split logic is okay.
* Comment time!

frank-trampe:
Why are you using g_strstrip, by the way? We're not using gchar.

jtanx:
[Reading the docs](https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strchug), I don't think it matters. g_strstrip modifies in place to remove the whitespace.

frank-trampe:
I think that they are functionally equivalent here, but I had to look it up to know that.

jtanx:
I'm not entirely sure  - is the concern about glib allocated memory?

frank-trampe:
No. It's just confusing to use the GLib function when the standard function is either equivalent or more appropriate.

jtanx:
Okay. I'm not aware of any standard function that does whitespace stripping.

frank-trampe:
I'm just talking about strstrip versus g_strstrip.

It's funny how a conversation like this that could happen in under 20 seconds face-to-face can take 20 minutes by GitHub.

jtanx:
I'm fairly certain `strstrip` isn't part of the standard C library. True that about the conversation :smile: 

I've also updated that gist with a few more comments. I've also removed `i` altogether in place of `has_1byte` which I think makes more sense.

frank-trampe:
* Ooh. You're right. It's just in the kernel, which is odd.
* The `i` change looks good.

jtanx:
I've done some preliminary testing and it seems to work. Funnily enough, the [sample](https://github.com/i-tu/Hasklig/blob/master/GlyphOrderAndAliasDB#L1552) I linked to suddenly switches to using space as the delimiter at line 1552. These entries are ignored. I'm not sure if we want to support both tabs and spaces, but I'd lean towards saying no.

frank-trampe:
That's reasonable. I'm ready for a pull request.

PR #2786, opened 2016-08-25
Redo ParseGlyphOrderAndAliasDB to work with arbitrarily long files

jtanx:
This supercedes #2785. cc @frank-trampe 

### Motivation and Context
- The current parser cannot handle more than 1024 entries in a GOADB file - after that it segfaults
- It does not adequately handle 3-column data or even 2-column data (newlines aren't stripped)

### Types of changes

### Description
Compared to https://gist.github.com/jtanx/1d81deb5344e33c7e461594ad5b593b6 :
* I've removed the special case handling of '.notdef' - I don't see why it shouldn't be named '.notdef' in the encoding.
* Unparseable lines now cause a warning to be displayed. These lines are also now added to the encoding, with no name and a value of -1
* I've fixed the dialogue being clipped:

    Before:
    encoding-before
    After:
    encoding-after

#### Limitations
When the encoding is saved by FF and then reloaded, information is lost. For [example](https://github.com/i-tu/Hasklig/blob/master/GlyphOrderAndAliasDB), when loading the following line at slot 3:

    CR CR uni000D

This gets assigned a unicode value of U+000D. But when FF saves the encoding, it gets dumped simply as:

     /CR

And there's no way to know that it should be mapped to U+000D

### DIFF
--- a/fontforge/encoding.c
+++ b/fontforge/encoding.c
@@ -576,66 +576,112 @@ return( NULL );
 return( item );
 }
 
-/* Parse the GlyphOrderAndAliasDB file format */
+/**
+ *  \brief Parses a GlyphOrderAndAliasDB encoding file
+ *
+ *  \param [in] file The file handle to the encoding file
+ *  \return The encoding, or NULL if none could be made
+ *
+ *  \details Each line in the file contains one glyph name (equating to one slot).
+ *           It is assumed that values are tab separated. There
+ *           are at least two columns, with a third column being optional.
+ *           The first column is the glyph name to be used on output. This
+ *           is what will normally be used to determine the unicode value for
+ *           this slot.
+ *           The second column is the 'friendly' name - this is what will
+ *           be used for the slot's name.
+ *           The third, optional column specifies the unicode value that the
+ *           slot should map to. When available, this will be used in preference
+ *           to the first column.
+ */
 static Encoding *ParseGlyphOrderAndAliasDB(FILE *file) {
-    char* names[1024];
-    char buffer[256];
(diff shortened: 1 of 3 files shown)


frank-trampe:
@jtanx, why would we add the invalid lines to the encoding?

jtanx:
Since each line corresponds to one slot, if we completely skipped invalid lines, we would alter the slot number for subsequent entries in the encoding - which I don't think is desirable.

frank-trampe:
Hmm. I'm not sure whether that would be a problem or not. @davelab6? @felipesanches?

frank-trampe:
@felipesanches, you started all this, and you must help us finish it.

davelab6:
I would suggest to reject invalid files outright, rather than trying to hotfix/heal them

frank-trampe:
I think that we do that, as for feature files, because sometimes people want to use the same standard file with different typefaces, some of which may not cover the entire encoding space.

frank-trampe:
But I don't see the point of having a bunch of `[-1, null]` entries unless somebody will be indexing in the table according to the line numbers in the file. And I don't see any use case for that.

davelab6:
I say try it and see who squeals

frank-trampe:
Okay. Let's disable the creation of the blank entries with a static flag for now. If somebody complains, we'll probably need to implement full round-tripping, but at least we'll have something.

jtanx:
Okay, I've commented out the part about adding in invalid entries - they're just skipped over now. I didn't bother with a flag since if we need to implement this it will need updating the code anyway.

PR #2788, opened 2016-08-26
Show glyph outline when fill is on, preview off

### LINKED ISSUE #2581: Show outline when fill is turned on
When fill is turned on, the outline disappears.  This seems to be relatively newly introduced - I don't recall seeing it before now, though I haven't been tracking the mainline version carefully.  If someone did this on purpose, I urge them to reconsider.  "Fill" is most useful when it is exactly that - darkening the interior of the outlines while the outlines remain visible.  If I didn't want to see the outlines, I would choose "Preview" instead.

### LINKED ISSUE #2719: Guides hidden when Foreground layer is active and Fill turned on
**OS:** Windows
**FontForge Version:** ?.?.?

**Steps to reproduce:**
1. Open a glyph.
2. Select the Guides layer.
3. Draw several guide lines.
4. Switch to the Foreground layer.
5. Turn on Fill so that the glyph outline is filled.

**Expected:**
Guides remain visible.

**Actual:**
Guides disappear.

**Workaround:**
If I turn off Fill, the guides become visible again.

jtanx:
### Types of changes

### Description
When fill mode is activated, show the outline. However,
do not show the outline if preview mode is also activated.

Fixes #2581, Fixes #2719

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -1684,7 +1684,8 @@ void CVDrawSplineSetSpecialized( CharView *cv, GWindow pixmap, SplinePointList *
 	if( strokeFillMode==sfm_stroke_trans )
 	    fgstrokeFillMode = sfm_stroke_trans;
 	if( shouldShowFilledUsingCairo(cv) ) {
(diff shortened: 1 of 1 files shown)


frank-trampe:
It looks safe to me.

PR #2791, opened 2016-08-27
File chooser: Make the right click context menu work

### LINKED ISSUE #2691: Segfault in file chooser when selecting 'Refresh file list' from context menu
* Right click in file chooser
* Select 'Refresh file list' from context menu
* Fontforge segfaults

Probably applies to all platforms

jtanx:
### Types of changes

### Description
Fixes #2691 (this bug has been present since forever, or at least since the 2012 release).

If initiated from a right click, the GGadget passed to the popup menu
handler will be the GFC GGadget itself. As so, its userdata would be the
initiator's user data, and NOT the GFC struct, as is normally expected.

Thus, we cannot call GGadgetGetUserData. However, since in the GFC struct,
its GGadget comes first, effectively the pointer to its GGadget will equal
the pointer to its GFC struct.

Therefore, ensure to always pass the GFC GGadget to the popup handler and
simply cast from this to the GFC struct.

### DIFF
--- a/gdraw/gfilechooser.c
+++ b/gdraw/gfilechooser.c
@@ -891,7 +891,14 @@ static int gotten=false;
 
 static void GFCPopupMenu(GGadget *g, GEvent *e) {
     int i;
-    GFileChooser *gfc = (GFileChooser *) GGadgetGetUserData(g);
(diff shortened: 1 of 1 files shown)


frank-trampe:
Okay.

PR #2792, opened 2016-08-27
Fix improper use of cu_copy on GMenuItem's. Fixes #1927.

### LINKED ISSUE #1927: ustring.c crash with gcc -O3
~~Somewhere between a7edc6d (HEAD~15) and 0b06871 (current HEAD) this crash was introduced.~~ Does not seem to be specific to the font.

**Edit: this seems to be dependent on compiler/optimization**

``` gdb
$ gdb --args fontforge /usr/share/texmf-dist/fonts/opentype/public/inconsolata/Inconsolatazi4-Regular.otf
Reading symbols from fontforge...done.
(gdb) run
Starting program: /usr/bin/fontforge /usr/share/texmf-dist/fonts/opentype/public/inconsolata/Inconsolatazi4-Regular.otf
warning: Could not load shared library symbols for linux-vdso.so.1.
Do you need "set solib-search-path" or "set sysroot"?
 Based on sources from 12:03 EST 13-Nov-2014-ML-TtfDb-D.
 Based on source from git with hash:0b06871f3d924d284fff87448488ebedd0c881b2
fontforge.hasUserInterface!
PyFF_onAppClosing() cnt:1
else argv[i]:/usr/share/texmf-dist/fonts/opentype/public/inconsolata/Inconsolatazi4-Regular.otf

cu_copy (pt=<optimized out>) at ustring.c:411
411     ustring.c: No such file or directory.
(gdb) bt full
#0  cu_copy (pt=<optimized out>) at ustring.c:411
        res = 0x8c6b80 "8\355d\367\377\177"
        rpt = <optimized out>
        n = <optimized out>
#1  0x00007ffff7a79638 in mb2DoGetText (mb=0x7ffff7aa0ac2) at windowmenu.c:329
No locals.
#2  0x00007ffff7a79610 in mb2DoGetText (mb=0x7ffff7aa0ac2) at windowmenu.c:333
No locals.
#3  0x00007ffff7a79610 in mb2DoGetText (mb=0x7ffff7aa0ac2,
    mb@entry=0x7ffff7d22700 <mblist>) at windowmenu.c:333
No locals.
#4  0x00007ffff79b617c in FontViewInit () at fontview.c:7360
No locals.
#5  FontView_Create (sf=0x8f0670, hide=0) at fontview.c:7452
        fv = 0x8e8360
        pos = {x = 1, y = 0, width = 0, height = 0}
        gw = <optimized out>
        wattrs = {mask = (unknown: 0), event_masks = 0, border_width = -22784,
          border_color = 4130380441, background_color = 0, cursor = ct_default,
          window_title = 0x8f0670, icon_title = 0x8a3400, icon = 0x0,
          nodecoration = 0, positioned = 0, centered = 0, undercursor = 0,
---Type <return> to continue, or q <return> to quit---
          noresize = 0, restrict_input_to_me = 0, redirect_chars_to_me = 0,
          is_dlg = 0, not_restricted = 0, redirect_from = 0x7fffffffcde0,
          transient = 0x7fffffffdf10,
          utf8_window_title = 0x7ffff698948d <LoadSplineFont+109> "M\205\355t\bL\211\357\350\246\304\337\377H\211\330[]A\\A]A^\303f.\017\037\204", utf8_icon_title = 0x0}
        gd = {pos = {x = 0, y = 0, width = -136380832, height = 32767}, box = 0x0,
          mnemonic = 0, shortcut = 0, short_mask = 0 '\000', cols = 0 '\000', cid = 0,
          label = 0x0, u = {list = 0x0, tabs = 0x0, menu = 0x0, menu2 = 0x0,
            boxelements = 0x0, matrix = 0x0, drawable_e_h = 0x0, completion = 0x0,
            sbinit = 0x0, col = 0, radiogroup = 0}, flags = gg_skip_hotkey_processing,
          popup_msg = 0x0, handle_controlevent = 0x0}
        gsize = {x = 0, y = 0, width = -160460568, height = 32767}
        icon = 0x0
        nexty = 0
        size = {x = 0, y = 0, width = 9057280, height = 0}
#6  0x00007ffff7a69142 in fontforge_main (argc=<optimized out>, argv=0x7fffffffdf08)
    at startui.c:1343
        buffer = "/usr/share/texmf-dist/fonts/opentype/public/inconsolata/Inconsolatazi4-Regular.otf", '\000' <repeats 942 times>
        pt = <optimized out>
        load_prefs = <optimized out>
---Type <return> to continue, or q <return> to quit---
        i = 1
        recover = <optimized out>
        any = 0
        next_recent = 0
        pos = {x = 200, y = 200, width = 379, height = 319}
        wattrs = {
          mask = (wam_events | wam_bordwidth | wam_backcol | wam_cursor | wam_positioned | wam_isdlg | wam_utf8_wtitle), event_masks = 4294967293, border_width = 2,
          border_color = 0, background_color = 16777215, cursor = ct_pointer,
          window_title = 0x0, icon_title = 0x0, icon = 0x0, nodecoration = 0,
          positioned = 1, centered = 0, undercursor = 0, noresize = 0,
          restrict_input_to_me = 0, redirect_chars_to_me = 0, is_dlg = 1,
          not_restricted = 0, redirect_from = 0x0, transient = 0x0,
          utf8_window_title = 0x7ffff7a9278a "FontForge", utf8_icon_title = 0x0}
        display = <optimized out>
        rq = {family_name = 0x0, point_size = 12, weight = 400, style = fs_italic,
          utf8_family_name = 0x74cd70 "dejavu serif,times,caslon,serif,clearlyu,unifont"}
        ds = 4
        ld = 0
        openflags = 0
---Type <return> to continue, or q <return> to quit---
        doopen = 0
        quit_request = <optimized out>
#7  0x00007ffff72d1040 in __libc_start_main () from /usr/lib/libc.so.6
No symbol table info available.
#8  0x000000000040073e in _start ()
No symbol table info available.
(gdb)
```

jtanx:
### Types of changes

### Description
In all the changed cases, the fields are actually just normal
strings and not the unichar variant. This stems from:
82322a4a4898aeb099b2a1f834bd111bf434ea3e

This bug was discovered after compiling at -O3 optimisation level.

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -13056,7 +13056,7 @@ return;
  	if( mblist_nomm[i].shortcut )
 	    mblist_nomm[i].ti.text_untranslated = copy(mblist_nomm[i].shortcut);
 	else
-	    mblist_nomm[i].ti.text_untranslated = cu_copy(mblist_nomm[i].ti.text);
+	    mblist_nomm[i].ti.text_untranslated = copy((char*)mblist_nomm[i].ti.text);
 
 	mblist_nomm[i].ti.text = (unichar_t *) _((char *) mblist_nomm[i].ti.text);
     }
--- a/fontforgeexe/windowmenu.c
+++ b/fontforgeexe/windowmenu.c
@@ -292,7 +292,7 @@ return;
 	    tmp[1] = (unichar_t)(0);
 	    mb[i].ti.text_untranslated = cu_copy(tmp);
 	} else
-	    mb[i].ti.text_untranslated = cu_copy(mb[i].ti.text);
+	    mb[i].ti.text_untranslated = copy((char*)mb[i].ti.text);
 	if ( mb[i].ti.text!=NULL ) {
 	    mb[i].ti.text = (unichar_t *) S_((char *) mb[i].ti.text);
 	    if ( mb[i].sub!=NULL )
@@ -326,7 +326,7 @@ return;
 	if( mb[i].shortcut )
 	    mb[i].ti.text_untranslated = copy(mb[i].shortcut);
 	else
-	    mb[i].ti.text_untranslated = cu_copy(mb[i].ti.text);
+	    mb[i].ti.text_untranslated = copy((char*)mb[i].ti.text);
 	if ( mb[i].ti.text!=NULL ) {
 	    mb[i].ti.text = (unichar_t *) S_((char *) mb[i].ti.text);
 	    if ( mb[i].sub!=NULL )
--- a/inc/intl.h
+++ b/inc/intl.h
@@ -28,6 +28,8 @@
 #ifndef _INTL_H
 #define _INTL_H
 
+#include "fontforge-config.h"
+
 #if !defined( HAVE_LIBINTL_H )
 
 # define _(str)			(str)


frank-trampe:
The field is typed as Unicode, and I must imagine that the GTextInfo handler touches it occasionally. Would it be better just to start using Unicode in these fields instead of casting/hacking?

jtanx:
Hmm I don't think so. From what I can see, these have always worked as being standard (if translated through gettext) strings, so I'm not even sure why those fields are marked as unichar_t's.

frank-trampe:
If we are fairly sure that the text field in GTextInfo is always used as a char pointer, we ought to change the type. How sure are you? I'm happy to do some more digging next week if necessary.

jtanx:
I did some further digging, and I'm confused. The predominant use case seems to be to cast a standard string to the `unichar_t` variant, but then in other cases, it's actually converted, as [here](https://github.com/fontforge/fontforge/blame/master/fontforgeexe/charview.c#L11258)

jtanx:
The best I could come up with is that if the `text_is_1byte` field is set, the field is treated as a UTF8 encoded string. But then in a lot of other cases, it's simply presumed that it's a normal string without checking for this flag.

jtanx:
In any case, the answer is:
* The casts in this PR are correct - if you look at surrounding code, e.g. [here](https://github.com/fontforge/fontforge/pull/2792/files#diff-085da67dad16e974616022e0856d5b58L13061), they're already cast to standard strings
* The text field may or may not be an actual  unichar_t string, most likely depending on the `text_is_1byte` field - as far as I can work out.
* I do not think we should rewrite to make it always unicode - that's a lot of work (and I mean a lot), and we probably can't anyway because it's needed to do the gettext translations.
* GDraw is bad

frank-trampe:
Okay. I suppose we can leave this ickiness for now. It hasn't caused many problems, or at least many visible problems.

PR #2404, opened 2015-07-03
Add AppVeyor support. Closes #2382

jtanx:
This adds automated native Windows builds, both 32-bit and 64-bit.

To cut down build time, I precompiled all [the dependencies](https://github.com/jtanx/fontforgelibs) and hosted them on [SourceForge](http://sourceforge.net/projects/fontforgebuilds/files/build-system-extras/fontforgelibs/), similar to the MSYS2 repository.

On each successful build, 'artifacts' are created on the AppVeyor page that are the portable 32-bit and 64-bit builds. Build time is around 20 minutes per build (so 40 minutes total for the 32-bit and 64-bit builds).

Currently these builds are slightly different to the ones I package manually due to some experimentation with GLib (#2290).

PR #2793, opened 2016-08-28
Add AppVeyor support

### LINKED ISSUE #2382: Consider adding AppVeyor support
Automated native MSYS builds would be nice

jtanx:
This is a follow-up to #2404. Closes #2382. This means that we'll have automatic, downloadable Windows builds (**NOTE** these are slightly different from the release builds in that they're completely vanilla. I've been patching release builds with Unicode filename support for some time - and this will be missing from these builds).

Primary differences to before:
* AppVeyor supports msys2 by default. Builds are also faster now
* In the interest of speed, only build the 32-bit version, since AppVeyor does not support concurrent builds
* Use Bintray as the primary repo for all the required supporting libraries - this should eliminate the download issues with using SourceForge as the download repository.

Build times take around 12 minutes.

For those interested:
Support libraries are compiled from [this repo](https://github.com/jtanx/fontforgelibs) and then uploaded to both [Bintray](https://bintray.com/jtanx/fontforgelibs) and [SourceForge](https://sourceforge.net/projects/fontforgebuilds/files/build-system-extras/).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,55 @@
+# The SourceForge servers are really unreliable to download from.
+# Half the time, the downloads time-out, causing the build to fail.
+# As such, I have made a bintray mirror, which should be more reliable.
+platform:
+  - x64
+branches:
+  only:
+  - master
+clone_depth: 10
+environment:
+  global:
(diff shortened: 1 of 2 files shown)


frank-trampe:
Fine with me.

PR #2804, opened 2016-09-06
Add boundary checks for allocating source/destination buffers when decompressing WOFF metadata

aalex-millerr:

### Important

### Motivation and Context
Fixes memory safety issues when reading .woff files

### Types of changes

### Description
Adds multiple checks to prevent dereferencing a null pointer because malloc fails and to prevent allocating a 0-sized chunk and writing in to it when attempting to read WOFF file metadata. Also checks that the alleged length of the compressed data as reported in the WOFF file is equal to or less than that of the uncompressed data because this will cause the uncompress() subroutine to return an error anyway.
### Final checklist

### DIFF
--- a/fontforge/woff.c
+++ b/fontforge/woff.c
@@ -348,11 +348,45 @@ return( NULL );
     }
 
     if ( sf!=NULL && metaOffset!=0 ) {
+	/*
+1.	Check that the length of the compressed data is equal to or less than the length of the uncompressed data
+	Even though uncompress() is memory safe it's good practice to ensure that there's always enough room in our destination buffer
+
+2.	Allocate metadata destination buffer
+	Check if malloc returned a null pointer (requested too much memory) or if we overflowed and are pointing to a "0-size" buffer
+	If there's an overflow set it to null, if there's a null pointer, return
+
+3.	Allocate metadata source buffer
+	Same checks as #2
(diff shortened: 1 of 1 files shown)


fontforge/woff.c
```diff
@@ -348,11 +348,49 @@ return( NULL );
...
+	* this. This check, in conjunction with #2, also prevents calling
+	* malloc(0) because 0xFFFFFFFF is the largest possible value for 
+	* an unsigned 32bit int and the case where both values are 0xFFFFFFFF 
+	* is prevented by check #2.
+	*
+	* We can safely pass sf->woffMetadata as a NULL pointer because
+	* it's never accessed anywhere else without a check for it being
+	* NULL first
+	*/
+	if(metaLenCompressed > metaLenUncompressed) { // Check #1 to prevent uncompress from returning an error
```
frank-trampe: It is possible for compressed data to be larger than the uncompressed data. Where would this cause a crash or a bad access?

aalex-millerr: Oops, I misinterpreted what the uncompress() function does. I had it in my head that the length of the source being longer than the length of the destination buffer would always cause uncompress() to return Z_BUF_ERROR but that doesn't appear to be the case.

PR #2809, opened 2016-09-07
Make sure scrollbars always have the same width

jtanx:
Not to be merged yet.

This is a follow-up to #2174 with the intention to update this branch as this issue is investigated further by whoever takes this on. cc @frank-trampe

### DIFF
--- a/gdraw/gscrollbar.c
+++ b/gdraw/gscrollbar.c
@@ -520,8 +520,7 @@ static void GScrollBarFit(GScrollBar *gsb) {
     minheight = 2*(gsb->thumbborder+gsb->arrowsize) + GDrawPointsToPixels(gsb->g.base,2);
 
     if ( gsb->g.vert ) {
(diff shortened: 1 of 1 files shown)


davelab6:
@frank-trampe this seems like it can be merged to me

frank-trampe:
No. This only hides an insignificantly small part of a deeper problem. We're keeping this open as a reminder that we need to get the root problem fixed.

davelab6:
What root problem do you suspect? :)

ctrlcctrlv:
@frank-trampe, can you please answer @davelab6's query? This PR is years old; I think it should be merged ASAP or else closed, there's no point in keeping it around. I tested it and it does work for me.

frank-trampe:
Okay. Fine with me. The root problem has since fallen far down the priority list and is unlikely to return to the top 10 anytime soon.

PR #2815, opened 2016-09-08
Build system: Cleanup a few options

### LINKED ISSUE #2425: Please don't autodetect Homebrew
It would be great if Homebrew-specific changes can be made in the Homebrew formula and not in fontforge's configure scripts; I'm patching it out as part of achieving the goals I describe in https://github.com/fontforge/fontforge/issues/2353#issuecomment-121009759. I figured I'd open an issue to avoid the situation where you attempt to defeat me attempting to defeat your detection code. ;) Let's see if we can agree on an approach.

jtanx:
### Types of changes
    - Removal of `--with-pythonbinary` option that was misleading (see description below). Whoever manages the Homebrew formula will want to remove this option if/when this is merged.

### Description
* Remove homebrew check. Fixes #2425.
* Remove bogus option `--with-pythonbinary`. This was only used for checking
  IPython - it's nonsensical to have this option given that Python has
  already been checked for by AM_PATH_PYTHON. If you want to override which
  Python to use, export the PYTHON variable before running the configure script.
* Cleanup the IPython detection code. the `use_python_cross_option` variable no
  longer exists. Also only check for IPython only if we're building with Python
  support.
* Remove the Python check in configure.ac - this is already handled by
  FONTFORGE_ARG_DISABLE_PYTHON_SCRIPTING. This allows FontForge to build without
  having Python installed and `--disable-python-scripting` is passed. Previously,
  even if --disable-python-scripting was specified, configure would fall over
  because of the unconditional AM_PATH_PYTHON check.
* Don't link libfontforge to Python on OS X - fixes #2553.
* Remove the non-standard way of compiling `macobjective.m`. Instead use a method more canonical with autotools. This method of compiling objective C files was stolen from my GDK PR.
* Force symlink creation (again, stolen from my GDK PR) so that repeated `make install`s on OS X work.
* Move program detection and `gl_EARLY` to run before `AM_PROG_AR`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -597,7 +597,6 @@ AC_SUBST([MY_CFLAGS],[${my_cflags}])
 #--------------------------------------------------------------------------
 # Create and Set FontForge Library Flags
 
-test x"${i_do_have_python_scripting}" = xyes && my_libs="${my_libs} ${PYTHON_LIBS}"
 test x"${i_do_have_giflib}" = xyes && my_libs="${my_libs} ${GIFLIB_LIBS}"
 test x"${i_do_have_libjpeg}" = xyes && my_libs="${my_libs} ${LIBJPEG_LIBS}"
 test x"${i_do_have_libpng}" = xyes && my_libs="${my_libs} ${LIBPNG_LIBS}"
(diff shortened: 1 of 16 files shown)


frank-trampe:
This looks okay, but I can't test with the Macintosh builder until next week.

jtanx:
@frank-trampe I've updated the PR to also fix #2847.

jtanx:
I've also replaced `finite` with `isfinite` as clang was complaining that `finite` is deprecated.

PR #2817, opened 2016-09-08
Remove dead code

gioele:
This PR removes some functions that are no longer reachable from the current code, commented out or if-zeroed out.

I tried to make the commits as small as possible, for the sake of review and to document the commits in which the original code has been removed.

### DIFF
--- a/contrib/fonttools/pfadecrypt.c
+++ b/contrib/fonttools/pfadecrypt.c
@@ -42,27 +42,6 @@ static int hex(int ch1, int ch2) {
 return( (ch1<<4)|ch2 );
 }
 
-static int hexline=0;
-static void tohex(FILE *out,int ch) {
-    int ch1, ch2;
-
-    ch1 = ch>>4;
-    if ( ch1<=9 )
-	putc('0'+ch1,out);
-    else
-	putc('A'-10+ch1,out);
-    ch2 = ch&0xf;
-    if ( ch2<=9 )
-	putc('0'+ch2,out);
-    else
-	putc('A'-10+ch2,out);
-    hexline += 2;
-    if ( hexline>70 ) {
-	putc('\n',out);
-	hexline = 0;
-    }
-}
-	    
 static unsigned short r;
 #define c1	(unsigned short) 52845
 #define c2	(unsigned short) 22719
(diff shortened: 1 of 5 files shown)


frank-trampe:
Hi there. Some of these things are likely to be useful in the future. I specifically recall writing one of the unused functions to aid future efforts of some sort. I'll take a closer look at these late next week.

jtanx:
On the other hand, I bet that some of that code has remained dead for many number of years.

I don't think we should be leaving dead code in there only on the off chance that it might be useful - that in part is what revision control is for.

gioele:
Hi! In this PR there are two classes of commits: one kind of commit removes obvious dead/broken code (declaration of non-existing functions), the one deletes code that has not been used for years. There are only two or three commits of the latter kind, and I have deleted only functions that have not been used for at least two years.

Inside each commit message there is a link to the commit where the last code that used that function has been removed.

gioele:
Ping. What is the consensus on this PR? Can I go on and merge this removal of dead code/unused declarations?

frank-trampe:
There are definitely things in there that I don't want removed. We'll get to this after the release.

gioele:
Is the reduced version #2880 more acceptable? :)

frank-trampe:
That's fine.

gioele:
Thank you for merging #2880. I have rebased this branch. Now it contains only the commits that have to do with function code.

davelab6:
@frank-trampe I don't see any reason not to merge this

JoesCat:
This looks okay, however, like @frank-trampe, I also hesitate at removing code quickly. The codebase is large, complex, and a lot of PRs past and present are "putting-out" immediate brush-fires. Sometimes there are PRs that have a thought in mind but never had a chance to complete...My last PR is an example, where it still needs equivalent python docs, or in native scripting converting error exits from hard exits to a safe return/cleanup/exit, or a lot of other code belonging to other developers as well as George, that got started, but hadn't had time to complete.
It's like a doctor in the operating room, it's 10x easier to cut something out, than to add something in, so, cut-out if it makes sense, but if there seems to have been reason for the code, maybe worth investigating and completing whatever was supposed to go in - in the end, you have better code instead of less code.

ctrlcctrlv:
At this point...this no longer can be merged without a conflict, it's years old, and @JoesCat seems to not want it. So I'm closing this as it is unlikely to ever be merged, primarily due to its age. It can be reopened if interest in it is ever renewed.

PR #2821, opened 2016-09-10
RFC: Move non-static function in own headers

gioele:
This is a request for comments for a future PR. It is incomplete and not meant to be merged, but it shows the general direction.

I tried to modify Fontforge so that it could give me more informative error reports. Diving in the code, I noticed that it is very hard to get around the flow of the functions and their organization. One thing that stroke me as odd (but understandable) is the fact that there is no clear distinction of which functions are just helper functions and which act as an API or internal API. In theory here is a distinction between the directories `fontforge/` and `fontforgeexe/`, in practice everything is highly coupled.

So, in order to make things easier for possible future contributors (but also to make the compilation faster, simpler and with fewer warnings), I'd like to
1. move all the declarations of non-static functions that can be found in `fontforgeexe/` to separate header files (for example, all the non-static functions in `foo.c` are declared in `foo.h`)
2. remove most of the declarations from `fontforge/views.h` and `fontforge/fontforgeui.h`
3. add the relevant `#include "foo.h"` directives in all the `.c` files that use functions from `foo.c`

(Obviously not only functions, but also structs and typedef should be moved, but functions are the lowest hanging fruit.)

This will give us a clear view of which file requires which other file.

Once this job is over, one can
1. find `extern` functions that were never meant to be part of the API, make them static in their own `.c` file and remove their declarations from the `.h` file;
2. rearrange the code to make more sense, separating for real all the library functions from the UI functions, opening the way to more testing and to a serious refactoring of the UI code.

Do you think this plan make sense? Should I continue developing this series of patches?


frank-trampe:
This is certainly the way we (and by that I mean people other than me) ought to have done things originally. But there are costs to change. Let me look at this a bit more.

mskala:
If we're going to do a major global rearrangement like this, I hope we can also take the opportunity to run the whole code base through indent and convert it to a less idiosyncratic style.

lemzwerg:
> If we're going to do a major global rearrangement like this, I hope
> we can also take the opportunity to run the whole code base through
> indent and convert it to a less idiosyncratic style.

This would be a good idea, to be done after the next major release.


   Werner

jtanx:
Yeah the header inclusion is kind of a mess. e.g. I recall that in at least one place, the order has to be very specific because of gnulib.

I think the current 'standard' has been to suffix private headers with `P`, like `gxdrawP.h`. Not sure if we still want to do that or not.

On the list of other things to look into - looking into the thousands of compiler warnings that get ignored because they're not enabled by default (like `-Wunused`).

One such fix that can be done is to change no argument function prototypes from

    extern void BitmapViewFinishNonStatic();

To

    extern void BitmapViewFinishNonStatic(void);

To avoid `warning: function declaration isn't a prototype [-Wstrict-prototypes]`

gioele:
Thank you for all the comments; please allow me to address all of them in one reply.

@frank-trampe: I see some costs, but the main cost I see is the time spend rearranging and testing things (and most of this is automated by a script I wrote). Except for this cost, I see only benefits with the first part of the proposal (grouping declaration by implementation file).

@mskala: this is all about the function definition, almost no indentation is needed. But yeah, the could could use some indentation love (at least a clear guideline for the new code).

@jtanx: all the headers in `fontforgeexe/` are private by definition. Untangling the headers in `fontforge/` is going to require some discussion about private headers and so on. But even the thought of it scares me. Let's solve the "simple" problems first.

@jtanx: I have another WIP branch in which I fix the no-arguments prototypes. (Fixing those warnings is where it all started. It will be straightforward they said.) From my POW it makes more sense to disentangle the declarations en mass first and then make more error-prone changes like fixing the prototypes later. I am not so sure that `()` == `(void)` across all the codebase.

gioele:
I went on and I have moved many other declarations to their own header files. The CI errors are due to the fact that I have not tested the build with the freetype debugger enabled yet. This PR is not meant to be merged yet.

These new patches show clearly that the bulk of `fontforge/views.h` consists of declarations that should live somewhere else and that, once this PR is landed, it will be easier to make sense of the dependences of the code, cut some loose ends and properly separate library functions from UI functions.

Let me know if there are strong objections before I go on and complete the last third of remaining files.

frank-trampe:
I have skimmed the changes. They do move in the direction of how I would have preferred to write things originally, but there are big risks and costs to big changesets. And this is a big changeset!

FontForge is full of edge cases, and it is unlikely that whatever script you've written could be aware of them. Somebody, probably either @jtanx or I, will need to look at it line-by-line and to review the new compiler warnings line-by-line as well.

We're entering the pre-release embargo period soon, so this is unlikely to happen at least until next month.

I'm curious as to why some functions are plucked out of the general header files and others of seemingly similar category remain.

fontforge/views.h
```diff
@@ -33,6 +33,7 @@
 #include <ggadget.h>
 #include "dlist.h"
 
+#include <delta.h>
```
frank-trampe: You probably want quotes here.

fontforgeexe/freetypeui.h
```diff
@@ -0,0 +1,74 @@
+#ifndef FONTFORGE_FONTFORGEEXE_FREETYPEUI_H
+#define FONTFORGE_FONTFORGEEXE_FREETYPEUI_H
+
+#include <fontforge-config.h>
```
frank-trampe: Try quotes.

fontforgeexe/freetypeui.c
```diff
@@ -24,13 +24,17 @@
  * OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
  * ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
+#include <fontforge-config.h>
```
frank-trampe: Quotes.

fontforgeexe/fontinfo.h
```diff
@@ -0,0 +1,21 @@
+#ifndef FONTFORGE_FONTFORGEEXE_FONTINFO_H
+#define FONTFORGE_FONTFORGEEXE_FONTINFO_H
+
+#include <ggadget.h>
```
frank-trampe: Quotes.

gioele:
@frank-trampe: thank you for the small review. I did not know about the embargo period. If you prefer I can stop the work here and resume it later.

> FontForge is full of edge cases, and it is unlikely that whatever script you've written could be aware of them. 

But please note that these are all only syntactic changes. If any application that was not fontforgeexe relied on the declaration moved in this patchset, then it would not be working know, because the linker would not find their implementation anywhere. And I cannot imagine any possible compilation difference between fontforge before and after this PR. Do you know any particular detail that I should look at?

> I'm curious as to why some functions are plucked out of the general header files and others of seemingly similar category remain.

Simple explanation: I am proceeding in alphabetical order (mostly). :) I take one `.c` file, I see which functions are not marked as static, I remove their prototypes from the global headers and put them in the respective local header. After that I fix whatever compilation error may come up. Once the patchset is complete there will be no collective `views.h` or `baseviews.h` any more.

fontforgeexe/fontinfo.h
gioele: From the point of view of the executable, `ggadget.` is a library/system header, not a local header, isn't it?

frank-trampe: It's about search paths. When looking for a bracketed header name, gcc skips those provided by `-I` unless they trail a `-I-`. So anything not installed at compile time gets referenced with quotes.

frank-trampe:
That would normally be the case. Unfortunately, FontForge emits a lot of warnings that we usually ignore, and we would not immediately notice certain problems with prototype mismatches or switching between use of two identically named functions.

fontforgeexe/freetypeui.h
gioele: I had a look at the other files in fontforge and most of them are using brackets instead of quotes... Maybe one could go and harmonize them?

frank-trampe: Hmm. No idea how that happened. @rrthomas?

If other files already do it that way and it passes Travis, there's no need to worry about it for now.

jtanx:
>  I did not know about the embargo period. If you prefer I can stop the work here and resume it later.

@gioele feel free to continue on the pr, it's just that it won't be merged at least until the next release.

fontforgeexe/fontinfo.h
jtanx: @frank-trampe hmm I'm not sure about that. I always thought that quotes vs angle brackets was more about if it was quotes, it would search the current folder before falling back to whatever search paths were specified. [This](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html) certainly seems to support that idea.

frank-trampe: It does, and I was wrong about `-I` not being used by default for system headers. The special option is `-iquote`.

I still think using it for non-system headers is poor form. [Elsewhere](https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html) in the gcc documentation, it says that that is for system headers.

gioele:
Closing the PR. It was just a RFC PR and the related series of PR that goes under the nickname of "consolidation" have been merged some time ago.

Issue #2820, opened 2016-09-10
segfault opening TTF file


Kimik0:
FontForge version 20160404-0ubuntu1~trusty (from PPA)

Trying to open a certain TTF file results in a segmentation fault when the dialog reaches "Reading Glyphs", apparently without any further info as to what caused it.

Here's a backtrace from gdb after installing the fontforge-dbg package:
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff5a117cb in readttfencodings (ttf=0x86a2a0, info=0x7fffffffb980, 
    justinuse=0) at parsettf.c:4782
4782   parsettf.c: No such file or directory.
(gdb) bt
#0  0x00007ffff5a117cb in readttfencodings (ttf=0x86a2a0, info=0x7fffffffb980, 
    justinuse=0) at parsettf.c:4782
\# 1  0x00007ffff5a195a7 in readttf (filename=0x0, info=0x7fffffffb980, 
    ttf=0x86a2a0) at parsettf.c:5581
\# 2  _SFReadTTF (ttf=ttf@entry=0x86a2a0, flags=flags@entry=0, 
    openflags=openflags@entry=(unknown: 0), 
    filename=filename@entry=0x8015e0 "/tmp/HGRLKSJ.TTF", fd=fd@entry=0x0)
    at parsettf.c:6284
\# 3  0x00007ffff5acbaec in _ReadSplineFont (file=0x86a2a0, file@entry=0x0, 
    filename=<optimized out>, openflags=openflags@entry=(unknown: 0))
    at splinefont.c:1095
\# 4  0x00007ffff5acd07c in ReadSplineFont (filename=<optimized out>, 
    openflags=openflags@entry=(unknown: 0)) at splinefont.c:1266
\# 5  0x00007ffff5acd1d0 in LoadSplineFont (filename=<optimized out>, 
    openflags=openflags@entry=(unknown: 0)) at splinefont.c:1324
\# 6  0x00007ffff598e97c in ViewPostScriptFont (filename=<optimized out>, 
    openflags=0) at fontviewbase.c:1325
\# 7  0x00007ffff7a719bc in fontforge_main (argc=2, argv=0x7fffffffe368)
    at startui.c:1350
\# 8  0x00007ffff72fef45 in __libc_start_main ()
   from /lib/x86_64-linux-gnu/libc.so.6
\# 9  0x00000000004006fe in _start ()

[...]

PR #2822, opened 2016-09-11
parsettf.c: Add bounds checking for SJIS decoding

jtanx:
### Types of changes

### Description
Partial fix for #2820 (prevents segfault). Does not perform
CP932 detection, hence if font is actually encoded as such,
user must manually force re-encode to CP932.

Related Coverity ID: CID1083343


frank-trampe:
Okay.

PR #2826, opened 2016-09-12
charview: Re-expose the whole wordlist area when charview is resized.

jtanx:
### Types of changes

### Description
This fixes a bug that I noticed while working on the gdk branch. It prevents a display bug when resizing the window.

Before (nb: on X11/master branch):
beforeexpose

After (ignore the flicker/weird areas in resize - this is on the GDK branch (and the flicker is what I consider a GDK3 bug)):
afterexpose

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -5969,6 +5969,8 @@ return;
 	  GGadgetResize(cv->charselector, new_charselector_width, charselector_size.height);
 	  GGadgetMove(cv->charselectorPrev, new_charselectorPrev_x, charselectorPrev_size.y);
 	  GGadgetMove(cv->charselectorNext, new_charselectorNext_x, charselectorNext_size.y);
+	  charselector_size.x = 0; charselector_size.y = cv->mbh; charselector_size.width = newwidth + sbsize; charselector_size.height = cv->charselectorh;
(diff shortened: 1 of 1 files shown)


fontforgeexe/charview.c
```diff
@@ -5969,6 +5969,8 @@ return;
 	  GGadgetResize(cv->charselector, new_charselector_width, charselector_size.height);
 	  GGadgetMove(cv->charselectorPrev, new_charselectorPrev_x, charselectorPrev_size.y);
 	  GGadgetMove(cv->charselectorNext, new_charselectorNext_x, charselectorNext_size.y);
+	  charselector_size.x = 0; charselector_size.y = cv->mbh; charselector_size.width = newwidth + sbsize; charselector_size.height = cv->charselectorh;
+	  GDrawRequestExpose(cv->gw, &charselector_size, false);
```
frank-trampe: Where/how was it getting exposed before?

jtanx: If you had rulers disabled (View->Show->Rulers), it wasn't. If you had rulers enabled, then the whole window is getting re-exposed by [this call](https://github.com/fontforge/fontforge/blob/master/fontforgeexe/charview.c#L3302).

frank-trampe:
Fine with me.

PR #2833, opened 2016-09-14
Use ffglib.h instead of 'define/undef'-ing GList

gioele:
Tiny clean-up, found during working on PR #2821.

### DIFF
--- a/fontforgeexe/pythonui.c
+++ b/fontforgeexe/pythonui.c
@@ -29,12 +29,7 @@
 // to get asprintf() defined from stdio.h on GNU platforms
 #define _GNU_SOURCE 1
 
-#define GTimer GTimer_GTK
-#define GList  GList_Glib
-#include <glib.h>
-#include <glib-object.h>
-#undef GTimer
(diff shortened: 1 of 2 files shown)


frank-trampe:
@jtanx?

jtanx:
I restarted the mac build which failed because the gnulib clone failed. gnu.org seems mighty unstable as of late.

jtanx:
Hmm, I see that `ffglib.h` is an installed and not private header. We have to be careful here, because we redefine GTimer and GList. Presumably we don't want someone to include our headers and then unexpectedly have those redefined to GTimer_GTK/GList_Glib

frank-trampe:
Since we're only changing the references in compiled files, I don't consider this a regression. Let's investigate and make an issue.

Why do we install any headers for the graphical interface? Nobody builds against them, does he?

gioele:
Thank you for the merge. The ui headers are installed because they and the library headers are deeply, deeply entangled. Disentangling them is the point of original PR.

PR #2839, opened 2016-09-15
Use include guards

gioele:
This PR adds include guards to all the headers in `fontforge/`.

There is a commit for each modified file. This helps minimizing the "conflict surface" for long-living branches like PR #2821.

### DIFF
--- a/fontforge/autowidth.h
+++ b/fontforge/autowidth.h
@@ -1,3 +1,5 @@
+#ifndef FONTFORGE_AUTOWIDTH_H
+#define FONTFORGE_AUTOWIDTH_H
 
 struct charone {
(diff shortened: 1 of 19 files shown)


frank-trampe:
I don't have a problem with this. It's unlikely to cause merge problems. @JoesCat?

JoesCat:
Thanks @gioele!
An ounce of prevention is worth more than a pound of cure.  ;-)

jtanx:
I've just noticed this now on the build: https://ci.appveyor.com/project/fontforge/fontforge/build/1.0.170#L2170

This warning about DictionaryFree wasn't there before.

Issue #2832, opened 2016-09-14
"PS Interpretion" dialog may be causing Python script to hang


robmadole:
FontForge version: 20160404
OS: Ubuntu 14.04.5 LTS

We are using FontForge to import SVGs and have ran across an icon that causes the "PS Interpretion" dialog to appear whenever we use the GUI. It also causes FontForge to crash if I dismiss the dialog.

To reproduce the behavior I create a new font, choose a glyph, and then choose File -> Import. Use the following SVG file:

[bad-icon.svg.zip](https://github.com/fontforge/fontforge/files/473448/bad-icon.svg.zip)

And this appears:

![dialog](https://cloud.githubusercontent.com/assets/132895/18529797/3644ec1a-7a94-11e6-8ae2-209e1ad181cf.png)

Since we are using this in a Python script I suspect that there is no way to dismiss this dialog and the script will hang indefinitely. If I could dismiss the dialog FontForge still crashes which is actually preferable to it hanging.

jtanx:
~~~~~
Program received signal SIGSEGV, Segmentation fault.
0x6ea279be in RemoveBackForthLine (ss=0x43edfa8) at splinestroke.c:3194
3194            sp = sp->next->to;
(gdb) bt
#0  0x6ea279be in RemoveBackForthLine (ss=0x43edfa8) at splinestroke.c:3194
#1  0x6ea27962 in RemoveBackForthLine (ss=0x0) at splinestroke.c:3181
#2  0x6ea27962 in RemoveBackForthLine (ss=0x42a2898) at splinestroke.c:3181
#3  0x6ea27a09 in SSRemoveBackForthLine (contours=0x42b5570) at splinestroke.c:3
212
#4  0x6ea29105 in ApproximateStrokeContours (c=0x60e788) at splinestroke.c:3577
#5  0x6ea29271 in SplineSet_Stroke (ss=0x42414d0, c=0x60e788, order2=0) at splin
estroke.c:3614
#6  0x6ea292fe in SplineSets_Stroke (ss=0x42414d0, c=0x60e788, order2=0) at spli
nestroke.c:3629
#7  0x6ea296d1 in SplineSetStroke (ss=0x42414d0, si=0x60e920, order2=0) at splin
estroke.c:3696
#8  0x6e92d23b in SplinesFromEntityChar (ec=0x60e9c8, flags=0x60ea18, is_stroked
=0) at psread.c:3135
[...]

PR #2840, opened 2016-09-16
Some splinestroke fixes

jtanx:
### Types of changes

### Description
There's two commits here: One is to prevent a segfault/infinite loop experienced in #2832, and the second is to fix a bug I noticed while debugging that issue with regards to stroking paths of very small widths. 

As copied from the commit message:
* splinestroke.c: Prevent segfault/hang if BreakLine produces invalid results

    I'm unsure if there is a better fix to this situation, but if the added condition occurs, then the result will definitely be invalid, and if execution of the function is allowed to continue, either a segfault or infinite loop will result. This is because the following code will free first2 and second2, so if first2 *is* second1 or second2 *is* first1, then we're essentially use-after-freeing first1/second1.
*  splinestroke.c: Reduce skip distance in HideStrokePointsCircle.

    This helps to prevent anomalies in the output, especially in cases where the stroke width is very small. For an example, see fontforge/debugfonts@2f17183.

cc @frank-trampe

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -864,7 +864,7 @@ static void HideStrokePointsCircle(StrokeContext *c) {
 		if ( dist1sq<400 )
 		    j -= (6-1);
 		else
-		    j -= .66667*sqrt(dist1sq)-1;
+		    j -= .5*sqrt(dist1sq)-1;
 		/* Minus 1 because we are going to add 1 anyway */
 	    }
 	}
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -3129,11 +3129,11 @@ return( NULL );
 		int isnext=sp->next->to==sp2, isprev=sp->prev->from==sp2;
 		BreakLine(sp->next,start,end,&first1,&second1);
 		BreakLine(sp2->next,end,start,&first2,&second2);
-		if ( first1->me.x!=second2->me.x || first1->me.y!=second2->me.y ) {
+		if ( first1->me.x!=second2->me.x || first1->me.y!=second2->me.y || first1==second2 ) {
 		    IError("Confusion reighns!");
 return( ss );
 		}
-		if ( second1->me.x!=first2->me.x || second1->me.y!=first2->me.y ) {
+		if ( second1->me.x!=first2->me.x || second1->me.y!=first2->me.y || first2==second1 ) {
 		    IError("Confusion regnas!");
 return( ss );
 		}


fontforge/splinestroke.c
```diff
@@ -864,7 +864,7 @@ static void HideStrokePointsCircle(StrokeContext *c) {
 		if ( dist1sq<400 )
 		    j -= (6-1);
 		else
-		    j -= .66667*sqrt(dist1sq)-1;
+		    j -= .5*sqrt(dist1sq)-1;
```
frank-trampe: Is this an optimal value? Is `0.33333` to small?

fontforge/splinestroke.c
```diff
@@ -3129,11 +3129,11 @@ return( NULL );
 		int isnext=sp->next->to==sp2, isprev=sp->prev->from==sp2;
 		BreakLine(sp->next,start,end,&first1,&second1);
 		BreakLine(sp2->next,end,start,&first2,&second2);
-		if ( first1->me.x!=second2->me.x || first1->me.y!=second2->me.y ) {
+		if ( first1->me.x!=second2->me.x || first1->me.y!=second2->me.y || first1==second2 ) {
```
frank-trampe: What does the error case `first1 == second2` mean?

frank-trampe: That seems like a case where sp2->next recrosses sp->next. Which we ought to handle without an error.

jtanx: Okay. I wasn't sure how to handle that case, because if you look at the code block following this, `first1` gets assigned some values from `second2` and `second1` some from `first2`. Then `first2` and `second2` get freed - so you see that if `first2==second1` or `second2=first`, we have a problem, because they've already been freed.

frank-trampe: Agh. I'll take a look next week.

fontforge/splinestroke.c
jtanx: It was chosen arbitrarily. If you reduce it by too much, it runs slower - O(n^2) to the number of points used.

fontforge/splinestroke.c
jtanx: Yep, I think I might need some help on this one. I think you can push to my branch, so feel free to add commits if you wish.

frank-trampe: What is the test case?

jtanx: This one from #2832 should do it: https://github.com/fontforge/fontforge/files/473864/j.zip or for a larger test case: https://github.com/fontforge/fontforge/files/473448/bad-icon.svg.zip

frank-trampe:
@jtanx, I dug for several hours and found problems faster than I could fix them. Since I already put a lot of work into the general overlap removal logic, it seems worthwhile to standardize on that, at least for now. I have pushed a corresponding commit to the branch `splinestroke_reversal_rules_20160926_1` and put in pull request #2874.

Could you resubmit just the tolerance adjustment commit?

frank-trampe:
Oh, and, by the way, even with the overlap/backtrack removal working, the sample j still drops all curved segments. Another problem for another day.

jtanx:
Oh, brilliant! I didn't realise we could just reuse the remove overlap code here.

PR #2842, opened 2016-09-16
Fix some memleaks as reported by Valgrind

jtanx:
### Types of changes

### Description
These were found while I was running FF under Valgrind to see if there were any leaks in my GDK code.

Also: Remove one instance of a shadowed variable being
used in charview.

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -872,7 +872,7 @@ static void DrawPoint( CharView *cv, GWindow pixmap, SplinePoint *sp,
 {
     GRect r;
     int x, y, cx, cy;
-    Color col = sp==spl->first ? firstpointcol : pointcol, subcol;
+    Color col = sp==spl->first ? firstpointcol : pointcol;
     int pnum;
     char buf[16];
     int isfake;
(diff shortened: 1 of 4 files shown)


fontforgeexe/charview.c
```diff
@@ -12848,7 +12848,7 @@ CharView *CharViewCreateExtended(SplineChar *sc, FontView *fv,int enc, int show
     wattrs.mask = wam_events|wam_cursor|wam_utf8_wtitle|wam_utf8_ititle;
     wattrs.event_masks = -1;
     wattrs.cursor = ct_mypointer;
-    wattrs.utf8_icon_title = CVMakeTitles(cv,buf,sizeof(buf));
+    wattrs.utf8_icon_title = (const char*)CVMakeTitles(cv,buf,sizeof(buf));
```
frank-trampe: Why is this typed const?

jtanx: `utf8_icon_title` is marked as `const`. Probably to indicate that `GDrawCreateTopWindow` won't modify it (and it doesn't).

frank-trampe: It's still poor form. (Not that I"m blaming you.)

frank-trampe:
It looks fine to me.

PR #2844, opened 2016-09-17
gmenu.c: Prevent unnecessary hotkey checking & move hotkey checking code to separate function

### LINKED ISSUE #2843: Hotkey checking is inefficient
If you press any key, it is checked to see if it is a hotkey. Part of that code is to iterate through the menu and to see if it matches an accelerator for a menu item. As it iterates through the menus, it must run the menu's `moveto` function, which essentially prepares it for viewing, such as disabling menu items that are unavailable. In particular, `cv_edlistcheck` is called. One of the options it provides is the Paste option, which makes repeated calls to GDrawHasSelection, to see if pasting is possible. This is potentially quite a slow action, because it usually has to make a round-trip from client to X11 server each time this function is called.

So that means, *each time a key is pressed*:
* The menus are iterated through
* Their `moveto` functions are called
* GDrawHasSelection is spammed

The code is pretty dumb in that it always run the check even if it can't possibly be a hotkey, e.g. only one key pressed with no modifier also pressed.

jtanx:
### Types of changes

### Description
This PR prevents unnecessary checking of hotkeys - if we haven't pressed a modifier (e.g. Ctrl,Alt,Shift,Cmd) then there's no need to check if it's a hotkey. Secondly, there was a block that checked against all the mnemonics - this code was redundant (and part of the cause of the performance degradation found in #2843) as there was an equivalent check right above it.

I've moved the hotkey checking code into a separate function for clarity.

Fixes #2843

cc @frank-trampe

### DIFF
--- a/gdraw/gmenu.c
+++ b/gdraw/gmenu.c
@@ -1853,64 +1853,16 @@ static int osx_handle_keysyms( int st, int k )
 int osx_fontview_copy_cut_counter = 0;
 
 
-
-
-int GMenuBarCheckKey(GWindow top, GGadget *g, GEvent *event) {
-    int i;
-    GMenuBar *mb = (GMenuBar *) g;
-    GMenuItem *mi;
-    unichar_t keysym = event->u.chr.keysym;
-
+static int GMenuBarCheckHotkey(GWindow top, GGadget *g, GEvent *event) {
 //    TRACE("GMenuBarCheckKey(top) keysym:%d upper:%d lower:%d\n",keysym,toupper(keysym),tolower(keysym));
-
-    int SkipUnQualifiedHotkeyProcessing = 0;
-    // see if we should skip processing
-    if( g )
-    {
+    // see if we should skip processing (e.g. no modifier key pressed)
+    GMenuBar *mb = (GMenuBar *) g;
 	GWindow w = GGadgetGetWindow(g);
 	GGadget* focus = GWindowGetFocusGadgetOfWindow(w);
-	if( GGadgetGetSkipHotkeyProcessing(focus))
+	if (GGadgetGetSkipHotkeyProcessing(focus) || event->u.chr.state == 0)
 	    return 0;
-	SkipUnQualifiedHotkeyProcessing = GGadgetGetSkipUnQualifiedHotkeyProcessing(focus);
-    }
 
-
-    if ( g==NULL || keysym==0 ) return( false ); /* exit if no gadget or key */
-
-    if ( (menumask&ksm_cmdmacosx) && keysym>0x7f &&
-	    (event->u.chr.state&ksm_meta) &&
-	    !(event->u.chr.state&menumask&(ksm_control|ksm_cmdmacosx)) )
-	keysym = GGadgetUndoMacEnglishOptionCombinations(event);
-
-    if ( keysym<GK_Special && islower(keysym))
-	keysym = toupper(keysym);
-    if ( event->u.chr.state&ksm_meta && !(event->u.chr.state&(menumask&~(ksm_meta|ksm_shift)))) {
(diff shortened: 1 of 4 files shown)


gdraw/gmenu.c
```diff
@@ -1853,65 +1853,17 @@ static int osx_handle_keysyms( int st, int k )
...
-    int SkipUnQualifiedHotkeyProcessing = 0;
-    // see if we should skip processing
-    if( g )
-    {
+    // see if we should skip processing (e.g. no modifier key pressed)
+    GMenuBar *mb = (GMenuBar *) g;
 	GWindow w = GGadgetGetWindow(g);
 	GGadget* focus = GWindowGetFocusGadgetOfWindow(w);
-	if( GGadgetGetSkipHotkeyProcessing(focus))
+	if (GGadgetGetSkipHotkeyProcessing(focus) || event->u.chr.state == 0)
```
jtanx: No point checking if it's a hotkey if no modifier is held down

gdraw/gmenu.c
```diff
@@ -1853,65 +1853,17 @@ static int osx_handle_keysyms( int st, int k )
...
-	    if ( mb->mi[i].ti.mnemonic == keysym && !mb->mi[i].ti.disabled ) {
-		GMenuBarKeyInvoke(mb,i);
-		return( true );
-	    }
-	}
-    }
 //    TRACE("GMenuBarCheckKey(2) keysym:%d upper:%d lower:%d\n",keysym,toupper(keysym),tolower(keysym));
 
-    /* First check for an open menu underscore key being pressed */
-    mi = GMenuSearchShortcut(mb->g.base,mb->mi,event,mb->child==NULL);
```
jtanx: This is the redundant block that was removed. Mnemonics are checked already here: https://github.com/fontforge/fontforge/pull/2844/files#diff-36bbf0dde11c76b3e8ff12e45265a75fR1995

gdraw/gmenu.c
```diff
@@ -2017,6 +1969,41 @@ int GMenuBarCheckKey(GWindow top, GGadget *g, GEvent *event) {
...
+    int i;
+    GMenuBar *mb = (GMenuBar *) g;
+    unichar_t keysym = event->u.chr.keysym;
+
+    if ( g==NULL || keysym==0 ) return( false ); /* exit if no gadget or key */
+
+    if ( (menumask&ksm_cmdmacosx) && keysym>0x7f &&
+	    (event->u.chr.state&ksm_meta) &&
+	    !(event->u.chr.state&menumask&(ksm_control|ksm_cmdmacosx)) )
+	keysym = GGadgetUndoMacEnglishOptionCombinations(event);
```
frank-trampe: What is that?

frank-trampe:
The code looks okay. Did you test this with the graphical interface?

gdraw/gmenu.c
jtanx: I don't know. I just moved the hotkey code into a separate function. The diff doesn't really show this too well.

jtanx:
Yep, tested and everything works fine (same behaviour as worked before).

frank-trampe:
Great. I suggest that we move ahead.

jtanx:
I was testing on mac and was wondering why mnemonics didn't work. The short answer it's not due to this PR - the same behaviour is observed before and after the PR. Instead, I had to enable "Option keys send as Alt_L and Alt_R" under XQuartz. 

The good news is that the PR *does* help to drop CPU usage - from roughly 13%-15% combined for XQuartz+FontForge to ~5%-6% combined while holding down e.g. the space key in charview. Likely because XQuartz has to poll the pasteboard each time FF queries if it has some data because it doesn't provide a mechanism to detect when changes occur to the pasteboard. 

Anyway, I'll go ahead and merge.

PR #2845, opened 2016-09-17
Do not add libzmq twice

gioele:
### Description

There is no need to add `$(LIBZMQ_LIBS)`: it is already added in the previous lines, if available:

```make
if LIBZMQ
libfontforge_la_LIBADD += $(top_builddir)/collab/libzmqcollab.la $(LIBZMQ_LIBS)
endif LIBZMQ
```

### Types of changes


jtanx:
This is already handled by #2815.

gioele:
Sorry @jtanx, I did not notice the overlap between the two PRs.

PR #2846, opened 2016-09-17
Move to cvundoes.h all the prototypes of non-static cvundoes.c functions

gioele:
This standalone commit has been extracted from the work being done for PR #2821.

The prototypes for the public functions implemented in `cvundoes.c` were peppered over many different header files. This commit moves them all to `cvundoes.h`. The few comments that were present have been preserved.

This commit also moves to `cvundoes.h` some `extern` prototypes weirdly contained in `.c` files. (I wonder if anybody ever successfully compiled an application against libfontforge.)

Thanks to this move, it is now possible to understand which parts of the library implementation depend on the undo logic implemented in `cvundoes.c`. (Unfortunately, too many. The undo logic belongs to the UI, not to the library.)

Please note that no code inside the `.c` files has been touched.

### DIFF
--- a/fontforge/baseviews.h
+++ b/fontforge/baseviews.h
@@ -279,43 +279,13 @@ enum fvformats { fv_bdf, fv_ttf, fv_pk, fv_pcf, fv_mac, fv_win, fv_palm,
 	fv_fig,
 	fv_pythonbase = 0x100 };
 
-extern enum undotype CopyUndoType(void);
-extern int CopyContainsSomething(void);
-extern int CopyContainsBitmap(void);
-extern int CopyContainsVectors(void);
-extern const Undoes *CopyBufferGet(void);
 extern RefChar *CopyContainsRef(SplineFont *);
 extern char **CopyGetPosSubData(enum possub_type *type,SplineFont **copied_from,
 	int pst_depth);
 extern void CopyReference(SplineChar *sc);
-extern void SCCopyLookupData(SplineChar *sc);
 extern void PasteRemoveSFAnchors(SplineFont *);
-extern void PasteAnchorClassMerge(SplineFont *sf,AnchorClass *into,AnchorClass *from);
-extern void PasteRemoveAnchorClass(SplineFont *sf,AnchorClass *dying);
 
-/**
- * Serialize and undo into a string.
- * You must free() the returned string.
- */
-extern char* UndoToString( SplineChar* sc, Undoes *undo );
-
-/**
- * Dump a list of undos for a splinechar starting at the given 'undo'.
- * msg is used as a header message so that a dump at a particular time stands
- * out from one that occurs later in the code.
- */
-extern void dumpUndoChain( char* msg, SplineChar* sc, Undoes *undo );
-
-extern void ClipboardClear(void);
-extern SplineSet *ClipBoardToSplineSet(void);
-extern void BCCopySelected(BDFChar *bc,int pixelsize,int depth);
-extern void BCCopyReference(BDFChar *bc,int pixelsize,int depth);
-extern void PasteToBC(BDFChar *bc,int pixelsize,int depth);
-extern void FVCopyWidth(FontViewBase *fv,enum undotype ut);
-extern void FVCopyAnchors(FontViewBase *fv);
 enum fvcopy_type { ct_fullcopy, ct_reference, ct_lookups, ct_unlinkrefs };
-extern void FVCopy(FontViewBase *fv, enum fvcopy_type copytype);
-extern void PasteIntoFV(FontViewBase *fv, int pasteinto, real trans[6]);
 extern void FVCopyFgtoBg(FontViewBase *fv);
 extern void FVSameGlyphAs(FontViewBase *fv);
 extern void FVClearBackground(FontViewBase *fv);
@@ -489,21 +459,9 @@ extern void BCMergeReferences(BDFChar *base,BDFChar *cur,int8 xoff,int8 yoff);
 extern BDFChar *BDFGetMergedChar(BDFChar *bc) ;
 extern void BCUnlinkThisReference(struct fontviewbase *fv,BDFChar *bc);
 
-extern int CVLayer(CharViewBase *cv);
-extern Undoes *CVPreserveStateHints(CharViewBase *cv);
-extern Undoes *CVPreserveState(CharViewBase *cv);
 extern Undoes *_CVPreserveTState(CharViewBase *cv,PressedOn *);
-extern Undoes *CVPreserveWidth(CharViewBase *cv,int width);
-extern Undoes *CVPreserveVWidth(CharViewBase *cv,int vwidth);
-extern void CVDoRedo(CharViewBase *cv);
-extern void CVDoUndo(CharViewBase *cv);
-extern void _CVRestoreTOriginalState(CharViewBase *cv,PressedOn *p);
-extern void _CVUndoCleanup(CharViewBase *cv,PressedOn *p);
-extern void CVRemoveTopUndo(CharViewBase *cv);
 extern void CopySelected(CharViewBase *cv,int doanchors);
-extern void CVCopyGridFit(CharViewBase *cv);
 extern void CopyWidth(CharViewBase *cv,enum undotype);
-extern void PasteToCV(CharViewBase *cv);
 extern void CVYPerspective(CharViewBase *cv,bigreal x_vanish, bigreal y_vanish);
 extern void ScriptSCEmbolden(SplineChar *sc,int layer,enum embolden_type type,struct lcg_zones *zones);
 extern void CVEmbolden(CharViewBase *cv,enum embolden_type type,struct lcg_zones *zones);
--- a/fontforge/autohint.c
+++ b/fontforge/autohint.c
@@ -25,6 +25,7 @@
  * ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 #include "fontforge.h"
+#include "cvundoes.h"
 #include <stdio.h>
 #include <math.h>
 #include "splinefont.h"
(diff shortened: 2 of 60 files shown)


frank-trampe:
I'm okay with this. @jtanx?

jtanx:
> (Unfortunately, too many. The undo logic belongs to the UI, not to the library.)

From my limited understanding, the collaboration feature makes use of undo serialisation to work (i.e. it sends undo/redo data across the wire which clients then replay to stay in sync). Whether or not the collab feature itself should be only in the gui is another question. 

-----

WRT the PR, I can kind of see why those declarations would be across multiple headers, because some affect the charview while others the fontview (based on the CV/FV prefix). 

So it seems like cvundoes is a misnomer if it's actually handling all of these functions.

fontforge/baseviews.h
```diff
@@ -279,40 +279,12 @@ enum fvformats { fv_bdf, fv_ttf, fv_pk, fv_pcf, fv_mac, fv_win, fv_palm,
...
-extern void ClipboardClear(void);
-extern SplineSet *ClipBoardToSplineSet(void);
-extern void BCCopySelected(BDFChar *bc,int pixelsize,int depth);
-extern void BCCopyReference(BDFChar *bc,int pixelsize,int depth);
-extern void PasteToBC(BDFChar *bc,int pixelsize,int depth);
-extern void FVCopyWidth(FontViewBase *fv,enum undotype ut);
-extern void FVCopyAnchors(FontViewBase *fv);
 enum fvcopy_type { ct_fullcopy, ct_reference, ct_lookups, ct_unlinkrefs };
 extern void FVCopy(FontViewBase *fv, enum fvcopy_type copytype);
 extern void PasteIntoFV(FontViewBase *fv, int pasteinto, real trans[6]);
```
jtanx: Why was this function copied into cvundoes.h but not removed here?

jtanx: I can see the same situation for `PasteIntoMV`.

frank-trampe:
I agree that it is a bit baffling that all of these functions are in `cvundoes.c`, but, if they are there, it makes sense to have the declarations in the matching header.

mskala:
I think saving files, in at least some formats, is also implemented in a way that depends on undoes.  Couldn't find a clear place to point to in the code just now, but it was certainly something that annoyed me when I tried to remove undoes from FontAnvil.  The idea is that there is some processing, possibly including autohinting or remove overlaps, that is supposed to happen before a file is saved, but is thought not to be desirable as a permanent side effect of having saved the file.  So FontForge does the processing, saves the file, and then calls the same "undo" function that would be used by the GUI.

gioele:
I think the basic rule should be "functions implemented in X.c should have prototypes in X.h". This commit (and the others from the same dev branch) tries to do that by moving the prototypes. One could move the implementation instead, but that has some repercussion on the VCS history and raises the problem of "we should take the chance to reformat the code", a bikeshedding-prone discussion that I'd avoid for the moment. ;)

As a stated in other commits, I'd like to have a phase in which the current mess is first made clear, then cleaned up. I am happy to see that this commit is kickstarting the discussion on where things should be implemented and where. At the same time, I'd be happy if we could finish the discovery phase before starting making serious changes. But this decision belongs to you, the experts of the codebase.

@mskala: there are actually many points in which the undo callbacks are called. My opinion is that the logic for where to call undo and how is better left completely to the UI, not to the font manipulation library.

@jtanx: at least half of the file names in fontforge are misnomers ;)

@jtanx: regarding the collaboration feature. From the point of view of my refactoring, the collab code is the most invasive part of the codebase. My humble suggestion would be to (after the next release), remove it completely, go on with the long needed refactor that separates the data/font-processing code from the UI code, then reimplement it more sanely in the UI. But this is a topic that deserves its own separate RFC PR.

I understand that this series of cleanup touch on many other, more profound problems but I'd like to ask you to judge commit separately from all the other issues. :)

fontforge/baseviews.h
gioele: Thank you for noticing. Script error, I'll update the commit.

JoesCat:
+1 agreement about the collab code being invasive.
That was the first piece of code that made it impossible for me to compile FontForge on my desktop computer. I attempted/succeeded to separate it out so that it still worked plus I could continue too (with my older desktop), but only ran into the same problem of having it thrown back in again.
Since that point in time, I've been compiling/building on an alternate computer (not my preferred setup), plus a couple more things got thrown in making it that much harder for older computers to compile FontForge.
I don't have objection to collab, plus some users do make use of it, but it can certainly be separated-out. If you are looking for a method, I note George had a system such as function(), _function(), or you can go fairly extreme and funnel everything through a few switch points (like unicodelibinfo.{c,h}.
It will be interesting to see how you tackle this, as you recognize (and verbalize) an issue here.
Glad to see you have some plan of action you are following.

jtanx:
@JoesCat since the collab feature is entirely optional (can be disabled with `--without-libzmq` or simply not having libzmq present), what exactly is preventing it from being compiled on your main computer?

JoesCat:
On September 18, 2016 05:29:40 PM Jeremy Tan wrote:
> @JoesCat since the collab feature is entirely optional (can be disabled
> with `--without-libzmq` or simply not having libzmq present), what
> exactly is preventing it from being compiled on your main computer?

I had to look back to see where I ran into trouble:
Seems was around 2013-10-13 where I had to separate collab code just to 
allow myself to compile....but looking at what I have...monkeyiq did fix it 
afterwards, otherwise we wouldn't have tag 20140101.
My error.

Looking at latest files for my main machine, seems I had to stop some time 
after 2014-08-12 due to libraries getting introduced that aren't a part of 
my distro for that time.
I did a couple patches after that by using an older set of files and 
patching them into the newer code. File substitution only goes so far 
before the differences get too far apart.

I'll have to dig further, but I think there was some FontForge code deleted 
and a new library introduced shortly before I had to stop working on 
FontForge in 2015, or late 2014.

The only thing done now (related to FontForge) on my main machine is 
libuninamelist and libspiro.

frank-trampe:
So, just to be clear, everybody agrees with the changes in this pull request, right?

jtanx:
Fine by me

JoesCat:
I prefer to see some commenting (added to fontforge/cvundoes.h) if the 
author takes some time to analyze some of the stuff that he is shifting.

Sometimes code (where it sits), has some meaning due to the context of 
where it sits, and loses some meaning if we seperate it from the hive where 
it came from....this is where git blame gives some idea of where to look 
for additional info using gitk to see what/why/patches.

To avoid losing context, and hopefully gain some info/commenting, I think 
this is probably one of the best times to capture some info and comment 
what looks like could use some commenting ...if it makes sense to do so.

@gioele, can you add some comments that can hopefully aid some future 
developer incrementally improve FontForge code further?

If yes, great!
If not, was hoping you could, but otherwise okay with this PR.

gioele:
@JoesCat 
> Sometimes code (where it sits), has some meaning due to the context of where it sits, and loses some meaning if we seperate it from the hive where it came from....

Well, in this case we are putting the context back by moving the prototype where there have been implemented. ;)

> To avoid losing context, and hopefully gain some info/commenting, I think this is probably one of the best times to capture some info and comment what looks like could use some commenting ...if it makes sense to do so. 

I think the best time to do that will be when, once the situation is clearer, we start moving/removing/rearranging functions' code. For example, `FVCopy` has now been moved to `fontforge/cvundoes.h`, but that code obviously belongs to `fontforgeexe/fontview.{c,h}`. When the code and the definitions will be moved, then it will make sense to document what the functions do and what is the bigger picture.

> @gioele, can you add some comments that can hopefully aid some future developer incrementally improve FontForge code further?

I'd love to (I would also learn much more about the code base), but the limited amount of free time I can volunteer to this project is more than absorbed by the almost hundred commits that I have in the pipeline (see PR #2821). With a very conservative estimate of one full day of work for each header, it would require at least 4 months just to document all the moved functions. And that is a very optimistic forecast. At the same time I think part of it would be wasted time, because many function prototypes should just be deleted (see PR #2817) or made static inside the `.c` files. But let's take one step at the time. ;)

frank-trampe:
@JoesCat, I don't think that there's a lot of knowledge developed here. All @gioele did was find declarations for functions defined in `cvundoes.c` and move them to `cvundoes.h`.

@gioele, I'm not sure that it would be super-useful, but it wouldn't take more than a few minutes to note the origins of the declarations. They only come from three files, right?

gioele:
@frank-trampe: would those annotations be useful at all? I'd see them as mostly noise. From the (very short) git history I see that they ended up where they are (e.g. `splinefont.h`) just because those headers happened to be included everywhere and no changes to the `.c` files were required.

frank-trampe:
Alright. We'll merge it as-is.

jtanx:
I think this missed some in the collab sources:

~~~~
collabclientui.c: In function ‘zeromq_subscriber_process_update’:

collabclientui.c:200:3: warning: implicit declaration of function ‘CVLayer’ [-Wimplicit-function-declaration]

   printf("________________________ READ ... 2... ly_fore:%d cv.layer:%d\n", ly_fore, CVLayer(cv));

   ^

collabclientui.c:206:3: warning: implicit declaration of function ‘CVDoRedo’ [-Wimplicit-function-declaration]

   CVDoRedo( cv );

   ^

collabclientui.c: In function ‘collabclient_sendRedo_SC’:

collabclientui.c:1203:2: warning: implicit declaration of function ‘dumpUndoChain’ [-Wimplicit-function-declaration]

  dumpUndoChain( "start of collabclient_sendRedo()", sc, sc->layers[layer].undoes );

  ^

collabclientui.c:1230:5: warning: implicit declaration of function ‘SCDoUndo’ [-Wimplicit-function-declaration]

     SCDoUndo( sc, layer );

     ^
~~~~

(I was playing around with travis and noticed these)

gioele:
@jtanx: well spotted, thanks. I added a commit to fix the missing include to PR #2878.

PR #2853, opened 2016-09-19
gmenu.c Don't skip hotkey checking even if no modifiers are pressed

jtanx:
### Types of changes

### Description
This backs out one of the changes from #2844. One of my assumptions was wrong: It turns out hotkeys can be specified without a modifier, e.g. the 'Delete' key to delete CPs in the charview.

 cc @frank-trampe


jtanx:
Or the other option is to program the charview to directly accept  the delete key for this action, since it's not a hotkey in the traditional sense. I think this is the only exception to where a ' hotkey' has been specified without a modifier.

jtanx:
Nope, looks like I was wrong. There's more than one place where a hotkey is being specified without a modifier.

PR #2854, opened 2016-09-19
gtextfield.c: Fix right click popup when text is selected

jtanx:
### Types of changes

### Description
This moves the code that displays the context menu to run before
the text selection is changed - otherwise it's impossible to select
text then cut or copy it.

Before:
tfbefore
After:
tfafter


frank-trampe:
It looks good to me.

Issue #2852, opened 2016-09-18
Layers window is broken in certain cases


jtanx:
### When reporting a bug/issue:
- [x] Screenshot

    ![layers](https://cloud.githubusercontent.com/assets/5137410/18616060/df8fab24-7de7-11e6-8608-53dee988d5cd.png)
- [x] Windows 10 64 bit (but applicable to all)

Open Source Sans Pro (UFO). Layers window is not sized correctly. Even if I undock the palette and manually resize, it does not resize the layout to match new window size.

frank-trampe:
If I recall correctly, there is an entirely different code path for the layer palette when the typeface has multiple exportable layers. It probably has not gotten the same attention as the normal layer palette.

frank-trampe:
@jtanx, you closed this via #2855, right?

jtanx:
Yep.

PR #2855, opened 2016-09-19
cvpalettes.c: Fixup the display of the layers2 palette

jtanx:
### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue) Fixes, mostly #2852

### Description
cvp21
Display as docked (note it does not auto resize depending on the longest layer name - I just made the default width wider)
cvp22
Undocked - it resizes correctly (or correct enough)
cvp23
Scrolling works (note I right clicked -> New layer) but the expose code complains with "Invalid layer" @frank-trampe ? Was this palette repurposed? The comments say it's for type 3 fonts.
cvp24
Scrolling also works when docked

* Display scrollbar only if necessary
* Make size more reasonable
* Position the layers next to the checkboxes correctly
* Actually position the scrollbar
* Get rid of a few constants and actually calculate them
* Allow scrolling on the layer buttons themselves and not just the
  scrollbar area

As with all things GDraw, this took waaaay longer than it should have. And then I just had to run my auto-indent by mistake, right before committing. Oh boy. I was not happy.

----

This needs some testing on normal builds - I've only tested this on my GDK branch so far. It's late.

### DIFF
--- a/fontforgeexe/cvpalettes.c
+++ b/fontforgeexe/cvpalettes.c
@@ -1033,7 +1033,7 @@ static IJ getIJFromMouse( CharView* cv, int mx, int my )
  * This draws the three-dimensional relief on a button. It gets the dimensions from the button image.
  */
 
-int cvp_draw_relief(GWindow pixmap, GImage *iconimg, int iconx, int icony, int selected) {
+void cvp_draw_relief(GWindow pixmap, GImage *iconimg, int iconx, int icony, int selected) {
 	extern Color cvbutton3dedgelightcol; // Default 0xe0e0e0.		
 	extern Color cvbutton3dedgedarkcol; // Default 0x707070.
 	int iconw = iconimg->u.image->width;
@@ -1560,11 +1560,13 @@ return( SplineCharAntiAlias(&dummy,ly_fore,24,4));
 }
 
 /**
- *  \brief Recalculate the number of visible layers and reposition the scrollbar
+ *  \brief Recalculate the number of visible layers,
+ *         reposition the visibility checkboxes, and
+ *         if requested, reposition/resize the scrollbar.
  *
  *  \param [in] cv The charview
  */
-static void CVLayers2Reflow(CharView *cv) {
+static void CVLayers2Reflow(CharView *cv, bool resize) {
     extern int _GScrollBar_Width;
     GGadget *scrollbar;
     GRect cvl2size;
@@ -1574,6 +1576,28 @@ static void CVLayers2Reflow(CharView *cv) {
     layer2.visible_layers = (cvl2size.height - layer2.header_height - 2 * CV_LAYERS2_LINE_HEIGHT) / CV_LAYERS2_LINE_HEIGHT;
     if (layer2.visible_layers < 0) {
         layer2.visible_layers = 0;
+    }
+
+    // Reposition the visibility checkboxes
+    int num_potentially_visible = 2 + layer2.visible_layers + 1;
+    if (num_potentially_visible > layer2.current_layers){
+        num_potentially_visible = layer2.current_layers;
+    }
+    for (int i = 2, first_visible = 2 + layer2.offtop; i < layer2.current_layers; i++) {
+        GGadget *vis = GWidgetGetControl(cvlayers2, CID_VBase + i - 1);
+        if (vis == NULL) {
+            break;
+        }
+        if (i >= first_visible && (i - layer2.offtop) < num_potentially_visible) {
+            GGadgetMove(vis, 5, layer2.header_height + (i - layer2.offtop) * CV_LAYERS2_LINE_HEIGHT);
+            GGadgetSetVisible(vis, true);
+        } else {
+            GGadgetSetVisible(vis, false);
+        }
+    }
+
+    if (!resize) {
+        return;
     }
 
     scrollbar = GWidgetGetControl(cvlayers2, CID_SB);
@@ -1642,7 +1666,7 @@ static void CVLayers2Set(CharView *cv) {
 	layer2.layers[i+1] = BDFCharFromLayer(cv->b.sc,i);
     layer2.active = CVLayer(&cv->b)+1;
 
-    CVLayers2Reflow(cv);
+    CVLayers2Reflow(cv, true);
 }
 
 static void Layers2Expose(CharView *cv,GWindow pixmap,GEvent *event) {
@@ -1884,6 +1908,7 @@ static void Layer2Scroll(CharView *cv, GEvent *event) {
 return;
     layer2.offtop = off;
     GScrollBarSetPos(GWidgetGetControl(cvlayers2,CID_SB),off);
+    CVLayers2Reflow(cv, false);
     GDrawRequestExpose(cvlayers2,NULL,false);
 }
 
@@ -1906,7 +1931,7 @@ return( true );
 	PostCharToWindow(cv->gw,event);
       break;
       case et_resize:
-        CVLayers2Reflow(cv);
+        CVLayers2Reflow(cv, true);
       break;
       case et_expose:
 	Layers2Expose(cv,gw,event);
--- a/fontforgeexe/cvpalettes.c
+++ b/fontforgeexe/cvpalettes.c
@@ -76,15 +76,16 @@ struct l2 {
     BDFChar **layers;     /* layer thumbnail previews              */
     int sb_start;         /* x pixel position of the scrollbar     */
     int column_width;     /* width of various indicator columns    */
+    int header_height;    /* height of the header in pixels before the first layer */
     int mo_col, mo_layer; /* mouse over column and layer           */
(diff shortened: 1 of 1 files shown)


fontforgeexe/cvpalettes.c
```diff
@@ -1558,6 +1559,50 @@ static BDFChar *BDFCharFromLayer(SplineChar *sc,int layer) {
...
+ */
+static void CVLayers2Reflow(CharView *cv) {
+    extern int _GScrollBar_Width;
+    GGadget *scrollbar;
+    GRect cvl2size;
+
+    GDrawGetSize(cvlayers2, &cvl2size);
+    // Magic! I verified symmetry in Paint.NET
+    layer2.header_height = GGadgetGetY(GWidgetGetControl(cvlayers2, CID_VGrid)) - 1;
+    // Minus 2 because we always have the 'Guide' and 'Back' entries
```
frank-trampe: This is at odds with the comment in the struct for `visible_layers`, which states that only the guide layer is immune from scrolling.

fontforgeexe/cvpalettes.c
```diff
@@ -1558,6 +1559,50 @@ static BDFChar *BDFCharFromLayer(SplineChar *sc,int layer) {
...
+        GGadgetResize(scrollbar, GDrawPointsToPixels(cv->gw, _GScrollBar_Width), cvl2size.height);
+        GGadgetSetVisible(scrollbar, true);
+
+        GScrollBarSetBounds(scrollbar, 0, layer2.current_layers - 2, layer2.visible_layers);
+        if (layer2.offtop > layer2.current_layers - layer2.visible_layers) {
+            layer2.offtop = layer2.current_layers - layer2.visible_layers;
+        }
+        if (layer2.offtop < 0) {
+            layer2.offtop = 0;
+        }
```
frank-trampe: Could you explain the math here a bit more? I thought that current_layers and visible_layers were not compatible numbers since current_layers includes the non-scrolling layers, and visible_layers excludes them.

fontforgeexe/cvpalettes.c
```diff
@@ -1605,30 +1650,28 @@ static void CVLayers2Set(CharView *cv) {
...
     int i, ll;
     const char *str;
-    GRect r;
+    GRect r, oldclip;
     struct _GImage base;
     GImage gi;
     int as = (24*cv->b.sc->parent->ascent)/(cv->b.sc->parent->ascent+cv->b.sc->parent->descent);
-
-    if ( event->u.expose.rect.y+event->u.expose.rect.height<CV_LAYERS2_HEADER_HEIGHT )
-return;
```
frank-trampe: Was there a reason to abort when the event was in the header? Is there a reason we are not doing that now?

fontforgeexe/cvpalettes.c
```diff
@@ -76,15 +76,16 @@ struct l2 {
     BDFChar **layers;     /* layer thumbnail previews              */
     int sb_start;         /* x pixel position of the scrollbar     */
     int column_width;     /* width of various indicator columns    */
+    int header_height;    /* height of the header in pixels before the first layer */
```
frank-trampe: Why are we making this dynamic?

fontforgeexe/cvpalettes.c
```diff
@@ -1641,22 +1684,24 @@ return;
     base.trans = -1;
     GDrawSetFont(pixmap,layer2.font);
 
-    for ( i=(event->u.expose.rect.y-CV_LAYERS2_HEADER_HEIGHT)/CV_LAYERS2_LINE_HEIGHT;
-	    i<(event->u.expose.rect.y+event->u.expose.rect.height+CV_LAYERS2_LINE_HEIGHT-1-CV_LAYERS2_HEADER_HEIGHT)/CV_LAYERS2_LINE_HEIGHT;
-	    ++i ) {
+    // +2 for the defaults, +1 to show one extra (could be half showing it)
+    layerCount = layer2.visible_layers + 2 + 1;
+    layerCount = (layerCount > layer2.current_layers) ? layer2.current_layers : layerCount;
```
frank-trampe: It's not worth changing here, but I think that having a standard conditional is more readable if the value is already set to one of the possibilities.

```
layerCount = layer2.visible_layers + 2 + 1;
if (layerCount > layer2.current_layers) layerCount = layer2.current_layers;
```

frank-trampe:
Homebrew...

fontforgeexe/cvpalettes.c
jtanx: This was actually what the current scroll code was doing, so I just mimicked its behaviour.

fontforgeexe/cvpalettes.c
jtanx: The height is based on the font used which necessitates it being dynamic

fontforgeexe/cvpalettes.c
```diff
@@ -1641,22 +1684,24 @@ return;
...
+    for (i = 0; i < layerCount + 2; ++i) {
 	ll = i<2 ? i : i+layer2.offtop;
```
jtanx: This was part of the original code - see how it skips the first two entries from scrolling

fontforgeexe/cvpalettes.c
jtanx: Ooh you're correct. It's actually being correctly limited in Layer2Scroll already anyway, so I'll just remove this block altogether (see https://github.com/fontforge/fontforge/pull/2855/files#diff-c7901426fe4e267e6c7087b9cfde6019R1883).

fontforgeexe/cvpalettes.c
jtanx: This is even done by the normal layers palette, although for some reason it's declared outside the struct as `layer_header_height`.

fontforgeexe/cvpalettes.c
jtanx: Oops, that shouldn't have been deleted. I guess it's trying to optimise since if it's in the header area, that's already drawn by the ggadget. Anyway, I'll add it back.

fontforgeexe/cvpalettes.c
jtanx: I'll change it to a normal if statement. That said, I really dislike one liner if statements

fontforgeexe/cvpalettes.c
```diff
@@ -2058,7 +2111,13 @@ return;
     if ( cv->showback[0]&1 ) gcd[4].gd.flags |= gg_cb_on;
     if ( cv->showfore ) gcd[5].gd.flags |= gg_cb_on;
 
-    GGadgetsCreate(cvlayers2,gcd);
+    // Calculate the header height. The '-1' is magic! Needed to obtain symmetry
+    layer2.header_height = GGadgetGetY(GWidgetGetControl(cvlayers2, CID_VGrid)) - 1;
```
jtanx: I've moved the header height calculation here, since there's no need to recalculate it with every call to `CVLayers2Reflow`.

fontforgeexe/cvpalettes.c
```diff
@@ -2058,7 +2111,12 @@ return;
     if ( cv->showback[0]&1 ) gcd[4].gd.flags |= gg_cb_on;
     if ( cv->showfore ) gcd[5].gd.flags |= gg_cb_on;
 
-    GGadgetsCreate(cvlayers2,gcd);
+    GGadgetsCreate(cvlayers2, gcd);
+    // Calculate the header height. The '-1' is magic! Needed to obtain symmetry
+    layer2.header_height = GGadgetGetY(GWidgetGetControl(cvlayers2, CID_VGrid)) - 1;
```
jtanx: I've moved the header height calculation here, since there's no need to recalculate it with every call to `CVLayers2Reflow`.

fontforgeexe/cvpalettes.c
```diff
@@ -1558,6 +1559,42 @@ static BDFChar *BDFCharFromLayer(SplineChar *sc,int layer) {
...
+    GDrawGetSize(cvlayers2, &cvl2size);
+    // Minus 2 because we always have the 'Guide' and 'Back' entries
+    layer2.visible_layers = (cvl2size.height - layer2.header_height - 2 * CV_LAYERS2_LINE_HEIGHT) / CV_LAYERS2_LINE_HEIGHT;
+    if (layer2.visible_layers < 0) {
+        layer2.visible_layers = 0;
+    }
+
+    scrollbar = GWidgetGetControl(cvlayers2, CID_SB);
+    // Check if we need the scrollbar or not.
+    if (layer2.current_layers - 2 <= layer2.visible_layers || layer2.visible_layers <= 0) {
```
jtanx: I've modified the check to not even bother showing the scrollbar if there isn't any room to display any modifiable layers (this can happen if the window is undocked and the user resizes to a height smaller than 3 entries high).

jtanx:
I've done some further testing against the master (X11) branch and everything seems to work fine. I've also added a commit that syncs the visibility checkboxes when you scroll.

PR #2869, opened 2016-09-24
Add 8859-16

JoesCat:
Prepping to update Unicode, noted this hasn't been included
all these years yet. Adding it to complete the list for 8859-*


Unicode/charset/iso_8859_16.c
```diff
@@ -0,0 +1,176 @@
...
+    allzeros,
+    allzeros,
+    allzeros,
+    allzeros,
+    allzeros,
+    allzeros,
+    i8859_16_from_unicode_20
+};
+
+struct charmap i8859_16_from_unicode = { 0, 32, (unsigned char **) i8859_16_from_unicode_, (unichar_t *) unicode_from_i8859_16 };
```
frank-trampe: Does this get used somewhere?

JoesCat: This is the way these 8859-* (plus other files) are built using the existing generate*.sh script.
Very little modification done beyond adding the missing 8859-16 chart.

This would probably be okay with older computers running a smaller MAXC, but it is showing itself as a bit wasteful of program binary/memory space. I haven't changed how it's built, but looking at these maps, plus unialt.c, I think these could be made a bit more memory/program efficient if we (eventually) convert these to function calls (is_Ligatures.c is an example where we go beyond MAXC to include the full Unicode chart).

In terms of 8859-16, it exists, plus there are keyboards for it as well (wikipedia). I'm a little puzzled why it wasn't added, maybe a trade-off due to memory limits in 2001-ish, which isn't such an issue these days.

frank-trampe: This really isn't a lot of data, even by 2001 standards.

frank-trampe:
Let's merge.

PR #2870, opened 2016-09-25
Add 8859-16 (dump.c)

JoesCat:
Missed these...generate.sh doesn't modify as many files as dump.c
(hand entered, need to come back to this later).


frank-trampe:
I don't understand. Is there some script that ought to have made these changes automatically?

JoesCat:
I did a diff between the originally alphabet.c vs one after running generate.sh and noted no differences, therefore did not follow-up on code based on alphabet.c

Ideally, if all went as expected, generate.sh would have upgraded alphabet.c and generated a new (inc/)chardata.h to include info for 8859-16, but it appears alphabet.c is more like a copy-n-paste at the moment. I should run dump.c (and dump.c from 20120731) to see what happens, then compare differences between the three programs. ...basically, investigate all fopen() in dump.c.

Abel went as far as he went, but it appears, his generate.sh doesn't cover everything.
The existing dump.c also affects alphabet.c and chardata.h
It's my error for not following-up with dump.c (and also dump.c from 20120731).

When a program is built under a single author (GW), then there's a pretty good idea of where things go, but at 100,000+ lines, there is also the probability of errors, especially when commenting is a bit light. Errors probably happen and things could get missed....

...then...

...When a group of us (new developers) hit a project of this magnitude, there is a high probability of patching quickly, but not quite in the right places, so even though (us) developers mean well, we might not patch in the right places to fix something (example, patching utype.h but in reality should patch makeutype.c). ...part of the problem is sparse commenting, such as... "this file was generated by program X", but this problem is getting fixed as we improve FontForge (example, makeutype.c adds "this file was generated by program X" in several places).

PR #2873, opened 2016-09-27
Edit Unicode/makeutype.c to align with existing inc/utype.h

JoesCat:
Back in approx 2012, sourceforge, or SVN had trouble with accepting
large SVN patches, therefore the patch that modified inc/utype.h
was either forgotten, or the patch was decided to leave-out since
existing code may have looked better.

This patch makes Unicode/makeutype.c create inc/utype.h code as it
exists now instead of what it would be if Unicode/makeutype.c is run.


frank-trampe:
Did you test and diff?

JoesCat:
Seeing your question, thought I'd best check FontForge master in case my fork/copy is different...
https://raw.githubusercontent.com/fontforge/fontforge/master/inc/utype.h
okay...so far. The only real difference ws removing two extra lines (blanks between 3 #define).

If you want to test it out, remember to use unicode6.3 at this point since FontForge hasn't been updated/upgraded since about 20140101. Then do a diff:
cd Unicode
gcc -s -I../inc -o makeutype makeutype.c
./makeutype
diff utype.h ../inc/utype.h

frank-trampe:
That's good enough for me.

PR #2874, opened 2016-09-28
Temporarily standardize overlap removal for path stroking.

frank-trampe:
Temporarily use the main overlap removal tool in place of the stroke-specific backtrack/overlap function when expanding strokes.

See #2840.


frank-trampe:
Oops. I found some test files, and the results were not good. I had added some incompatible wedge logic. It is gone in #2875.

PR #2877, opened 2016-09-29
Unicode 9.0

JoesCat:
Update from unicode charts 6.3 to unicode charts 9.0


frank-trampe:
I can't verify this myself. Is there some way you can test it?

JoesCat:
On September 28, 2016 05:37:21 PM Frank Trampe wrote:
> I can't verify this myself. Is there some way you can test it?

It's an ugly duckling at the moment.

Only thing that's really self testing at the moment is is_Ligatures.c

and I'm reasonably satisfied with the current behaviour of makeutype.c

To satisfy yourself with makeutype.c, use unicode text files from 6.3 to 
confirm that the existing files match what makeutype.c builds, then reload 
with unicode files for 9.0 to update the affected 5 files (utype.h, 
Arabicforms.c, is_Ligatures.c, unialts.c, utype.c (you will need to 
copy/move utype.h to the inc directory).

It was fairly broken with dump.c, makeutype.c, makebuilds.c by the time we 
hit 20140101, and then Abel introduced a major change of direction with 
Unicode/charsets.

makeubuild.c, I did a kdiff3 between the existing master vs what was 
produced. I believe some bugfixes were thrown into gdraw after the fact, but 
I haven't chased that in makebuild yet, so the edits are handmerged in ATM.

If you're going to play with charset/generate.sh, you'll need to redirect 
to another place for fetching adobe files, I found files at ftp.netbsd.org 
but it appears it's blacklisting my "ancient" desktop browser, so, I think 
this may be another copy here...
https://www.mirrorservice.org/sites/ftp.netbsd.org/pub/pkgsrc/distfiles/adobe-
cmaps-20060615/
Looking at that link, we can shorten it...

in generate.sh, add # in front of a="ftp...."
and add a new line:
a="ftp.netbsd.org/pub/pkgsrc/distfiles/adobe-cmaps-20060615/"
This should get you the adobe tar.z files you need to run generate.sh

to play with the existing dump.c you'll want to copy the files from 
Unicode/charsets/raw/*.TXT into Unicode/
The uglier ones are doing a copy and rename of 
Unicode/charsets/raw/a???/cid2.... to appropriate a.....cid2 files.
read the comments in README.txt and generate.sh

I think I better edit dump.c or the README.txt

looking at all this mess, there's still something missing since I haven't 
seen anything here affecting tables for unicoderanges.c ...so I suspect I 
still need to look at 20140101, or 20120731

...maybe better to leave this PR open a little longer, but at least it 
gives you something to look at for the moment.

frank-trampe:
@JoesCat, the ligatures and fractions test failed on the Travis build.

JoesCat:
Thanks for the approval @cicku.
@frank-trampe, if you want to merge now, go ahead, but I think fontforge/unicoderange.c still needs to be upgraded to work properly with 9.0. This will probably have to be done by hand (like for 6.3) tonight. I think there might be a program that does the table (preferred method). Having seen the Travis workarounds, I'm not sure adding the ftp address in generate.sh was such a great idea....maybe better commented out....plus, there are probably more uptodate info on github now that Adobe moved several files to BSD+3 (follow the adobe redirect for dingbats).

frank-trampe:
I would prefer to hold this until it is finished, which seems to be soon.

JoesCat:
...used libuninameslist nameslist.c for reference to fill-in unicoderange (if anyone wants to compare.
...will have to find an automatic method (later if it doesn't already exist).
@frank-trampe  ...ready to merge after checks done

PR #2878, opened 2016-09-29
Consolidate fontforge/ headers, part1

gioele:
Here is the first bunch of standalone cleanups for the headers in `fontforge/`.

Similary to PR #2846, the commits in this PR move all the `extern` prototypes of non-static functions of various `.c` files to their respective `.h` files. In this first part, the headers of the files from `fontforge/activeinui.c` to `fontforge/dumppfa.c` (in alphabetical order) have been consolidated. `splinefont.h` is starting to clean up and some implicit dependencies are now explicit, hinting at where the future code clean-ups and refactorings should start from.

As with the previous PR, no implementation code has been touched (except a tiny change in `autosave.c`).

My plan is to consolidate about 30 files in the next PR (a third of the total). If you prefer, I can instead make a single huge PR with all the ~100 commits in it.


gioele:
(Rebased to fix a small clash between `sd.h` and POSIX `grp.h` found by Travis.)

JoesCat:
Makes the code a bit more wordy, but at the same time it gets rid of some ugly kludges.

These are affecting a lot of files at one time and can potentially cause merge issues. 
Let's not make a ~100 patch PR, as it will probably affect a few coders along the way.

PR #2879, opened 2016-09-29
Enable Travis-CI on all branches, not only master

gioele:
This PR removes the Travis-CI setting that limits builds to the master branch. The rationale for this is:

1. Trying to build a pushed branch always provides useful information, even for WIP branches.
2. It is hard for contributors to test their own branches before submitting PRs, leading to brown-bag rebases like [this](https://github.com/fontforge/fontforge/pull/2878#issuecomment-250617450) ;).

One could remove that setting their own fork, but keeping diverging master branches is not fun.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ compiler: gcc
 sudo: required
 dist: trusty
 osx_image: xcode8
-branches:
-  only:
-    - master
 os:
   - linux
   - osx


jtanx:
I rebase all the time when I make PRs. E.g. see my GDK pr. When you submit a PR, it has the same effect as running the build on travis, so why not just submit a wip pr to get the travis builds if you need them?

Otherwise I guess this might be ok. Don't do this for AppVeyor as you're only allowed one build at any one time.

frank-trampe:
I think it's fine just to test the active pull requests.

PR #2880, opened 2016-09-29
Remove dead prototypes (A less controverse version of PR #2817)

gioele:
This PR is a reduced version of PR #2817 that only removes prototypes of functions that have been removed years ago or never implemented.


frank-trampe:
Okay.

PR #2887, opened 2016-10-03
Adjust the build in order to suit Xenial.

frank-trampe:
libtoolize now refuses to override files without the `--force` option, which is a problem for repeated builds and also for building from the Debian source packages, which are bootstrapped and configured as necessary to generate the Makefile that generates the source package.

The default version of automake is 1.15, but the build system is hardwired to use 1.11 and fails when it is missing. We now specify the version in the Debian control file.


frank-trampe:
@jtanx, take a look.

frank-trampe:
Actually, I'm just going to merge this.

Issue #2892, opened 2016-10-05
Can't open .sfd files


raphaz:
FontForge doesn't let me open several .sdf files I created in the last couple days. 
I was able to open and close them before. 

I crashed a couple times when working in FontForge yesterday and today, usually when zooming in or moving points, on rather random occasions. 

When trying to open the files FontForge is asking to recover it. When I click "yes" or "yes to all" nothing happens. If I click "skip for now" an "Untitled1.sfd" glyph window opens (not the actual file); when I then open the respective file through the file/open dialog from here I go through the same recovery routine like mentioned above, to no avail.

Is there any way I can recover my work from the last couple days?!

Thank you in advance!
Rapha

frank-trampe:
@raphaz, are you saying that the old release handled them correctly and the new release crashes? Or that the files got corrupted somehow and cause crashes?

What version of FontForge are you using, and on what platform?

raphaz:
Sorry, I guess I wasn't specific enough.
I can't open .sdf files anymore, the same files I was able to open yesterday and a couple hours ago today. 
This happened after FontForge crashed a couple times.
I didn't change or update XQuartz or FontForge since I last opened the files.

macOS Sierra 10.12X; Quartz 2.7.9 (xorg-server 1.17.4); FontForge 20160402 

I just restarted the Mac and I still can't open any .sdf files at all.

When I open XQuartz I get a xterm window with bash-3.2$

![ffopendialogue](https://cloud.githubusercontent.com/assets/22640974/19120814/405e08c6-8af2-11e6-8b4b-0b2371d12923.jpg)
![ffopenoptions](https://cloud.githubusercontent.com/assets/22640974/19120812/404d5bf2-8af2-11e6-8667-c6fb9c50044a.jpg)
![xquartz_xterm](https://cloud.githubusercontent.com/assets/22640974/19120813/405a574e-8af2-11e6-8fa6-0ce408e0610c.png)

frank-trampe:
Your autosave files might be in `~/.config/fontforge/autosave`. See what's in there.

Do you absolutely need to recover the files? If so, you'll probably need to provide a copy of your autosave directory. If not, just move or delete it, and you ought to be able to open your original files.

If you can get FontForge to crash consistently, we would also appreciate getting the specific steps necessary to make it do so.

raphaz:
Strangely enough I managed to open a couple older files and one of today's files through the FF open dialog but I still can't open any of the files I worked on the last couple hours today.

[...]

PR #2893, opened 2016-10-05
Fix a crash from loading an empty sfd.

frank-trampe:
Fix a crash that results from dereferencing a blank encoding map resulting in turn from loading an sfd with no glyphs.

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -8808,6 +8808,8 @@ exit( 1 );
 	SFD_AssignLookups((SplineFont1 *) sf);
     if ( !d.hadtimes )
 	SFTimesFromFile(sf,sfd);
+    // Make a blank encoding if there are no characters so as to avoid crashes later.
+    if (sf->map == NULL) sf->map = EncMapNew(sf->glyphcnt,sf->glyphcnt,&custom);
 
     SFDFixupUndoRefs(sf);
 return( sf );


frank-trampe:
This came up in #2892.

frank-trampe:
@jtanx, take a look.

jtanx:
@frank-trampe see here: https://github.com/fontforge/fontforge/issues/2892#issuecomment-251824329

Are we sure this isn't fixed by 3b18b01ad983e5b4fd6fef6de0cd6010e3bf165e?

jtanx:
Ok. It seems a bit odd that in some cases, a null map is okay, but in others it's not.

frank-trampe:
Hmm. I patched it this way in order to simplify the fix (rather than tracking down all the places map might not be null-checked). Looking at your patch, though, I think that it's important that we make a decision as to whether it is okay for a loaded SplineFont passed to a FontView to have a null map and to fix accordingly. What do you think?

jtanx:
@frank-trampe hmm yes this would probably require a bit more thinking about where it's okay or not to have a NULL map, but in the meantime, I think this fix is ok.

frank-trampe:
Upon further study, I have decided that it does make more sense to require that every SplineFont have an encoding map. Otherwise, we would need to a support for an indeterminate state.

PR #2896, opened 2016-10-07
cvruler.c: Fix placement of prev_rect. Fixes #2895.

### LINKED ISSUE #2895: Ruler tool measurements misaligned [linux]
ff2016-10-ruler-number-glitch

Started working with the new October '16 release on Ubuntu (Xenial / 16.04). When using the ruler/measure tool in a glyph window, the segment-measurement numbers are not aligned to the measurement line; they are offset by a significant number of pixels (this is what's depicted in the screenshot attached).

So far, this is reproducible for me on any glyphs that has actual contours (components are also triggering an offset, it seems, but that's less of an issue up front). The numbers are supposed to sit next to the line segments that the tool measures; as it is, you can sort of work out which number goes with which measurement -- unless they are tightly spaced, in which case the numbers crowd over each other and may not be readable.

jtanx:
### Types of changes

### Description
This is CID1082720. The prev_ruler variable was getting instatiated
on every run of the for loop, defeating the purpose of this variable.

This is a very long standing bug (present since at least 2013), that
seems to only show under higher optimisation levels (-O2? -O3?).

cc @frank-trampe

### DIFF
--- a/fontforgeexe/cvruler.c
+++ b/fontforgeexe/cvruler.c
@@ -964,6 +964,7 @@ return;
 	int len;
 	int charwidth = 6; /* TBD */
 	Color textcolor = (cv->start_intersection_snapped && cv->end_intersection_snapped) ? measuretoolcanvasnumberssnappedcol : measuretoolcanvasnumberscol;
+	GRect prev_rect;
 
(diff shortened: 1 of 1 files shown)


frank-trampe:
Frank approves.

PR #2897, opened 2016-10-08
scripting.c: Don't free format_spec; free tmp.

jtanx:
### Types of changes

### Description
Just addressing some things that Coverity noticed.

CID1364088, CID1357888. It's no longer necessary to do so, because format_spec
will never point to a buffer that needs freeing - all the dynamically
allocated buffers are freed elsewhere.

### DIFF
--- a/fontforge/scripting.c
+++ b/fontforge/scripting.c
@@ -2358,6 +2358,7 @@ static void bExport(Context *c) {
 #endif
     if (( format>=4 && c->a.argc!=3 ) || (format<4 && c->a.argc==3 )) {
 	c->error = ce_wrongnumarg;
+	free(tmp);
 	return;
     }
     bdf=NULL;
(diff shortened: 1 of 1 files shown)


fontforge/scripting.c
```diff
@@ -2374,8 +2375,6 @@ static void bExport(Context *c) {
 	if ( c->curfv->selected[i] && (gid=c->curfv->map->map[i])!=-1 &&
 		SCWorthOutputting(c->curfv->sf->glyphs[gid]) )
 	    ScriptExport(c->curfv->sf,bdf,format,gid,format_spec,c->curfv->map);
```
frank-trampe: I notice that there is code above that might point `format_spec` to `tmp = pt`. Are we sure that `format_spec != tmp` here?

jtanx: I'm not sure it matters. The whole purpose of `tmp` is to simply store the address of the dynamically allocated buffer, such that it can be freed later.

frank-trampe: Right. But if `format_spec = pt` and `tmp = pt` and `free(tmp)` happens before this line, ...

jtanx: But it can't? `free(tmp)` is only called before the function returns  (line 2361) or at the end of the function (out of the for loop at line 2378)

frank-trampe: Wow. I must have been really tired!

frank-trampe:
It looks good to me.

PR #2904, opened 2016-10-11
Adding support for https when openning font url

dmvieira:
### Motivation and Context
- [X] Why is this change required? What problem does it solve?

### Types of changes
- [X] New feature (non-breaking change which adds functionality)

### Description
Now we can load fonts with https urls

### Final checklist

### DIFF
--- a/fontforge/http.c
+++ b/fontforge/http.c
@@ -568,7 +568,7 @@ return( false );
 	    password=copy("FontForge");
     } else if ( password==NULL )
 	password = copy("");
-    
+
     sprintf(cmd,"USER %s\r\n", username);
     if ( ftpsendr(soc,cmd,databuf,datalen)== -1 ) {
 	ff_progress_end_indicator();
@@ -594,7 +594,7 @@ return( false );
 	ChangeLine2_8(_("Requesting font..."));
     else
 	ChangeLine2_8(_("Transmitting font..."));
-	
+
     if ( ftpsendpassive(soc,&data_addr,databuf,datalen)<= 0 ) {
 	ff_progress_end_indicator();
 	close( soc ); free(filename); free(databuf);
@@ -663,7 +663,7 @@ return( true );
 FILE *URLToTempFile(char *url,void *_lock) {
     FILE *ret;
 
-    if ( strncasecmp(url,"http://",7)==0 )
+    if ( strncasecmp(url,"http://",7)==0 || strncasecmp(url,"https://",8)==0 )
 return( HttpURLToTempFile(url,_lock));
     else if ( strncasecmp(url,"ftp://",6)==0 ) {
 	if ( FtpURLAndTempFile(url,&ret,NULL))


fontforge/http.c
```diff
@@ -663,7 +663,7 @@ return( true );
 FILE *URLToTempFile(char *url,void *_lock) {
     FILE *ret;
 
-    if ( strncasecmp(url,"http://",7)==0 )
+    if ( strncasecmp(url,"http://",7)==0 || strncasecmp(url,"https://",8)==0 )
```
frank-trampe: This seems very unlikely to work. HttpURLToTempFile also checks the protocol and requires `http://`, and then it performs a raw socket write of the request.

dmvieira: You are right, I'll fix it, but it really needs double validation?

I'm new here... How could I run tests? I don't know how could I test it

jtanx: I'd rather not even have the ability for FontForge to open files directly from the internet. FF's implementation is horribly insecure/naive at best.

frank-trampe: @jtanx, we can't pull features without upsetting people, and nobody is under the illusion that FontForge is secure. Besides, it doesn't look like fixing the request code is that difficult if we really wanted to do it.

frank-trampe: @dmvieira, what are you trying to accomplish? There may be a better way to do it than using this feature.

dmvieira:
Thanks you for your attention @frank-trampe 

My real problem is how fontforge open font as a file. 

I'm using fontforge as a python module inside my service: https://github.com/the-fontain/fontcut

Now I need to do some IOs openning and generating new font file and it is very bad for a web service.

My perfect solution should be using fontforge to read font content from string or bytes and write the same.

Now I took a look the code and see that fontforge only works with files or URLs... My font URLs are at most https, so I tried to optmize at last it...

frank-trampe:
@dmvieira, it's funny you write about such a thing. As we speak, I'm writing a wrapper for a JavaScript function (from a binding for GraphicsMagick) that deals only with files so that it can take and make buffers (using the tmp library in order to make temporary files). Sometimes, this is the best way to do things, and I think that it is in your case.

FontForge uses a lot of putc and fprintf and fwrite statements, and so it would be difficult to capture output directly into memory. So maybe try something like I'm doing in JavaScript in Python. Write a Python function that takes bytes and some arguments as input and gives bytes as output. Have it write the input bytes to a temporary file provided by tempfile, and make an empty temporary file. Have FontForge read from the one and write to the other. Read the output file into memory, delete the temporary files, and return the output.

fontforge/http.c
frank-trampe: @jtanx, that would be great. But don't wait on me to write that patch.

dmvieira:
@jtanx got my point... Thanks @frank-tramp, I'll try to do this patch and close this PR right now... I'll need some help, but I'll try :)

frank-trampe:
@dmvieira, I wasn't saying that you need to submit a patch to FontForge that does these things. That would be a lot of work since it would need to handle a lot of cases. I was suggesting that you write a wrapper in your software that handles just the case you need.

fontforge/http.c
jtanx: @frank-trampe I meant don't even have the ability to open URLs in FF. Just use whatever Python provides to download the file to some location, then pass that location to FF for it to load/process.

frank-trampe: Oh. Well we can't do that. I'd get some angry e-mail, and then we'd be forced to reimplement the functionality securely.

dmvieira:
@frank-trampe I already use this wrapper in that project, but with fontforge I need to do a lot of IOs and it's terrible for a web server. I need performance and I cant get it Reading from disk and writing to disk all the time.

I'll try to do New methods in fontforge like "loadFromStream" and "generateToStream" that uses only memory, buy I see this needs a lot of work

jtanx:
Quite a while ago, I had an idea of switching out all the file operations in FF to use GIO file streams, which would:

* Eliminate the file opening issues on Windows when using unicode paths/paths that contains characters not in the user's current locale encoding
* Eliminate the use of temporary files, because it can then work on in-memory files

I guess it's pretty similar to @mskala's file system abstraction idea, although using a standardised library. Fairly hesitant to actually do this, mainly because it is a large change, but also because you would then have to use GIO for all file operations instead of standard C operations. Eh.

frank-trampe:
There are two possible causes for the performance issues that @dmvieira is experiencing. Either it's the overhead from the system calls, or it's the actual disk interaction.

* If it is from system calls, the overhead becomes insignificant when performing any meaningful work in FontForge, such as parsing a file into a SplineFont.
* If it is from actual disk interaction, there are easy solutions such as enabling write caching or just using a RAM disk.

Either way, I don't see this issue as a weight on the scale for restructuring file operations. If we were to do that, we would probably need to implement a wrapper and to make it modular, and, as @jtanx mentions, it would be a very, very large change and would probably break patch compatibility between FontForge and all existing forks.

jtanx:
Uhh what forks are we talking about here? Sortsmill has gone on its own way, as has FontAnvil. The sort of change I mentioned would be no more breaking than the current situation is like anyway. It's just that it is a large change, and you would need to think very carefully if this library is suitable and worthwhile.

frank-trampe:
It's perhaps not the big deal it would have been a few years ago, but we merged a patch from Sortsmill just a few months ago. I don't think it's worth the time and the effort to pursue this sort of change.

PR #2905, opened 2016-10-12
fix tests/fetched-fonts/ on .gitignore

MichinariNukazawa:
Fix a download font file is not included in .gitignore, from `$ make check`.

Description to tests/.gitignore add "/fetched-fonts/*" line.


frank-trampe:
These issues are normally invisible to me since my source tree is full of extraneous files. Was this really the only .gitignore aberration you found?

MichinariNukazawa:
Is it not generated even if you perform this in new tree?
'$make check'
'$git status'
"working directory clean" in your tree?

PR #2909, opened 2016-10-13
fix duplicate download font file for test.

MichinariNukazawa:
fix duplicate download font file for test.
## Reproduction

run `$make check`
download "tests/fonts/MunhwaGothic-Bold" and "tests/fetched-fonts/MunhwaGothic-Bold" is duplicate file.
## fix

remove downloading "tests/fonts/MunhwaGothic-Bold".
and depended tests changed.
## concerned
- Now using path separator is '/'. I don't know open this file path in Windows.
  (Do run the this test in Windows?)
- test006.pe is already 'skipped'.
  (test003.pe and test006.pe is using "MunhwaGothic-Bold".)


frank-trampe:
@MichinariNukazawa, there was a reason for this, not necessarily a very strong one, but a reason.

Launchpad blocks general network access, so we needed to include things in the source bundle that the build system would download. But I wanted to implement that in a separate layer so that the bundled/downloaded distinction would be abstracted out of the tests themselves.

The file in question is not very big.

JoesCat:
I note that sometimes @frank-trampe and @jtanx have to occasionally fix Travis or redirect to another webpage due to blocking (which is why I wouldn't recommend in adding Unicode/make* or Unicode/charset/generate.sh as part of makecheck or makedist).

I see there is a separate folder in FontForge with this font file too.
Would redirecting here: https://raw.githubusercontent.com/fontforge/debugfonts/master/MunhwaGothic-Bold be any better? or do we end-up also having blocking problems too (example a distro doing make and pulling files out of github (trading one problem for another))?

jtanx:
Nah the Travis issues wrt. this file should be gone now. The only reason it was an issue was because it used to download from the netbsd mirror. But they did strange things to their mirror and started moving the files around quite a few times. So I gave up and self-hosted it on Github. Since Travis basically only runs if Github is up, it's no longer an issue.

MichinariNukazawa:
Do we add MunhwaGothic-Bold to fontforge/fontforge repository?
All download code is delete.

MunhwaGothic-Bold is 563kb.
tests/fonts/NimbusLGCUni-Regular.sfd is 559kb
// alreary included font file max size. (but this is text format.)

jtanx:
I'd prefer not to - this repository is already fairly large as is.

jtanx:
This should be fixed in current master with #3111 and #3120

Issue #2823, opened 2016-09-11
Compiling with float vs double for the real type produces different results


jtanx:
This is a follow-up to #1747. As noted at the end, compiling with `--enable-real=float` (i.e. the setting for the 2012 release) results in different output as compared to `--enable-real=double` (every release since 2012). Another case would be #2774 and #2739.

frank-trampe:
@copyninja, what is the problem that you get building Pecita with double-precision floating point enabled?

copyninja:
@frank-trampe I've done a normal build as of now so it fails in validate() function. Does the fontforge upstream recommend use of `--enable-real` flag? if so which version should I use `float` or `double`

frank-trampe:
@copyninja, any idea as to what problem, specifically, is causing the validation failure? What's the output?

There are almost certainly edge cases where `float` works better than `double`, but we develop with `double` now and intend to keep moving towards that as the standard.

copyninja:
No error as such this is what I see when I run compile.pe provided by font upstream
```
./compile.pe 
Copyright (c) 2000-2014 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Based on sources from 20161010-ML-D.
 Based on source from git with hash:
Error: Validation error
Called from...
 ./compile.pe: line 1967
```
Its failing in validate call below

```
if (Validate() == 0)
 Generate('Pecita.otf')
else
 Error('Validation error')
endif
```
Is there anyway to make Validate output more details?

copyninja:
Also is `double` type default or do I need to specifically pass `--enable-real=double` to configure?. 

Sorry not so much expert in autotools but I do see in m4/fontforge_arg_enable.m4 that default value is indicated as double

frank-trampe:
As far as I know, the default value is `double`.

Perhaps try exporting anyway, opening in graphical FontForge, and running validation there? Otherwise, we probably need to use `gdb`.

copyninja:
OK I opened the generated sfd using fontforge graphical interface and clicked on validate and I got following [error](http://imgur.com/a/QTj9c). In console I've following
```
Copyright (c) 2000-2014 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
[...]

PR #2910, opened 2016-10-13
remove build option FONTFORGE_CONFIG_USE_DOUBLE

MichinariNukazawa:
remove build option FONTFORGE_CONFIG_USE_DOUBLE (FONTFORGE_ARG_ENABLE_REAL).

Is there the person using & testing this?

todo : remove defined "real" and "bugreal" in fontforge/splinefont.h


frank-trampe:
We can't do this.

jtanx:
Yeah unfortunately not. There are unresolved differences between compiling with float or doubles (see #2823).

PR #2915, opened 2016-10-16
replace type boolean int to bool in fontforgeexe/startui.c and other.

MichinariNukazawa:
replace int type of boolean like use.
change fontforgeexe/startui.c and relation files.

I don't touched "int splash".

concerned : 
The int splash variable is not changed.
Because extern in fontforgeexe/prefs.c and pointer convert to void * .

Todo : 
The int openflag variable is exist enum type.

### DIFF
--- a/fontforge/autosave.c
+++ b/fontforge/autosave.c
@@ -78,13 +78,13 @@ return;
 }
 
 
-int DoAutoRecoveryExtended(int inquire)
+bool DoAutoRecoveryExtended(int inquire)
 {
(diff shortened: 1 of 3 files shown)


frank-trampe:
I really don't want to start down this path. It would give us conflicting coding styles, clutter the changelog, confuse blame, and decrease patch mergeability. And it doesn't really offer any tangible benefit.

MichinariNukazawa:
OK.
Thank you for your check.

PR #2916, opened 2016-10-17
Travis improvements

### LINKED ISSUE #2859: Travis cleanup
Wishlist

* Simplify the Travis build procedure
    * Currently the Linux build essentially runs twice - the first time to compile/install FontForge, and the second when it runs distcheck. Not sure if we can do something about that.
* Make it make automated builds
* Make it possible to run automated Coverity builds

jtanx:
This greatly simplifies the Travis CI process while bringing a number of improvements too.

* Build has been largely consolidated back to `.travis.yml` instead of being spread out over multiple files
* Build times are reduced
    * Switched to container based builds means we can leverage the cache feature of Travis
    * Once the cache is established, the Linux build usually takes around 7-9 minutes. Otherwise around 12-14 minutes. The OSX build takes around 9-11 minutes with cache, and 13-16 minutes without.
* Coverity builds are now automated - just update/push master to the `coverity_scan` branch and it should run a new scan. Just make sure to cancel the OSX build because that also runs when it shouldn't...
* It's now building automated OSX bundles<sup>1</sup> that get pushed to Bintray: https://dl.bintray.com/fontforge/fontforge/
    * Quite a few of the 'features' are left out - Breakpad doesn't work, and Even isn't bundled. And all the other cruft that the bundle attempts to build in. Ugh. I don't like the idea of bundling everything and the kitchen sink into the .app. Very hard to maintain.
    * Not sure if the process exits cleanly
    * Bundles are only pushed when the CI build is being run for the `master` branch - not on PRs. This is a limitation of the deploy feature.

Resolves #2859


frank-trampe:
Well, the important thing here is that the Travis builds succeed, and they do. I looked at the code also, but not terribly closely.

PR #2918, opened 2016-10-19
ufo: Avoid empty format strings

gioele:
This change fixes the warning

```
ufo.c: In function 'injectNumericVersion':
ufo.c:109:3: warning: zero-length gnu_printf format string
[-Wformat-zero-length]
   if (versionMajor == -1) asprintf(textVersion, "");
      ^
```

### DIFF
--- a/fontforge/ufo.c
+++ b/fontforge/ufo.c
@@ -101,7 +101,7 @@ static void extractNumericVersion(const char * textVersion, int * versionMajor, 
 
 static void injectNumericVersion(char ** textVersion, int versionMajor, int versionMinor) {
   // We generate a version string from numeric values if available.
-  if (versionMajor == -1) asprintf(textVersion, "");
+  if (versionMajor == -1) asprintf(textVersion, "%s", "");
   else if (versionMinor == -1) asprintf(textVersion, "%d", versionMajor);
   else asprintf(textVersion, "%d.%d", versionMajor, versionMinor);
   return;


frank-trampe:
Per [this thread](https://patchwork.ozlabs.org/patch/67421/), the warning serves no legitimate purpose whatsoever and is forcing us to generate unnecessarily verbose code (either applying this patch or explicitly disabling the warning). Is there any way we could lobby to have the warning removed from `gcc`?

frank-trampe:
@davelab6, you know Richard Stallman, right?

gioele:
Well, the alternative is something like

```
if (versionMajor == -1) { textVersion = malloc(1); textVersion[0] = '\0'; }
```

I would not worry much about performances and verbose code in the current Fontforge codebase. ;)

frank-trampe:
@gioele, unless you're sure that this is the last case, I think that it would be vastly better to change the build options.

davelab6:
@frank-trampe huh? :)

frank-trampe:
If not, disregard, but I thought you knew him somehow. As much as this bothers me, I don't want to go wading into forums over it, so any opportunity for direct contact would be great.

davelab6:
I do but what does this have to do with him? :)

gioele:
@frank-trampe: this is the only occurrence of this problem. I am reluctant to disable any warning in the fragile fontforge codebase, any nudge from the compiler is welcome. ;)

Issue #2912, opened 2016-10-14
Glyphs exported as SVG have unexpected viewBox width


robmadole:
FontForge version: built from Git commit hash fe1d84b7d05178885a36ba0adce0559a82a493d5

Steps to reproduce:
1. Open FontAwesome.otf (in attached Zip file)
2. Select glyph `/resize_vertical`
3. Export to SVG
4. Open the SVG in a text editor

Expected:

The viewBox on the SVG will be `0 -144 432 1008`.

Actual value:

`0 -144 1008 1008`

This problem makes the SVG unusable in situations where a correct viewBox is required. One thing of note is that this apparently only occurs with glyphs that are narrower than the EM unit. (If you export `camera` the viewBox is correct.)

Here is the icon in FontForge:

![resize_vertical_ff](https://cloud.githubusercontent.com/assets/132895/19404014/726d56c4-9230-11e6-9593-cfef63fa525a.png)

Here is the icon opened in Illustrator:

<img width="720" alt="opened_in_illustrator" src="https://cloud.githubusercontent.com/assets/132895/19404034/9a4d0b76-9230-11e6-8008-25e23d93e714.png">

OTF file used to export SVGs:

[FontAwesome.otf.zip](https://github.com/fontforge/fontforge/files/530526/FontAwesome.otf.zip)

robmadole:
I think this behavior is caused by [this line](https://github.com/fontforge/fontforge/blob/f868b218516c734b3f8573019d05548b437391f5/fontforge/svg.c#L1040).

frank-trampe:
It looks wrong to me, but there must have been a reason for it. @davelab6, @mskala, @jtanx?

davelab6:
No idea, sorry :(

frank-trampe:
Okay. No objections.

@robmadole, would you submit a pull request with the line commented out and a note explaining why?

robmadole:
@frank-trampe let me put up a PR of what I'm currently running

skef:
Fixed by #2919

PR #2919, opened 2016-10-19
Use glyph height and width instead of bounding box when exporting as SVG

robmadole:
Possible solution to #2912. See that issue for the context behind why this change is being proposed.


frank-trampe:
Fine with me. @mskala, any objections?

mskala:
Looks okay to me.  I don't have a strong sense of what's going on with the SVG code, though, and I haven't had much attention on FF development in general for several weeks because I'm in the middle of organizing my intercontinental move.  Current status:  apartment rented, will move in Nov. 3, probably won't get my main computer for a couple weeks after that.

frank-trampe:
@mskala, I've never needed to move furniture over an ocean, but I'm sure it's no fun. Best of luck with all that.

PR #2921, opened 2016-10-21
Mark three small functions static

gioele:
These two commits mark three small functions as static. They were never exported and are used only in the file where they are defined.

Found while working on part 2 of PR #2878.


JoesCat:
Thanks!

Issue #2920, opened 2016-10-21
Glyph name confusion


kilobyte:
When trying to build [3270font](https://github.com/rbanffy/3270font) with new fontforge, there's a regression wrt glyph names, and a validation failure.
Because of spurious dependencies, it's easiest to try it from my tree:
```
git clone git@github.com:kilobyte/3270font.git;git checkout 4978e16;make test
```
Fontforge versions 20120731 and 20160404 worked ok, 20161005 interprets "Omega" as deprecated U+2126 OHM SIGN rather than regular U+03A9 GREEK CAPITAL LETTER OMEGA.
```
Attempt to unget two characters
CHECKING     3270Medium.otf
The glyph named Omega is mapped to U+03A9.
  But its name indicates it should be mapped to U+2126.
PASS         3270Medium.otf

Copyright (c) 2000-2014 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Based on sources from 20161015-ML-D.
 Based on source from git with hash: 
Attempt to unget two characters
CHECKING     3270Medium.pfm
ERROR     22 Two glyphs have the same Unicode code point
FAIL         3270Medium.pfm
```
In this particular case, there's an easy workaround: explicitly writing Omega.greek instead of Omega makes the failure go away, but I'm reporting in case this affects more fonts that are harder to change.

frank-trampe:
@JoesCat, this looks like an encoding issue.

JoesCat:
Hi @kilobyte,
Whether this bug is fixed or not, I do recommend that you do use the longer more detailed definitions, such as Omega.greek, and not the shorter definitions to avoid these sorts of potential bugs (Upstream bug here: https://github.com/rbanffy/3270font/pull/25 ). I'm +1 on recommending the extra wording.

Doing a grep and searching for terms such as "3A9", "2126" shows some fluidity within FontForge (example: https://github.com/fontforge/fontforge/commit/20b9ce1cbcfdc7d620b2b11ab5f7f075987fbad4 ), as well as with upstream charts with new characters introduced and others deprecated.
Looking at other tables such as Tex encoding or other spots with variables like ".1992", ".1980", suggests we may need to add another variable, like YEAR or VERSION when it comes to import/export

Thanks for the recommendation of follow-up using your git tree, this should help narrow it down.

[...]

PR #2922, opened 2016-10-23
Remove characters from is_Ligature.c lookup tables

JoesCat:
makeutype.c looks for ligatures and fractions, but these are characters,
therefore exclude these characters and remove them from lookup tables.

### DIFF
--- a/Unicode/is_Ligature.c
+++ b/Unicode/is_Ligature.c
@@ -79,8 +79,7 @@ const uint16 ____ligature16[] = {
   0xfef6, 0xfef7, 0xfef8, 0xfef9, 0xfefa, 0xfefb, 0xfefc};
 
 const uint32 ____ligature32[] = {
-  0x00011176, 0x0001f670, 0x0001f671, 0x0001f672,
-  0x0001f673};
+  0x00011176};
 
 const uint16 ____vulgfrac16[] = {
   0x00bc, 0x00bd, 0x00be, 0x2150, 0x2151, 0x2152, 0x2153, 0x2154,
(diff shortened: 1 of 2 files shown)


frank-trampe:
@JoesCat, is this related to #2920?

JoesCat:
No, this is unrelated (and written before issue #2920 was brought up), but is related to improving the list of ligatures and fraction definitions in "is_Ligatures.c".
I was looking at functions for expanding ligatures and fractions, and on closer inspection noted the "fraction" was a character and needed to take it out of the table. On further inspection I also found ligature characters that would make more sense calling them characters instead of ligatures (you can follow the added "if" statements in makeutype.c). is_ligatures.c tables are updated with the removed chars.

frank-trampe:
@JoesCat, this is all sort of difficult to check. Could we, not necessarily now but on some future rewrite, start using the sizeof operator?

JoesCat:
I'm thinking that older code going back to 2000ish may be less likely to contain sizeof() operators. Compilers weren't smart enough to reduce stuff like sizeof(char) to 1....and if smart enough, take that and use ++value or --value (which is a code reduction and a speed improvement over value=value+1).
This is not much of a big deal when working with CPUs like the Intels and AMDs, but more of an issue with older/slower CPUs like 386, sparcs, mips, arms, etc... ...makes sense if you think of it this way.
Patch above has no references to sizeof(), so I'm not sure if you're pointing out old pre-existing code or new recently-introduced code.

frank-trampe:
@JoesCat, the sizeof operator, as far as I know, has always generated a fixed value at compile time.

The reason I suggest it here is to avoid mismatches between the size of the array as declared/initialized and the size as used.

PR #2933, opened 2016-11-02
File descriptor leak fixed + maps deference fixed

### LINKED ISSUE #2903: File descriptor leak in fontforge 2.0.20161004
There is a file descriptor leak in fontforge 2.0.20161004. When converting a large number of fonts you will eventually run out of open file descriptors.

The problem is in"redohead" in tottf.c

If you redo the head, the previous file handle is not closed. You need to change 

```
static void redohead(struct alltabs *at) {
    at->headf = tmpfile();
```

to

```
static void redohead(struct alltabs *at) {
    if (at->headf) fclose(at->headf); // file descriptor leak
    at->headf = tmpfile();
```

davidhedley:
tottf.c:
If you redo the head, the previous file descriptor is not closed
`Closes #2903`

tottfgpos.c:
If maps[cnt] is NULL you print an error message but then immediately deference it = crash. I've seen this with Arial Unicode MS

### DIFF
--- a/fontforge/tottfgpos.c
+++ b/fontforge/tottfgpos.c
@@ -1847,11 +1847,16 @@ static void dumpGSUBmultiplesubs(FILE *gsub,SplineFont *sf,struct lookup_subtabl
 	putshort(gsub,offset);
 	if (maps[cnt] == NULL) {
 		fprintf( stderr, "maps[%d] is null; glyphs[%d] is \"%s\"; lookup name is \"%s\".\n" , cnt , cnt , (glyphs[cnt]->name ? glyphs[cnt]->name : ""), sub->subtable_name) ;
-	}
-	for ( gc=0; maps[cnt][gc]!=NULL; ++gc );
+            gc = 0;
+	} else {
+    	    for ( gc=0; maps[cnt][gc]!=NULL; ++gc );
+        }
 	offset += 2+2*gc;
     }
     for ( cnt = 0; glyphs[cnt]!=NULL; ++cnt ) {
+        if (maps[cnt] == NULL) {
+            break;
+        }
 	for ( gc=0; maps[cnt][gc]!=NULL; ++gc );
 	putshort(gsub,gc);
 	for ( gc=0; maps[cnt][gc]!=NULL; ++gc )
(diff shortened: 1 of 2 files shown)


fontforge/tottfgpos.c
```diff
@@ -1847,11 +1847,16 @@ static void dumpGSUBmultiplesubs(FILE *gsub,SplineFont *sf,struct lookup_subtabl
...
+            gc = 0;
+	} else {
+    	    for ( gc=0; maps[cnt][gc]!=NULL; ++gc );
+        }
 	offset += 2+2*gc;
     }
     for ( cnt = 0; glyphs[cnt]!=NULL; ++cnt ) {
+        if (maps[cnt] == NULL) {
+            break;
+        }
```
frank-trampe: Could we push this condition into the for loop condition?

frank-trampe: It would make the code simpler and more readable.

fontforge/tottfgpos.c
```diff
@@ -1847,11 +1847,16 @@ static void dumpGSUBmultiplesubs(FILE *gsub,SplineFont *sf,struct lookup_subtabl
 	putshort(gsub,offset);
 	if (maps[cnt] == NULL) {
 		fprintf( stderr, "maps[%d] is null; glyphs[%d] is \"%s\"; lookup name is \"%s\".\n" , cnt , cnt , (glyphs[cnt]->name ? glyphs[cnt]->name : ""), sub->subtable_name) ;
-	}
-	for ( gc=0; maps[cnt][gc]!=NULL; ++gc );
+            gc = 0;
+	} else {
+    	    for ( gc=0; maps[cnt][gc]!=NULL; ++gc );
+        }
```
frank-trampe: It's not a big deal, but the indentation is off.

JoesCat:
Hi @davidhedley - in future, you will want to create a "branch" where you don't mess-up your master. The added benefit is you can also work on other ideas on other branches while waiting for a PR (like this one) to get accepted or not.

Working from your Master, tends to bring you to a grinding-halt waiting for a PR to get approved, or worse yet, denied (meaning you'd have to start a fresh new git clone to get past your existing patch that got denied).

PR #2940, opened 2016-11-04
Simpler unicode

gioele:
Currently `makeutype` regenerates both the Unicode data arrays and the support code. This is more cumbersome that it should be, leading to problems like that in <https://github.com/fontforge/fontforge/commit/253709778b214089961049d60b4ab662dab596e7#commitcomment-19673406>.

In this PR I removed all the magic constants calculated by `makeutype` and replaced them with compile-time calculated expressions. No run-time impact and a complete separation between generated Unicode data and code that uses that data.

Now `makeutype` does not generate `utype.h` and `is_Ligature.c` anymore. It still generates `utype.c` and the other files plus a new `is_Ligature_data.h` header file.


gioele:
Hi @JoesCat, thank you for the heads up, I did not know you were extending this part of the codebase. I had a cursory look at your WIP code in the commit https://github.com/JoesCat/fontforge/commit/5ea90d0c2b9d3ab2d89c00725bfb961dc8cef32c.

In general, I think it will be easy to move these functions from the current style (going through `makeutype`) to the new style of this PR. More or less the only change that is needed is using the `ELEMENTS_IN_ARRAY` macro instead of hard-coded magic numbers like `19 `. For example, in `int Ligature_get_altC(int n)`,

```patch
-    if ( n<0 || n>=19 ) 
+    size_t ligature_max = ELEMENTS_IN_ARRAY(ligature16) + ELEMENTS_IN_ARRAY(ligature32);
+    if (n < 0 || n >= ligature_max)
```

(We could also have another macro that automatically sums the elements in `array16` and `array32` and makes this kind of code more concise, but that should be a followup commit to this PR.)

You will also save time by not having to run `makeutype` after every change. ;)

Talking about the new tables in your commit with manually extracted data, I think that we could have a `is_Ligature_data_manual.h` header where all the manually extracted constants reside and write in the maintainer notes that

1. for the moment, after every Unicode release one has to manually modify that data and
2. in the future the extraction of those pieces of data will be moved to `makeutype` and the `_manual.h` no longer used.

WRT the "Add include guard to combiners.h" commit, I have moved it to a separate PR #2942. If it is fine, please merge it or greenlight it so I can merge it myself (I find it bad taste to merge my own PR).

JoesCat:
There are some commits I like above, but there are others that will have issues with this preliminary branch "is_Ligature.c expand ligatures/vulgar/fractions ". I'll halt at the moment so that we compare code.

I would recommend bringing-out "Add include guard to combiners.h" as it is another topic unrelated to the subject line, and can therefore be added as a separate issue.

JoesCat:
The problem I had is that the unicode tables do change in size as new revisions appear, but your use of sizeof() makes this go nicer in separating the code and the data.

FYI - My preliminary work still needed fontforge/python.c and doc/html/python.html additions, but it should be easy to figure that out and add the associated code.

Some ideas are worth abandoning some branches with better ideas.

Due to the proximity of the work - If you want to work on the expanded part, please go ahead (there's other stuff I can look at).

gioele:
@JoesCat, what about this plan: we merge this PR, then in a separate branch I adapt your current WIP code in the branch [prelim_ligature_expansion](https://github.com/JoesCat/fontforge/compare/prelim_ligature_expansion) over to the new style and then I give it back to you for the finishing touches and the final commit?

gioele:
@JoesCat: if you are OK with this branch, could you please merge it? In this way I can start porting your  `prelim_ligature_expansion` branch over to it. I'd like to wrap up some pull requests before embarking in more code porting. :)

JoesCat:
Hi @gioele,
When reviewing code via github, I did not notice at first, but you need to change your TAB spacing from your current +4 to the common/standard +8.
It's a common number that has existed for a long time going back probably further back than CPM.
Your current +4 TAB setting is messing-up with existing alignment.
We shouldn't need to go back after a fix to fix a fix.
https://raw.githubusercontent.com/gioele/fontforge/6cb2f23b6dd26f59fbd76882eff97e1e17307ab8/Unicode/makeutype.c

JoesCat:
If you want to seperate the renaming of __constants to another pull request, we could resolve the smaller PR first...., this should help reduce the ligature PR size ....but do take note of TAB spacing.

gioele:
@JoesCat, I have updated this PR to make all the constants and the comments properly align regardless of the settings used for indentation.

Moving the renaming to another PR is impossible because it is part of the very first commit in the series. Talking about other patch series, I have other small refactoring and cleanups for this directory but they are more focused and in other independent feature branches in my own repo (`remove-usprintf` and `do-not-redefine-isalpha-and-co`).

JoesCat:
Thanks for the effort @gioele - this is a big PR, and these tend to get heavily commented due to size alone. I got hung-up looking at combiners.h and see that is looking very much like in need to a similar fix like with is_Ligature.c/h (the data appears based on 3.0.0 and there is a large markup going into 9.0, but combiners.h also appears to deviate from unicode.org too, so I suspect there is going to need to be overrides and maybe some bugs to send upstream). I'll try look at this PR in a few hours....

gioele:
Not really, #2962 and #2964 cover only part of this PR. In addition to these two already merged PRs, one also needs

* PR #2970 by @JoesCat (currently non mergeable),
* the missing commits from my branches
  + [`simpler-unicode-ligatures`](https://github.com/fontforge/fontforge/compare/master...gioele:simpler-unicode-ligature) and
  + [`simpler-unicode-utype`](https://github.com/fontforge/fontforge/compare/0ac6ccc80db76ed5ce9a473356c9866c56b4e780...gioele:simpler-unicode-utype).

@JoesCat: how do you want to proceed? I think the best way forward is for me to close this PR and let you cherry-pick the missing commits once PR #2970 will be integrated. How does that sound?

JoesCat:
On January 2, 2017 01:37:47 PM Gioele wrote:
> @JoesCat: how do you want to proceed? I think the best way forward is
> for me to close this PR and let you cherry-pick the missing commits
> once PR #2970 will be integrated. How does that sound?

Sounds okay.
I'm still chewing through stuff I need to get done here that is completely 
unrelated to Fontforge. I've already downloaded a copy of your FontForge 
branch for this particular issue, so even if you close the issue and delete 
the entire branch so that you can start clean (for yourself) - that's okay 
with me.

I've still got 5 particular hot things to get out of the way first before I 
even look at FontForge - sorry about being missing in action for now.

JoesCat:
Closed by #3013.

Issue #2711, opened 2016-04-27
Please do not put timestamps in generated files


jre-wine:
cc @frank-trampe 
### When you open an issue for a change/improvement/feature request:
- [x] A description of the problem you're trying to solve, including _why_ you think this is a problem
- [ ] If the feature changes current behavior, reasons why your solution is better
- [x] (optional) Possible solution/fix/workaround

I'd like to propose to not put any timestamps in the generated .ttf (or any other) files. If this is not wanted generally, I'd like to propose an option which allows this.
### Reason:

The [Wine project](https://www.winehq.org/) uses a [forked fontforge](http://source.winehq.org/git/fontforge.git/) (this issue is about 1 of their 4 changes). The generated files are stored in a git repository. Recreating the files with an unpatched fontforge creates spurious diffs after every rebuild.
Generally this is related to reproducible builds (see below).
### Related things:

This has already been partly requested in issue #2389 (Generate: Don't include FFTM) by @davelab6, but afaik timestamps also land in the tables "head" (and "bhed"). I understand that while FFTM is fontforge specific, the latter two were [specified by Apple](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6head.html). I assume/hope this doesn't mean that timestamps are strictly required in order to not break 3rd party consumers.
But even in this case I can confirm that Wine works happily without timestamps.

This is also related to the **[reproducible builds effort](https://reproducible-builds.org/)**:
-  #2490 (propagate creation and modification times to TTF files, [Debian bug #774148](https://bugs.debian.org/774148)) is already fixed (95a470e941d9a20fbdaca51334e8b6b9d93cfae4).
- But there's still [Debian bug #774274](https://bugs.debian.org/774274) (fontforge: please use SOURCE_DATE_EPOCH for reproducible font modification time; was: please allow scripts to set the font modification time). This request would be obsolete if you dropped timestamps altogether.
  On the other side, [implementing SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/) would also solve the problem for Winehq maintainers if a sensible workflow is found. It's just not as trivial as for a Debian maintainer. If it helps I may file a separate issue about that. I have no patch to offer, but there are some instructions in the link.
### For reference:

We discussed the Winehq fork:
https://www.winehq.org/pipermail/wine-devel/2016-April/112893.html

The original Winehq hacks (based on 20081224):
[...]

PR #2943, opened 2016-11-06
Use SOURCE_DATE_EPOCH in version timestamps

gioele:
This PR pushes upstream a patch carried by Debian since long time. It is safe to apply, does not interfere with the normal packaging workflow and make Fontforge a good citizen. :)

This is unrelated to issues #2490 and #2711: those issues are about timestamps in generated font files, this is PR has to do with the packaging of Fontforge itself.

Commit message:

> To allow package to build reproducibly use SOURCE_DATE_EPOCH as
> mentioned in <https://reproducible-builds.org/specs/source-date-epoch/>.


gioele:
The current `date` invocation is not compatible with the OSX `date` command. I will revise this PR later to make it work on both GNU and OSX systems.

gioele:
I updated the PR with the necessary changes for Mac OS X. Now Travis is happy again.

frank-trampe:
Okay.

PR #2944, opened 2016-11-06
Streamline Makefile.am

gioele:
(Simple PR extracted from the always ongoing work on fixing the mess that is the current header situation and the library/application separation border, see PR #2878.)

The commits in this PR put some order in the various `*_HEADERS` and `*_SOURCES` automake variables. This cleanup will make the next commits much more concise and understandable.

### DIFF
--- a/fontforge/Makefile.am
+++ b/fontforge/Makefile.am
@@ -38,29 +38,71 @@ noinst_HEADERS = \
 	autohint.h \
 	autosave.h \
 	autotrace.h \
-	autowidth.h \
 	autowidth2.h \
+	autowidth.h \
 	bitmapchar.h \
 	bitmapcontrol.h \
 	bvedit.h \
+	charview_private.h \
 	clipnoui.h \
+	configure-fontforge.h \
 	crctab.h \
 	cvexport.h \
 	cvimages.h \
+	cvundoes.h \
 	dumpbdf.h \
 	dumppfa.h \
-	fontforgeui.h sftextfieldP.h configure-fontforge.h views.h 	\
-	flaglist.h strlist.h charview_private.h cvundoes.h tables.h
+	flaglist.h \
+	fontforgeui.h \
+	sftextfieldP.h \
+	strlist.h \
+	tables.h \
+	views.h
 
-pkginclude_HEADERS = autowidth2.h configure-fontforge.h fontforge.h		\
-	libffstamp.h psfont.h stemdb.h autowidth.h delta.h fontforgevw.h	\
-	lookups.h savefont.h ttf.h baseviews.h fvmetrics.h mm.h			\
-	scriptfuncs.h ttfinstrs.h edgelist2.h namehash.h scripting.h		\
-	uiinterface.h bezctx_ff.h edgelist.h groups.h nonlineartrans.h		\
-	sd.h unicoderange.h bitmapcontrol.h encoding.h ofl.h search.h		\
-	usermenu.h fffreetype.h PfEd.h sfd1.h ffpython.h			\
-	plugins.h sflayoutP.h print.h splinefont.h cvruler.h ffglib.h           \
-	glif_name_hash.h
+pkginclude_HEADERS = \
+	autowidth2.h \
+	autowidth.h \
+	baseviews.h \
+	bezctx_ff.h \
+	bitmapcontrol.h \
+	configure-fontforge.h \
+	cvruler.h \
+	delta.h \
+	edgelist2.h \
+	edgelist.h \
+	encoding.h \
+	fffreetype.h \
+	ffglib.h \
+	ffpython.h \
+	fontforge.h \
+	fontforgevw.h \
+	fvmetrics.h \
+	glif_name_hash.h \
+	groups.h \
+	libffstamp.h \
+	lookups.h \
+	mm.h \
+	namehash.h \
+	nonlineartrans.h \
+	ofl.h \
+	PfEd.h \
+	plugins.h \
+	print.h \
+	psfont.h \
+	savefont.h \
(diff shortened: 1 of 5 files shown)


jtanx:
I'm not a fan of the one file per line expansion here. Given the large number of files that needs to be included, having it like so just makes the makefile unnecessarily verbose and harder to see at a glance all the files that are part of the target.

gioele:
Hi @jtanx! Please note that

1. every time one entry is changed the diff is much harder to read and the change is lost in the noise;
2. because of point 1, the git log is harder to read and follow;
3. most of these files are in the wrong entry anyway (some `noinst_` should be installed and the other way around);
4. there has been at least one case of a file that has been duplicated and placed in more than one automake variable without anybody noticing;
5. the fact that the entries were not ordered made it so that people added new files randomly at the end or in the middle;
6. the Makefile cannot be too verbose, it is still the same number of included files. ;)

I think the excessively long list of files shows that the current code is a mess from the point of view of the organization. This PR identifies and makes clear that we have a problem. Let's accept this fact and work on restructuring the codebase to solve this problem rather than sweeping this unpleasant fact under the rug.

And I must admit that it is a pleasure to be able to remove these files one at the time (as I do, for example, at the end of the branch [remove-configure-fontforge-h](https://github.com/gioele/fontforge/tree/remove-configure-fontforge-h)). :tada:

jtanx:
1. Github often highlights which part of the line changed. It's also not exactly all that often that a file gets added/removed so I don't follow the idea that it's that hard to make out what happened from the diff
2. See before
3. Okay that should be fixed
4. Ditto
5. Having files separately line by line isn't exactly any more likely to make further additions be alphabetical. 
6. I'd argue that it does - in your change, you literally have to scroll the page to see which files are included, vs. looking at 2-3 lines.

Also I don't think that the number of files included has much bearing on anything being messy, merely a reflection on the fact that FontForge is made up of many parts, and at times, it is complex.

Yes there are problems with the makefile, but no I don't think that increasing the verbosity by placing each source file on a single line makes the situation any better.

gioele:
@jtanx, I usually think more in terms of git rather than github. Most pieces of the git infrastructure, for example the merging strategies or `git log -G`, all work in terms of lines.

WRT order, as the title says, I not only put them all on a single line, I have also reordered them. Making sure that new files end up in the proper places is now very simple, also at review time.

About the verbosity. If the list of files is tool long (and it is), that means that we should reorganize the files, not the list. :)

For example, all the CharView files and functions should be moved away from the `fontforge/` directory into something like `fontforgeexe/charview`. But doing it is not an easy task and requires a lot of small and careful changes, together with some scripts to isolate the existing parts and their dependencies. I am doing all this in PR #2878 and other followup branches. Having streamlined Makefile.am files would allow me to do that job even faster.

jtanx:
@gioele, I still don't think this is a good idea.

With regards to git log, as mentioned before, I just don't see how it would be all that difficult to pick up on what's changed, especially as the source files included doesn't change all that often. 

It's nice that you reordered them, but the same could be done without listing each file on one line, which breaks the conciseness and ease of reading of the makefile.

Unless we're talking hundreds of files, no I don't think that the long list of files is any indicator that re-organisation is necessary, especially in this case and for the size of a project that FontForge is. 

Files should not be moved for the sake of making a makefile concise - in particular, I also don't support the idea of moving charview functions into another folder. Folders should be used at a module level, and not at arbitrary sub grouping - *this* would definitely make the makefile more complex than it needs to be, especially in its current recursive design.

JoesCat:
Like @jtanx, I too prefer to see code without having to scroll up-n-down too much, but fortunately, or not, I don't tend to visit Makefiles too often.
I have a preference for compressed format, but I'm okay with these changes if we don't need to come visiting Makefiles again. I believe @mskala would like to have files move to a standard-ish format, therefore could use his input.

What I would like to indicate, is check that 'make distro' and 'make distrocheck' keep all files that should be kept. I was doing some of that in 20120731.... 20140101 but didn't check everything, so there may be some stragglers that still need to be added/removed/cleaned-up. ...some of this can be done by comparing the results of make distro to your HEAD/MASTER and looking for differences.

mskala:
> I believe @mskala would like to have files move to a standard-ish format, therefore could use his input.

I agree having the files in a standard-ish format would be good.  I agree having filenames in alphabetical order is good.  I don't have strong feelings either way on one per line or more than one per line.  In the past I've used alphabetical order, hand-wrapped to fit in 80-character lines (usually more than one filename per line), but it's true that this creates a bit of a version control mess when filenames are added or deleted and the lines have to be re-wrapped, worse if people other than me touch the code and don't respect the existing convention; after reading this thread I'm considering re-doing some of my files to be one filename per line to make version control work better.

JoesCat:
fontforge\Makefile.am is already partially converted, some of the other Makefile.am are a little stale and have the '\' out of alignment.

@gioele, can you look at Unicode/Makefile (as an example) and tab the '\' ?

gioele:
Hi @JoesCat, `fontforge/Makefile.am` got some additions in this style via PR #2878 that got merged some days ago. I was planning on making some noise on that PR after this got merged or rejected, so we could have a single style, but @frank-trampe was faster. :)

WRT the tab, I fear they will not help: tabs _after_ variable-length strings are not going to align anyway, because everybody uses a different tab setting. In fact, the `\`  do not align for me at the moment.

frank-trampe:
I'm personally indifferent on the line-breaking question, but @jtanx and @JoesCat deal with the build system a lot, so let's keep the current formatting as they prefer.

gioele:
I'm closing this PR. Too little gain and too many changes to the various `Makefile.am` in the meantime.

PR #2945, opened 2016-11-06
Remove configure-fontforge.h

gioele:
This PR carefully removes all the pieces of the unused file `configure-fontforge.h`.

Even though all its content is not currently used, I think it is not right to simply delete it. I have thus moved all the unused macros to the files where they are at least referenced. This preserves all the comments and makes it possible to follow the history of these macros via tools like `git log --follow`.

Now, if we think those macros are useless and we want to remove support for them, we can do it directly in the respective files.

(This PR is a revised version that does not depend on PR #2944.)


frank-trampe:
I don't see why this is necessary. It just scatters unused things that were once centralized, no?

gioele:
(Hi @frank-trampe, thank you for merging the first chunk of consolidated headers! I updated this PR to fix the merge conflict in the Makefile.am.)

`configure-fontforge.h` was where options were set before `automake.ac` and the rest of the autotools have been introduced. People slowly removed defines from here until only unused macro/options remained and nobody took the last step of just removing the file.

I moved the macros to the places where there are referenced, but I think we could just remove them and the associated `#ifdef`'d chunks of code. Probably nobody has used them in ages. If there is consensus, we will do that in future commits, so that we will be able to maintain a clean history of what went where and when.

For the macros `RECENT_MAX` & `FORMER_MAX`, they do not belong to fontforge-the-library in the first place. Last, `BACK_LAYER_MAX` is not really an adjustable setting, just a constant that has to do with spline fonts and ended up being set placed near other constants.

gioele:
If there are no strong objections to this removal, I'd merge it to clean up the PR situation and prepare for part 2 of the PR on refactoring.

jtanx:
Hmm, I tend to agree with @frank-trampe; I see little value in this. I prefer having configuration `#define`s all in one header rather than spread out across the files - that's the point of a config header. I also see you've had to define `RECENT_MAX` twice, which is less than ideal.

If you really want to remove this header, then perhaps try migrating them all into the autoconf generated `config.h`/`fontforge-config.h` header.

gioele:
Adding even more options to `configure.ac` seems to go against my aim of removing cruft from fontforge. :wink:

About @jtanx's comment on `RECENT_MAX`: well, the whole `fontforge/noprefs.c` is an almost perfect duplicate of `fontforge/fontforgeui.h` (see the surrounding code). It makes sense to duplicate also that line.

Actually my plan is to remove all these unused constants and the related code. None of these constants have been enabled/compiled for ages a none of the Linux distros carry patches to enable them. For example, `FONTFORGE_CONFIG_CVT_OLD_MAC_FEATURES` has never been enabled since 2003.

Should I go on and add commits to remove all these unused constants and the related dead code?

gioele:
Hi all, I have updated this PR. Now all the code that has been disabled for more than 10 years has been deleted, the associated `#define` guards have been removed as well. One constant has been turned into an autoconf configure option and the few remaining constants have been moved to the relevant headers (and one duplicated in `noprefs.c`, but that is the original intent of that file).

frank-trampe:
@gioele, that doesn't make it better. There is a significant amount of disabled code in FontForge, some of which has never been enabled, but much of it exists/persists for a reason. A lot of disabled code was written to make maximum use of existing knowledge and ideas that might help future efforts. We don't have the resources to finish even a small fraction of the things that we want to do, but we can't start abandoning them now.

gioele:
I understand that some code has historical value; I doubt any any of the code I am deleting here has any. This is code that the original author has left in then forgot about. Just have a look at the commit messages I reference in my commits or the comments left by the author.

Anyway, these snippets of ifdef'd-out code are not going to disappear: they will live forever in the git repository. Actually, they would even be hightlight now, thanks to all the information contained in the deletion commits.

If you want, we could create a `historical-code.md` file where we make a list of dead-ends that have been found in the code with a small explanation and a link to the relevant commits. That would be much more useful than having theses random pieces of abandoned code interspersed with the current code.

gioele:
Hi, I have rebased this PR on top of the smaller PR #2977 (the one with uncontroversial removals).

ctrlcctrlv:
At this point...this no longer can be merged without a conflict, it's years old, and many seem to not want it. So I'm closing this as it is unlikely to ever be merged. It can be reopened if interest in it is ever renewed.

gioele:
hi @ctrlcctrlv: deleting this is PR was definitely the right choice. :)

Maybe you could add a link to this PR at the beginning of `configure-fontforge.h` to remind future developers of this cleanup effort and the difficulties surrounding it?

jtanx:
Funny story, I'm actually getting rid of this header with my cmake changes. However it's more 'fold it into the current config header' than remove all the code.

PR #2946, opened 2016-11-06
Remove debian-old directory

gioele:
The packaging files for Debian are stored in `Packaging/debian` since
commit 58a4dfbd6ea4d99c8be8eea6857e05f0eb50dcfa.

The Debian maintainers confirmed that that directory is not used by them,
see the thread "debian-old directory in fontforge: somehow useful?". [1]

[1] http://lists.alioth.debian.org/pipermail/pkg-fonts-devel/2016-November/018587.html


JoesCat:
I'm plus one for removing it as we should ideally have the ./configure stuff correctly setup for "everyone", however, since I'm not the one building compiled releases, this is best answered by @frank-trampe and @jtanx.
This pull request may need some things to happen with configure.ac and the m4 directory before safely removing it. I would recommend checking Debian's fixes, Fedora's, and the BSD distros for internal fixes that can probably be pulled in, ...homebrew and macports may also have some fixes too.

gioele:
Hi @JoesCat, I also thought about all those things, but it turns out that that directory is just an old snapshot of the Debian packaging stuff.

Years ago (commit 97843f0cf5453ab93183f894ab0c17462dad3f1b) @frank-trampe disconnected that directory from the whole build system and has been left there to take dust. As I wrote in the commit message, the Debian maintainers reassured me in http://lists.alioth.debian.org/pipermail/pkg-fonts-devel/2016-November/018587.html that...

> Ah. Feel free to delete in upstream :-). Debian teams create their own
> packaging stuffs. We ignore packaging from upstream anyway.

WRT integrating patches, there is only one non-Debian specific patch that I have found in the Debian repository (and not in `debian-old`), the one in PR #2943 (unfortunately broken on OS X, will fix it soon when I get access to my OS X machine).

frank-trampe:
As far as I remember, I never used the contents of the directory, and I'm not sure whether they were ever used upstream.

JoesCat:
Would it be worth moving fontforge.xpm into the desktop/icon-older directory?

JoesCat:
Hi @gioele - i think i probably broke this merge by moving a file out (for old nostalgic reasons).
Would you mind regenerating this so it will merge without problems?
Sorry about creating extra work.
I don't think there appears to be any objections, so, once it appears okay, just go ahead and merge it.

Thanks,
Joe

Issue #2950, opened 2016-11-15
Build failure with CZMQ 4.0.1: error: unknown type name 'zctx_t'


ilovezfs:
```
make[4]: Nothing to be done for `all-am'.
Making all in collab
[...]

PR #2956, opened 2016-11-23
Do not accept libczmq version 4

gioele:
The deprecated `zctx` API has been removed in version 4 of libczmq.

As reported in #2950, this causes compilation problems on recent versions of MacOS and homebrew.

Until the collabclient code is updated to the new `zsock` API, we cannot compile against libczmq version 4 and later, so let's limit ourselves to version 3.

This build will fail on Travis because of an unrelated MacOS issue: `The command "if [ "$TRAVIS_OS_NAME" == "osx" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then travis-scripts/ffosxbuild.sh $PREFIX $TRAVIS_COMMIT; fi" exited with 1.` @jtanx, @frank-trampe, could you have a look at it?

### DIFF
--- a/m4/fontforge_arg_with.m4
+++ b/m4/fontforge_arg_with.m4
@@ -401,7 +401,7 @@ AC_DEFUN([FONTFORGE_ARG_WITH_ZEROMQ],
 [
 FONTFORGE_ARG_WITH([libzmq],
         [AS_HELP_STRING([--without-libzmq],[build without libzmq])],
-        [ libczmq >= 2.2.0 libzmq >= 4.0.4 ],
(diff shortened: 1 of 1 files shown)


gioele:
Well, it seems it passes all Travis checks. Somehow the error fixed itself (or did someone fix up on Travis?)

Anyways, I propose we merge this as a "temporary solution" to the problem of collabclient being outdated. In practical terms it means still having the collab feature on Linux, Mac and Windows, but not on Mac + Homebrew.

jtanx:
Seems reasonable to me.

PR #2963, opened 2016-11-28
Update combiners.h from 3.0.0 to 9.0

JoesCat:
Manual update, concentrated mostly on anything with "COMBIN" in the
range of {0...0xFFFF}. This is better done searching line-by-line
using a program as it is easy to miss codes without "COMBIN" in it.


JoesCat:
Needs attention later.
See reference combiners table here:
http://www.unicode.org/reports/tr44/#Canonical_Combining_Class_Values

PR #2964, opened 2016-11-28
Simpler code with extra fixes

JoesCat:
This is a partial set of patches based on #2940 with additional edits based on review of code.

Hi @gioele, are okay with these edits (up to this point)?
I have not added the other patches yet, but am looking at further edits.


gioele:
Hi @JoesCat, these changes look good, but what are the additional edits? The only differences I can see is the different way to produce newlines in 16 and 32-bit arrays and the use of parenthesis around `return` statements (that I see as noise, but not a big deal). Are there other issues I did not notice?

Just wondering, what is the aim of this separate PR? Is there something wrong with PR #2964?

I have a branch with a trimmed down version of it (with only the changes to `is_Ligature` and not `utype.h`). Here you can find the relevant commits <https://github.com/fontforge/fontforge/compare/master...gioele:simpler-unicode-ligature>.

If you want you can just use that instead of these commits. Otherwise I can go on and merge this PR.

frank-trampe:
I like these.

JoesCat:
Hi @gioele, I just wanted to check with you if you were okay with these edits before merging. Thanks.

Yes, there were other changes not noticed too (between #2940 and the recent PR, plus this current PR) that goes beyond the return() noise.

When I made PR #2762, it had 60+ comments and edits before it got accepted into mainline code, when I asked 3 questions/requests concerning PR #2940, nothing was changed and I was told it was impossible to change. A few days ago, I thought there actually were edits made, but it only appeared to be a rebase, so I concluded it easier to simply go and fix what looked like needed fixing.

I'll comment on your new code edits and point-out things that need attention, meanwhile, this looks ok to merge.  Thanks @frank-trampe for the +1

PR #2969, opened 2016-12-07
Consolidate fontforge/ headers, part2

gioele:
The second bunch of standalone cleanups for the headers in `fontforge/`, a followup to PR #2878. I made this PR smaller than previously stated in the hope it will be easier to review.

Similary to the previous PR, these commits all the `extern` prototypes of non-static functions of various `.c` files to their respective `.h` files. In this second PR, the headers of the files from `fontforge/effects.c` to `fontforge/nonlineartrans.c` (in alphabetical order) have been consolidated. `splinefont.h` and `baseview.h` getting smaller and smaller. The most important thing is that many implicit dependencies are now explicit, hinting at where the future code clean-ups and refactorings should start from.

No implementation code has been touched and this PR. Comments, where available, have been preserved and moved near to the function declarations.

### DIFF
--- a/fontforge/lookups.c
+++ b/fontforge/lookups.c
@@ -26,6 +26,7 @@
  * ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 #include "fontforgevw.h"
+#include "fvfonts.h"
 #include <chardata.h>
 #include <utype.h>
 #include <ustring.h>
--- a/fontforge/lookups.c
+++ b/fontforge/lookups.c
@@ -25,6 +25,9 @@
  * OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
  * ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
+
+#include "lookups.h"
+
 #include "fontforgevw.h"
 #include "fvfonts.h"
 #include <chardata.h>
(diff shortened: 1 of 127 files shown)


frank-trampe:
This is certainly something worth doing. I will look at the specific changes soon.

fontforge/lookups.c
```diff
@@ -25,7 +25,12 @@
  * OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
  * ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
+
+#include "lookups.h"
```
frank-trampe: Why did this need to move?

fontforge/nonlineartrans.c
```diff
@@ -60,7 +62,7 @@ return;
     chunkfree(e,sizeof(*e));
 }
 
-static int gettoken(struct context *c, real *val) {
+static int gettoken(struct expr_context *c, real *val) {
```
frank-trampe: Was there a name conflict?

frank-trampe:
There are a lot of stray line breaks in the include blocks.

fontforge/lookups.c
gioele: `lookups.h` contains the prototype declarations for the functions implemented in `lookups.c`. It should be the very first file to be included so that we are sure that it is self-contained, i.e. includes all the headers that contains the datatypes used in those prototype declarations. This allows us to make sure that the compilation of any files does depend on the order in which other headers are included.

gioele:
Thank you for the review @frank-trampe. The empty line between the include blocks are there on purpose: they separate the `#include` for  the prototypes of the implemented functions are declared from the `#include`s used to bring in the prototypes of functions implemented elsewhere.

PR #2970, opened 2016-12-08
Simpler code

JoesCat:
Simplify code by generating only the data.
This code is further refined from the earlier PR and improved refinements that @gioele provided.


JoesCat:
Can you see if anything looks like needing further fixes and improvements?

Unicode/is_Ligature.c
```diff
@@ -1,104 +1,21 @@
 /*
 Copyright: 2012 Barry Schwartz
+    Create is_Ligature.c (using a python script) to test for Vulgar fractions
 Copyright: 2016 Joe Da Silva
-License: BSD-3-clause
+    Re-write is_Ligature.c to test for ligatures, Vulgar and Other-fractions.
+Copyright: 2016 Gioele Barabucci
+    Simplify makeutype.c code. Split is_Ligature.c, create is_Ligature_data.h
+License: GPL-3+
```
frank-trampe: Why are we changing the license?

Unicode/is_Ligature.c
```diff
@@ -1,104 +1,21 @@
...
 Contributions:
 */
 
-/* This file was generated using the program 'makeutype' for Unicode_version 9.0 */
+#ifndef FONTFORGE_UNICODE_IS_LIGATURE_C
```
frank-trampe: Why do we have an include guard in a C file?

Unicode/is_Ligature.c
mskala: I thought it was project policy to use GPL3+ on new code, and only preserve existing licenses on old code.  So if this really _is_ a re-write and not just a modification of what was there, GPL3+ would be appropriate.

frank-trampe: Thanks! I do appreciate that.

As to the more general issue, I wasn't equating copyright and licensing. My point was that since we don't have a means to consolidate copyright ownership/control (which would allow one party to issue the code under new licenses in the future), the only way for us to retain licensing and relicensing flexibility is to use permissive/flexible licenses.

frank-trampe:
We also have a merge conflict, which I think is keeping Travis from running.

Unicode/is_Ligature.c
gioele: just chining in. I did not make any change to the header/licensing block in my original commits, but I am fine with any license you end up choosing for this code.

gioele:
@JoesCat, could you please rebase this PR against the current master? This branch has been created against your old PR that has then been merged and superseded by other commits.

Unicode/Makefile.am
```diff
@@ -67,6 +67,15 @@ libgunicode_la_SOURCES =	\
 	gwwiconv.c		\
 	$(NULL)
 
+libgunicode_la.$(OBJEXT):	\
```
gioele: Hi @JoesCat, $(OBJEXT) is unnecessary at best and possibly harmful for future maintainability. I wrote why as a comment to a previous commit, but maybe it got lost. Here my original comment (from <https://github.com/fontforge/fontforge/commit/fe13eca58f262fd4e0d4c84070bddd086ee0399e#commitcomment-19995303>)

> hi @JoesCat, I think I am not getting what you are saying.
>
> Just having a `.h` file in `_SOURCES` will never lead to the creation of a separate object file. (And even if it did it would be completely empty, because (at the end of the PR) all the variables in it are declared static.)
>
> You can see that the only rule that depends on `is_Ligature_data.h` or that makes use of it is in `.deps/libgunicode_la-is_Ligature.Plo`:
>
> ```Makefile
> libgunicode_la-is_Ligature.lo: is_Ligature.c is_Ligature_data.h \
>  ../inc/basics.h ../inc/fontforge-config.h ../lib/stdio.h \
>  /usr/include/stdio.h /usr/include/features.h \
> ```
>
> but that is just a declaration of dependency. Only one `.lo` file will be produced. And it will be produced using this ad-hoc rule generated by the autotools (split in multiple lines for readability):
>
> ```Makefile
> libgunicode_la-is_Ligature.lo: is_Ligature.c
> 	$(AM_V_CC)$(LIBTOOL) $(AM_V_lt) --tag=CC $(AM_LIBTOOLFLAGS) $(LIBTOOLFLAGS) --mode=compile \
> 	$(CC) $(DEFS) $(DEFAULT_INCLUDES) $(INCLUDES) $(libgunicode_la_CPPFLAGS) $(CPPFLAGS) \
> 	$(libgunicode_la_CFLAGS) $(CFLAGS) -MT libgunicode_la-is_Ligature.lo -MD -MP -MF \
> 	$(DEPDIR)/libgunicode_la-is_Ligature.Tpo -c -o libgunicode_la-is_Ligature.lo \
> 	`test -f 'is_Ligature.c' || echo '$(srcdir)/'`is_Ligature.c
> ```
>
> After running `make` you can see that in `.libs/` only one `.o` file for `is_Ligature` has been produced and no traces of `is_Ligature_data.h`. Autotools and libtool may be convoluted and do weird things, but not that weird. ;)

In addition to that comment, I should also point out that all the `../inc` files (and many others) are automatically added as dependencies by Automake. You can see this inspecting the deps file `Unicode/.deps/libgunicode_la-is_Ligature.Plo`.

Please just add `is_Ligature_data.h` to `_SOURCES`. Let's make the autotools do some work for us. ;)

Unicode/is_Ligature.c
JoesCat: @gioele, it's been a long time since I've looked at it, but RedHat has/had a contribution guide, and one of the things that stuck in my mind was, smallish patches of about 10ish lines of code was considered a contribution, and large edits needed a copyright. Although there are many patches in this list you originally created, the sum total has some structural changes of significance in relation to the size of the file and what they do.
There are 2 somewhat recent Debian issues (prior to pulling in this new post-20120731-FontForge) which is also asking about clarifying copyrights (example Unicode/charset/* ), so you'll likely see more of these headers added after examining file histories.
I don't think renaming variables, editing space/formatting, or shuffling functions count, and there is a lot of gray area in this too.

Unicode/Makefile.am
JoesCat: kind of rushed it out, but all the feedback is good.

Unicode/is_Ligature.c
frank-trampe: @gioele?

frank-trampe:
@JoesCat, particularly when working on a multi-person branch, it seems wise to keep it in the FontForge repository so that we can all work on it. Could you perhaps push this branch there and start a new pull request so that @gioele and I can help?

Unicode/is_Ligature.c
gioele: hi @frank-trampe, I do not know why that header guard is in that C file; it is not in any of my commits. I suppose it somehow got there while @JoesCat was mixing my patches with his work on Unicode 9. Weird things happen all the time with git. ;)

JoesCat:
Resolved by #3013

PR #2974, opened 2016-12-15
Fix OS/2.fsSelection field calculation

catharanthus:
Currently, when exporting a TTF font, FontForge will analyze the font name, and if it contains the "Obli" substring, it will set the OBLIQUE bit (bit 9) and clear the ITALIC bit (bit 0) of the fsSelection field of the OS/2 table. However, the OpenType specification suggests that both ITALIC and OBLIQUE bits are set for oblique fonts (see https://www.microsoft.com/typography/otspec/os2.htm#fss).

This behavior creates issues on Windows, where if a family contains Oblique fonts, not all fonts from the family will be displayed in the Control Panel.

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -3377,7 +3377,6 @@ static void setos2(struct os2 *os2,struct alltabs *at, SplineFont *sf,
 	os2->fsSel |= 8;
     if ( os2->version>=4 ) {
 	if ( strstrmatch(sf->fontname,"Obli")!=NULL ) {
-	    os2->fsSel &= ~1;		/* Turn off Italic */
 	    os2->fsSel |= 512;		/* Turn on Oblique */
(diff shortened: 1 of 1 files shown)


jtanx:
The relevant part from the documentation appears to be:

> In most cases, if OBLIQUE is set, then ITALIC will also be set, though this is not required.

This change looks ok to me.

JoesCat:
Hi @catharanthus, I'm currently updating AUTHORs, do you want your name added, or leave it like it is?
Thanks for the patches - Joe

catharanthus:
Hi @JoesCat, my name is Alexey Kuznetsov.

JoesCat:
> Hi @JoesCat, my name is Alexey Kuznetsov.

Excellent - and many thanks for your help improving FontForge.

PR #2977, opened 2016-12-18
Reduce configure-fontforge.h

gioele:
While we sort out how what is the best way to go about removing `configure-fontforge.h` in PR #2945, I isolated in this smaller PR a couple of low-hanging fruits: turning the last still used configuration option to a proper autoconf option and moving some constants near other similar constants in the appropriate headers.

A reminder about the duplicated constant in `fontforge/noprefs.c`: that file contains, on purpose, hundreds of duplicated declarations. Life is interesting in fontforge's land. ;)


frank-trampe:
That's fine.

Issue #2987, opened 2017-01-06
Native scripting: DebugCrashFontForge won't run without a font loaded


jtanx:
Running

```
fontforge -lang=ff -c "DebugCrashFontForge()"
```

Will complain that:

```
DebugCrashFontForge: This command requires an active font
Called from...
 <command-string>: line 1
```

This shouldn't be the case (why would you need to load a font just to crash FF?)

That said, having a function just to cause a segfault is pretty neat 😉

JoesCat:
in scripting.c
```c
static struct builtins {
    const char *name;
    void (*func)(Context *);
    unsigned int nofontok: 1;	/* 0=require active font */
    unsigned int argcnt: 4;	/* 0=defined in function */
    unsigned int argtype: 4;	/* 0=defined in function */
```

Change the 0 to a 1 on line:
```c
-    { "DebugCrashFontForge", bDebugCrashFontForge, 0,0,0 },
+    { "DebugCrashFontForge", bDebugCrashFontForge, 1,0,0 },
```
Following git blame, the reason for this function seems to be to test if you can run a crash debugger (in Windows, MAC, Linux, otherOS) in the first place. Can't be complete if you can't crash it on purpose ;-P

JoesCat:
Fixed it, then added more (patch 2990).
Reading another article about COW, think it best to leave this off as default.
http://advisories.mageia.org/MGASA-2016-0347.html

If you think we should add a configure toggle, like --with-debugcrash then it could be done, but recommend default action is to leave it off.

JoesCat:
Seems issue resolved.

PR #2990, opened 2017-01-12
Native scripting: DebugCrashFontForge won't run without a font loaded

JoesCat:
Fixes issue #2987 to allow instruction to work without a font loaded.
Not all CPUs are able to detect a segfault, but several are able to
detect a divide by zero, therefore made divide by zero default crash.

### DIFF
--- a/fontforge/scripting.c
+++ b/fontforge/scripting.c
@@ -8235,11 +8235,54 @@ static void bValidate(Context *c) {
     c->return_val.u.ival = SFValidate(c->curfv->sf, ly_fore, force );
 }
 
-static void bDebugCrashFontForge(Context *UNUSED(c))
-{
+/* #define _DEBUGCRASHFONTFORGE 1 */
+#ifdef _DEBUGCRASHFONTFORGE
+static int bDebugCrashFontForgeS(int s) {
+    return( bDebugCrashFontForgeS(s+1) );	/* crash call stack */
+}
+#endif
+
+static void bDebugCrashFontForge(Context *c) {
+/* This function was introduced to crash FontForge on purpose for debugging. */
+/* To test your debugger, try: fontforge -lang=ff -c "DebugCrashFontForge()" */
+#ifdef _DEBUGCRASHFONTFORGE
+    int a, b, d;
+    char buffer[2], *ptr = NULL;
+
+    a = 0; b = d = a + 1;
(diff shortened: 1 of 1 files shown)


fontforge/scripting.c
```diff
@@ -8235,11 +8235,54 @@ static void bValidate(Context *c) {
...
+	c->return_val.u.ival = b / a;	/* crash with divide by zero */
+    else if ( d==2 )
+	b = bDebugCrashFontForgeS(0);	/* crash the call stack */
+    else if ( d==3 )
+	buffer[(d+d)] = 1;		/* crash on stack buffer overflow */
+    else if ( d==4 )
+	*ptr = 1;			/* crash on *ptr==NULL */
+    else if ( d==5 ) {
+	ptr = (void *)1; --ptr;
+	*ptr = 1;			/* crash on *ptr==0 address */
```
frank-trampe: Do we support any platforms where `0 != NULL`? And what is the advantage of this construction of ptr versus `ptr = (void *)0`?

JoesCat: > Do we support any platforms where 0 != NULL?

Hi @frank-trampe, personally I don't think anyone of us have machines where 0 != NULL
I looked around for some info based on stackflow remarks and ran into this:
http://c-faq.com/null/machexamp.html - the closest I worked with was an HP3000x, but I highly doubt it would have ran fontforge ;-)
On the other hand, some bug fixes added into libspiro and fontforge were due to users that run MIPs machines, and getting code to run on them (so in one way we contributed without having actual machines), and we've also had some contributors with fixes based on trying to compile code for 390 machines (don't see those often but there are patches in the code thanks to those users).

> And what is the advantage of this construction of ptr versus ptr = (void *)0?

A smart compiler could interpret that as NULL during compile time. Selecting 1 (not NULL) and then moving back one to zero makes it a CPU binary execution (hopefully outsmarts a smart compiler). I was reading a few stackexchange and stackoverflow, and the --ptr; method was mentioned - this appeared reasonable.

Realistically, a program shouldn't have ways to purposely crash, but reading Ben's original reason for implementing it seems like a necessary evil for the current time until we whittle-down the number of bugs. Not sure if we should have it turned on by default, or just for a debug version for debugging reasons. What do you think - any thoughts/direction?

PR #2994, opened 2017-01-14
Update de.po

Atalanttore:
Improved German translation.

### Important

### Motivation and Context
- Improve German translation


JoesCat:
Thanks for the improvements @Atalanttore 

The German language file has not been touched in quite a while.
If you want to update further, please read issue #2769 where we introduced a small little helper.
For German, you could use:
make FontForge-Update-po 1=de

If you have trouble with this, we'll update the file for you.
Thanks for the update.

Atalanttore:
Okay.

PR #2975, opened 2016-12-18
Consolidate fontforge/ headers, part 3

gioele:
The third installation of the cleanups of the headers in `fontforge/`, a followup to PR #2969 (part 2) and #2878 (part 1).

As usual, I tried to kept this PR as "uneventful" as possible and I have moved moved the comments (if available) near to the function declarations. There are only two small things to note:

* The introdution of `mem.c` and the associated `mem.h`. I moved there the various `memlong`, `memushort`, etc function that are used by many parsing functions in many files but whose implementation was scattered among various `.c` files. (BTW, I suppose all these functions could be removed in favor of standard endian-correct bitshuffling functions, but that is for a separate PR).

* The duplication of the `#define real` block in `mem.h`. Avoid that duplication amounts to solving the whole float/double problem of FontForge. Definitely outside the scope of this series of cleanup.

PR #3000, opened 2017-01-18
Consolidate fontforge headers, part 4

gioele:
The fourth installation of the cleanups of the headers in `fontforge/`, a followup to PR #2975 (part 3), #2969 (part 2) and #2878 (part 1).

Just like the previous PRs, this PR is quite "uneventful": only declarations have been moved and the comments (where available) preserved. The only things to note are two drive-by commits that mark some functions as static (both functions have never been exported and are only used inside the same file in which they are found).

### DIFF
--- a/fontforge/Makefile.am
+++ b/fontforge/Makefile.am
@@ -79,6 +79,7 @@ noinst_HEADERS = \
 	spiro.h \
 	splinefill.h \
 	splineorder2.h \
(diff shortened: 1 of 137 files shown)


gioele:
The Travis-CI failure is caused by the homebrew/glib problem reported in issue #3001.

frank-trampe:
It looks fine to me.

PR #3011, opened 2017-01-28
fix warning: res>=0

JoesCat:
A lot of warnings in various files. 1 down, many to go....


JoesCat:
other libs still show a slew of warnings when clicking "X" to see error list

PR #3021, opened 2017-02-03
Fix "Unicode 2.0, full repertoire" cmap encoding.

### LINKED ISSUE #2728: cmap encodingRecord generated with broken (platformID, platformSpecificID)
FontForge version 20120731 running on Linux Mint 17.3 Cinnamon 64-bit, kernel 3.13.0-37-generic.

When non-BMP characters are present in a font FontForge generates a cmap (0, 10) encodingRecord. As far as I can tell from the documentation this combination of (platformID, platformSpecificID) does not exist. I believe the correct encodingRecord should have been (0, 4).

I have attached a font (LinLibertine_R.otf) that has what I believe are the correct encodingRecords. If you load this font in FontForge and generate another OpenType font from it you will see that the generated font has different encodingRecords. I have attached such a font (LinLibertine_R_generated.otf) .

Both attached files are in the following archive: [LinLibertine_R.zip](https://github.com/fontforge/fontforge/files/279317/LinLibertine_R.zip)

nigeltao:
Previously, FontForge would emit a cmap table with Platform ID = 0,
Platform Specific Encoding ID = 10. The 0/10 combination is bogus. It
should be 0/4 instead.

Both these specs:
  - https://www.microsoft.com/typography/otspec/name.htm
  - https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6cmap.html
lists 0/4 and 3/10 as valid combinations, but not 0/10. The first number
is the PID, and means Unicode (0) or Microsoft (3). The second number is
the PSID, but its meaning depends on the PID.

Fixes #2728.

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -4939,17 +4939,13 @@ static void dumpcmap(struct alltabs *at, SplineFont *sf,enum fontformat format) 
     if ( hasmac&1 ) {
 	/* big mac table, just a copy of the ms table */
 	putshort(at->cmap,0);	/* mac unicode platform */
-	putshort(at->cmap,3);	/* Unicode 2.0 */
+	putshort(at->cmap,3);	/* Unicode 2.0, BMP only */
 	putlong(at->cmap,mspos);
     }
     if ( format12!=NULL ) {
 	/* full unicode mac table, just a copy of the ms table */
 	putshort(at->cmap,0);	/* mac unicode platform */
-        if( map->enc->is_unicodefull ) {
-	    putshort(at->cmap,10);	/* Unicode 2.0, unicode beyond BMP */
-	} else {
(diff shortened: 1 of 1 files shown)


davelab6:
@frank-trampe is away at the moment, but this is a pretty trivial fix and looks good to me. Could anyone else review it and confirm it looks good to them, so we can merge quickly? :)

jtanx:
What needs to be checked is is FF actually spitting out a UCS-2 encoded table in this case? I believe that the `is_unicodefull` path is there to indicate a UCS-4 encoded path - which should be 3/10 (platform/encoding)

jtanx:
Looking at it a bit further, I do see 3/10 further down. So I guess this patch is ok.

frank-trampe:
Yeah. I'm sort of swamped right now and can't verify the specification-compliance. If you both think it's okay, let's merge it.

davelab6:
Thank you!!

Issue #3001, opened 2017-01-19
Builds on MacOS/Travis-CI fail because of wrong option `-arch i386`


gioele:
Current MacOS builds on Travis-CI, for example [the build for PR #3000](https://travis-ci.org/fontforge/fontforge/builds/193218939), are failing with the following error because of an issue in the way the Homebrew recipe has been set up.

```
In file included from ./splinerefigure.c:34:
In file included from ./splineutil.h:6:
In file included from ./views.h:32:
In file included from ./ffglib.h:27:
In file included from /usr/local/Cellar/glib/2.50.2/include/glib-2.0/glib.h:30:
In file included from /usr/local/Cellar/glib/2.50.2/include/glib-2.0/glib/galloca.h:32:
/usr/local/Cellar/glib/2.50.2/include/glib-2.0/glib/gtypes.h:423:3: error: '_GStaticAssertCompileTimeAssertion_0' declared as an array with a negative size
  G_STATIC_ASSERT(sizeof (unsigned long long) == sizeof (guint64));
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/local/Cellar/glib/2.50.2/include/glib-2.0/glib/gmacros.h:232:103: note: expanded from macro 'G_STATIC_ASSERT'
#define G_STATIC_ASSERT(expr) typedef char G_PASTE (_GStaticAssertCompileTimeAssertion_, __COUNTER__)[(expr) ? 1 : -1] G_GNUC_UNUSED
^~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [splinerefigure.lo] Error 1
```

This problem has been reported in issue homebrew/homebrew-core#5404. The root cause seems to be the `-arch i386` option erroneously added to CFLAGS on x86_64 platforms by some packages (perl? maybe required by glib?).

Linux builds are not affected by this problem.

PR #3023, opened 2017-02-05
Force MacOS builds to target only x86_64

gioele:
Commit 25c51f592df2c69f12ed5fc1d39c7917ef829dcf introduced in 2014 the following compile flags on MacOS

    -arch x86_64 -arch i386

These flags were used to create universal binaries for OS X up to version 10.12. In the current MacOS version and, it seems, in the current Travis + Homebrew installations, only x86_64 makes sense.

This PR fixes all the recent Travis failures, as described in issue #3001.

### DIFF
--- a/m4/fontforge_platform_specifics.m4
+++ b/m4/fontforge_platform_specifics.m4
@@ -23,7 +23,7 @@ AS_CASE([$host],
        MACAPP=""
       fi
 
-      RAW_COMPILE_PLATFORM_CFLAGS=" $CFLAGS -arch x86_64 -arch i386 "
+      RAW_COMPILE_PLATFORM_CFLAGS=" $CFLAGS -arch x86_64 "
       AC_SUBST([RAW_COMPILE_PLATFORM_CFLAGS])
 
       dnl fink puts stuff under /sw


frank-trampe:
I'm not sure whether there's been a change not reflected in [the documentation]( https://developer.apple.com/library/content/documentation/Darwin/Conceptual/64bitPorting/building/building.html ), but that was certainly valid at the time, and building universal binaries was important. Mac OS 10.8 was the first to run exclusively in 64-bit mode, and thus the first for which one could be assured that a 64-bit-only binary would run. That may seem like a long time ago, but I caught flak last year for something that somebody else did that broke IRIX 6.5.22 support.

It's normally poor form to specify compile options that would not allow running on the building machine (building 64-bit-only binaries on a 32-bit machine). Since Mac OS is such a closed platform, and since Apple so aggressively deprecates old hardware and software, we might be able to break 32-bit builds without irritating anybody, but let me see whether there's an easier fix.

jtanx:
As per Homebrew/homebrew-core#5404 it appears to be an upstream issue, so it should hopefully be fixed by GLib, or otherwise Homebrew should be patching the header.

gioele:
Thank you for the pointer to the correct documention @frank-trampe. In my mind `-arch` was just an Apple synonym for `-march`. I realize now that it was used to create universal binaries (I supposed a second compilation pass was used in that case).

I have updated the commit message to reflect this info, in case no better solution can be found and it gets merged.

cicku:
+1

frank-trampe:
@jtanx, is there an estimated timeline on the GLib fix?

frank-trampe:
@gioele, actually, it all gets compiled in a single pass. A single Mach-O file can contain instructions for multiple architectures, and each compilation/linking step hits all the bases. The result is one binary. This technology was very useful during the PowerPC to x86 transition.

jtanx:
I did a little bit of searching, and it seems no one has reported it upstream(?????)....

I don't have time right now, but I'll see if I can report it some time this week.

gioele:
@frank-trampe any update on this?

@jtanx: have you had a chance to report the root cause of this problem upstream? I can also do if, if needed. Let me know.

frank-trampe:
@DomT4, could you help us here?

jtanx:
Sorry @gioele, that would be great if you could submit upstream - I've become a bit incapacitated for the moment

gioele:
And now we have an upstream report: <https://bugzilla.gnome.org/show_bug.cgi?id=780238>. :)

DomT4:
Sorry for the delay on my reply to the ping here. I presume from the conversation the desire here is primarily to avoid dropping 32-bit support? If you want my opinion though, I think you'd be perfectly safe doing so these days. Even going from 10.8 that's approaching 5 major macOS/OS X releases ago.

It is an upstream issue. MacPorts have just been patching it forever, but MacPorts _(for better or worse)_ are more liberal than Homebrew on the adopting patches upstream aren't aware of or actively don't want front.

It's possible Homebrew _might_ take the patch now it has been reported upstream, but I can't make promises given that I left the team there. FWIW Homebrew has pretty much killed off support for `universal` options in the core formulae.

jtanx:
Hmm, well it seems no one upstream wants to do anything about this. I'm fine merging this PR.

jtanx:
Last call before I merge this!

frank-trampe:
No. I'll find another way to fix this before the next release.

frank-trampe:
What happens if we remove the arch flag entirely?

jtanx:
I presume it will build the default arch, probably `x86_64`. I don't see the need to support this if the last potential 32-bit OS is Snow Leopard, and especially if no one upstream is willing to fix it either.

Issue #2866, opened 2016-09-23
SVG table support


KrasnayaPloshchad:
* Provide a general summary of the issue in the **Title** above.  
* Before you open an issue, please check if a similar issue already exists or has been closed before.

### Important
Mark with [x] to select. Leave as [ ] to unselect.

### When reporting a bug/issue:
- [ ] Screenshot
- [ ] The FontForge version and the operating system you're using
- [x] The behavior you expect to see, and the actual behavior
- [ ] Steps to reproduce the behavior 
- [ ] \(optional) Possible solution/fix/workaround

### When you open an issue for a change/improvement/feature request:
- [ ] A description of the problem you're trying to solve, including _why_ you think this is a problem
- [ ] If the feature changes current behavior, reasons why your solution is better
- [ ] \(optional) Possible solution/fix/workaround

[Several days ago](https://www.microsoft.com/typography/otspec/changes.htm), OpenType Specification introduced 'SVG ' tag to produce colour glyph, this tag should inplemented in FontForge.
https://www.microsoft.com/typography/otspec/svg.htm

KrasnayaPloshchad:
Firefox rejected to support SVG fonts in favor of supporting embedding SVG glyph into OpenType fonts (see [bug 119490](https://bugzilla.mozilla.org/show_bug.cgi?id=119490) at their Bugzilla), at this point we should do something to responce.

tshinnic:
Mentioning SVG, fonts, and FireFox together in one paragraph is bound to enrage some people.  It is Mozilla that sank [SVG 1.1 SVG fonts](https://www.w3.org/TR/SVG/fonts.html).  That cited bug report, extending over 11 years duration and with repeated pleading for Mozilla to quit stonewalling (because all other browsers had support!) - and including offers to fund and even staff the development effort - are a very good example of an organization which has ceased responding to its public.  That the attitude demonstrated by Mozilla belonged to the years B.C. (before Chrome) should not be overlooked.

Having been successful at "make the whiners go away" a couple of the perpetrators threw together a face-saving proposal, in combination with Adobe.  That proposal was embedding SVG glyphs within Opentype font files, the so-called SVG-in-Opentype.  That this was only a marginal improvement on the original SVG fonts - and not really well thought out - has become apparent now that font variations has come to fruition, as there is no path for using variations with embedded SVG glyphs.

[...]

PR #3043, opened 2017-03-08
don't write a second unicode attribute out to svg xml

carmatdaw:
When exporting SVG files, if the font used variation selectors, the file written out would have two 'unicode' attributes for each glyph that has a variation.  This causes problems for string XML interpreters.  The code I commented out has existed for a very long time, but I don't know if it's a bug that has existed for that long and gone unnoticed or if there is a legitimate reason to export invalid XML code for SVG's.

I'm going to assume it's a long standing bug that no one noticed. 

My change comments out two lines which will result in invalid XML whenever variation selectors are used.

### Motivation and Context

### Types of changes

### Description

### Final checklist

### DIFF
--- a/fontforge/svg.c
+++ b/fontforge/svg.c
@@ -788,8 +788,8 @@ static void svg_scdump(FILE *file, SplineChar *sc,int defwid, int encuni, int vs
 	    fprintf( file, "unicode=\"&#x%x;\" ", alt[0]);
 	else
 	    fprintf( file, "unicode=\"&#x%x;\" ", encuni);
-	if ( vs!=-1 )
-	    fprintf( file, "unicode=\"&#x%x;\" ", vs);
+	//if ( vs!=-1 )
+	//    fprintf( file, "unicode=\"&#x%x;\" ", vs);
     }
     if ( sc->width!=defwid )
 	fprintf( file, "horiz-adv-x=\"%d\" ", sc->width );


frank-trampe:
It's possible that the variation case ought to be part of the main if statement, but I'm not sure where it goes in the fallback hierarchy. @davelab6?

frank-trampe:
Pinging @davelab6.

davelab6:
> a legitimate reason to export invalid XML code for SVG's.

I don't think there is a legitimate reason to export invalid XML code for SVG's :)

> I'm going to assume it's a long standing bug that no one noticed.

I agree that assumption is reasonable. 

Please merge

davelab6:
Maybe mention this URL in the comment?

frank-trampe:
@davelab6, you didn't answer my question about the logic.

frank-trampe:
Pinging @davelab6.

adrientetar:
Yes I think the variation selector should be appended to the unicode value.

davelab6:
Thank you Adrien :) I don't have an opinion

khaledhosny:
The spec link FWIW: https://www.w3.org/TR/SVG/fonts.html#GlyphElementUnicodeAttribute, though I think SVG fonts are deprecated and useless, dropping support for writing SVG fonts would be even better.

tshinnic:
@khaledhosny [SVG fonts](https://www.w3.org/TR/SVG/fonts.html) are *effectively* deprecated, and now desupported by all browsers, greatly due to the stonewalling by Mozilla, as I mentioned in my comment to #2866 SVG table support.  The replacement SVG-in-Opentype was years in coming and has its own problems (like foot-dragging by Chrome).

Please consider that import/export of old-style SVG fonts can continue to be a good aid for people transitioning to SVG-in-Opentype from their old-style files and programs.  I would hate to have Fontforge drop such an aid just when it is becoming useful again.  Later when SVG-in-Opentype is well-established and has additional tooling, such as possibly here, perhaps the old-style SVG fonts will again become truly 'useless' and can be dropped.

khaledhosny:
Mozilla did not support SVG fonts for good reasons, they were basically just a collection of glyphs with any proper text layout support (either for complex scripts or even typography) so practically useless for anything but the simplest of uses.

Continuing to support importing SVG fonts is fine, but IMHO any time spent supporting SVG font export is wasted. Just my 2 qirsh, I’m not part of the FontForge team.

frank-trampe:
This is resolved in #3102.

Issue #3042, opened 2017-03-07
fontforge: PyFF_OpenFont(): fontforge killed by signal 6 


pnemade:
1) Use a simple fontforge script where it will open the sfd file e.g. say generate.ff

```
import fontforge
from sys import argv

argc = len( argv )

if argc > 1:
	for i in range( 1, argc ):
		f = fontforge.open( argv[i] )
		otfile = f.fontname + ".otf"
		print("Generating OpenType file ", otfile)
		f.close()
```

2) get any sfd file e.g. FreeSans.sfd from gnu-free-fonts upstream tarball or get it directly from here https://pnemade.fedorapeople.org/FreeSans.sfd

3) run following command
/usr/bin/fontfore -lang=py -script generate.ff FreeSans.sfd

This will cause core dump. I suspect issue somewhere in python3.6 version.  I am getting this core dump on Fedora 26 system where we have fontforge-20161012-5.fc26.x86_64, python3-3.6.0-20.fc26.x86_64, gcc-7.0.1-0.10.fc26.x86_64 packages installed.

Truncated backtrace:
```Thread no. 1 (7 frames)
 #6 PyFF_OpenFont at python.c:1013
 #7 _PyCFunction_FastCallDict at /usr/src/debug/Python-3.6.0/Objects/methodobject.c:234
 #8 _PyCFunction_FastCallKeywords at /usr/src/debug/Python-3.6.0/Objects/methodobject.c:295
 #9 call_function at /usr/src/debug/Python-3.6.0/Python/ceval.c:4788
 #10 _PyEval_EvalFrameDefault at /usr/src/debug/Python-3.6.0/Python/ceval.c:3275
 #11 _PyEval_EvalCodeWithName at /usr/src/debug/Python-3.6.0/Python/ceval.c:4119
 #12 PyEval_EvalCodeEx at /usr/src/debug/Python-3.6.0/Python/ceval.c:4140
```

More information can be obtained from https://bugzilla.redhat.com/show_bug.cgi?id=1429574

pravins:
This has been affecting build failure for Lohit series as well.  
* https://koji.fedoraproject.org/koji/taskinfo?taskID=18248306  
* Build Log: https://kojipkgs.fedoraproject.org//work/tasks/8306/18248306/build.log

* Script: https://raw.githubusercontent.com/pravins/lohit/master/scripts/apply_featurefile.py

frank-trampe:
[...]

PR #3046, opened 2017-03-10
Fix many of the free() calls to PyMem_Free() for the python3.6 requirement

pnemade:
This is also based on the discussion happened in upstream issue
https://github.com/fontforge/fontforge/issues/2992

This commit is a fix for the reported two problems in issue
https://github.com/fontforge/fontforge/issues/3042

The upstream documentation says
PyArg_ParseTuple() will allocate a buffer of the needed size, copy the
encoded data into this buffer and adjust *buffer to reference the newly
allocated storage. The caller is responsible for calling PyMem_Free()
to free the allocated buffer after use.


frank-trampe:
This looks good to me.

pnemade:
@frank-trampe Thank you for your review.

PR #3076, opened 2017-06-02
Clean up ufo.c

gioele:
This small PR declares a series of function as `static` and removes a couple of functions that have never been used since their introduction in 2014.

### DIFF
--- a/fontforge/psread.c
+++ b/fontforge/psread.c
@@ -93,7 +93,7 @@ static void dictfree(struct pskeydict *dict) {
 		dict->entries[i].type==ps_lit ) {
 	    free(dict->entries[i].u.str);
 	    dict->entries[i].u.str = NULL;
-    }
+	}
 	else if ( dict->entries[i].type==ps_array || dict->entries[i].type==ps_dict )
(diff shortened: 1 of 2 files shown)


frank-trampe:
I wrote the two functions that you have flagged for deletion when I was implementing the data structures that they use. They are necessary to support generating a clean UFO, which is something that we will almost certainly need to do.

As for adding static qualifiers, I don't think that it's worthwhile cluttering the changelog to provide them, particularly since we would then need to revert the change to add scripting support for some of those functions (like SplineFontHasUFOLayerNames) later.

ctrlcctrlv:
I'm not sure why this was reopened, but as it's already been explicitly rejected, no longer can be applied cleanly, and is years old, I'm putting it to bed. If there is interest in renewing this effort it can be reopened.

gioele:
Thanks for the taking care of these old PRs @ctrlcctrlv! :)

ctrlcctrlv:
No problem, someone has to do it&hellip;usually people hate closing PRs, no matter how old. At some point we just have to say, this isn't likely to ever be merged.

PR #3102, opened 2017-07-01
Avoid writing a duplicate Unicode value in svg output.

frank-trampe:
This is a slight adjustment of #3043.


frank-trampe:
@carmatdaw, @adrientetar, take a look.

Issue #3107, opened 2017-07-04
Accessing unused encoding slots in Python throws TypeError, instead of returning False for "font[slot].isWorthOutputting()". 


HinTak:
While re-writing a script in the legacy scripting interface to the python scripting interface 
( https://github.com/fontforge/fontforge/issues/3085#issuecomment-312740591 ), I came across this wierdness of the python interface: access an encoding slot that has not been used, results in a TypeError,
instead of a glyph which is false for "font[slot].isWorthOutputting()". This is not or poorly documented from in the slot accessing documentation, or the "isWorthOutputting()" documentation.

Also, this makes testing for unused slots (i.e. Not WorthOutputing()) a bit bizzare, as throwing an TypeError is the expected behavior while Not Throwing is the surprise and Not Throwing is therefore a error condition.

khaledhosny:
From the documentation:

> isWorthOutputting 	() 	Returns whether the glyph is worth outputting into a font file. Basically a glyph is worth outputting if it contains any contours, or references or has had its width set.

I don’t think it does what you think it is doing.

HinTak:
My (intended) usage of the "{Is}WorthOutputting()" is simply to confirm that I am copying from a source used slot into a destination unused slot.This works with the legacy scripting interface. But in the python code, the expected outcome (unused slot) throws a TypeError, while the surprise outcome, for which the user needs to be alerted, is when it DOES NOT throw an TypeError exception.
So I needed to both suppress the TypeError, and also deliberately throw another Error (currently I just print a message and continue) when fontforge does not throw a TypeError.

Retrieving unused slots - a glyph position that has not been drawn on - gives a TypeError is certainly surprising. It should just be a largely empty glyph, rather than throwing a TypeError.

This is also very confusing when the python routine shares almost the same name (except the "Is" at the beginning) with the legacy scripting routine but behaves substantially differently.

khaledhosny:
The `TypeError` is if coming from trying to index the font with a non-existing glyph; i.e. `font[slot]`, the index should be a glyph name or a glyph encoding (font specific, a fontforge thing not to be confused with the unicode value). `isWorthOutputting()` method is a member of the glyph class, you just can’t use it without a glyph.

HinTak:
That still does not help. I know what the current behavior is. However, let me repeat, that the goal is to test for unused destination (to paste into). Unfortunately, the desired outcome is an error condition in fontforge, and the no-error situation is the undesirable outcome that
[...]

PR #3109, opened 2017-07-07
Adding a new selection-based method, isWorthOutputting, for the pytho…

HinTak:
…n font object; and documentation

Thanks Frank Trampe for the suggestion.

Fixes
https://github.com/fontforge/fontforge/issues/3107

### Important

### Motivation and Context

### Types of changes

### Description

### Final checklist


JoesCat:
Although the subject is python oriented, it is worth making an equivalent "nativescript" command too, either now, or eventually later. Some complaints about FontForge have been how some instructions exist in nativescript, but not python, or python, but not nativescript, or GUI but not scripting.
There are users that exclusively use nativescript and not python.
Can you add a nativescript version?

HinTak:
You don't get it - this addition is to mimic the native script's WorthOutting() behavior, which is selection based when used without an argument, and glyph-based when used with argument. The python interface is simply lacking compared to the native interface. There is nothing to add in the native scripting interface. This is the whole point of this addiion: to make the python interface better match the native one in functionality.

HinTak:
It is a bit late for the naming discussion. The python glyph-based method did not follow the native name, so I think the font/selection-based method should follow python-convention, i.e. the glyph-based method naming. The native method simply have the functionality of both, depending on whether you pass an argument in; that's all, and with a slightly different name.

It is just unfortunate that the python implementation did not try to map first all the functionality of the native scripting interface. Or at least, put a placeholder in (and throw UnimplementedException for the time being, until somebody fill it in).
As such, it is difficult to migrate from one to the other.

frank-trampe:
@HinTak, the compound function in native scripting and the native FontForge function that the per-glyph Python binding exposes are different things and justifiably (if originally by accident) have different names. This wrapper corresponds directly to the the wrapper function from native scripting and not to the native FontForge function, and it ought to be named accordingly.

If your goal is to ease the transition between native and Python scripting, which seems to me to be the only reason to add this Python binding at all, it seems wise to keep the names consistent between the two rather than arbitrarily naming this function after another function that behaves differently.

This is what I am proposing.

| Name | Logic in C | Logic in FF Scripting | Logic in Python |
|--|--|--|--|
| isWorthOutputting | Take a glyph as input, output whether the glyph is significant. | (Not implemented.) | Take a glyph as input, output whether the glyph is significant. |
| WorthOutputting | (Not implemented.) | Take a slot or a selection as input, output whether there is an assigned glyph and whether it is significant. | Take a slot or a selectionas input, output whether there is an assigned glyph and whether it is significant. |

This is what you are proposing.

| Name | Logic in C | Logic in FF Scripting | Logic in Python |
|--|--|--|--|
| isWorthOutputting | Take a glyph as input, output whether the glyph is significant. | (Not implemented.) | If called on a glyph, take the glyph as input and output whether the glyph is significant. If called on a splinefont, take a slot or a selection as input and output whether there is an assigned glyph and whether it is significant. |
| WorthOutputting | (Not implemented.) | Take a slot or a selection as input, output whether there is an assigned glyph and whether it is significant. | (Not implemented.) |

HinTak:
There are some fundamental differences between the two scripting languages: the python variant is object-oriented, whereas the native variant is largely context-oriented, or based on the current selections. I don't think there needs to be any relationship between the scripting languages and the C internals; but I see consistency within each language more important - hence a uniform 'IsWorthOutputting()' method across the python variant, and does different but related things depending on the object.

The point of this change is that,

Select(x);
# check that the destination is empty before pasting
if (! WorthOutputting())
    Paste()
else
    # throw error
    Print("Destination Full")

does not require doing something awful like this in python:

            font.selection.select( x )
            try:
                font[x]
            except TypeError:
                # expect this!
                font.paste()
                pass
            else:
                # throw error
                print("Destination full")


I think arguing the scripting functions' name should match the C internals is a bit weak - the whole point of the scripting language is that one doesn't want to touch the C internals... if that's a valid argument, python's len() methods should be named strlen() or something, and python's str() method should be named sprintf() !

ctrlcctrlv:
At this point...this no longer can be merged without a conflict, it's years old, and @frank-trampe seems to not want it. So I'm closing this as it is unlikely to ever be merged, primarily due to its age. It can be reopened if interest in it is ever renewed.

HinTak:
Updated version of this pull in the New https://github.com/fontforge/fontforge/pull/4522

PR #3111, opened 2017-07-09
Fixup dist target included/cleaned files

### LINKED ISSUE #3110: lots of left-over files and not exactly clean which of make *clean wll clean it.
After `make`, a lot of files appears, and `make clean` does not really clean up much. Also there are a lot of confusing *clean targets in the generated Makefile. There is a distclean target but it does not clean up either. It is not exactly clear which will clean the tree to the original state.

I did

```
rm -rf *
git reset --hard
```

a few times already. That's really barbaric and should not be a requirement to do anything with the fontforge code base.

jtanx:
* Don't include `fontforge-config.h` and `Info.plist` in the dist package (these are generated files)
* Testsuite: Don't delete `testsuite` on `make distclean` - otherwise you'd need autoconf to regenerate it
* Bundle fetched fonts when `make dist` is called. Fixes the files being re-downloaded anyway when running `make check` from a dist tarball. 
* Separate fetched fonts from generated fonts; remove generated fonts on `make distclean`, only remove fetched fonts on `make maintainer-clean`
* Add in simple TTC loading check

Closes #3110.

<strike>I also noticed that the test that depends on mingliu.ttc fails now; something's changed that's causing FontForge to no longer be able to read the names of the fonts within the mingliu ttc.</strike> actually no, it looks like the source for that mingliu.ttc is not good.

### DIFF
--- a/inc/Makefile.am
+++ b/inc/Makefile.am
@@ -34,10 +34,11 @@ DISTCLEANFILES=fontforge-config.h stamp-h1 stamp-h2
 # they should require a "fontforge/" prefix.
 
 pkginclude_HEADERS = basics.h chardata.h charset.h dynamic.h	\
-	fileutil.h fontforge-config.h gdraw.h gfile.h ggadget.h gicons.h gimage.h gio.h		\
+	fileutil.h gdraw.h gfile.h ggadget.h gicons.h gimage.h gio.h		\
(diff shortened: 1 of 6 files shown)


JoesCat:
Hi @rrthomas, hope you have a minute to review a change to some lines concerning distclean?

rrthomas:
Looks good to me!

Issue #3123, opened 2017-07-29
Multiple CVE's assigned against fontforge


copyninja:
Hi,

There are multiple CVE's assigned aginst fontforge as can be seen in [Debian security tracker](https://security-tracker.debian.org/tracker/source-package/fontforge). All these issues are reported by @gnehsoah and are as follows.

- [x] #3088
- [x] #3091 
- [x] #3096 
- [x] #3090 
- [ ] #3098 
- [x] #3092 
- [x] #3087 
- [ ] #3097 
- [x] #3093 
- [x] #3089 

I'm opening this issue to track these CVE progress. 

CC: @jtanx  @frank-trampe @JoesCat @davelab6

davelab6:
I can't speak for anyone else but from what I can see, nowadays FontForge
is "diy". If you can't fix your problem yourself, you can pay someone like
Frank to do it, but no one is too actively fixing things other people care
about.

frank-trampe:
I'll look at these eventually, but I've been super busy, and we have a mailing list crisis right now that's diverting my attention.

copyninja:
@frank-trampe July release contains fixes for above in it?.

copyninja:
@jtanx is #3089 fixed?. I did not see it in any commits though you have manually closed bug using PR #3124

jtanx:
@copyninja #3089 was fixed as a consequence of fixing one of the other reported issues - specifically from this commit: https://github.com/fontforge/fontforge/pull/3124/commits/4de0c58a01e5e30610c200e9aea98bc7db12c7ac#diff-dbea8a90f6a82c8f5c9b4970183d95c6R3668

copyninja:
@jtanx  Thanks for clarification. Any ETA for #3097 and #3098. Just asking as I need to prepare a upload to Stable Security of Debian.

VictorRodriguez:
Hi team 

I wonder if the CVE-2017-17521 is a valid CVE or is a won't fix for you : 


Description | uiutil.c in FontForge through 20170731 does not validate strings before launching the program specified by the BROWSER environment variable, which might allow remote attackers to conduct argument-injection attacks via a crafted URL, a different vulnerability than CVE-2017-17534.

https://security-tracker.debian.org/tracker/CVE-2017-17521

JoesCat:
Hi @VictorRodriguez ,
There are many things to bulletproof in FontForge, but I wonder if this is a valid CVE ?

[...]

Issue #3098, opened 2017-06-14
global buffer overflow in ValidatePostScriptFontName (parsettf.c)


fatal0:
```bash
# fontforge -lang=ff -c 'Open($1)' ./poc/ValidatePostScriptFontName-in-parsettf.c-global-buffer-overflow.otf
Copyright (c) 2000-2014 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Based on sources from 09:14 UTC 13-Jun-2017-D.
 Based on source from git with hash: b8e5ff8f24955f4d7d59ac73c903cc088b21bdb6
Warning: Mac and Windows entries in the 'name' table differ for the
 SubFamily string in the language English (US)
 Mac String: Medium
Windows String: Regular
Warning: Mac and Windows entries in the 'name' table differ for the
 Fullname string in the language English (US)
 Mac String: CFF_Type-1_0x0d_expl Medium
Windows String: CFF_Type-1_0x0d_expl
=================================================================
==21454==ERROR: AddressSanitizer: global-buffer-overflow on address 0x000001081a64 at pc 0x00000051b6be bp 0x7ffd652858e0 sp 0x7ffd652858d0
READ of size 4 at 0x000001081a64 thread T0
    #0 0x51b6bd in ValidatePostScriptFontName /root/afl_fuzz/project/test/fontforge/fontforge/parsettf.c:1459
    #1 0x526ab9 in readcfftopdict /root/afl_fuzz/project/test/fontforge/fontforge/parsettf.c:3006
    #2 0x52ae2a in readcfftopdicts /root/afl_fuzz/project/test/fontforge/fontforge/parsettf.c:3323
    #3 0x532df4 in readcffglyphs /root/afl_fuzz/project/test/fontforge/fontforge/parsettf.c:3934
    #4 0x54687c in readttf /root/afl_fuzz/project/test/fontforge/fontforge/parsettf.c:5527
    #5 0x552652 in _SFReadTTF /root/afl_fuzz/project/test/fontforge/fontforge/parsettf.c:6266
    #6 0x7e8756 in _ReadSplineFont /root/afl_fuzz/project/test/fontforge/fontforge/splinefont.c:1114
[...]

Issue #3097, opened 2017-06-14
global buffer overflow in umodenc (parsettf.c)


fatal0:
```bash
# fontforge -lang=ff -c 'Open($1)' ./poc/umodenc-in-parsettf.c-global-buffer-overflow.otf
Copyright (c) 2000-2014 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Based on sources from 09:14 UTC 13-Jun-2017-D.
 Based on source from git with hash: b8e5ff8f24955f4d7d59ac73c903cc088b21bdb6
Warning: Mac and Windows entries in the 'name' table differ for the
 SubFamily string in the language English (US)
 Mac String: Medium
Windows String: Regular
Warning: Mac and Windows entries in the 'name' table differ for the
 Fullname string in the language English (US)
 Mac String: CFF_Type-1_0x0d_expl Medium
Windows String: CFF_Type-1_0x0d_expl
In CFF_Type-1_0x0d_expl, in glyph .notdef, 'CFF ' advance width (565) and
  'hmtx' width (629) do not match. (Subsequent mismatches will not be reported)
Bad encoding information in 'cmap' table.
=================================================================
==21440==ERROR: AddressSanitizer: global-buffer-overflow on address 0x000000f0e6b8 at pc 0x0000005367d5 bp 0x7ffc5395e050 sp 0x7ffc5395e040
READ of size 4 at 0x000000f0e6b8 thread T0
    #0 0x5367d4 in umodenc /root/afl_fuzz/project/test/fontforge/fontforge/parsettf.c:4237
    #1 0x53c51d in readttfencodings /root/afl_fuzz/project/test/fontforge/fontforge/parsettf.c:4762
    #2 0x546e63 in readttf /root/afl_fuzz/project/test/fontforge/fontforge/parsettf.c:5563
    #3 0x552652 in _SFReadTTF /root/afl_fuzz/project/test/fontforge/fontforge/parsettf.c:6266
    #4 0x7e8756 in _ReadSplineFont /root/afl_fuzz/project/test/fontforge/fontforge/splinefont.c:1114
    #5 0x7ea217 in ReadSplineFont /root/afl_fuzz/project/test/fontforge/fontforge/splinefont.c:1285
[...]

PR #3124, opened 2017-07-30
Fix a few cves

jtanx:
So the long and the short of it is that FontForge was not designed with security in mind. I would not be surprised at all if a simple permutation on the test cases resulted in another buffer overflow/out of bounds read/etc. There's just too many cases to cover.

Anyway, this fixes most of the *specific* test cases as listed in #3123. I've avoided #3097 and #3098 which involves changing the codegen around the unicode library. For any other issue in that list where there isn't a specific commit, it's probably been fixed as a side effect of one of the other commits.

### DIFF
--- a/fontforge/parsettf.c
+++ b/fontforge/parsettf.c
@@ -3533,7 +3533,7 @@ static void readcffset(FILE *ttf,struct topdicts *dict,struct ttfinfo *info) {
 	    for ( i = 1; i<len; ) {
 		first = dict->charset[i++] = getushort(ttf);
 		cnt = getc(ttf);
-		for ( j=0; j<cnt; ++j )
+		for ( j=0; j<cnt && i<len; ++j )
 		    dict->charset[i++] = ++first;
 	    }
 	} else if ( format==2 ) {
--- a/fontforge/parsettf.c
+++ b/fontforge/parsettf.c
@@ -3116,6 +3116,10 @@ static struct topdicts *readcfftopdict(FILE *ttf, char *fontname, int len,
 	  case (12<<8)+24:
 	    LogError( _("FontForge does not support type2 multiple master fonts\n") );
 	    info->bad_cff = true;
+	    if (sp < 4) {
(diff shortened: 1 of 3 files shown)


frank-trampe:
Just in time. I'm cutting a release tomorrow. I'll review as soon as I get a chance.

fontforge/parsettf.c
```diff
@@ -2796,6 +2796,10 @@ return( 3 );
 	pt = buffer;
 	do {
 	    ch = getc(ttf);
+		if ((pt-buffer) > (sizeof(buffer) - 3)) {
```
gioele: Hi @jtanx, the rest of the CVE-related commits LGTM, but here I would add a comment about what does it mean that `(pt-buffer) > (sizeof(buffer) - 3)`.

jtanx: Thanks for reviewing! I've slightly modified the check in addition to adding in the comments (technically, it should only check for `sizeof(buffer) - 2`).

fontforge/parsettf.c
```diff
@@ -3320,6 +3333,10 @@ return( NULL );
 	offsets[i] = getoffset(ttf,offsize);
     dicts = malloc((count+1)*sizeof(struct topdicts *));
     for ( i=0; i<count; ++i ) {
+	if (fontnames != NULL && fontnames[i] == NULL) {
+		LogError(_("Number of CFF font names is less than dict size: %d < %d"), i, count);
+		break;
+	}
```
frank-trampe: This leaves dicts with random data at and above index `i`, no?

frank-trampe:
I made one inline comment/query. It otherwise looks fine.

fontforge/parsettf.c
jtanx: Not quite; `dicts` is NULL terminated, which happens at the conclusion of the loop (line 3347).

frank-trampe: Ah, okay.

frank-trampe:
All good.

PR #3127, opened 2017-08-01
Fix public headers

gioele:
Some headers installed by `make install` refer to other FontForge headers that are not installed. This breaks external C and C++ applications that use FontForge as a library.

This PR fixes all the issues of this kind that I have found.

Side note: some of the fixed problems are now, but some have been around for years. I wonder how (or if) people manage to compile applications that depend on FontForge.


gioele:
@jtanx, @JoesCat, @frank-trampe any objections to these changes?

jtanx:
Yeah, it looks okay I guess. What uses FontForge as a compile-time dependency? I only know of pdf2htmlex which uses some old branch of FF.

frank-trampe:
Looks good to me.

PR #3131, opened 2017-08-03
coverage: Add script to create a coverage report using gcov/lcov

gioele:
This commit contains a small (but cute ;)) script that generates an HTML coverage report. This report describes how many (and which) lines of code are executed by the tests performed by `make check`.

A demo: https://gioele.github.io/fontforge/report-1501802471/

Sadly the coverage is quite low. Only 20% of the total lines (~18% of the total functions). :( But now that we have a quantifiable metric, we will make it much better! ;)

This is the first step of the following plan:

1. generate a local HTML report of the coverage
2. upload the HTML report to a `gh-pages` branch by pushing a certain directory to Travis-CI
3. setup a diff-based workflow to show how coverage is increasing or decreasing.

PS: I tried setting up the code coverage via autoconf/automake but I ultimately abandoned the idea: setting up gcov in autoconf/automake is be quite convoluted and requires the `Makefile.am` files to be extremely clean and lean, something that the current FontForge repo does not offer at the moment. If others want to try, there is already a m4 macro package [ax_code_coverage](https://www.gnu.org/software/autoconf-archive/ax_code_coverage.html). Good luck integrating it and removing '-O2' from the `CFLAGS`. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,14 @@ env:
   global:
     # For Coverity
     - secure: "MeTS1Pqa5gzx1nn/peW/9a5kq84bba3XYUljOfkCUqzuyGiERk/nmok+RW7skrgzboBlKxnNG8+ykKqHMwK9s9M83ezFxvEWXBcKEpmEQKkqXPI5hpMs6jGLTgpeuheSIzqHA3danV8iircp1GOiTLWA0pt/AOsNLZiaYBh0OiE="
+  matrix:
+    - TRACE_CODE_COVERAGE=false
+    - TRACE_CODE_COVERAGE=true
+
+matrix:
+  exclude:
+    - os: osx
+      env: TRACE_CODE_COVERAGE=true
 
 addons:
   apt:
@@ -25,7 +33,7 @@ addons:
       - deadsnakes
     packages: [autoconf, automake, autotools-dev, bzip2, gcc-4.9, libtool,
       libjpeg-dev, libtiff4-dev, libpng12-dev, libfreetype6-dev, libgif-dev,
-      libx11-dev, libxml2-dev, libpango1.0-dev, libcairo2-dev, python3.5-dev]
+      libx11-dev, libxml2-dev, libpango1.0-dev, libcairo2-dev, python3.5-dev, lcov]
   coverity_scan:
     project:
       name: "fontforge/fontforge"
@@ -38,6 +46,9 @@ addons:
 cache:
   directories:
     - $TRAVIS_BUILD_DIR/travisdeps
+
+before_install:
+  - if $TRACE_CODE_COVERAGE ; then pip --quiet install --user cpp-coveralls==0.3.12 ; fi
 
 #before_install:
 #  - if [ "$TRAVIS_OS_NAME" == "osx" ]; then brew update || brew update; fi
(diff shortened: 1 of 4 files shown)


jtanx:
Interesting. Any thoughts about coveralls and the like (i.e. the ones with Travis/Github integration)?

gioele:
First, an explanation: steps 2 and 3 are by-products of another private project I am working on. In that project, non-FLOSS hosted services like Coveralls or Code Climate cannot be used.

That said, uploading gcov traces to Coveralls for fontforge is definitely a possibility.

The need for this script, however, remains it we don't want to make our Travis builds twice as slow as they are now. As I wrote, integrating gcov support into the normal compilation procedure is a pain because of how the FF `Makefile.am` files have been written. (If somebody wants to go on on a yak-shaving expedition refactoring the automake files, I can offer some pointers about what needs to be cleaned up.) Because gcov is not enabled during the normal compilation, one additional compilation cycle is needed (`make clean`, `make`, `make check`) in order to gather the coverage data. Our Travis builds are already extremely slow. I'd avoid making them twice as slow.

If there is consensus on having our code coverage analysed, I can add a commit to this PR to hook up this script to Travis and upload the results to a `gh-pages` branch. After that we can easily switch to Coveralls by modifying the "upload" part of the script.

jtanx:
>  Our Travis builds are already extremely slow. I'd avoid making them twice as slow.

It should be doable to have some sort of matrix build to run the code coverage build in parallel to the main build, so it's unlikely to slow it down overall (the Mac build is by far the slowest, and that's with the Linux build already doing a 2x build for distcheck haha).

In terms of uploading to gh-pages, I just get the feeling that that's not really the right place to be putting reports that are effectively tied to a build (well okay, for FontForge this probably doesn't matter given how infrequently the test suite changes). But since we already have access to coveralls, we might as well go straight to using that.

On Makefile integration, well yeah the Makefiles are a bit crap, but something like https://github.com/curl/curl/commit/f6e0f4556eabc8c2629d16ae599a153392d30d05 doesn't look _too_ bad... 

And as for the "-g -O2" default CFLAGS, that appears to be coming directly from autoconf/autotools itself. I'd suggest that when doing a coverage build that the CFLAGS be overridden manually, e.g.

```
CFLAGS="-g -O0" ./configure
```

or use

```
./configure --enable-debug            (does the same thing, although might enable some other debugging code via the `FONTFORGE_DEBUG` define)
```

gioele:
Setting up the matrix to do a parallel build is a good idea! Not only it is faster, it also solves the problem of rebuilding with different CFLAGS and AM_* vars.

About the `profit` step :), Coveralls can be integrated quite painlessly via <https://github.com/eddyxu/cpp-coveralls#linux>. Could you set up a coveralls account for fontforge?

jtanx:
I've added a project for FontForge onto coveralls <https://coveralls.io/github/fontforge/fontforge>.

If you want to test anything on travis, you can try pushing to the `travis` branch (if it exists, it's usually ok to force push over it/do whatever).

gioele:
Hi @jtanx: I have moved this PR over to your branch and I modified `.travis.yml` to produce the coverage report and upload it to coveralls.

It works: see <https://coveralls.io/builds/12800271> and <https://coveralls.io/github/fontforge/fontforge>.

jtanx:
Awesome, great job getting it working!

Issue #2389, opened 2015-06-24
Generate: Don't include FFTM


davelab6:
It is annoying to have to remove `FFTM` tables from `.ttf` fonts made by FontForge. 
- [ ] The Generate dialog should have an option to not create them
- [ ] native scripting `Generate()` should have this option
- [ ] pythong scripting `font.generate()` should have this option

ctrlcctrlv:
...why?

http://pfaedit.org/non-standard.html#FFTM

It's just a timestamp, I'm not sure who it's hurting by leaving it there.

I'm firmly in the "more options are always better" camp (interfaces designed by committee are a list of checkboxes and I am quite okay with that) but I'm just curious why this annoys you.

davelab6:
Unneccessary extra bytes should not exist in web fonts. Serving billions of copies of a font per day, it all adds up.

ctrlcctrlv:
Gotcha :+1:

dscorbett:
@ctrlcctrlv, this should be reopened.

PR #3134, opened 2017-08-12
support reproducible builds

### LINKED ISSUE #2711: Please do not put timestamps in generated files

volth:

### Important

### Motivation and Context
- [x] It blocks propagation of environment entropy (such as the current time or the Linux user name) to the produced fonts when it is not desirable. It is detected by presence ```$SOURCE_DATE_EPOCH``` in the environment which is set by Debian and NixOS build script when they want to do a reproducible build.
- [x] Closes #2711 #2389

### Types of changes

### Description
if ```$SOURCE_DATE_EPOCH``` is detected then its value is used instead of system time or last modification time of the source files (which is often close to the current time if they were just patched or pre-processed). Also presence of ```$SOURCE_DATE_EPOCH``` blocks propagating of the current Linux user name to the ```author``` field of the fonts. It is often non-deterministic (like ```build-user-4```, etc)

It is not relevant to https://github.com/fontforge/fontforge/pull/2943, which is about ```$SOURCE_DATE_EPOCH``` when fontforge is being compiled from C sources, this fix about ```$SOURCE_DATE_EPOCH``` when fontforge compiles fonts.

### Final checklist


jtanx:
This kind of feels like a hack... This is better done as one of:

* Check `SOURCE_DATE_EPOCH` once in one location and set some variable accordingly. Use said variable where needed
* As before, but take it as an input argument to fontforge instead of an environment variable
* As before, but as a user preference.

gioele:
Hi @volth, thank you for the PR. As @jtanx said, the changes you are proposing are OK, but a bit too "verbose". Fontforge's code base is already _hackish_ enough. :)

I went ahead and created added two functions to `gutils.c` that wrap `time` and `st_mtime`. You can find the code in my branch [`reproducible-time-author`](https://github.com/gioele/fontforge/tree/reproducible-time-author).

Could you please update this commit to use these two functions?

frank-trampe:
@volth, thanks for pushing this ahead by submitting the pull request. This has been on my list for a long time. @gioele, any chance you could just make a new superseding pull request with your modifications? I like the simple fix.

gioele:
@frank-trampe: I'm afraid I don't understand what you are suggesting; the English parser in my brain gets clogged every now and then. ;)

Are you suggesting to merge @volth commit now and then clean up using my two functions? Or are you suggesting that I make another PR with just my changes, merge it and then ask @volth to reimplement the suggested changes on top of the new master? Or something else altogether?

frank-trampe:
@giole, I was hoping that you might just be able to make one pull request that accomplishes what we need. It simplifies the merge logistics.

JoesCat:
Can you review GetTime() found in [libspiro call-test.c](https://github.com/fontforge/libspiro/blob/master/tests/call-test.c)
This GetTime() version has already gone through edits to work with older operating systems (as old as DOS), and also works okay with current operating systems. Some patching was done to revert a fail for an older openBSD 7or8, but HEAD works okay with Mac, so I believe it's okay now too.
`SOURCE_DATE_EPOCH` is a recent variable that doesn't exist with older operating systems and I suspect it will be a problem for backwards compatibility.

jtanx:
From what I can see, SOURCE_DATE_EPOCH  is an environment variable, so unless the os doesn't support that, I don't see why it wouldn't work.

JoesCat:
On September 3, 2017 03:06:26 PM Jeremy Tan wrote:
> From what I can see, SOURCE_DATE_EPOCH  is an environment variable, so
> unless the os doesn't support that, I don't see why it wouldn't work.

Yes, you're right - my error.

The "} else {" section of the if is the part that should be modified to deal 
with recent and older operating systems.
When I upgraded libspiro, it definitely needed to use the older code style 
for older operating systems, and it needed the newer style for newer 
operating systems.

The earliest use of SOURCE_DATE_EPOCH appears to be about 2015-10-26
https://reproducible-builds.org/docs/timestamps/
It appears to be fairly linux/debian centered:
https://wiki.debian.org/ReproducibleBuilds/TimestampsProposal#Reading_the_variable

The suggestion I made works as far back as 1990, maybe even earlier, and is 
what I should have meant for the "} else {" section.

frank-trampe:
@JoesCat, just to be clear, your wrapper only deals with fetching the actual time, not with parsing a string to an integer, right? If so, these two things (adding support for getting a timestamp from an environment variable and enhancing support for getting the current time) are entirely disjoint issues, such that we can handle them at separate times. It wouldn't hurt to work on our legacy support, but it's not clear to me that adding a smarter time wrapper would actually allow us to target older platforms. The gnulib switch essentially foreclosed on all of those possibilities, and it's tough to go back now.

volth:
There are still reproducible problem: fonts compiled on MacOS differ from fonts compiled on Linux: (Travis CI failure in https://github.com/NixOS/nixpkgs/pull/28216).
I have no MacOS here so cannot investigate :(

gioele:
hi @volth, could you please the Travis script of the [failing built](https://travis-ci.org/NixOS/nixpkgs/jobs/279055625) to store the generated font somewhere or to output it to the standard output? Something like the following would be enough.

```
- echo "Dump of the generated file" && hexdump < generated-font.ttf
```

That will give us a file to play with in order to discover the root cause of this.

volth:
@gioele If I would know how to download from Travis CI, I could fix the issue. It is not mine Travis, it is the one affiliated with github which does build the pull requests

The script to build the font is
https://github.com/NixOS/nixpkgs/blob/master/pkgs/data/fonts/libertine/default.nix
The resulting sha256 is different on Linux and MacOS

gioele:
@volth: you can change the Travis script without being the repo owner. It's quite easy. :) Just add

```
after_failure: echo "Dump of the generated file" && hexdump < GENERATED-FONT.ttf
```

to `.travis.yml` after the `script:` line in the `osx` stanza. Then make a commit in the branch linked to the PR and push it.

Travis will display the content of the font in the log and from that we can read it back.

volth:
@gioele nice hint, thank you!

volth:
There is something really weird on MacOS: not only the font files are different, but there are one font file missing (65 files produced on Linux and 64 on MacOS)

volth:
Apart from Travis issues (with lost files), there are difference in order of same-name features in GSUB table (I built the same fonts on Hackintosh in VirtualBox):



I am not sure should the same-name features be ordered or merged or dropped.. Perhaps it is a topic for another issue.

frank-trampe:
I can't look at this until later in the week. I'm not terribly concerned about Macintosh, though, as industrial workflows of the sort that require consistency probably run on Linux these days.

volth:
This is likely gcc vs. clang issue than linux vs. macos

gioele:
hi @volth, I have just pushed PR #3176 with full support for `SOURCE_DATE_EPOCH`.

Could you please close this PR and file a separate bug with all the information about the mismatch between Linux and MacOS fonts, possibly with attachments?

PR #3147, opened 2017-09-09
startnoui.c: Include header that defines FindOrMakeEncoding

### LINKED ISSUE #3145: FontForge segfaults when instantiating a new font if compiled with the UI disabled
Hello,

I got a new macbook and tried to run fontcustom, but I always get the error `error  `fontforge` compilation failed. Try again with --debug for more details.`. After some research it seems that it is still working with the version `20170104`. How can I downgrade to this version? If I do `brew search fontforge` I only see the latest version.

currently I got:
fontcustom 1.3.8
fontforge 20170806

Thanks!

jtanx:
Failure to include this was causing an implicit pointer to int conversion
which results in an invalid pointer address on 64-bit builds.

Fixes #3145.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: c
 compiler: gcc
 sudo: false
-dist: precise
+dist: trusty
 osx_image: xcode8
 
 branches:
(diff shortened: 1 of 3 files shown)


jtanx:
I think #2992 is happening again (https://travis-ci.org/fontforge/fontforge/jobs/273513179)

I downgraded back to python 3.5 since it's not really related to this issue

frank-trampe:
Was this already merged into Homebrew somehow?

jtanx:
I got a patch merged in, since this really only affects versions of FontForge compiled with the GUI/X11 disabled.

frank-trampe:
The changes look good to me. If it works there and works in Travis, I say we merge it.

PR #3150, opened 2017-09-12
python: Use PyMem_Free wherever PyArg_ParseTuple+es is used

jtanx:
This fixes the crash I was seeing against python 3.6 when running the test suite. Still not sure if this covers all the cases :S (follow up to #3147)

https://docs.python.org/2.0/ext/parseTuple.html

Travis will probably fail because of zeromq.


frank-trampe:
The code looks fine, and the Macintosh Travis job shows Python importing correctly, so you can merge this if you're happy with it.

jtanx:
I've updated Travis to actually use Python 3.6, and I've also just disabled that warning that was preventing czmq from being built.

PR #3152, opened 2017-09-18
Allow to override build date

bmwiedemann:
in order to make font builds reproducible
such as the kde-oxygen-fonts openSUSE package.

See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.


jtanx:
Is this already covered by #3134?

bmwiedemann:
yes. I'll have a look if I can help get it merged.

bmwiedemann:
@frank-trampe I cherry-picked the nice infrastructure commit from @gioele 
GetTime is now used in the 4 places that I found to be relevant for openSUSE font packages to build reproducibly.
I did some testing and therefore am confident that it works.

gioele:
Thanks @bmwiedemann, it looks good to me.

I sugesst we merge this for the moment, then we can go after all the other failure cases pointed out by @volth in #3134.

volth:
For reproducible builds it also needed to overwrite ```GetAuthor```, there is often Linux user name which is not stable but something like "nixbld4". Anyway, it is good to make the first little step towards reproducible builds.

gioele:
@volth: indeed, that is the idea. Now this basic cases, later all cases.

@jtanx, @frank-trampe: if there are no objections I could merge this code (overriding `time()` to make it reproducible) in the next days.

gioele:
Merged, thanks @bmwiedemann.

PR #3153, opened 2017-09-20
Always use ffglib.h in place of glib.h

### LINKED ISSUE #2841: Rename structures conflicting with GLib/GIO/GDK
Since we control the libraries (and we're the only ones using it), it would be ideal to rename some of our GDraw components, like `GList`, `GTimer` and `GMenuItem` to something else, like `GDList` , `GDTimer` and `GDMenuItem` to avoid conflicts with GLib/GIO/GDK. This means that we can eliminate `ffglib.h` and avoid having to write things like `GList_Glib` when we actually want to use the glib variant of GList.

For example, right now I have to back out @gioele's changes to `startui.c` in my [GDK PR](https://github.com/fontforge/fontforge/pull/2782/commits/c34919d0044a5ff14874eb3e919316571e67a29d) because I need to include the gdk header.

jtanx:
This moves `ffglib.h` out of `fontforge` and into `inc`, and further replaces all includes to `glib.h` with `ffglib.h`, which standardises how glib is imported and its functions name mangled (where necessary).

This should resolve #2841 as an alternative solution (on second thought, renaming our own structs is too invasive). This is in prep for #2782.


frank-trampe:
Fine with me. Merge when ready.

jtanx:
Thanks

PR #3158, opened 2017-09-28
Travis-CI: Run coveralls only if the corresponding account has been set up

gioele:
The current Travis script pushes the coverage results even when the user has no account on Coveralls. This means many failed builds for all the external contributors that are not using the main fontforge repo to do tests (this includes me ;)).

With this commit, the coverage stuff is run only if a Coveralls account for the pushing user exist.

On a related note, I am working on a simplification of the Travis script so that in the future the build will be skipped entirely. Let's help the nice Travis people save some electricity. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ os:
 
 env:
   global:
+    - COVERALLS_IS_SET_UP=true # assume coveralls is set up for the moment, will be detected later
     # For Coverity
     - secure: "MeTS1Pqa5gzx1nn/peW/9a5kq84bba3XYUljOfkCUqzuyGiERk/nmok+RW7skrgzboBlKxnNG8+ykKqHMwK9s9M83ezFxvEWXBcKEpmEQKkqXPI5hpMs6jGLTgpeuheSIzqHA3danV8iircp1GOiTLWA0pt/AOsNLZiaYBh0OiE="
   matrix:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -69,6 +70,9 @@ install:
 
     if [ ! -z "$TRACE_CODE_COVERAGE" ] ; then
         FFCONFIG="$FFCONFIG --enable-code-coverage --enable-debug"
+        if ! curl --head --fail --silent "https://coveralls.io/github/${TRAVIS_REPO_SLUG}" > /dev/null ; then
+            COVERALLS_IS_SET_UP=false
```
jtanx: I think you could just

```bash
unset TRACE_CODE_COVERAGE
```

and not have `COVERALLS_IS_SET_UP` at all.

jtanx: Actually thinking about this a bit more, you could perhaps have the coveralls build as an allowable failure via the `allow_failures` setting, and just exit early in this case

gioele: hi @jtanx, I discarded `allow_failures` because using it would mean that failures run the coverage script or to upload to coveralls even from the official repo/branch would be "ignored", something I'd like to avoid.

But yes, using `allow_failures` would save us some line compared to this solution.

jtanx: I think it's somewhat saved by the fact that if nothing gets uploaded to coverity, then eventually it will register the coverage as being 0%, which marks the build as a failure. 

In any case, you shouldn't need the extra `COVERALLS_IS_SET_UP` variable.

frank-trampe:
This one is concurrently non-invasive, highly testable, and more in your areas of expertise than mine. If you two are happy, and the change passes Travis, just merge it.

gioele:
Thanks for the review @jtanx  and @frank-trampe. I will close this PR now without merging it. A similar functionality is now part of the more general PR #3168.

PR #3160, opened 2017-09-29
Headers for tables

gioele:
This PR moves definitions of shared constants to separate headers.

There are probably other constants that could help such a treatment, but these are those that we need to fix in order to fix some warnings/errors spotted by gcc 7 while working on gutils for the reproducible builds.

One question: does `gimageclut.c` really needs to include the file with the constant itself under MINGW32? Isn't the inclusion of the header enough?

```diff
 #if defined(__MINGW32__)
-#  include "../gutils/divisors.c"
+#  include "divisors.c"
 #endif
```


jtanx:
Okay, I think I see what's going on here.

div_tables is only used by gdraw. Previously it probably needed to be included under Windows, for some wacky DLL reason (likely [this](https://stackoverflow.com/questions/19373061/what-happens-to-global-and-static-variables-in-a-shared-library-when-it-is-dynam)). Since you've moved div_tables into gdraw, you no longer need this block, so remove it:

```c
 #if defined(__MINGW32__)
 #  include "divisors.c"
 #endif
```

Secondly, since div_tables is used internally to gdraw itself, don't put its header under `inc`, as it's private to gdraw. Just add the definition to `gdrawP.h` and remove `divisors.h`.

gioele:
hi @jtanx, I removed `divisors.h` and made the necessary adjustments. Also AppVeyor seems happy now.

frank-trampe:
This all looks okay to me. It achieves a significant reduction in code tangles with minimal disruption.

jtanx:
Not entirely sure it was necessary to rename `divisors.c` to `div_tables.c` but eh.

PR #3165, opened 2017-10-01
psread.c: dictfree: NULL entries after freeing

jtanx:
This prevents a double-free condition if the entry is also in the 'tofrees' garbage list.

### DIFF
--- a/fontforge/psread.c
+++ b/fontforge/psread.c
@@ -90,8 +90,10 @@ static void dictfree(struct pskeydict *dict) {
 
     for ( i=0; i<dict->cnt; ++i ) {
 	if ( dict->entries[i].type==ps_string || dict->entries[i].type==ps_instr ||
-		dict->entries[i].type==ps_lit )
+		dict->entries[i].type==ps_lit ) {
 	    free(dict->entries[i].u.str);
+	    dict->entries[i].u.str = NULL;
+    }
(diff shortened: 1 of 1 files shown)


fontforge/psread.c
```diff
@@ -90,8 +90,10 @@ static void dictfree(struct pskeydict *dict) {
 
     for ( i=0; i<dict->cnt; ++i ) {
 	if ( dict->entries[i].type==ps_string || dict->entries[i].type==ps_instr ||
-		dict->entries[i].type==ps_lit )
+		dict->entries[i].type==ps_lit ) {
 	    free(dict->entries[i].u.str);
+	    dict->entries[i].u.str = NULL;
+    }
```
frank-trampe: @gioele already merged this, but try to match the whitespace type in the future.

jtanx: @frank-trampe I'm not sure what you're referring to; the whitespace was matched to the surrounding code.

gioele: @jtanx: @frank-trampe is referring to the closing brace: it is out of place. I have pushed a fix, it was indeed misleading when seen in the surrounding code (this was one of those rare cases when context diff is better than unified diff).

Let's have a minute of silence in remembrance of all the hours mankind has wasted dealing with GNU-style indentation. :)

Issue #3162, opened 2017-09-30
Index in Title wont change resulting in navigation inconsistency to previous and next glpyhs 


Anonrate:
### FontForge and Operating System
Fontforge:
- Based on sources from 19:04 UTC 29-Sep-2017-ML-TtfDb-D.
- Based on source from git with hash: b07d93cd1a158b2f8288694140d1661464b9f0ef

Operating System:
- Debian Stretch with backport 12.1.6 kernel

### Issue
- [ ] <sup><b>[Issue 1]</b></sup> Index in Title of GlyphView stays the same as the first glyph opened and wont change when going to the next or previous glyph via `View > Next | Prev Glyph`, **BUT** if you save with the GlyphView open, close FontForge, reopen it, when going to the next or previous glyphs, the index changes in the title, but issue reocures if you close the GlyphView then open another glyph.
- [ ] <sup><b>[Issue 2]</b></sup> This correlates to issue *Issue 1*, when issuing command `View > Find In Font View`, the Glyph that is highlighted is that of what index is shown in the title, despite of what the actually current Glyph is.
- [ ] <sup><b>[Issue 3]</b></sup> If trying to open a glyph that has already been opened in to the GlyphView, it just switches to the tab that it was originally opened in, but wont actually change to the glyph you're trying to open again.
- [ ] <sup><b>[Issue 4]</b></sup> What happens in *Issue 1* is the same result that happens when navigating the CharInfo Dialog.

### Expected results
- [ ] <sup><b>[Issue 1]</b></sup> For the Index in the Title to correlate to that of the current Glyph so navigation is accurate when going to the next and previous glyphs.  <sup>See why I think this is happening.</sup>
- [ ] <sup><b>[Issue 2]</b></sup> For the correct glyph to be highlighted in the FontView when trying to *Find* it from the GlyphView.
- [ ] <sup><b>[Issue 3]</b></sup> For the correct Glyph to be opened into the GlyphView from the FontView, even if the Glyph has already been opened.
- [ ] <sup><b>[Issue 4]</b></sup> For the correct *Next* or *Previous* Glyph to be shown when navigating through the CharInfo dialog.

### Why do I think it's happening?
What all these issues have in common is that the index that is displayed in the Title, seems to be what determines what current Glyph is active, but internally for the most part, as the rest of the title corresponds to the active glyph.

[...]

PR #3167, opened 2017-10-04
This is a 'quickfix' for Issue 3162.

Anonrate:
I've been looking around to try and fix issue #3162, and so far that's what I've been able to come up as for a *QuickFix*.  The index in the title seems to finally correspond.   However, if you try to open a glyph that has already been opened, it will just goto the tab that is was opened in.  It wont switch to that particular glyph though.


jtanx:
From what I recall, there was some wordlist modification shenanigans happening around here: https://github.com/fontforge/fontforge/blob/467c1763911132d2506b0706016c1c223f894361/fontforgeexe/charview.c#L7489, something about it not doing the same thing as `CVChangeChar`

But I didn't take a close enough look to see what needed fixing.

Anonrate:
I've just recently started looking at the FontForge code, so I'm still getting use it and where everything is located.  My ctags doesn't seem to be saving the location for where types are defined and declared, so I have to manually find them.  It took me forever to find where `charview` was defined.

PR #3168, opened 2017-10-04
Make Travis build 25%+ faster (and cleaner)

gioele:
This PR divides the Travis builds in 3 different modes: compilation, distcheck and coverage.

The average build time is now around 7 minutes vs the average 10 minutes for the current Travis builds. The fastest build completes now in about 5 minutes.

The bottleneck remains the MacOS builds (up to 30 minutes before the MacOS VMs are even lauched these days), but these changes make my test cycle quite shorter anyway.

In the future I'd like to

1. use the default GCC (4.8.4). Not installing GCC 7 and the associated PPA would save about 1 minute per build. (How come GCC 4.9 and now GCC 7 are used? only for the colored diagnostics?)
2. cache the state after `./bootstrap` and `./configure`. This would save an additional 2 minutes per build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ os:
 
 env:
   global:
+    - MAKEFLAGS="-j4"
     # For Coverity
     - secure: "MeTS1Pqa5gzx1nn/peW/9a5kq84bba3XYUljOfkCUqzuyGiERk/nmok+RW7skrgzboBlKxnNG8+ykKqHMwK9s9M83ezFxvEWXBcKEpmEQKkqXPI5hpMs6jGLTgpeuheSIzqHA3danV8iircp1GOiTLWA0pt/AOsNLZiaYBh0OiE="
   matrix:
@@ -93,11 +94,11 @@ install:
         wget --tries 1 "http://download.savannah.gnu.org/releases/freetype/freetype-$FTVER.tar.gz" || \
           wget "https://sourceforge.net/projects/freetype/files/freetype2/$SFFTVER/freetype-$FTVER.tar.gz"
 
-        pushd zeromq-4.1.5 && ./configure --prefix=$DEPSPREFIX && make -j4 && make install && popd
-        pushd czmq-3.0.2 && CFLAGS="-Wno-format-truncation" ./configure --prefix=$DEPSPREFIX && make -j4 && make install && popd
-        pushd libspiro-0.5.20150702 && ./configure --prefix=$DEPSPREFIX && make -j4 && make install && popd
-        pushd libuninameslist-20160701 && ./configure --prefix=$DEPSPREFIX && make -j4 && make install && popd
-        pushd libunicodenames-1.1.0_beta1 && ./configure --prefix=$DEPSPREFIX && make -j4 && make install && popd
+        pushd zeromq-4.1.5 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
+        pushd czmq-3.0.2 && CFLAGS="-Wno-format-truncation" ./configure --prefix=$DEPSPREFIX && make && make install && popd
+        pushd libspiro-0.5.20150702 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
+        pushd libuninameslist-20160701 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
+        pushd libunicodenames-1.1.0_beta1 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
         tar -zxf freetype-$FTVER.tar.gz -C $DEPSPREFIX
         popd
       fi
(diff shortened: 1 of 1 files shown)


jtanx:
At the moment, making the build time shorter is moot, because of that Mac build; it's only as fast as the slowest build. Paradoxically, in its current form, this PR increases the number of build jobs to 4 linux builds and 1 Mac build, which *increases* the build time, if other builds are queued - there's a limit of 5 concurrent jobs over all the active requests.

.travis.yml
```diff
@@ -134,15 +136,12 @@ install:
...
-  - ./bootstrap
-  - ./configure $FFCONFIG
-  - make -j4
-  - make -C contrib/fonttools -j4
-  - make install
-  - if [ ! -z "$TRACE_CODE_COVERAGE" ]; then make check ; coveralls --gcov-options '\-bulp' --gcov $(which $GCOV) ; fi
-  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then make distcheck -j4; fi
-  - fontforge -version
-  - $PYTHON -c "import fontforge; import psMat; print(fontforge.__version__, fontforge.version()); fontforge.font();"
+  - if [ -n "$MODE" ] ; then ./bootstrap && ./configure $FFCONFIG ; fi
```
jtanx: Why is this conditional on $MODE being nonempty?

.travis.yml
```diff
@@ -15,29 +15,30 @@ os:
...
-    - os: linux
-      env: OSX_ONLY=true
+      env: MODE=coverage
 
 addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
       - deadsnakes
-    packages: [autoconf, automake, autotools-dev, bzip2, gcc-7, libtool,
```
jtanx: This was like so, to act as a guide as to what packages are required to build FontForge.

.travis.yml
gioele: `MODE` is unset when we want to skip the whole build, for example when in coverage mode but coveralls is not set up (see commit c1bfb98abdbb94c0b274f9e17617fdc7def108cc).

gioele:
> Paradoxically, in its current form, this PR increases the number of build jobs to 4 linux builds and 1 Mac build, which increases the build time, if other builds are queued - there's a limit of 5 concurrent jobs over all the active requests

Well, we can always remove the `--without-x` build and go back to 4. ;)

One of your recent commits removed the standard build with support for X11 on Linux. I reintroduced it because I thought it was strange not to build the most used part of FontForge. :)

jtanx:
> One of your recent commits removed the standard build with support for X11 on Linux. I reintroduced it because I thought it was strange not to build the most used part of FontForge. :)

The x11 build is covered by the coverage build, so although on paper it may take longer to complete, it's doing more for that sole build.

gioele:
> The x11 build is covered by the coverage build, so although on paper it may take longer to complete, it's doing more for that sole build.

The problem with conflating coverage testing and the full build is that coverage is going to fail (gracefully or vociferously) for all users that have no coveralls account. That means no feedback on the full build for anybody that does not use the main repo to develop (e.g.… me :)).

.travis.yml
gioele: Sounds reasonable and costs only a couple of seconds. Then please exclude commit d096e0e0ad627af730a7869aa204a7074a501801 during the merge.

.travis.yml
```diff
@@ -134,15 +136,12 @@ install:
...
-  - make -C contrib/fonttools -j4
-  - make install
-  - if [ ! -z "$TRACE_CODE_COVERAGE" ]; then make check ; coveralls --gcov-options '\-bulp' --gcov $(which $GCOV) ; fi
-  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then make distcheck -j4; fi
-  - fontforge -version
-  - $PYTHON -c "import fontforge; import psMat; print(fontforge.__version__, fontforge.version()); fontforge.font();"
+  - if [ -n "$MODE" ] ; then ./bootstrap && ./configure $FFCONFIG ; fi
+  - if [ "$MODE" == "compilation" ] || [ "$MODE" == "coverage" ] ; then make && make -C contrib/fonttools && make install ; fi
+  - if [ "$MODE" == "coverage" ] ; then make check ; coveralls --gcov-options '\-bulp' --gcov $(which $GCOV) ; fi
+  - if [ "$MODE" == "distcheck" ] ; then make distcheck ; fi
```
jtanx: Doing a distcheck build separate from 'compilation' is pointless, because to do a distcheck, it needs to run the build first anyway.

jtanx:
> The problem with conflating coverage testing and the full build is that coverage is going to fail (gracefully or vociferously) for all users that have no coveralls account. That means no feedback on the full build for anybody that does not use the main repo to develop (e.g.… me :)).

I'm still not fully convinced that this is a good way to go about solving this problem. You could do something like:

```
coveralls --gcov-options '\-bulp' --gcov $(which $GCOV) || true
```

Supposing this fails on FF, then that's fine, it'll eventually get picked up by coveralls because no coverage data will have been sent to it, marking the build as 0% covered (and failing the build)

.travis.yml
gioele: @jtanx? distcheck does not need nor does run the build first...

The first thing it does is creating the dist tarball, then untar it and only then run the full compilation cycle inside the release directory. Only `./configure` and `make check` is run twice.

You can check that no build is done before creating the tarball in the logs of one of those distcheck-only builds: <https://api.travis-ci.org/jobs/283364498/log.txt> and <https://travis-ci.org/fontforge/fontforge/jobs/283364498>.

Removing distcheck from the main build allows the main build to pass/fail faster. Same for the distcheck build itself.

jtanx:
Superseded by #3171. Feel free to push changes onto that branch.

PR #3170, opened 2017-10-04
Remove unused file utils.c

gioele:
The file `fontforge/utils.c` has been unused and basically empty since its introduction in 2004.

The redundant struct definition in it has been introduced in commit c397c35288685a9424e4e1aeffb81143a397a88f and commented out 20 minutes later in commit f611c0c351fec9790ea604eb661e0f88b77d5034.


frank-trampe:
@gioele, it was not introduced in that commit. It had been present before. Reuben had made a lot of mass-scripted commits, but I was unable to review the 30-thousand lines or so of changes, so we ended up reverting many (but not all) of them and then reapplying the ones that were worth verifying. This one was apparently orphaned by one of the unreviewed changes that we failed to revert.

There was a lot of good code broken during the unmanaged years (2012-2014, roughly), and the fact that a reference to a piece of historical code was broken during that time is not a good indicator of whether it is unnecessary. Much more productive than merely finding ten unused functions would be finding one such function and determining when it was orphaned and why. Then we can either restore it to its former use or verify that it is in fact unnecessary.

gioele:
OK, let's go through the history of the never used `utils.c`. Pro bono, no consulting fees. :)

* 2004 (7fdad80fae7ef3fe29475a3af22681f98b4c4e21): the initial revision from GW contains a basically empty `utils.c`, GW probably created it with the intent to add utility functions there but forgot about it.

* 2005-2012 (d1c0a6f048493f3e015ae0fad0e5ead261f9535a, 0a7920209cb11022f7b2a165b9b38255ca4443d5, e7f8e9d9656a55af8de2391baf20079f8ad48def, 3a842df07dada4494d17bc58387eb14de5fdd4ed, 8f8e0fe04030d16f4fabc9f7f8487803f5ee81e2, e32266e73a25441fba6118dee205e7d03b3a67b8, 5beb4b7b0508f35f130a622dbe8d30c236af58e6, 54fcb78b586cfd139a4febe1e18c870dee1bd578, 4ba25b662da85534a0aa2ab9ae53384fcc9a6cd9): GW updates the year in the copyright statement and changes the name into `fontforge` from `pfaedit`, nothing else is changed.

* 2014 (e05e5254b77fd5471a673e75a347cbe80d9ec0b1, 7ee9023b84c74761c170dfcd8673b305d549e016): Ruben just removes the files, Frank puts it back. No changes in content. The file is still empty.

* 2014-05-29 16:40 (c397c35288685a9424e4e1aeffb81143a397a88f): Frank adds some code to utils.c...

* 2014-05-29 17:00 (f611c0c351fec9790ea604eb661e0f88b77d5034): ... only to comment it out with`#if 0` about 20 minutes later, noting that it was broken. (The reason for the brokenness is that it duplicates a struct that is already present in other UI-related files.)

* Nothing is changed until this very day.

I know that the FF history is quite tangled and it is worth preserving, but in this case there is absolutely nothing in this file. We are just wasting electricity parsing headers and producing a empty object file. :)

jtanx:
Okay, so it was removed in 03fc77a871c3e509a96f95180fcfe27aa67b9e08, then reintroduced with c397c35288685a9424e4e1aeffb81143a397a88f, but the definition was moved from exelibstamp.c to utils.c

frank-trampe:
The problem was not that it duplicated code from gtkui. As far as I know, we don't build that. The problem was that the declaration of struct library_version_configuration was missing. That also makes libstamp unusable. The value of checking compatibility between library and executable is not entirely clear to me in this age of package management. If we need that, we ought to fix the underlying problem.

frank-trampe:
I have no memory of library version mismatches (the original reason for the now-broken code) being problematic, and I don't see a significant likelihood of them occurring in the future given that most of our users install monolithic bundles or version-managed packages. Obviously, they would cause spurious crashes if they happened and would be hard to trace, but I don't see how the underlying condition would happen these days. @jtanx, @JoesCat, do you recall any recent problems of this sort?

JoesCat:
@gioele,
You looked at the file, you checked out the history, you saw that this file seems to have no current use, and appears to not be anything that has potential use or appears to be in midflight (partially built but needs improving).
I'm with your decision here, let's remove it, and thanks for doing all the homework. Excellent!

In terms of the rest of this thread...
For myself, I somewhat use 20120731 as a homebase or reset-point for the code I'm looking at, because like @frank-trampe indicates, code had changed dramatically from that point, some we can agree with, others not, but it is what it is ATM. So... ATM, keep what makes sense, fix what needs fixing.

In terms of the garbage collector,
I did not realize it was such a massive edit (30-thousand lines?). Fontforge code made use of two sorts of memory access https://github.com/fontforge/fontforge/blob/v20120731-b/inc/basics.h we all know about alloc(), but then there is also galloc() (and friends), this memory space is a reusable memory space that avoids time penalties and increasing memory fragmentation. This works great in smaller/slower machines under a single user, I'm not sure how this behaves under multithreading and possible multiusers (https://en.wikipedia.org/wiki/Multiseat_configuration). Pre-20120731, all Fontforge code was mastered by GW, post-20120731, the rest of us jumped in and lacking documentation, etc, we tried what we could... and I'm sure there are probably a couple cases where we mixed alloc() vs galloc() methods, if not already mixedup earlier 8-(

...we eventually turned-off galloc() to behave like alloc().

Lack of documentation is one thing, but another notable thing about Fontforge code is plenty of alloc(), but very few free(), this works okay on a DOS computer, or a computer you tend to shut-off at the end of the day, but if you run a computer 24x7, you can see lost and leaked memory left behind whenever you use Fontforge. The garbage collector was a fix to try and attempt to clean house, but I don't think Fontforge is something you can use a garbage collector on. In my opinion, I think it is best to setup memory at the beginning of a process, and clean-house at the end of a process to avoid leaked memory. Absolutely avoid longjumps, but you can safely use goto fairly effectively to avoid code clutter as long as you "goto" within the same function (if you want examples, several gutil functions have goto, and there's a couple big functions in the fontforge lib directory, just grep "goto" to find them). This memory leak problem is easier if you take a bottom-up approach, starting with single functions, then move-up to functions calling those functions. Doing a top-down approach, you'll quickly get lost in the forest. If we continue to clean-house within function processes, garbage-collection and leaked memory becomes an non-issue later (Don't forget to handle realloc() properly).

In terms of libstamp,
When users began noticing the Fontforge development activity post-20120731, the bugs and issues mushroomed (currently 800+ now), libtimestamp wasn't enough, and a buildhash helped a lot in weeding-out bugs.
The new configure.ac also reset all libraries to "2.0" pre-20140101, which also added to the confusion.
Even though a build hash exists, we still need to remember to update the version number of the various libraries when we generate new releases.
Versioning is found at the top of the configure.am so it's all in one place. Fixes and improvements to library time stamps likely needs attention to configure.ac, Makefile.am, config.h, plus the various other locations where you'll find more stamps, however, nobody has really paid deep attention to correctly versioning the libraries "correctly".
"Correctly" meaning, go up 0.1 if you remain backwards compatible, or go up 1.0 if you break backwards compatibility.
@lemzwerg pointed-out an error how timestamps was generated in libuninameslist https://github.com/fontforge/libuninameslist/issues/7, this has already been also added into libspiro, but I don't think it's been added yet into Fontforge.

I don't know what the best solution is here... go with one big massive library, or continue with several specialized libraries - each direction has tradeoffs.... better to stay one method than jump back-n-forth.

@frank-trampe, we haven't had recent timestamp problems since we fixed the "2.0" timestamp error, but this may be masked by larger problems ATM, and most users bringing up issues ATM tend to compile from source. I haven't paid much attention ATM, but we still should add @lemzwerg fix into Fontforge if not already added.

frank-trampe:
@JoesCat, this code only existed, as far as I can tell, to check for a version mismatch between the FontForge executable and the FontForge library, which would obviously cause odd problems if it were to happen. The question is whether such a thing (somebody reporting a crash due to mismatched executable and library versions) would ever plausibly happen given how we distribute FontForge. If so, I'll try to get it working again. If not, I'd like to get this pull request merged.

JoesCat:
> The question is whether such a thing (somebody reporting a crash due to mismatched executable and library versions)

There were a couple of issues reported pre/post2013dec, some titles won't match because we did not really know what the problem was at the time, but when we added [Library version values upgraded accordingly ](https://github.com/fontforge/fontforge/commit/d0826fc152794539f7ada3758099d2da46a6ba0d), this sort of problem faded away.

+1 merge.

frank-trampe:
Good enough for me.

PR #3171, opened 2017-10-05
Fix travis builds

jtanx:
This supersedes #3168.

We always want to run `make check`, which runs the test suite against that matrix build's configured/built version.

Picked up from #3168, we only run distcheck on one of the branches, which means that the nox build is sped up to around 7-9 minutes (depending on if the build cache is present). That's good enough for me, if you're after a quick indicator of if the full build will fail (actually, since they more or less do the same thing up until the distcheck step, then they're likely to fail at around the same time into the build).

Yes, the full build is still around 15 minutes, but I'd say having two linux builds vs. four is better, especially given the concurrency limits. We can revisit this if ever it becomes slower than the mac build.

I've also separated out the 'upload coverage' step to the `after_success` event, and made it always succeed, so it works when built on forked travis builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ os:
 
 env:
   global:
+    - MAKEFLAGS="-j4"
     # For Coverity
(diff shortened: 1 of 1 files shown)


gioele:
I like that we will always run `make check` from now on. Safety first. :)

PR #3176, opened 2017-10-08
Honor SOURCE_DATE_EPOCH in more cases

### LINKED ISSUE #2711: Please do not put timestamps in generated files

gioele:
This PR is an extension of the merged PR #3152, making sure that SOURCE_DATE_EPOCH is used everywhere the current time is serialized in a file.

@volth, @bmwiedemann: could you please test this to see if further reduces or eliminate problems with reproducible build on Linux?

@volth: The MacOS issues will need to be taken care of separately.

WRT Author names, `GetAuthor` already honors the user-configurable env variable `USER`.

Closes #3134
Fixes #2711

### DIFF
--- a/contrib/fonttools/Makefile.am
+++ b/contrib/fonttools/Makefile.am
@@ -35,6 +35,8 @@ noinst_PROGRAMS = findtable pfadecrypt rmligamarks stripttc woff	\
 
 woff_LDADD = $(ZLIB_LIBS)
 dewoff_LDADD = $(ZLIB_LIBS)
+ttf2eps_LDADD = $(top_builddir)/Unicode/libgunicode.la \
+                $(top_builddir)/gutils/libgutils.la
 
 EXTRA_DIST = README-ttf2eps
 
--- a/fontforge/macbinary.c
+++ b/fontforge/macbinary.c
@@ -1423,7 +1423,6 @@ static int DumpMacBinaryHeader(FILE *res,struct macbinaryheader *mb) {
 
 	/* Creation time, (seconds from 1/1/1904) */
     now = mactime();
-    time(&now);
     *hpt++ = now>>24; *hpt++ = now>>16; *hpt++ = now>>8; *hpt++ = now;
 	/* Modification time, (seconds from 1/1/1904) */
     *hpt++ = now>>24; *hpt++ = now>>16; *hpt++ = now>>8; *hpt++ = now;
--- a/fontforge/macbinary.c
+++ b/fontforge/macbinary.c
@@ -51,8 +51,8 @@
 #include <math.h>
 #include <sys/types.h>
 #include <sys/stat.h>
-#include <time.h>
 #include <ustring.h>
+#include <gutils.h>
 #include "ttf.h"
 #include "psfont.h"
 #if __Mac
@@ -1359,9 +1359,8 @@ static void DumpResourceMap(FILE *res,struct resourcetype *rtypes,enum fontforma
 }
 
 long mactime(void) {
-    time_t now;
+    time_t now = GetTime();
 
-    time(&now);
     /* convert from 1970 based time to 1904 based time */
     now += (1970-1904)*365L*24*60*60+((1970-1904)>>2)*24*60*60;
     /* Ignore any leap seconds -- Sorry Steve */
--- a/fontforge/print.c
+++ b/fontforge/print.c
@@ -1157,13 +1157,17 @@ static void dump_pdfprologue(PI *pi) {
     fprintf( pi->out, "  /Creator (FontForge)\n" );
     fprintf( pi->out, "  /Producer (FontForge)\n" );
     now = GetTime();
-    tm = localtime(&now);
+    if (!getenv("SOURCE_DATE_EPOCH")) {
+	tm = localtime(&now);
+    } else {
+	tm = gmtime(&now);
+    }
     fprintf( pi->out, "    /CreationDate (D:%04d%02d%02d%02d%02d%02d",
 	    tm->tm_year+1900, tm->tm_mon+1, tm->tm_mday, tm->tm_hour, tm->tm_min, tm->tm_sec );
 #ifdef _NO_TZSET
     fprintf( pi->out, "Z)\n" );
 #else
-    if ( timezone==0 )
+    if ( timezone==0 || getenv("SOURCE_DATE_EPOCH") )
 	fprintf( pi->out, "Z)\n" );
     else {
 	if ( timezone<0 ) /* fprintf bug - this is a kludge to print +/- in front of a %02d-padded value */
--- a/fontforge/print.c
+++ b/fontforge/print.c
@@ -39,6 +39,7 @@
 #include "splinesaveafm.h"
 #include "splineutil.h"
 #include "tottf.h"
+#include <gutils.h>
 #include <stdlib.h>
 #include <math.h>
 #include <time.h>
@@ -1155,7 +1156,7 @@ static void dump_pdfprologue(PI *pi) {
 	fprintf( pi->out, "  /Title (Character Displays from %s)\n", pi->mainsf->fullname );
     fprintf( pi->out, "  /Creator (FontForge)\n" );
     fprintf( pi->out, "  /Producer (FontForge)\n" );
-    time(&now);
+    now = GetTime();
     tm = localtime(&now);
     fprintf( pi->out, "    /CreationDate (D:%04d%02d%02d%02d%02d%02d",
 	    tm->tm_year+1900, tm->tm_mon+1, tm->tm_mday, tm->tm_hour, tm->tm_min, tm->tm_sec );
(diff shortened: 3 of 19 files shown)


contrib/fonttools/Makefile.am
```diff
@@ -35,6 +35,8 @@ noinst_PROGRAMS = findtable pfadecrypt rmligamarks stripttc woff	\
 
 woff_LDADD = $(ZLIB_LIBS)
 dewoff_LDADD = $(ZLIB_LIBS)
+ttf2eps_LDADD = $(top_builddir)/Unicode/libgunicode.la \
```
jtanx: Seems a bit heavy to add a dependency on these dlls for just the one function (and considering the standalone nature of the rest of the contrib tools)

fontforge/macbinary.c
```diff
@@ -1423,7 +1422,6 @@ static int DumpMacBinaryHeader(FILE *res,struct macbinaryheader *mb) {
 
 	/* Creation time, (seconds from 1/1/1904) */
     now = mactime();
-    time(&now);
```
jtanx: What's up with this?

fontforge/print.c
```diff
@@ -1155,23 +1156,11 @@ static void dump_pdfprologue(PI *pi) {
...
     fprintf( pi->out, "  /Producer (FontForge)\n" );
-    time(&now);
-    tm = localtime(&now);
+    now = GetTime();
+    tm = gmtime(&now);
     fprintf( pi->out, "    /CreationDate (D:%04d%02d%02d%02d%02d%02d",
 	    tm->tm_year+1900, tm->tm_mon+1, tm->tm_mday, tm->tm_hour, tm->tm_min, tm->tm_sec );
-#ifdef _NO_TZSET
     fprintf( pi->out, "Z)\n" );
-#else
```
jtanx: Why is this removed?

fontforge/macbinary.c
gioele: Oops, I forgot to add the background story in the commit. :)

That `time(&now)` has been added by GW during a mass commit. If the comment is right then using `time` there is plainly wrong, as `time` counts from 1970. On the other hand, `mactime` counts from 1904, just like the comment says. In addition, in the following line the modification time is also stored and there only `mactime` is used.

I trust the comment and the overall context, thus removed that `time`. However I have put it in a separate comment, so that it will be easy to bisect should any problem arise.

contrib/fonttools/Makefile.am
gioele: That dependency isn't already there only because there is a lot of code copied and pasted in `ttf2eps`. :) Same for `showttf`

We really should go though it and remove all the duplicated and outdated code. This would leave only the core ttf2eps logic there. That is the point of a `libfontforge`, isn't it? ;)

fontforge/print.c
gioele: That part of the code was there only to calculate and output the timezone offset. With no timezone there is no need to calculate it. ;)

BTW, I think it is broken on MacOS (can't remember the details, I had a brief look before realizing it could just be deleted).

frank-trampe:
Why is it necessary to drop timezone support?

bmwiedemann:
I think the plan is to always use UTC everywhere so that users with different timezone settings can still get the same result from fontforge calls.

gioele:
hi @volth, hi @bmwiedemann, I added some code to make sure that the env variable `USER` is preferred to GECOS fields if `SOURCE_DATE_EPOCH`. I cannot see any other externally controlled source of non-reproducibility. Does this work for you?

I did quite a bit of testing and everything seems to work fine when `SOURCE_DATE_EPOCH` is set as well as when it is not set.

Any objections to merging this?

volth:
I am not sure about Debian, but NixOS does not a set fake stable ```$USER``` for the build, and I anticipate the need to forge ```$USER``` may meet some objections in the community

gioele:
What does the build system of NixOS does when an author/user _must_ be set? Does it default to something?

Or is there an equivalent `SOURCE_DATE_EPOCH` for `USER`? Something like `SOURCE_USER`? If such a thing exist we can easily support it.

volth:
> What does the build system of NixOS does when an author/user must be set?

I guess it is something very new and fontforge-specific. Are the other build tools besides ```fontforge``` with the concept of "```author/user must be set```" ?

gioele:
> > What does the build system of NixOS does when an author/user must be set?
>
> I guess it is something very new and fontforge-specific. Are the other build tools besides fontforge with the concept of "author/user must be set" ?

Too many. :) tar, for example, must put something in the user/group of each file it records in a tar ball. Git/SVN/Hg/CVS also require you to have an user/email set and will store it in the metadata files.

I think that the more general problem (that surely has already been dealt with by NixOS) is: "How do you deal with env variables that end up being encoded in a compiled file?".

frank-trampe:
We can't drop timezone support without getting feedback from a lot of people. Let's either trigger UTC output when SOURCE_DATE_EPOCH is set or, if that doesn't work for some reason, set another environment variable like TIMESTAMP_TZ.

gioele:
hi @frank-trampe, you have a point: there is a slight chance of things breaking because of the missing timezone (I can picture a script that checks against the local time using strings instead of proper time routines). Assuming sloppy code out there seems a sensible default. ;)

I rebased the branch against master, reverted most of the commit that forced UTC and added some code to conditionally use UTC if `SOURCE_DATA_EPOCH` is set.

contrib/fonttools/Makefile.am
jtanx: True, but I can also see the reasoning; they have no dependencies on anything else, so they can be shipped independently of FF itself. 

For example, if I'm reading this right, the [fontforge-extras](https://packages.debian.org/sid/fontforge-extras) package ships with no dependency on fontforge itself.

frank-trampe:
@gioele, I'd like to rebase (perhaps in a new branch and pull request) with the "Output UTC" and its revert dropped. I can do it, but it's easier for you to do since you own the branch.

gioele:
hi @frank-trampe, I rebased the branch and squashed away the "Output UTC" commit/revert commit.

h01ger:
the implemented logic and decissions look good to me - though I'm not familar with this code base :)

Thank you!

gioele:
hi @frank-trampe and @jtanx, are there any outstanding objections to merging this PR?

frank-trampe:
Go ahead!

frank-trampe:
Or I'll do it.

contrib/fonttools/Makefile.am
frank-trampe: I forgot to comment here, but our repackaging dropped fontforge-extras. I'm hoping that Debian will follow suit.

gioele:
Well, you've been faster. Thanks :)

PR #3181, opened 2017-10-12
Remove support for HTTP and FTP

### LINKED ISSUE #3175: HTTP support broken since (at least) 2014: remove?
I was working on replacing the insecure (at least 2 serious Coverity reports) and outdated (no HTTPS) `fontforge/http.c` file with a curl-based implementation when I noticed that, basically, support for downloading HTTP files has not been working since, at least, the 20141230 release.

Should we just remove the support for HTTP and FTP?

Arguments pro removal:

* It does not support HTTPS. Good luck finding a website that is serving files over plain HTTP. (I had to use my own HTTP server to test the original code.)
* It has not been working for many years and many releases (at least since the 20141230 release, probably also earlier).
* It is highly insecure: a malicious server could force an exploitable crash by sending malformed headers/content.
* downloading files does not belong to the library, the application should take care of that (trivial in python, the Linux binary could use GIO, the MacOS will probably have some native easy way of doing it).

To test try the following

    $ echo 'Open("http://test.svario.it/mingliu.ttc");' | fontforgeexe/fontforge -lang ff -script -

(The remote file will disappear in few weeks.)

gioele:
This PR removes the (since long broken) code for HTTP and FTP and the totally unused `gioftp` library. As discussed in issue #3175,

* The current code does not support HTTPS. Good luck finding a website that is serving files over plain HTTP. (I had to use my own HTTP server to test the original code.)
* It has not worked properly for many years and many releases (at least since the 20141230 release, probably also earlier).
* It is highly insecure: a malicious server could force an exploitable crash by sending malformed headers/content.
* Downloading files is a task that does not belong to a library, the application should take care of it.

There is also a new small test to make sure that the opening files using `file://` URLs works fine and that future changes will not break that. Thanks @jtanx for pointing it out that `file://` must be supported.

Side note: this code (as well as the old code) allows relative `file://` URLs that, in theory, are forbidden by the relevant RFCs. I think that being a bit loose is OK in this case.

Closes #3175. (And a dozen of Coverity issues, see the commit logs.)


tests/testsuite.at
```diff
@@ -105,6 +107,7 @@ check_pedit([Test generation of TrueType instructions],[test127.pe],[Ambrosia.sf
 check_pedit([Test StrSplit overrun],[test128.pe])
 check_pedit([Ligatures and Fractions Table Lookups],[test129.pe])
 check_pedit([TTC loading],[test130.pe],[NotoSans-Regular.ttc])
+check_pedit([file:// protocol],[test131.pe],[${abs_srcdir}/fonts/Ambrosia.sfd])
```
jtanx: Why doesn't passing `[Ambrosia.sfd]`  work?

jtanx:
Is this a full removal? I still see references in gdraw and gio (e.g. `giohosts.c`)

tests/testsuite.at
gioele: `file://` URLs are supposed to be always absolute, see <https://tools.ietf.org/html/rfc8089#page-4>. File managers will also always pass URLs with absolute paths.

Given that "support for `file://`" means just stripping the path, in practice a simple relative path would work, but just by accident. :)

jtanx: But the whole `check_file` macro ensures that it always gets transformed to an absolute path

gioele:
This is a removal of the features exposed by the `http.c` file and `gioftp.c`.

Removing the unused parts of GIO is also on my TODO list, but more invasive and will requires more careful testing. Let's start with the low hanging fruits. ;)

tests/testsuite.at
gioele: @jtanx: you are absolutely right. Somehow at a certain point I needed that modification (forgot why :confused:), but, indeed, it is not needed now. I'll edit the commit after the review.

frank-trampe:
Per my comment on the issue, I'd like to pause any non-trivial non-constructive pull requests for a few months so that we can focus on open issues. In an ideal world, I'd be able to do all of those things and spend lots of time reviewing other pull requests, but work has been unprecedentedly busy. We've essentially sidelined all efforts to improve the user experience for several months now, and I'm under pressure to work through some of that backlog.

jtanx:
Let's get the ball rolling on this, as part of #3512.

jtanx:
Note that I'm targeting a merge for this this weekend.

frank-trampe:
Okay.

PR #3183, opened 2017-10-14
More tests: Compare output of exported fonts

gioele:
This PR adds some regression tests that make sure that fonts are properly exported. This will give us confidence, for example, that we will not be breaking generated fonts while fixing other issues. :relieved: 

These tests cover many parts of the code for which we had zero test coverage before (see the Coveralls report).

This PR introduces a new class of test in `testsuite.at`: content-based tests. In these test the output of a generated file is compared against a reference known-good output,  produced with `SOURCE_DATE_EPOCH=1483326245` (= 2017-01-02 03:04:05 UTC) and `USER=FontForgeTestsuite`.

Other similar tests for other parts of the code will follow in the future, but let's get the foundations right first with these small examples. ;)

Please note that this PR is based on the branch `gioele:reproducible-time-author` (see PR #3176) because comparing the output requires the output to be bit-exact/reproducible. The commits to review are, thus, only 4d08e4adbc457b7eef61ce4f72b83f2abfc7d3de, fea19abe793baad4fe1389b717df13f74b50257d, 23d325a01f2bf4003c92f82cf0ce3ee242e1c21e, c0b9c7aa319e53d2137429bc07f9ec6f4f9d525c, 901f098e5f10db655c1af257c0d0286650d18cc8.


frank-trampe:
@gioele, are the failed tests a reflection of things broken in FontForge?

gioele:
Indeed, these tests have already uncovered a couple rounding errors in the MacOS code for EPS export.

```diff
--- f-Ambrosia-g-k.eps  2017-10-15 07:07:48.000000000 +0000
+++ /Users/travis/build/gioele/fontforge/tests/references/f-Ambrosia-g-k.eps    2017-10-15 06:52:54.
000000000 +0000
@@ -33,7 +33,7 @@
 %0008FFFFE3333333333333210000000000
 %0008FFFFFFFFFFFFFFEFFFFFC700000000
 %0008FFFFFFFFFFFD61016EFFFFC1000000
-%0008FFFFFFFFFF81000002DFFFFA000000
+%0008FFFFFFFFFF81000002DFFFFB000000
 %0008FFFFFFFFE3000000003FFFFF400000
 %0008FFFFFFFD200000000009FFFF900000
 %0008FFFFFFC1000000000005FFFFB00000
@@ -41,9 +41,9 @@
 %0008FFFFE200000000000007FFFFC00000
 %0008FFFFE00000000000000BFFFFA00000
 %0008FFFFE00000000000003FFFFF700000
-%0008FFFFE0000000000000CFFFFF300000
-%0008FFFFE0000000000008FFFFFC000000
-%0008FFFFE000000000004FFFFFF4000000
+%0008FFFFE0000000000000BFFFFF300000
+%0008FFFFE0000000000006FFFFFC000000
+%0008FFFFE000000000002EFFFFF4000000
 %0008FFFFE00000000001DFFFFF90000000
 %0008FFFFE0000000000AFFFFFC10000000
 %0008FFFFE0000000008FFFFFC100000000
@@ -51,9 +51,9 @@
 %0008FFFFE00000003EFFFFC10000000000
 %0008FFFFE0000002DFFFFB100000000000
 %0008FFFFE000000CFFFFA0000000000000
-%0008FFFFE000008FFFF900000000000000
-%0008FFFFE00004FFFF8000000000000000
-%0008FFFFE0001DFFFD0000000000000000
+%0008FFFFE000009FFFF900000000000000
+%0008FFFFE00005FFFF8000000000000000
+%0008FFFFE0001EFFFD0000000000000000
 %0008FFFFE000AFFFFE1000000000000000
 %0008FFFFE005FFFFFF8000000000000000
 %0008FFFFE00DFFD9EFE100000000000000
@@ -77,7 +77,7 @@
 %0008FFFFE00000000000000001DFFFFFA0
 %000AFFFFF000000000000000004FFFFF90
 %000DFFFFF4000000000000000005FFFF80
-%007FFFFFFC1000000000000000003DFF80
+%007FFFFFFC1000000000000000003DFF70
 %0CFFFFFFFFE40000000000000000016C60
 %%EndPreview
 %%EndProlog
```

jtanx:
I'd say it's more likely that FontForge is relying on undefined behaviour which clang is doing differently to gcc.

jtanx:
Actually, on float rounding/arithmetic, I think that kind of depends on the optimisation level. So I'm not entirely sure if this is a bug, or if it's just because the test is really brittle.

frank-trampe:
EPS isn't terribly important, but we ought to strive for accurate round-tripping. More importantly, we can't enable a check that fails unless we plan to fix the problem soon. Otherwise, checking for regressions via the automatic Travis build becomes a lot more difficult.

jtanx:
I can reproduce this. 

This is from a Sierra VM:
[a.zip](https://github.com/fontforge/fontforge/files/1386853/a.zip)

This is from Windows, with freetype 2.8.1:
[g-k-f-Ambrosia-gu-006b006B-%.zip](https://github.com/fontforge/fontforge/files/1386970/g-k-f-Ambrosia-gu-006b006B-.zip)

Note that for Windows, I had to do `diff -uZ` (ignore trailing whitespace) due to EOL differences (CR LF vs LF). There's also other tests failing on Windows, but I don't think it's related; I'll look into that later.

The difference is in the rasterisation; I wouldn't be surprised it was caused by using a different FreeType version than the reference was generated on. So the test is over-constrained.

gioele:
@jtanx: thanks for testing it again and pointing out the newline problem, I'll fix it in a later update to the branch.

First the direct question: Should I remove the failing test and move it to a separate issue?

Now some more elaborate comments about your and @frank-trampe's remarks:

* About adding failing tests. If this branch is to be integrated I'd push only working tests and attach failing test to a separate issue. I agree that in normal circumstances Travis should always succeed.

* Whether the test is too strict. That's the whole point of it! ;) Joking apart, one could come up with a much complex testing harness that can deal with slight differences, but do we want it?

I think that it is important to deliver cross-platform bit-exact results. That means: using all rasterization API in the proper way, setting the proper compiler flags and avoiding numerically instable code. A slight misalignment between MacOS and Windows could have an avalanche effect in browsers where a lot of sub-pixel calculations are done. At the same time, these amplified rendering mistakes could go unnoticed for a long time. Designers (mostly on the MacOS side) will never know that users (mostly on the Windows side) see something completely distorted on their screens.

In this very case, I think that there is just some funny code in the float-to-int-to-hex code that the EPS exporter uses. If the error was caused by the FreeType code, then why aren't the PDF and the SVG tests also failing in the same spot?

jtanx:
> Whether the test is too strict. That's the whole point of it! ;) Joking apart, one could come up with a much complex testing harness that can deal with slight differences, but do we want it?

If the test is always so strict that it fails most of the time, then I question the usefulness of that test.

> I think that it is important to deliver cross-platform bit-exact results.

This only makes sense when all external variables are fixed - including that all your external dependencies are the same. Across platforms, and outside of a made-to-purpose reproducible build environment, that's simply not true.

> In this very case, I think that there is just some funny code in the float-to-int-to-hex code that the EPS exporter uses. If the error was caused by the FreeType code, then why aren't the PDF and the SVG tests also failing in the same spot?

I think there's some confusion here; in the EPS format, FF is rasterising the glyph using FreeType and hex dumping that in the preview block. I take it that SVG and PDF outputs don't perform any such rasterisation and hence are not even touching FT2 at all. I'm guessing that getting the output the same would require the compiled version of FF to use the same version of FT2. But at the end of the day, I don't think anyone really cares about the pre-rendered preview being different - which is why I'm saying that test is over-constrained.

Note however I did see the PDF output being different on Windows, but that's likely a separate issue or cause.

jtanx:
Ah, yeah I think I know why the PDF output is different, it's due to the CRLF ending throwing out the size parameters. Noted, the release Windows builds of FontForge are [patched](https://github.com/jtanx/fontforge/tree/2017-august-release) (have been patched for 2 years!) and specifically [here](https://github.com/jtanx/fontforgebuilds/blob/master/patches/gnulib.patch), so that all files are read/written in binary mode, which makes it more consistent with Linux (LF endings instead of CRLF).

I should probably do something about trying to get most of those patches back in/reduce the number of patches necessary.

ctrlcctrlv:
At this point...this no longer can be merged without a conflict, it's years old, and there are questions as to whether it's really needed. So I'm closing this as it is unlikely to ever be merged, primarily due to its age and all the conflicting files. It can be reopened if interest in it is ever renewed.

gioele:
I would be a bit hesitant to drop this PR. Fontforge badly needs a testing mechanism, especially in the form of data-driven tests. Have other forms of comparison-based testing been added in the past years?

Please note that this PR has not been merged because it was too successful! :sunglasses: Straight away it did uncover a rounding artifact in MacOS's EPS export function.

I'm reopening this PR just to make sure that this comment does not get lost in the flood of notifications due to the current cleanup action. Feel free to close it as soon as other developers have chimed in.

ctrlcctrlv:
Sure; let's leave it open another week, and if there are still no comments, close it out.

PR #3184, opened 2017-10-14
Just reformatted and some compile warning fixes

Anonrate:
This isn't really a fix to anything but just reformatting the code.  I've got a lot of spare time so I decided I'm going to go through and and clean up the code.  I read an issue a few weeks ago saying that this needs to be done, so  I've taken the initiative in doing so.  Not sure if the coding style will be accepted or not, but I thought it's worth a shot and if it is and the pull request is accepted, I will continue on sharing the results.


frank-trampe:
@Anonrate, we normally don't push commits that focus on whitespace as it makes the blame tool less useful and decreases merge compatibility with forks. There seem to be some substantive changes mixed in with the whitespace changes. Could you submit just those for review with a bit more explanation?

Anonrate:
No problem.  I'm going to hard reset my fork and changing my vim setting to use tabs.  The only reason why I was using whitespaces was to make it look the same for everyone as not everyone has their tabstop set to 8.

I will do a small file and submit that with an explanation as to why I've chosen the codestyle.  There were some part where it was very unclear weather something was part of a code block or not due to multiple statements on one line and no curly braces.

I shall even proceed to format existing whitespaces into tabs in the current code.

frank-trampe:
That wasn't exactly what I meant. What I mean is that we normally don't reformat code except when substantively rewriting it. I'm hoping that you can submit just the non-format changes for review (with a little more explanation of the specific changes, if possible).

In my usage at least, whitespace includes spaces, tabs, and linebreaks.

Anonrate:
So in other words, you're not looking for anyone to clean up the code in some areas?
What about fixing compile warnings?

frank-trampe:
We always want to fix warnings and substantive problems, just not deviant whitespace outside the realm of the actual changed code.

PR #3185, opened 2017-10-15
Informative macro request

Anonrate:
**DISCLAIMER**  I would first like to mention that I'm getting use to this whole git **Fork**, **Pull Request**, **Remote** etc thing.  I'm getting use to the commands which is why there is 3 commits and one basically undoing the other and so forth.

### Type of Request

### What made me request this feature
Looking through the code, more specifically to fix Issue #3162, there has been multiple times where I had added debug messages using the `TRACE` macro, and at the time I was just becoming familiar with the layout and location of everything in the code, so in order to know where one of my messages were coming from, I had used the predefined macros `__FILE__`, `__func__` and `__LINE__` as extra information so I knew precisely where and what messages were being output.

Having put multiple debug message using those macros, it had become very tedious and rather annoying to keep repeating:
```C
TRACE("ISSUE 3162: SplineCharChange at %s:%s:%d\n", __FILE__, __func__, __LINE__);
```
Up to the creation of this pull request, I have still done my debugging accordingly.

### Advantages of feature
Everyone who helps contribute to FontForge, weather you're new and just started to look over the code, or you're old (*Not your age*), and have become very familiar with the code, not only will this be more convenient for newcomers and allow them to work more fluidly and efficiently, I'm sure it will even give aid to the *Veterans* that have dedicated their time to this project, and make their workflow even more smooth.

### To the point (The feature) Knowledge Is Power
The first feature of <sup>[2]</sup> which are both macros, is simply called `INFO`

#### But what it do?
Macro `INFO` is just like the `TRACE` macro except it provides more information as to where this debug message is coming from such as the *Function* name, the *File* the Function is located in, and the *Line* of which called the macro.

#### Will it break stuff?
Unless there is already some other function, macro etc..  defined by the name `INFO`..  The answer is **No** it will not break stuff.  The only issue that I can see happening is if your compiler does not support the these three predefined **Standard** macros
- `__FILE__` Self explanatory as to what it expands to.
- `__func__` The name of the function calling the macro.  <sup><b>There is a macro defined as `__FUNC__`, but I do believe that it is not portable.</b><sup>
- `__LINE__` This expands the line of which the macro was called from.  **NOT** the line that corresponds to that of where `__func__` is defined **or** declared.

#### How does one use this contraption?
Don't like it?  Sounds to difficult and complex?  Have no idea what the "`__`" in those macros are for and why they are there on all three?

Suck it up.  <sup><em>This is just humor appointed to the "Noobies" and I mean no disrespect.</em></sup>  If you've used `TRACE`, you can use `INFO`.  The parameters are the **Exact** same.

### Second Feature
I'm not going to go to indepth and explain this one, because I don't even know what this implementation is called.  I just know how to use and I've used it in some of my previous projects, in particular my own debugger I had worked on long ago.

## Not in enum
I'm requesting that the macro `NOTINENUM` be implemented into the source code as I firmly do believe that it will be of good use.  It is very simplistic and can save a lot of time debugging and finding issues.

I'm aware that there are compiler flags for what this macro does (`-Wswitch` and `Wswitch-default`) (*Explanation is soon.  Be patient.  Jeezus.*), but what if someone forgets to set the `--enable-gcc-warnings` flag, and just has the `--enable-debug` flag set?

#### Shutup and tell me what it do
In a `switch` statement, there are `case labels`, and those labels are of at least one value that you're expecting...  **BUT** if you get a value that you weren't expecting...  You wont even know about it.  Now being as big as FontForge is, I'm sure there have been arguments given to the wrong functions and all that, but no **Error/Warning** was thrown, or there was but just missed because you had the `--enable-gcc-warnings` flag set and you were flooded with words.

#### How to use
`NOTINENUM` has two parameters.  The first parameter specifies the value that was not in the expected enum.  The second parameter is the enum that was expected.

What the result is the name of the invalid value, the value and the enum that was expected, will be output.

Last minute thoughts is that the second parameter may not work, or may need to be given a special way.  I can't exactly remember.  It's been a while since I had done things like this.  But it **Should** not break anything.


gioele:
Hi @Anonrate, thanks for the PR. Having all this extra info in the logs would, indeed, be very useful.

I have two questions/suggestions:

1. Why a separate `INFO`? You could just change the definition of `TRACE`. Or the other way around: `INFO` outputs just a string, `TRACE` a string plus details about the current function. This is also how common logging systems differentiate between "info" and "trace" levels, with "trace" being _more_ verbose than "info".

2. Could you show, in a separate commit, how `NOTINENUM` is used in practice? This would be helpful in understanding if its implementation matches its intended use.

About the Git/commit thing:

* You can use `git commit --amend` to modify the last commit if you spotted a mistake.
* You can use `git rebase -i HEAD~5` to ask git to show you the last five commits and guide you though the process of editing a single one, changing a commit message or _squash_ two commits into one.
* After that you can update the PR with `git push --force`. It is OK to do this on PRs, at least in this project.

Try them out! No fear, with Git you cannot make irreversible mistakes. ;)

jtanx:
Hi, so as mentioned, you can do a rebase to clean up branch history before merging.

In terms of the change itself, I actually do have something a bit more substantive (https://github.com/jtanx/fontforge/blob/2b7478d07615dec481b482dd0e3eac47c90ac160/gdraw/ggdkdrawlogger.c, this was mostly for debugging GDK at the time, although I guess it can be made generically available). 

I think in general, there's more than enough logging macros already. There's the TRACE macro, and there's also the ff_post_warning (and other similar macros that I forget the name of). There's also direct calls to printf (not so good). So I'd be hesitant to add more, instead of modifying the ones we already use.

For `NOTINENUM`, hmm, not really a fan of that one - in particular, peppering the code with NOTINENUM calls when there's a valid compiler switch to detect such cases. You'll find however that there are *many* cases that aren't handled in switch statements but is completely as expected it was just the coding standard in use at the time). And in the case where it truly may be warranted, then I think writing it out with TRACE is perfectly fine. 

P.S: 

image

That... doesn't look right. I think it's just a bug in GitHub.

frank-trampe:
If this is only for per-case debugging, such that the references to `INFO` would not get committed, why do we need to define the macro in the FontForge codebase? You can just add it to some universal header in your debug environment, right? And maybe post it here for others to use in the future?

I'm puzzled by the `NOTINENUM` thing and would also like to see a code sample.

Anonrate:
Sorry for the delay in response.  I have just quickly coded this to demonstrate an example of `NOTINENUM'.  I have also given the output and provided comments.

## Example of `NOTINENUM'
```C
#include <stdio.h>
#include <stdbool.h>

#define NOTINENUM(x,y)                                                              \
    printf("\n%s:%s:%d\n", __FILE__, __func__, __LINE__);                           \
    printf("case label: '%s(%d)' is not valid member of '%s'\n", (#x), (x), (#y))

enum valid {
    valid_vala,
    valid_valb,
    valid_valc
};

/*
 * I was going to use "invalid", but I thought prepending "not" would make it more clear as to what I'm trying to
 * demonstrate.
 */
enum notvalid { notvalid_vala = 3 };

int
machine(enum valid arga)
{
    switch (arga)
    {
        /* 'wanted_vala' is expected because we have a case label for it. */
        case valid_vala:
            printf("\n%d is contained in 'enum valid' and was expected.\n", arga);
            return arga;

        /*
         * Even though 'valid_valb' and 'valid_valc' are both contained in 'enum valid', we weren't expecting one of
         * those values.  We were expecting 'valid_vala'.
         *
         * If we are informed that a correct value was given by 'arga', but not expected, that gives us something to
         * start from in fixing whatever issue maybe present.  We will also know that we are on the right track because
         * 'arga' is contained in the expected enum, which in this case was 'enum valid'.
         */
        case valid_valb:
        case valid_valc:
            printf("\n%d is contained in 'enum valid', but was not expected.\n", arga);

            /*
             * This is just in case if we want the return value of 'machine' to only return 'arga', if 'arga' is both
             * a valid member of 'enum valid' and was expected.  Now if we return the magicnumber '0', and 'arga' is
             * equal to '0'..  Even though it's wrong..  It's right..  But actually wrong.  So we make sure that that
             * 'arga' is not equal to '0'.
             */
            return (!arga ? -1 : 0);

        /*
         * Any integral value can be given, but we are only looking for specific ones, and those are that of what we
         * have 'case labels' for, and each of those 'case labels' are contained in a specific enum which is 'enum
         * valid'.
         *
         * I would say that something like is more common if someone is passing integral value themselves, rather than
         * an enum member.
         *
         * For example:  Say someone takes a quick look at the enum declaration (Or would it be considered
         * definition?..) of 'enum valid'.  They see 3 members, now maybe they are new to coding and just learning, or
         * maybe their exhausted so they just think quickly and pass '3' rather than 'valid_valc' because they like
         * magicnumbers.  The value '3' is not even in the range of 'enum valid', it is in the range of 'enum notvalid'.
         *
         * If we aren't informed that not only was 'arga' not expected..  But also 'notvalid'(See what I did there?  xD)
         * , it could effect our workflow and make finding something that could have been so easy, extremely difficult,
         * especially if the 'switch statement' consists of 'defined constants' rather than members of an enum.
         */
        default:
            NOTINENUM(arga, enum tmpe);
            return (!arga ? -1 : 0);
    }
}

int
main(void)
{
    printf("0 is %s\n",             (machine(0)             == 0)               ? "Good" : "Bad");
    printf("valid_vala is %s\n",    (machine(valid_vala)    == valid_vala)      ? "Good" : "Bad");
    printf("valid_valb is %s\n",    (machine(valid_valb)    == valid_valb)      ? "Good" : "Bad");
    printf("2 is %s\n",             (machine(2)             == 2)               ? "Good" : "Bad");
    printf("notvalid_vala is %s\n", (machine(notvalid_vala) == notvalid_vala)   ? "Good" : "Bad");
    printf("false is %s\n",         (machine(false)         == false)           ? "Good" : "Bad");

    return 0;
}
```

### Output
```
0 is contained in 'enum valid' and was expected.
0 is Good

0 is contained in 'enum valid' and was expected.
valid_vala is Good

1 is contained in 'enum valid', but was not expected.
valid_valb is Bad

2 is contained in 'enum valid', but was not expected.
2 is Bad

ex.c:machine:95
case label: 'arga(3)' is not valid member of 'enum tmpe'
notvalid_vala is Bad

0 is contained in 'enum valid' and was expected.
false is Good
```

I have also included a link to the exact same code tested on **Ideone** so you can see the output yourself.   [Here ya go][1]

[1]: https://ideone.com/aHVeQC

Anonrate:
I didn't mean to close this.  I just reset my fork to the upstream.

skef:
@anonrate: you said you did not mean to close this but have not reopened it. Is this still an active proposal?

PR #3218, opened 2017-12-12
Fix typo in python.c and sflayout.c.

movie-travel-code:
1.The 'struct PyFF_Point' in 'points = malloc(self->pt_cnt*sizeof(struct PyFF_Point *))' is a forward declaration, which is different from the 'PyFF_Point'(typedef ff_point). 2.I believe that 'openfont_str' is a typo and there is no definition of 'openfont_str' in fontforge.

### DIFF
--- a/fontforge/sflayout.c
+++ b/fontforge/sflayout.c
@@ -402,7 +402,7 @@ void LayoutInfoRefigureLines(LayoutInfo *li, int start_of_change,
     double scale;
 
     if ( li->lines==NULL ) {
-	li->lines = malloc(10*sizeof(struct opentype_str *));
+	li->lines = malloc(10*sizeof(struct opentype_str **));
 	li->lineheights = malloc(10*sizeof(struct lineheights));
 	li->lines[0] = NULL;
 	li->lmax = 10;
--- a/fontforge/sflayout.c
+++ b/fontforge/sflayout.c
@@ -504,7 +504,7 @@ void LayoutInfoRefigureLines(LayoutInfo *li, int start_of_change,
     li->pcnt += pdiff;
 
     if ( li->lmax <= li->lcnt+lcnt - (le-ls) + 1 ) {
-	li->lines = realloc(li->lines,(li->lmax = li->lcnt+30+lcnt-(le-ls+1))*sizeof(struct openfont_str **));
+	li->lines = realloc(li->lines,(li->lmax = li->lcnt+30+lcnt-(le-ls+1))*sizeof(struct opentype_str **));
 	li->lineheights = realloc(li->lineheights,li->lmax*sizeof(struct lineheights));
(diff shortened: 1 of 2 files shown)


jtanx:
Good spot, thanks. Interestingly enough, semantically it doesn't actually matter, since it's just the size of a pointer, which is fixed (I didn't know you could put *any* name there and gcc would accept it, but I guess it makes sense.)

fontforge/sflayout.c
```diff
@@ -504,7 +504,7 @@ void LayoutInfoRefigureLines(LayoutInfo *li, int start_of_change,
     li->pcnt += pdiff;
 
     if ( li->lmax <= li->lcnt+lcnt - (le-ls) + 1 ) {
-	li->lines = realloc(li->lines,(li->lmax = li->lcnt+30+lcnt-(le-ls+1))*sizeof(struct openfont_str **));
+	li->lines = realloc(li->lines,(li->lmax = li->lcnt+30+lcnt-(le-ls+1))*sizeof(struct opentype_str **));
```
jtanx: Could you please also change this to `struct opentype_str *`?

frank-trampe: Did this somehow get fixed without dropping the comment?

movie-travel-code: Thank you for your review, jtanx. 

- Of course, these two typos do not change the behavior of the program.
- To be honest,  I am not an expert in fontforge :). I found these two problems when I perform a static analysis on fontforge. I believe that changing `struct opentype_str**` to `struct opentype_str*` requires someone more familiar with fontforge. And this work can be done in the next commit. Sorry for this.

movie-travel-code: Thank you for your comment, frank-trample.
What do you mean? Should I add comments for this change?

jtanx: @movie-travel-code there's nothing to analyse, in this context, `sizeof struct opentype_str*` is what it should be.

movie-travel-code: @jtanx In `li->lines = realloc(ptr, length*sizeof(struct openfont_str **));`, the type of `li-lines` is **`struct opentype_str ***`**, see https://github.com/fontforge/fontforge/blob/master/fontforge/sflayoutP.h#L85, so I think `sizeof(struct opentype_str **) ` is more appropriate than `sizeof(struct opentype_str *)`. Of course, whatever it is, it doesn't matter.

Thank you again for taking the time to review the code.

jtanx: @movie-travel-code sorry, you're right. I was looking at https://github.com/fontforge/fontforge/pull/3218/files#diff-fe9fdb4418dbd4edec366102cd4f55a8R405 for reference, but that one is wrong.

PR #3236, opened 2017-12-23
Add an option to not generate duplicated Mac names

khaledhosny:
### Motivation and Context
Most macOS applications will use name entries from MS platform if the Mac platform is missing.

### Types of changes

### Description
This option to not generate the legacy Mac names especially that they use 8-bit encodings that does not cover the whole Unicode.


khaledhosny:
I don’t need this anymore actually.

PR #3248, opened 2018-01-30
Comment out some debug peintf’s

khaledhosny:
Programs should not be printing messages to the terminal that the user does not need to deal with, it just makes spotting real warnings or error messages hard.


frank-trampe:
It looks good to me. Feel free to merge when the tests finish.

khaledhosny:
I do not have write access to the repository (not for a while now).

PR #3250, opened 2018-02-03
Scale xHeight and capHeight in SFScaleToEm()

khaledhosny:
When changing the *Em Size* in *Element → Font info → General* (or from Python/scripting), `SFScaleToEm()` is called to scale the glyphs and other font-wide metrics, but the relatively new `os2_xheight` and `os2_capheight` were not scaled like other similar fields.


JoesCat:
Thanks

PR #3251, opened 2018-02-04
Remove another debug printf

khaledhosny:


frank-trampe:
I'd personally like to keep the start-up messages, but not enough to rework the pull request.

Perhaps we ought to set the "quiet" option by default?

khaledhosny:
I cleaner way would use and improve FontForge’s existing logging machinery and allow the user to select verbosity level (debug, info, warning, error).

PR #3254, opened 2018-02-09
Fix PcfReadBitmaps() in some cases

maksqwe:


JoesCat:
I don't have fontforge setup at the moment, but looking at your patch, I looked at fourbyte swap and think it may also have an error. This looks like it could probably be i+=4, not i+=2
```C
1645: static void FourByteSwap(uint8 *bitmap,int sizebitmaps) {
1646:    int i, t;
1647:
1648:    for ( i=0; i<sizebitmaps-1; i+=2 ) {
```
is it an error?

maksqwe:
Seems to be yes. Also added fix for FourByteSwap()

JoesCat:
Thanks!

PR #3263, opened 2018-03-09
Fix crash when loading a duplicate name list

khaledhosny:
Loading a name list with a title of an existing one causes a crash, now
it is rejected with an error message instead.

### Types of changes


frank-trampe:
Looks good to me.

PR #3270, opened 2018-03-17
Don’t mark UFO fonts with layers multilayer

khaledhosny:
All fonts in FontForge can have layers, so there is no need to do any thing special for UFO here. More importantly, FontForge’s multilayer mode means something very different:
https://fontforge.github.io/multilayer.html

One consequence of marking a font multilayer is that it can’t be hinted, which is wrong for UFO.

### Types of changes


frank-trampe:
Hmm. I wrote that code, and I knew what multi-layer mode meant at the time. No idea how it happened. In any case, this fix looks correct to me.

PR #3286, opened 2018-05-13
Add ./configure --enable-fontforge-extras

JoesCat:
Bug 827962 - acorn2sfd not in fontforge package.
Very old bug, but only saw this recently.
acorn2sfd now builds, plus added acorn2sfd.1 man page.
Moved acorn2sfd.c out of the libfontforge directory since it was an
orphan program (most users will want just fontforge), and moved it
into fonttools where it now can be part of fontforge-extras.
Interestingly ARM stands for Acorn RISC Machine.

Bug 754414 fix - showttf.c: "Vender URL" should be "Vendor URL"
merged patch shown in launchpad by Brian Murray.

Bug #736221 - showttf -verbose prints help message "showttf -verbose"
instead of being verbose
Reference to a man page? found one, but also needed extra options added.
Enabled building showttf to see what was the problem with this bug here.
Wasn't able to reproduce bug, but further checking saw this bug was
already patched by Alan Coopersmith (merged by Michal-n 2010-04-15).

import/sync "fontforge-extras" showttf.c and this showttf.c code_of
interest to update this version. 'diff' showed some missed edits or
additions worth adding. Then did a lot of editing to reduce the
large number of warnings and errors showing during compiling.

Add missing Encodings.ps.gz (found in fontforge-extras).
put it in doc/html since it is referenced in html docs.
TODO: fix FontForge to get this file in native scripting.

Updated contrib files so they compile using option:
./configure --enable-fontforge-extras.
Default is currently no install. Can be changed later.

Updated stripttc with error checks.
converted README-ttf2eps into ttf2eps.1

Fixed configure.ac to pay more attention to glib-2.0 pkg
and flags. This also involved check headers for -pthreads
because GLIB_CFLAGS outputs pthreads and GLIB_LIBS outputs
several libraries. There also seems to be another type of
GIO-2.0 called GIO-UNIX-2.0 distributed by some distros.

PThreads checking needs to be careful of which platform it
is on, building on linux is easy, but some of the other
host platforms expect other libraries, therefore kept the
checks minimal to just lib function and pthreads.h check.

cidmap/src programs are now built (but not installed).

Modified travis.yml to avoid the make && make contrib.


contrib/fonttools/acorn2sfd.c
```diff
@@ -330,7 +333,7 @@ static void ReadChunk(FILE *file,struct Outlines *outline,int chunk) {
 	flag = r_getint(file);
     else if ( outline->version==6 )
 	flag = (1<<7);
-    if ( !flag&0x80000000 )
+    if ( (!flag)&0x80000000 )
```
jtanx: This change makes no sense, you're effectively ANDing a 1 or a 0 against that bitflag now.

contrib/fonttools/showttf.c
```diff
@@ -1,3 +1,6 @@
+#if 0
```
jtanx: ???

contrib/fonttools/showttf.c
```diff
@@ -307,8 +311,8 @@ exit(0);
 	length = getlong(ttf);
 	cs = filecheck(util,offset,length);
  printf( "%c%c%c%c checksum=%08x actual=%08x diff=%x offset=%d len=%d\n",
-	     tag>>24, (tag>>16)&0xff, (tag>>8)&0xff, tag&0xff,
-	     checksum, cs, checksum^cs,
+	     (int)((tag>>24)&0xff), (int)((tag>>16)&0xff), (int)((tag>>8)&0xff), (int)(tag&0xff),
```
jtanx: shouldn't those first casts be to chars

.travis.yml
```diff
@@ -37,7 +37,7 @@ addons:
     project:
       name: "fontforge/fontforge"
       description: "Free (libre) font editor for Windows, Mac OS X and GNU+Linux"
-    build_command_prepend: cov-configure --comptype gcc --compiler $CC --template && ./bootstrap --force && ./configure $FFCONFIG
+    build_command_prepend: cov-configure --comptype gcc --compiler $CC --template && ./bootstrap --force && ./configure $FFCONFIG --enable-fontforge-extras
```
jtanx: Why not just move that flag into FFCONFIG

contrib/cidmap/Makefile.am
```diff
@@ -1,5 +1,59 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, see http://www.gnu.org/licenses/.
 
--include $(top_srcdir)/git.mk
```
jtanx: Yeah you shouldn't have removed these, and all your new makefiles needs it too.

```
$git status
On branch master
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        contrib/Makefile
        contrib/Makefile.in
        contrib/admintools/Makefile
        contrib/admintools/Makefile.in
        contrib/cidmap/Makefile
        contrib/cidmap/Makefile.in
        contrib/cidmap/src/.deps/
        contrib/cidmap/src/.libs/
        contrib/cidmap/src/Makefile
        contrib/cidmap/src/Makefile.in
        contrib/cidmap/src/mapcns1.exe
        contrib/cidmap/src/mapcns1.o
        contrib/cidmap/src/mapgb1.exe
        contrib/cidmap/src/mapgb1.o
        contrib/cidmap/src/mapjapan1.exe
        contrib/cidmap/src/mapjapan1.o
        contrib/cidmap/src/mapjapan2.exe
        contrib/cidmap/src/mapjapan2.o
        contrib/cidmap/src/mapkorean.exe
        contrib/cidmap/src/mapkorean.o
        contrib/fonttools/.deps/
        contrib/fonttools/.libs/
        contrib/fonttools/Makefile
        contrib/fonttools/Makefile.in
        contrib/fonttools/acorn2sfd-acorn2sfd.o
        contrib/fonttools/acorn2sfd.exe
        contrib/fonttools/dewoff.exe
        contrib/fonttools/dewoff.o
        contrib/fonttools/findtable.exe
        contrib/fonttools/findtable.o
        contrib/fonttools/pcl2ttf.exe
        contrib/fonttools/pcl2ttf.o
        contrib/fonttools/pfadecrypt.exe
        contrib/fonttools/pfadecrypt.o
        contrib/fonttools/rmligamarks.exe
        contrib/fonttools/rmligamarks.o
        contrib/fonttools/showttf.exe
        contrib/fonttools/showttf.o
        contrib/fonttools/stripttc.exe
        contrib/fonttools/stripttc.o
        contrib/fonttools/ttf2eps-ttf2eps.o
        contrib/fonttools/ttf2eps.exe
        contrib/fonttools/woff.exe
        contrib/fonttools/woff.o
```

.travis.yml
JoesCat: I wasn't sure about if it should be default or not.

Do you think that maybe this should be default build, disable to not include extras?

contrib/cidmap/Makefile.am
JoesCat: then we add it in in a patch.

contrib/fonttools/acorn2sfd.c
jtanx: But this change literally makes no sense. When you do `(!flag)`, you're making that value either exactly 1 or 0. So when you AND it with `0x80000000`, you'll always get 0. 

You either want !(flag&0x80000000) or *maybe* `(~flag)&0x80000000`.

contrib/fonttools/showttf.c
jtanx: How could it have already been there when the diff shows it as being added?

contrib/fonttools/showttf.c
jtanx: What compiler were you using to test this? I find it a bit dubious that gcc wouldn't accept chars for `%c`.

frank-trampe:
@JoesCat, we have a review process in place for a reason. Please try to follow it. We can't afford regressions.

.travis.yml
jtanx: No, definitely this should remain be enable to include. 

What I mean is that this stuff used to get tested under CI, which is fine. But what you did here was to basically only run it when we do coverage builds, which is very infrequent.

contrib/cidmap/Makefile.am
jtanx: Did you compile this locally at all? It would have been fairly easy to spot this...

frank-trampe:
If there are no objections, in the interest of not having broken code in the commit history, I'm going to force push a rebase on master that removes these commits. We can then make a new pull request, fix it, review it, and merge it at the appropriate time. Does that work for everybody?

contrib/fonttools/acorn2sfd.c
JoesCat: yes, better choice == (~flag)&0x80000000, avoids the possible ! error

contrib/fonttools/showttf.c
JoesCat: merged from here https://packages.debian.org/sid/fontforge-extras

jtanx:
I'd prefer not to force-push on master, revert commits are ok. Having said that, this is all in contrib code, so not sure we need to do a revert if it'll just be addressed soon after. Happy either way.

contrib/fonttools/acorn2sfd.c
jtanx: Do you have a spec reference or something, because that's fairly different semantics to what it was doing before.

contrib/fonttools/showttf.c
jtanx: So we took an older version of the source ([1](https://github.com/fontforge/fontforge/blob/5aa251a3230006ce20d68ba533fc59dff16c6c79/contrib/fonttools/showttf.c), [2](http://metadata.ftp-master.debian.org/changelogs/main/f/fontforge-extras/fontforge-extras_0.3-4_changelog)) and copy pasted it back in?

Either way, I don't think it's great to be adding in redundant code.

contrib/admintools/copyright.c
```diff
@@ -53,7 +53,7 @@ static void ProcessFile(char *filename,FILE *output, char *dir) {
     time_t now;
     struct tm *tm;
 
-    src = fopen(filename,"r");
+    src = fopen(filename,"rs");
```
jtanx: What's with adding this 's' flag to all the fopen commands?

It's not in any [spec](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fopen.html) that I can find, and [glibc](https://www.gnu.org/software/libc/manual/html_node/Opening-Streams.html) makes no mention of it.

.travis.yml
JoesCat: Ok,
How do you suggest it be integrated in configure.ac and Makefile.am ?
the distcheck has a lot of stuff disabled, commented out. these should be checked too.
How can we test with this as default off, but also enabled? Add another test after linux-nox ?

jtanx: Don't do anything there. Just add it to FFCONFIG, as suggested before: https://github.com/fontforge/fontforge/blob/master/.travis.yml#L61

That will cause that option to be enabled on all the Linux and Mac CI builds.

contrib/admintools/copyright.c
JoesCat: You mentioned problems with "r" and asked if it was ok to go with "rb"
"s" would be for shared.

jtanx: I can understand "rb". But where does "s" come from, and what even supports that flag? As far as I can tell, it's nonstandard, so has no effect.

contrib/cidmap/Makefile.am
JoesCat: Compiled locally (in a cloned directory, then files transferred back to master)
...then compiled again on the pull request - this came up all checkmarks good.
never dirtied-up the master, so never had to run that command - new to me.

.travis.yml
JoesCat: Checked enabled on in #3284, came up all green.
FFCONFIG tests on on all builds.
...but we're looping back to the same question? How can you also test with it off?

contrib/admintools/copyright.c
JoesCat: Ok, we take it out.
"s" is pretty old, in an operating system known as DOS, and other variants where it began to make sense to read and share at the same time. "r" by itself meant lock others out, so it wasn't shareable until the file was released.

.travis.yml
jtanx: Why do we need to test both configurations? If you've seen it work without the option enabled then there's not much that can go wrong, it's not like the build system is in a constant state of flux.

It's good enough to just enable it in ci so that any obvious build issues with the code itself are not missed.

contrib/fonttools/showttf.c
JoesCat: gcc --version
gcc (Mageia 5.5.0-1.mga6) 5.5.0

```bash
$ cp -pR ff ff_test
$ cd ff_test
[ff_test]$ ./bootstrap; ./configure --enable-fontforge-extras --enable-gcc-warnings; make
```
Seems a bit spotty on what it warns.
I removed the (int) around that line and nothing warned this time except for unsigned ints
I went elsewhere and removed it from another line, now I have a bunch...
```code
    6629: printf("\t  Justification priority data for '%c%c%c%c' script, '%c%c%c%c' lang.\n",
	    (stag>>24)&0xff, (stag>>16)&0xff,
	    (stag>>8)&0xff, stag&0xff,
	    (ltag>>24)&0xff, (ltag>>16)&0xff,
	    (ltag>>8)&0xff, ltag&0xff );
```
```bash
showttf.c: In function ‘readttfjustlangsys’:
showttf.c:6629:12: warning: format ‘%c’ expects argument of type ‘int’, but argument 2 has type ‘uint32 {aka unsigned int}’ [-Wformat=]
     printf("\t  Justification priority data for '%c%c%c%c' script, '%c%c%c%c' lang.\n",
            ^
showttf.c:6629:12: warning: format ‘%c’ expects argument of type ‘int’, but argument 3 has type ‘uint32 {aka unsigned int}’ [-Wformat=]
showttf.c:6629:12: warning: format ‘%c’ expects argument of type ‘int’, but argument 4 has type ‘uint32 {aka unsigned int}’ [-Wformat=]
showttf.c:6629:12: warning: format ‘%c’ expects argument of type ‘int’, but argument 5 has type ‘uint32 {aka unsigned int}’ [-Wformat=]
showttf.c:6629:12: warning: format ‘%c’ expects argument of type ‘int’, but argument 6 has type ‘uint32 {aka unsigned int}’ [-Wformat=]
showttf.c:6629:12: warning: format ‘%c’ expects argument of type ‘int’, but argument 7 has type ‘uint32 {aka unsigned int}’ [-Wformat=]
showttf.c:6629:12: warning: format ‘%c’ expects argument of type ‘int’, but argument 8 has type ‘uint32 {aka unsigned int}’ [-Wformat=]
showttf.c:6629:12: warning: format ‘%c’ expects argument of type ‘int’, but argument 9 has type ‘uint32 {aka unsigned int}’ [-Wformat=]
showttf.c: In function ‘readttfjstf’:
```

jtanx: char is probably the same as int on your system. Since the specifier is %c, then if needed, you should be casting to char. Try that and see if the warning goes away.

.travis.yml
jtanx: Sorry, but it's just not practical to test every build option under the sun in ci. As mentioned before, I'm more  concerned about catching any regressions on the code, and less so with the build system. If you're making changes to that, then I'd expect you'd test any build config changes to see if they work. Adding a build just to see if it builds properly without contrib code adds very little value while just blowing out the ci build time. 


As for your other build system changes, I'm not sure how it's related to gdk, considering that all ci builds at the moment still build without it too.

contrib/fonttools/showttf.c
JoesCat: looks like (char) is fine.
gcc needs a bug fixed.
6691 already has (char)
There's a lot of them that probably could go in a function.
Main task was to get it plus other programs to compile - and it/they does now.

.travis.yml
JoesCat: All green checkmarks on #3286, and #3284.
"make dist" copied all the dist files with or without extras enabled.
wiki on glib and gdk shows dependencies...so for now, I'll try things that don't have gdk, harfbuzz, cairo, pango, X.
different topic, but related to this last comment... #3290 built ok locally, but came-out as "X" on fontforge pull request, so that needs more attention as to why it broke.

contrib/fonttools/acorn2sfd.c
JoesCat: riscosfonts.zip

PR #3290, opened 2018-05-16
Add libuninameslist Names2 support

JoesCat:
Add native scripting and python access to libuninameslist >= ver10.9 Names2 functions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,8 @@ install:
         wget 'https://github.com/zeromq/zeromq4-1/releases/download/v4.1.5/zeromq-4.1.5.tar.gz' -O - |tar -zxf -
         wget 'https://github.com/zeromq/czmq/releases/download/v3.0.2/czmq-3.0.2.tar.gz' -O - |tar -zxf -
         wget 'https://github.com/fontforge/libspiro/releases/download/0.5.20150702/libspiro-0.5.20150702.tar.gz' -O - | tar -zxf -
-        wget 'https://github.com/fontforge/libuninameslist/releases/download/20160701/libuninameslist-20160701.tar.gz' -O - | tar -zxf -
+        # wget 'https://github.com/fontforge/libuninameslist/releases/download/20160701/libuninameslist-20160701.tar.gz' -O - | tar -zxf -
+        wget 'https://github.com/fontforge/libuninameslist/releases/download/20180408/libuninameslist-dist-20180408.tar.gz' -O - | tar -zxf -
         wget 'https://bitbucket.org/sortsmill/libunicodenames/downloads/libunicodenames-1.1.0_beta1.tar.xz' -O - | tar -Jxf -
         wget --tries 1 "http://download.savannah.gnu.org/releases/freetype/freetype-$FTVER.tar.gz" || \
           wget "https://sourceforge.net/projects/freetype/files/freetype2/$SFFTVER/freetype-$FTVER.tar.gz"
@@ -91,7 +92,8 @@ install:
         pushd zeromq-4.1.5 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
         pushd czmq-3.0.2 && CFLAGS="-Wno-format-truncation" ./configure --prefix=$DEPSPREFIX && make && make install && popd
         pushd libspiro-0.5.20150702 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
-        pushd libuninameslist-20160701 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
+        # pushd libuninameslist-20160701 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
+        pushd libuninameslist-20180408 && ./configure --enable-pscript --prefix=$DEPSPREFIX && make && make install && popd
         pushd libunicodenames-1.1.0_beta1 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
         tar -zxf freetype-$FTVER.tar.gz -C $DEPSPREFIX
         popd
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,6 @@ install:
         wget 'https://github.com/zeromq/zeromq4-1/releases/download/v4.1.5/zeromq-4.1.5.tar.gz' -O - |tar -zxf -
         wget 'https://github.com/zeromq/czmq/releases/download/v3.0.2/czmq-3.0.2.tar.gz' -O - |tar -zxf -
         wget 'https://github.com/fontforge/libspiro/releases/download/0.5.20150702/libspiro-0.5.20150702.tar.gz' -O - | tar -zxf -
-        # wget 'https://github.com/fontforge/libuninameslist/releases/download/20160701/libuninameslist-20160701.tar.gz' -O - | tar -zxf -
         wget 'https://github.com/fontforge/libuninameslist/releases/download/20180408/libuninameslist-dist-20180408.tar.gz' -O - | tar -zxf -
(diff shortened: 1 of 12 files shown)


JoesCat:
Hi @jtanx, @frank-trampe,
This is getting pretty close to a squash-n-merge once the python part is checked-out.
At the moment I don't have python dev loaded, but I noted that enabling it in configure failed to stop configure, so, in summary, the python part of configure needs checking first before I complete this.
If you have python dev package installed and want to give this a whirl, you can try at the commandline:
fontforge -lang=py script tests/test1090.py
it should perform similarly to
fontforge script tests/test131.pe

fontforge/python.c
```diff
@@ -856,6 +856,56 @@ static PyObject *PyFF_UnicodeNamesListVersion(PyObject *UNUSED(self), PyObject *
...
+}
+
+static PyObject *PyFF_UnicodeNames2FrmTabFromLib(PyObject *UNUSED(self), PyObject *args) {
+    long val;
+    char *temp;
+
+    if ( !PyArg_ParseTuple(args,"|i",&val) )
+	return( NULL );
+    if ( (temp=unicode_name2FrmTab(val))==NULL ) {
+	temp=malloc(1*sizeof(char)); *temp='\0';
```
jtanx: Not a big fan of this. Surely you could just

```c
return Py_BuildValue("s", "");
```

gutils/unicodelibinfo.c
```diff
@@ -133,6 +140,7 @@ char *unicode_name(int32 unienc) {
      * revisit later.
      */
     if( ( unienc >= 0xAC00 && unienc <= 0xD7A3 ) && ( name_data == NULL ) ) {
+	/* replaced code below to reduce library dependencies
```
jtanx: Can we please not do this, we already are using glib so just use it.

tests/test1010.py
```diff
@@ -0,0 +1,54 @@
+# test some libuninameslist functions (if the library is included).
+
+import sys, fontforge
+
+cnt = UnicodeBlockCountFromLib()
+if ( cnt < 1 ):
```
jtanx: This is python, not C, remove the brackets.

tests/test1010.py
```diff
@@ -0,0 +1,54 @@
...
+if ( cnt < 25 ):
+  print("Test Names2 functions. Must be libuninameslist 10.9 or newer.")
+  raise ValueError("Expected 25 or more values that have Names2.")
+
+# none of these are -1 if library loaded
+u0 = fontforge.UnicodeNames2NxtUniFromLib(0)
+u1 = fontforge.UnicodeNames2NxtUniFromLib(1)
+u2 = fontforge.UnicodeNames2NxtUniFromLib(2)
+print("Sample of internal table has ",u0,", ",u1,", ",u2," as the first 3 unicode values.")
+if ( u0 < 0 || u1 < 1 || u2 < 2 ):
```
jtanx: `or`, not `||`. Did you run this locally?

jtanx:
If you're going to add this for python, then please get a python build working and test it locally before continuing.

tests/test1010.py
JoesCat: Still need to get python dev loaded. This can get removed in a squash-merge.

tests/test1010.py
JoesCat: Didn't want it to be confused as a variable :-)
...but it appears to be common:
https://stackoverflow.com/questions/19747371/python-exit-commands-why-so-many-and-when-should-each-be-used
...even here, there is an example where you need to say quit() exit(), not quit or exit.

gutils/unicodelibinfo.c
JoesCat: I'm ok with it in the executables, since you need it for the GUI anyways, but if we have an opportunity to reduce dependencies in the libraries, they can be a little more portable.
I'd like to reduce dependencies on the nox if it can be done.

fontforge/python.c
JoesCat: Nice replacement. There's a lot of places to replace stuff like this in python.c.

.travis.yml
```diff
@@ -83,15 +83,17 @@ install:
...
+        wget 'https://github.com/fontforge/libuninameslist/releases/download/20180408/libuninameslist-dist-20180408.tar.gz' -O - | tar -zxf -
         wget 'https://bitbucket.org/sortsmill/libunicodenames/downloads/libunicodenames-1.1.0_beta1.tar.xz' -O - | tar -Jxf -
         wget --tries 1 "http://download.savannah.gnu.org/releases/freetype/freetype-$FTVER.tar.gz" || \
           wget "https://sourceforge.net/projects/freetype/files/freetype2/$SFFTVER/freetype-$FTVER.tar.gz"
 
         pushd zeromq-4.1.5 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
         pushd czmq-3.0.2 && CFLAGS="-Wno-format-truncation" ./configure --prefix=$DEPSPREFIX && make && make install && popd
         pushd libspiro-0.5.20150702 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
-        pushd libuninameslist-20160701 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
+        # pushd libuninameslist-20160701 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
```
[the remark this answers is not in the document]
JoesCat: ditto - removed on squash-n-merge

.travis.yml
```diff
@@ -83,15 +83,17 @@ install:
         wget 'https://github.com/zeromq/zeromq4-1/releases/download/v4.1.5/zeromq-4.1.5.tar.gz' -O - |tar -zxf -
         wget 'https://github.com/zeromq/czmq/releases/download/v3.0.2/czmq-3.0.2.tar.gz' -O - |tar -zxf -
         wget 'https://github.com/fontforge/libspiro/releases/download/0.5.20150702/libspiro-0.5.20150702.tar.gz' -O - | tar -zxf -
-        wget 'https://github.com/fontforge/libuninameslist/releases/download/20160701/libuninameslist-20160701.tar.gz' -O - | tar -zxf -
+        # wget 'https://github.com/fontforge/libuninameslist/releases/download/20160701/libuninameslist-20160701.tar.gz' -O - | tar -zxf -
```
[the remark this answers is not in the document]
JoesCat: Sounds good. Will upgrade.

tests/test1010.py
jtanx: Functions have nothing to do with if statements. Having brackets where they are unnecessary in such statements is un-pythonic. Just remove them.

gutils/unicodelibinfo.c
jtanx: This code depends on `xasprintf`, which comes from gnulib, and is required whether or not you build with the gui.

Even if you wanted to avoid that dependency, the way you have rewritten it is not good. Instead of adding yet more magic arrays, with memcpys with magic indexing (which is hard for me and others reading this to verify that it's correct), you could have used some form of `malloc`+`sprintf`, or `asprintf`, instead of making it this unreadable and relying on yet more magic arrays

Anyway, just revert it back to how it was before, it compiled fine with or without the gui.

JoesCat:
ready to squash-n-merge.

frank-trampe:
Okay. Do it.

JoesCat:
Hi @frank-trampe - I was about to head-out-the-door.
I see @khaledhosny patch getting merged now - so it's going to still be checking before I'm out.
If this and the others look ok to you, please go ahead and merge - these will take time to get the check (vs the X).

PR #3291, opened 2018-05-18
contrib and --fontforge-extras

JoesCat:
Additional edits and improvements to contrib directory build.

### DIFF
--- a/contrib/fonttools/acorn2sfd.c
+++ b/contrib/fonttools/acorn2sfd.c
@@ -332,7 +332,7 @@ static void ReadChunk(FILE *file,struct Outlines *outline,int chunk) {
     if ( outline->version>=7 )
 	flag = r_getint(file);
     else if ( outline->version==6 )
-	flag = (1<<7);
+	flag |= (1<<7);
     if ( (~flag)&0x80000000 )
 	fprintf( stderr, "Bad flag at beginning of chunk %d\n", chunk );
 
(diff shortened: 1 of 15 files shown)


contrib/fonttools/acorn2sfd.c
```diff
@@ -333,7 +333,7 @@ static void ReadChunk(FILE *file,struct Outlines *outline,int chunk) {
 	flag = r_getint(file);
     else if ( outline->version==6 )
 	flag = (1<<7);
-    if ( (!flag)&0x80000000 )
+    if ( (~flag)&0x80000000 )
```
jtanx: So I read [the spec](http://web.archive.org/web/20070303153914/http://www.pinknoise.demon.co.uk/Docs/Arc/Fonts/Formats.html) which says

```
  Chunk data:
                4       flag word (not present before file format version 7):
                              bit 0 set => horizontal subpixel placement
                              bit 1 set => vertical subpixel placement
                              bit 7 set => dependency byte(s) present (see 
                                           below)
                              bits 2..6 and 8..30 must be 0
                              bit 31 must be 1
```

So I guess this works, although I'd say it's easier read as `if (!(flag&0x80000000))`, but whatever.

Anyway, I guess that'd mean that if `version==6`, this check would always fail, so line 335 should probably read

```c
flag |= (1<<7);
```

JoesCat: It's also easier to write for some functions "return( NULL )", and then do a test of "if ( val )", however, there are some machines where NULL is not zero, so sometimes, it makes more sense to say "return( 0 )" and "if ( val )", or else write "return( NULL )" and "if ( val!=NULL )"

JoesCat: I had put priority on compiling acorn2sfd, glib caused a bit of fuss.
Good catch for "flag |= ( 1<<7)"

PR #3292, opened 2018-05-18
clarify python-dev requirement and force a halt if user wants python-scripting

JoesCat:
Fix a long standing bug where configure does not differentiate between python (the interpreter), and python-dev which is required for adding scripting and extensions. This clarifies that configure
will check and quietly continue if no python is requested.
If user requests python scripting during configure, but there is no python dev package, then configure will HALT with an error.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -916,7 +916,7 @@ Summary of optional features:
 			UseIt?	HaveIt?
   real (floating pt)		${my_real_type}
   programs		yes	${i_do_have_programs}
-  native scripting	yes	${i_do_have_native_scripting}
+  native scripting	yes	${i_do_have_native_scripting}${i_want_python_dev}
   python scripting	${enable_python_scripting}	${i_do_have_python_scripting}
   python extension	${enable_python_extension}	${i_do_have_python_extension}
   freetype debugger		${i_do_have_freetype_debugger}
--- a/configure.ac
+++ b/configure.ac
@@ -916,8 +916,8 @@ Summary of optional features:
 			UseIt?	HaveIt?
   real (floating pt)		${my_real_type}
   programs		yes	${i_do_have_programs}
-  native scripting	yes	${i_do_have_native_scripting}${i_want_python_dev}
-  python scripting	${enable_python_scripting}	${i_do_have_python_scripting}
+  native scripting	yes	${i_do_have_native_scripting}
+  python scripting	${enable_python_scripting}	${i_do_have_python_scripting}${i_want_python_ver}
   python extension	${enable_python_extension}	${i_do_have_python_extension}
   freetype debugger		${i_do_have_freetype_debugger}
   raw points mode		${i_do_have_debug_raw_points}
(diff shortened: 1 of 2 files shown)


JoesCat:
This pull request is probably better done as a squash-n-merge to clean-out some changes that aren't necessary, however, seeing the Monaco bug, and reference to 2.17, there's another patch to add to this tthread - will have it out shortly.
Questions:
Seeing reference to ipython and python even in configure.ac, it might be worth bringing those two tests into the python-scripting function, plus add arg_enable or arg_with that interact/switch that switches the enable_python_scripting switch from check to "yes" - or leave them as-is where they are now?
I'll leave it out of the next patch, but something to think about...

frank-trampe:
The changes so far look fine; we can always make a squash branch when this one is finished. What do you mean by the python-scripting function?

JoesCat:
> What do you mean by the python-scripting function?

Seems the center of attention involves the i_do_have_python_scripting function, which made sense to also add i_do_have_python_extension in file m4/fontforge_arg_enable.m4

I'll attach an incomplete additon here - squashing will clean it up.

JoesCat:
@jtanx - Understood.
-enable-python-scripting == run fontforge, load script, understands "command()"
-enable-python-extension == run python, load script, import fontforge, understands "fontforge.command()"
Looking at the pyhook directory, it might need a local-uninstall to clean-up for make uninstall.

At the moment, seems I can only run AM_PATH_PYTHON() once.
Ran into problem attempting to try python3 and python2+pythondev2, no pythondev3.
If run as -enable-python-scripting=2, it would find this configuration above, but failed if using -enable-python-scripting (without=2, seems because AM_PATH_PYTHON() was called 2x, for 3.3 first, then 2.7 afterwards.).
travis.yml test appears ok with python3+pythondev3 (linux), python2+pythondev2 (mac, windows).
Only condition left to try is a PC with python3+pythondev3 AND python2+pythondev2

frank-trampe:
Ping.

JoesCat:
Tried a live distro with both python 2 an 3 dev packages enabled.
Results are:
--enable-python-scripting=2 will configure for 2.7 or higher,
--enable-python-scripting=3 will configure for 3.3 or higher,
--enable-python-scripting will default to 2.7 or higher.

jtanx:
Uhh... I'm fairly sure that changes the behaviour of defaulting to Python 3 over 2.

JoesCat:
...well, the intended target was aiming for python3 first (if you have dev3), otherwise fetch python2 (if you also have dev2 at same time). Unfortunately using configure.ac am_python(2.7) appears to mean start from 2.7 and go up (would have preferred try 3.3 and go down). configure seemed to have problems with more than one am_python checked, so trying am_python(3.3) and failing that try am_python(2.7) didn't work (even though it would have solved the preference of 3 over 2). :-/

right now, this patch allows you to choose 2.7+ or 3.3+ (or default==2.7+ if you got both loaded), 
...accepting the am_python() dual use as a problem, at least we've got more info and control doing a ./configure  :-)

...going offtopic on the python-scripting, got 2 questions:
1) good/bad idea? - move pycontrib/* into contrib/py/* and enable it on --fontforge-extras
2) good/bad idea? - --python-extensions=py to use python to install/uninstall python programs, otherwise default to current behaviour?  ...use libuninamelist/py/Makefile.am as example

PR #3298, opened 2018-05-31
Fix erroneous "rb" in strcmp

jtanx:
Introduced from 12f7ca9240affc892bc3e136df01f20a11de1733


JoesCat:
Hi @jtanx,
I recall you mentioned rb is desired for win PCs. Maybe something in configure.ac based on $host, and then defining a variable as r or rb. maybe something similar to libspiro/configure.ac:109..114 might help both conditions.

Issue #3200, opened 2017-11-03
Installing Fontforge (Tag:20170731 ) in AMI Amazon Linux 2016.09


sarathprasath-krishnaswamy:
### When reporting a bug/issue:
- [ ] Screenshot
- [x] The FontForge version and the operating system you're using -- Fontforge (Tag: 20170731), AMI Version: 2016.09
- [x] The behavior you expect to see, and the actual behavior -

I tried with the steps mentioned in the https://github.com/fontforge/fontforge/blob/master/CONTRIBUTING.md#building-a-red-hat-source-package-rpm, as I thought it will align with yum package manager

I tried `./configure --without-x` but it throws error like below:

`[root@machine1 fontforge]# PKG_CONFIG_PATH=/usr/local/lib/pkgconfig ./configure --without-x;`
`checking build system type... x86_64-pc-linux-gnu`
`checking host system type... x86_64-pc-linux-gnu`
`checking for a BSD-compatible install... /usr/bin/install -c`
`checking whether build environment is sane... yes`
`checking for a thread-safe mkdir -p... /bin/mkdir -p`
`checking for gawk... gawk`
`checking whether make sets $(MAKE)... yes`
`checking whether make supports nested variables... yes`
`checking how to create a pax tar archive... gnutar`
`checking whether to enable maintainer-specific portions of Makefiles... yes`
`checking for style of include used by make... GNU`
`checking for gcc... gcc`
`checking whether the C compiler works... yes`
`checking for C compiler default output file name... a.out`
`checking for suffix of executables...`
`checking whether we are cross compiling... no`
`checking for suffix of object files... o`
`checking whether we are using the GNU C compiler... yes`
`checking whether gcc accepts -g... yes`
`checking for gcc option to accept ISO C89... none needed`
`checking dependency style of gcc... gcc3`
`checking how to run the C preprocessor... gcc -E`
`checking for grep that handles long lines and -e... /bin/grep`
`checking for egrep... /bin/grep -E`
`checking for ANSI C header files... yes`
`checking for sys/types.h... yes`
`checking for sys/stat.h... yes`
`checking for stdlib.h... yes`
`checking for string.h... yes`
`checking for memory.h... yes`
`checking for strings.h... yes`
`checking for inttypes.h... yes`
`checking for stdint.h... yes`
`checking for unistd.h... yes`
`checking minix/config.h usability... no`
`checking minix/config.h presence... no`
`checking for minix/config.h... no`
[...]

Issue #2968, opened 2016-12-07
fatal error: 'ltdl.h' file not found


WendyAnjanette:
ff version: the lastest release
os: osx el capitan
I tried installing fontforge with this configuration
./configure \
--disable-programs \
--enable-native-scripting \
--without-x \
--without-cairo \
--without-giflib \
--without-libjpeg \
--without-libtiff \
--without-libpng \
--without-libspiro \
--without-libuninameslist \
--without-libunicodenames \
--without-iconv \
--without-libzmq \
--without-libreadline \
--disable-python-scripting \
--disable-python-extension;

but I got an error: 
In file included from plugins.c:29:
../inc/pluginloading.h:31:10: fatal error: 'ltdl.h' file not found

is that a bad configuration? is there anyway I could fix this error?

frank-trampe:
Hi, Wendy.

That suggests that you haven't installed the libltdl developer package. Is
there something wrong with the .app bundle that we provide that requires
you to build from source?

WendyAnjanette:
Hi Sir Frank, 

thanks for replying, I tried install libltdl dev using homebrew but it seems like there's no libltdl there, isn't libltdl included in libtool? I'm just testing something out.

WendyAnjanette:
isn't there anyway I could fix this? or do I really need to use homebrew to be able to install fontforge?

JoesCat:
Hi @WendyAnjanette,
I don't have a Mac, but understand it is based on BSD, and therefore may have some links in another location.
In configure.ac you will see a line that shows:
dnl AC_SEARCH_LIBS([dlopen], [dl dld])
please change it and add
AC_SEARCH_LIBS([dlopen], [dl dld libc])
AC_CHECK_LIB([dl],[dlopen])
AC_CHECK_HEADER([ltdl.h])
AC_CHECK_HEADERS([dlfcn.h])
AC_CHECK_LIB([dl],[dlopen])

run
./bootstrap
./configure
watch for messages refering to the lines mentioned above and see if they say found or yes.

WendyAnjanette:
Hi JoesCat, will try that, thanks!

WendyAnjanette:
got this error from ./bootstrap
glibtoolize:   error: 'libltdl/COPYING.LIB' exists: use '--force' to overwrite
glibtoolize:   error: 'libltdl/Makefile.am' exists: use '--force' to overwrite
glibtoolize:   error: 'libltdl/README' exists: use '--force' to overwrite
[...]

Issue #1463, opened 2014-06-27
Fontforge Mingw ufo.c build issue


jatmew:
I am trying to build the latest fontforge master on mingw. I am running into an issue where make fails with ufo.c

First,  I ran into issue error of unidentified object PyObject on ufo.c:194. I had ran configure with --disable-python-scripting and --disable-python-extension. I notice that the code around it was within ifndef so I put the entire function PythonLibToXML within ifndef __NO_PYTHON condition.

Now, I am running into undefined reference to ... in ufo.c with snippet below:

```
fined reference to `xmlNodeListGetString'
.libs/libfontforge_la-ufo.o: In function `SFReadUFO':
C:\Pdf2htmlEx\msys\1.0\home\JTOLBERT\fontforge-master\fontforge/ufo.c:2468: unde
fined reference to `xmlStrcmp'
C:\Pdf2htmlEx\msys\1.0\home\JTOLBERT\fontforge-master\fontforge/ufo.c:2408: unde
fined reference to `xmlStrcmp'
C:\Pdf2htmlEx\msys\1.0\home\JTOLBERT\fontforge-master\fontforge/ufo.c:2472: unde
fined reference to `xmlStrcmp'
C:\Pdf2htmlEx\msys\1.0\home\JTOLBERT\fontforge-master\fontforge/ufo.c:2410: unde
fined reference to `xmlStrcmp'
C:\Pdf2htmlEx\msys\1.0\home\JTOLBERT\fontforge-master\fontforge/ufo.c:2474: unde
fined reference to `xmlStrcmp'
.libs/libfontforge_la-ufo.o:C:\Pdf2htmlEx\msys\1.0\home\JTOLBERT\fontforge-maste
r\fontforge/ufo.c:2412: more undefined references to `xmlStrcmp' follow
.libs/libfontforge_la-ufo.o: In function `SplinePointListInterpretGlif':
C:\Pdf2htmlEx\msys\1.0\home\JTOLBERT\fontforge-master\fontforge/ufo.c:2834: unde
fined reference to `xmlParseFile'
C:\Pdf2htmlEx\msys\1.0\home\JTOLBERT\fontforge-master\fontforge/ufo.c:2836: unde
fined reference to `xmlParseMemory'
.libs/libfontforge_la-ufo.o: In function `UFOLoadGlyphs':
C:\Pdf2htmlEx\msys\1.0\home\JTOLBERT\fontforge-master\fontforge/ufo.c:2112: unde
[...]

Issue #142, opened 2012-09-28
Improve libltdl configuration


davelab6:
http://old.nabble.com/ltdl-improvements-to34492432.html

PR #3301, opened 2018-06-02
Do a header check for ltdl.h and halt if not found

JoesCat:
This notifies user to add ltdl.h since it might not be loaded if you add libtool packages.
Required by startui.c or startnoui.c, and plugins.


JoesCat:
Should help, close or improve issues #142, #1463, #2968, #3200.

PR #3307, opened 2018-06-16
Add WOFF2 support

### LINKED ISSUE #1503: Support WOFF2
Following https://github.com/fontforge/fontforge/pull/1501, add support for WOFF2 (related, https://github.com/fontforge/fontforge/issues/1457)

jtanx:
Requires libwoff-dev

Closes #1503.


fontforge/woff.c
```diff
@@ -233,11 +233,7 @@ SplineFont *_SFReadWOFF(FILE *woff,int flags,enum openflags openflags, char *fil
     rewind(woff);
     {
         int signature = getlong(woff);
-        if ( signature==CHR('w','O','F','2') ) {
```
frank-trampe: Was this being called on woff2 files?

fontforge/woff2.cc
```diff
@@ -0,0 +1,193 @@
...
+#include <string>
+#include <vector>
+#include <woff2/decode.h>
+#include <woff2/encode.h>
+
+/**
+ * Fontforge's headers aren't C++ includable, so forward-declare
+ * everything ourselves.
+ */
+extern "C" {
```
frank-trampe: Why C++ at all?

fontforge/woff.c
jtanx: No, this is dead code; `ReadSplineFont` already checks the signature, so the only way you could have reached this function is if the signature was `wOFF`.

fontforge/woff2.cc
jtanx: Because the woff2 library is c++

fontforge/woff2.cc
```diff
@@ -0,0 +1,193 @@
...
+    }
+    return ret;
+}
+
+int _WriteWOFF2Font(FILE *fp, SplineFont *sf, enum fontformat format, int32_t *bsizes, enum bitmapformat bf, int flags, EncMap *enc, int layer)
+{
+    FILE *tmp = tmpfile();
+    int ret = 0;
+    if (tmp) {
+        if (_WriteTTFFont(tmp, sf, format, bsizes, bf, flags, enc, layer)) {
```
frank-trampe: Well that's clever!

fontforgeexe/openfontdlg.c
```diff
@@ -135,11 +146,7 @@ struct openfilefilters def_font_filters[] = {
     { NU_("ΤεΧ Bitmap Fonts"), "*{pk,gf}" },
     { N_("PostScript"), "*.{pfa,pfb,t42,otf,cef,cff,gai,pf3,pt3,gsf,cid}{.gz,.Z,.bz,.bz2,.lzma,}" },
     { N_("TrueType"), "*.{ttf,t42,ttc}{.gz,.Z,.bz,.bz2,.lzma,}" },
-#ifdef _NO_LIBPNG
```
frank-trampe: What obsoleted the NO_LIB_PNG logic? And how did it relate to woff?

jtanx: I'm guessing that the `_NO_LIBPNG` check is because the woff (not woff2) code depends on (maybe?) zlib compression that would have been present with libpng. Or something like that.

fontforge/woff2.cc
frank-trampe: Oh. I missed the "namespace" line in the headers. I guess we don't have much choice unless we use a special tool or maintain our own headers.

fontforgeexe/openfontdlg.c
frank-trampe: Is that obsolete now? Does plain WOFF fail safely elsewhere without libpng?

jtanx: No, it's still needed for the WOFF code. https://github.com/fontforge/fontforge/blob/master/fontforge/woff.c#L67

jtanx:
This builds and runs locally, but travis is complaining of

```
 OBJCLD   libfontforgeexe.la
libtool: link: unable to infer tagged configuration
libtool: link: specify a tag with `--tag'
make[3]: *** [libfontforgeexe.la] Error 1
```

So just need to massage the build system a bit...

fontforgeexe/openfontdlg.c
frank-trampe: Okay. I guess moving it is the best approach then.

gdraw/gfilechooser.c
```diff
@@ -138,6 +138,9 @@ return( NULL );
 return( NULL );
 	++pattern;
     }
+    if (pattern > eop) {
+        return NULL;
```
frank-trampe: What was happening here?

frank-trampe:
This all looks good to me, but the build system macros can be a bit blinding. @JoesCat, do you want to take a look too?

gdraw/gfilechooser.c
jtanx: This code is super buggy. I was having trouble getting it to explicitly support both `woff` and `woff2. Really, `pattern` should never run past `eop` (end of pattern). This is just good enough to get it working.

fontforge/woff2.cc
JoesCat: Groan...c++...another dependency ...but it is what it is - would've preferred C. :-/
...on the plus side, the library is maintained by others and kept up to date. :-)

fontforgeexe/openfontdlg.c
JoesCat: Some time ago, someone indicated zlib is pretty well a requirement, so fontforge requires zlib. libpng depends on zlib, but doesn't necessarily need to be included.
WOFF requires zlib, but appears  has nothing to do with png, so it would appear ok to leave-out the lib_png dependency when speaking of WOFF. I just glanced quickly through this page but didn't dig in... https://www.w3.org/blog/2018/03/woff-2-0-the-inside-scoop/

JoesCat:
yeah - Autotools is a bit of a monster at times. The CXX assumption reminds me of libuninameslist and taking-out the am_python in configure.ac
I think bootstrap tends to behave differently between a clean start vs a dirty directory, so if dealing with autoconf I find it easier to cp -pR master --> test, then go into test and ./bootstrap then if modified again, cp -pR master --> testN, then go into testN ..... tedious, but seems easiest way to avoid some bootstrap whoas.
ATM my spare time is targetting an eventual pull-request for unicode 11.0.

fontforgeexe/openfontdlg.c
jtanx: Good point, I see it's a hard dependency [here](https://github.com/fontforge/fontforge/blob/master/configure.ac#L371), I'll remove conditionals on LIBPNG where it only requires zlib.

jtanx:
Alright, I'm done with all changes; I'll merge this soon if there's no objections.

frank-trampe:
@jtax, let's rebase into another branch (keeping this one for historical purposes) before merging. We'll be glad to have it later.

jtanx:
@frank-trampe do you mean to squash all the commits, but leave the history on another branch? I was just planning on doing a merge.

Anyway, I've moved the majority of the implementation into C, because of some issues with gnulib and C++ on Windows.

frank-trampe:
@jtanx, I like to squash in a new branch in a new pull request that closes the old one so that comments, questions, and answers retain their context.

PR #3309, opened 2018-06-23
Parsing bug of subtable rules with coverage type

Ishotihadus:
This parsing bug causes segfault when sfd or ttf / otf file is generated.


JoesCat:
Hi @Ishotihadus, do you have a sample(s) file that can be tested.
If cannot be public, can you send to my email digital at joescat dot com

frank-trampe:
Sorry we missed this one. The change makes sense to me.

PR #3310, opened 2018-06-27
Added German translations

chris-je:


JoesCat:
Thanks for the update @chris-je, I've update de.po further so you may find more words to translate.
You may want to test some of the updates you made, for example:
```c
msgid "Poin_ts too close"
-msgstr ""
+msgstr "Punkte zu nah"
```
This might need to be "Punk_tezu nah" so that you use t to select (the '_' is used to highlight key to press).
t might be a better choice since then you can say t in english or german when explaining to users how to do something.

Please check your updates - and thanks for your contribution.

PR #3323, opened 2018-07-29
Revert "Apply a patch from Adrien for handling underline position better sans spurious whitespace changes."

khaledhosny:
This reverts commit 5aa59fb77c642ee80b4ed16f9136182a3220576b.

The commit broke reading the rest of the post table for CFF fonts since
the two fields are not properly skipped when read. Furthermore in
OpenType the values in post table are authoritative over the CFF table,
so the old behaviour before this commit was correct.

### Types of changes

### DIFF
--- a/fontforge/parsettf.c
+++ b/fontforge/parsettf.c
@@ -5175,21 +5175,9 @@ static void readttfpostnames(FILE *ttf,struct ttfinfo *info) {
 	fseek(ttf,info->postscript_start,SEEK_SET);
 	format = getlong(ttf);
 	info->italicAngle = getfixed(ttf);
-    /*
-     * Due to the legacy of two formats, there are two underlinePosition
-     * attributes in an OpenType CFF font, one being stored in the CFF table.
-     * FontForge due to its pfa heritage will only keep the PostScript/CFF
-     * underlinePosition in the SplineFont so we'll calculate that here if we
-     * are indeed working on a TTF.
-     * If we have a CFF font, cffinfofillup() has already read the appropriate
-     * data and so we don't rewind it (if info->uwidth is odd we are possibly
-     * introducing a rounding error).
-     */
-	if (info->cff_start==0) {
 	info->upos = (short) getushort(ttf);
 	info->uwidth = (short) getushort(ttf);
-	info->upos -= info->uwidth/2;		/* 'post' defn of this field is different from FontInfo defn and I didn't notice */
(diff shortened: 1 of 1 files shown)


khaledhosny:
Without this pull request, running `fontlint` on any font with `CFF` table gives spurious warnings for nearly every glyph in the font, due to incorrectly read `isFixedPitch` value:
```
The advance width of A (695) does not match the font's advanceWidthMax (6272) and this is a fixed pitch font
```

frank-trampe:
I won't be able to look at this in appropriate depth for over a week. In particular, I want to dig up the test case the prompted the original change. @adrientetar, do you have any thoughts on this in the interim?

adrientetar:
I don't remember but, makes sense that the getushort calls shouldn't be in a conditional

khaledhosny:
Ping

frank-trampe:
Apologies for the delay. I'll look at it tomorrow.

frank-trampe:
The specification definitely doesn't have a conditional on whether those table values are present. I thought that maybe there were different versions and that the presence of a CFF table might have been a shortcut to guess the version, but that seems not to be the case. Unfortunately, the change to invert the underline adjustment was actually correct, so I'll need to split hairs on this one and submit a new pull request. Stand by.

frank-trampe:
I accidentally pushed the fix to master. The commit is 9e9278c66ecc47290accb4e9e4aff82e8578ce1e. I will await test results from @khaledhosny before closing this.

khaledhosny:
Why not changing the `info->upos -= info->uwidth/2;` back to `+=`? The code in `tottf.c:dumppost()` still writes `sf->upos-sf->uwidth/2`. Also the comment makes does not reflect what the code is doing now, so it should either be edited or dropped.

frank-trampe:
@khaledhosny, the [specification from Microsoft](https://docs.microsoft.com/en-us/typography/opentype/spec/post) says to subtract, and we are not inverting the axis here, so it seems appropriate to subtract. Maybe the other code is wrong. Or maybe I'm missing something.

khaledhosny:
IIUC, FontForge’s internal structure deals with underline position differently than the `post` table does (see the comments on the reading and writing lines), so the addition when reading and subtraction when writing is to compensate for this. If you are going to align FontForge’s, then more places will need to be adjusted. Right now the data do not roundtrip.

PR #3330, opened 2018-08-28
Allow passing a font object to mergeFonts()

khaledhosny:
Allow passing a font object to `mergeFonts()`, so that one can merge an already opened font instead of having to save it to the disk first.

### Types of changes


JoesCat:
Thanks @khaledhosny  - much appreciated.

PR #3342, opened 2018-10-02
Show missing glyph name in error message

khaledhosny:


JoesCat:
Nice detail - Thanks!

PR #3350, opened 2018-10-22
 Support single subs in ligature and alternate features and subtables

skef:
### Types of changes
- [ x] Bug fix (non-breaking change which fixes an issue)

### Description
- [x ] Describe your changes in detail.

Ligature substitution subtables substitute a single glyph for multiple glyphs, as in:
```
lookup fracb_sub {
    sub \one \slash \four  by \onequarter;
    sub \one \slash \two  by \onehalf;
    sub \three \slash \four  by \threequarters;
} fracb_sub;
```
It is legal both in Adobe's format and in FontForge for such a table to also contain
single-equivalent substitutions, as in:
```
    sub \uni2215  by \slash;
```
But support for this is missing in the current code, so loading a feature table with such
entries results in the error `All entries in a lookup must have the same type on line ...`

Happily, multiple substitution tables also support single-equivalent substitutions, and
that *is* handled in the current FontForge code, so it was easy enough to copy and adapt
the bit of code to do the same for ligature tables. 

Note that this commit is based off my previous pull request rather than the current master. 

### Final checklist
- [ x] My code follows the code style of this project found [here](https://github.com/fontforge/fontforge/blob/master/CONTRIBUTING.md#coding-style).
- [ x] I have read the [**CONTRIBUTING**](https://github.com/fontforge/fontforge/blob/master/CONTRIBUTING.md) guidelines.

### DIFF
--- a/fontforge/featurefile.c
+++ b/fontforge/featurefile.c
@@ -5024,7 +5024,7 @@ static void fea_ParseLookupDef(struct parseState *tok, int could_be_stat ) {
     struct feat_item *item, *first_after_mark;
     enum otlookup_type lookuptype;
     int ret;
-    int has_single, has_multiple;
+    int has_single, has_multiple, has_ligature, has_alternate;
 
     /* keywords are allowed in lookup names */
     fea_ParseTokWithKeywords(tok, false);
@@ -5110,23 +5110,37 @@ return;
     }
     fea_end_statement(tok);
 
-    /* Now a multiple substitution may have a single destination. In which case*/
-    /*  it will look just like a single substitution. So if there are both */
-    /*  multiple and single subs in a lookup, translate all the singles into */
-    /*  multiples */
-    has_single = has_multiple = false;
+    /* Any of a multiple, alternate, or ligature substitution may have */
+    /* a single destination. In either case it will look just like a single */
+    /* substitution. So if there are both multiple/alternate/ligature and */
+    /* single subs in a lookup translate all the singles into the */
+    /* corresponding type */
+    has_single = has_multiple = has_ligature = has_alternate = false;
     for ( item=tok->sofar ; item!=NULL && item->type!=ft_lookup_start; item=item->next ) {
 	enum otlookup_type cur = fea_LookupTypeFromItem(item);
 	if ( cur==gsub_multiple )
 	    has_multiple = true;
+	else if ( cur==gsub_alternate )
+	    has_alternate = true;
+	else if ( cur==gsub_ligature )
+	    has_ligature = true;
 	else if ( cur==gsub_single )
 	    has_single = true;
     }
-    if ( has_multiple && has_single ) {
-	for ( item=tok->sofar ; item!=NULL && item->type!=ft_lookup_start; item=item->next ) {
-	    enum otlookup_type cur = fea_LookupTypeFromItem(item);
-	    if ( cur==gsub_single )
-		item->u2.pst->type = pst_multiple;
+    if ( has_single ) {
+	enum possub_type psttype = pst_pair;
+	if ( has_multiple && !( has_alternate || has_ligature ) )
+	    psttype = pst_multiple;
+	else if ( has_alternate && !( has_multiple || has_ligature ) )
+	    psttype = pst_alternate;
+	else if ( has_ligature && !( has_multiple || has_alternate ) )
+	    psttype = pst_ligature;
+	if ( psttype!=pst_pair ) {
+	    for ( item=tok->sofar ; item!=NULL && item->type!=ft_lookup_start; item=item->next ) {
+		enum otlookup_type cur = fea_LookupTypeFromItem(item);
+		if ( cur==gsub_single )
+		    item->u2.pst->type = psttype;
+	    }
 	}
     }
 
@@ -5354,7 +5368,8 @@ static void fea_ParseFeatureDef(struct parseState *tok) {
     uint32 feat_tag;
     struct feat_item *item, *size_item = NULL;
     int type, ret;
-    int has_single, has_multiple;
+    enum otlookup_type lookuptype;
+    int has_single, has_multiple, has_ligature, has_alternate;
 
     fea_ParseTag(tok);
     if ( tok->type!=tk_name || !tok->could_be_tag ) {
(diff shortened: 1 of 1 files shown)


skef:
In #3355 @frank-trampe asked:
> Can we just add ligature searching as a third prong in the previous block (alongside the single/multiple check)?
> 
> And maybe we could add error reporting for multiple-to-multiple substitutions? You've reviewed the code in question more recently than I have; tell me if this is unnecessary.

My revised commit to this request (c5049fe, which Github is experiencing some hopefully temporary confusion about as I write this) bundles the multiple and ligature code into a single set of loops as requested. 

As for error reporting, in the routine I modified before (`fea_ParseLookupDef`) that is already handled starting at line 5136 (after my changes) starting at the comment "/* Make sure all entries in this lookup of the same lookup type */". 

However, looking into these questions I realized that there is a different routine (`fea_ParseFeatureDef`) that had a version of the multiple check. So this new commit adds the ligature check to that code as well. 

That routine did not have an error check on conflicting types. So I adapted the error check in the original routine to that one, while omitting the check that the type is defined (because features can be used for things lookups can't, so just mandating that every feature has a lookup type is almost certainly wrong.)

Here is a feature file that verifies the logic and should be usable on any Basic Latin-containing `.sfd`:

[ligmulttest.zip](https://github.com/fontforge/fontforge/files/2543460/ligmulttest.zip)

Note that I am not an expert on feature files, so it may be possible that some feature can have conflicting lookup types while separate lookup tables cannot. Given the existing multiple check that seems unlikely, but it might be good to confirm with your experts before accepting the new error logic in `fea_ParseFeatureDef`.

skef:
On reflection I realize I may have misunderstood this question:

> And maybe we could add error reporting for multiple-to-multiple substitutions?

and that you may have meant something like this:

    lookup multmulttest {
        sub \w \x by \y \z;
    } multmulttest;

The `sub` line here isn't syntactically legal and gets dropped with an error earlier in the parsing process, which then gets this marked as an empty lookup:

    Unparseable glyph sequence in substitution on line 2 of multmulttest.fea
    This lookup has no effect, I can't figure out its type on line 3 of multmulttest.fea

If there were other `sub` lines they would be accepted and processed in the normal way.

skef:
Can we revive this conversation? The more recent code should address the question asked in #3355

skef:
It turns out I tried to revive this last week a little too soon. I hit the same problem with alternate substitution tables (for the 'salt' feature and similar), so I have added that type to the check as well. That change made the compressed style I was using untenable, so the code is also easier to read now. 

This raises the question of what further table types a single substitution can mix in with. However, I'm not an expert on feature files in general; I'm just submitting fixes for the problems I hit and can't easily work around.

I would like to see this in the pipeline soon if possible, as I am writing scripts to update the features in an open source font, and the maintainers will want to verify they work with some version of fontforge. Since I'm pretty much done with the conversion, it's unlikely I'll hit additional problems of this particular type (table mismatches).

fontforge/featurefile.c
```diff
@@ -5489,24 +5504,53 @@ return;
...
+	    }
+	}
+    }
+
+    /* Make sure all entries in this lookup of the same lookup type */
+    lookuptype = ot_undef;
+    for ( item=tok->sofar ; item!=NULL && item->type!=ft_feat_start; item=item->next ) {
+	enum otlookup_type cur = fea_LookupTypeFromItem(item);
+	if ( cur==ot_undef )	/* Some entries in the list (lookupflags) have no type */
+	    /* Tum, ty, tum tum */;
```
frank-trampe: It's probably okay, but this is the sort of thing that throws some compilers for a loop, and it would probably be clearer just to double-set `lookuptype` here anyway. Not worth changing.

frank-trampe:
This all looks good to me.

fontforge/featurefile.c
```diff
@@ -5489,24 +5504,53 @@ return;
...
+	if ( psttype!=pst_pair ) {
+	    for ( item=tok->sofar ; item!=NULL && item->type!=ft_feat_start; item=item->next ) {
+		enum otlookup_type cur = fea_LookupTypeFromItem(item);
+		if ( cur==gsub_single )
+		    item->u2.pst->type = psttype;
+	    }
+	}
+    }
+
+    /* Make sure all entries in this lookup of the same lookup type */
```
jtanx: Are we sure this applies to feature definitions too, and not just lookup definitions? Because this check is causing a feature file which previously parsed OK to now fail ([here](https://github.com/BlackFoundry/InriaFonts/blob/master/masters/INRIA-SANS/InriaSans-Regular.ufo/features.fea#L140))

skef: The answer seems to be "no, we're not sure". And looking into this the news is better than I was expecting at first, but it's still not clear what to do. 

Convert the font with `ufo2ft`, output as OpenType, and open in FontForge and you wind up with four `frac` subtables. Interestingly two of those are single substitution tables -- the one for the slashes is separate from the one for numerator substitution. Then there's a ligature table and a contextual table. So the idea seems to be to just dynamically sort out the mixture of subtables needed when reading a feature, rather than restricting it to a given type. 

With FontForge, remove all this code, including the old multiple loop, and open the UFO in FontForge and you get three `frac` subtables. This is the good news -- I'm surprised there's any "sorting" going on at all. The result looks similar to the OTF except that there's only one table for single substitutions.

At first glance, at least, both seem to work in the Metrics window. 

But here's the problem: FontForge seems to just be sorting the entries by type and then putting them in *some* order -- maybe the order in which a given type is first encountered. `ufo2ft`, and therefore probably fontTools, seems to be creating a new subtable each time there is a switch in type. That's likely to cause different results in some cases, although I'm not positive about that. So CW was right to leave out the check here, and I was wrong to add it back in, but it's not yet clear to me what changes are needed. 

@khaledhosny can you give any advice on this?

skef: IndraSerif-Regular's 8 is ugly, so it may not be necessary to fix this.

skef: [Aha](https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#7b-ordering-of-lookups-and-subtables):
> A lookup in the OpenType font will be created from each named lookup block (§4.e) or each run of rules with the same feature, script, language, lookupflag and lookup type attribute.

So neither the old or my replacement code was right -- it should all be removed. But there's a bit more work to do in either the parsing or application stage, because this isn't quite what FontForge is doing.

PR #3352, opened 2018-10-27
Sensibly extend python interface for point selection. 

skef:
### Types of changes
- [] Bug fix (non-breaking change which fixes an issue)
- [] New feature (non-breaking change which adds functionality)

### Description

Most of the time you perform a transformation on or otherwise fiddle with a subset of points, you select a set of "base" points and then use all the control points that come with them (I'm taking the "base" terminology from the `Get Info...` point dialog). This is how the `Transform ...` dialog works, for example. 

The existing python "Point" interface says this about the `selected` flag:

> Whether this point is selected in the UI. If an off-curve point is selected it means the preceding (interpolated) on-curve point is selected.

The "interpolated" is a clue that this documentation is quadratic-contour specific. In fact, for cubic
contours only base points are marked selected and control points never are. 

It would be easy to extend the `selected` flag to correspond to whether a control point is selected. But most of the time that's not what you want to know -- you want to know whether the control point's base point is selected. You can get that information but it's a pain -- you have to scan back and forward in the contour's point list, deal with missing control points on straight lines, and so forth. Until the interface is extended more radically so that a control point has a reference to its base point, it's better to just supply the information needed most of the time. 

What I have done in this push is:

1. Extended the meaning of `selected` for control points to be true if the base point is selected and false otherwise.
2. Added a `highlighted` flag that is always false for base points but true for a control point if it is selected.

Selected is also writable for base points and highlighted for control points. Here are some demonstration menu items:

```
import fontforge
import psMat
import math

def regGlyph(fn, ms, sc = None):
    fontforge.registerMenuItem(fn, None, None, "Glyph", sc, ms)

def getSF():
    ss = fontforge.askString("Scale Factor", "Enter one 1-based scale factor or two x y scale factors separated by space")
    sa = ss.split(' ')
    if len(sa) == 1:
        sx = sy = float(sa[0])
    else:
        sx = float(sa[0])
        sy = float(sa[1])
    return (sx, sy)

def scaleSelectedbyContour(junk, glyph):
    l = glyph.layers[1]
    glyph.preserveLayerAsUndo(1)
    sf = getSF()
    for c in l:
        cbb = c.boundingBox()
        tr = psMat.translate(-(cbb[0]+cbb[2]/2), -(cbb[1]+cbb[3])/2)
        tr = psMat.compose(tr, psMat.scale(sf[0], sf[1]))
        tr = psMat.compose(tr, psMat.translate((cbb[0]+cbb[2]/2), (cbb[1]+cbb[3])/2))
        for p in c:
            if p.selected or p.highlighted:
                p.transform(tr)
    glyph.layers[1] = l

def revHighlight(junk, glyph):
    l = glyph.layers[1]
    glyph.preserveLayerAsUndo(1)
    for c in l:
        for p in c:
            if not p.on_curve:
                p.highlighted = not p.highlighted
    glyph.layers[1] = l

def revSelect(junk, glyph):
    l = glyph.layers[1]
    glyph.preserveLayerAsUndo(1)
    for c in l:
        for p in c:
            if p.on_curve:
                p.selected = not p.selected
    glyph.layers[1] = l

regGlyph(scaleSelectedbyContour, "Scale Selected by Contour")
regGlyph(revHighlight, "Reverse selection of control points")
regGlyph(revSelect, "Reverse selection of base points")
```
`scaleSelectedByContour` will prompt for one or two dimensions (1 rather than 100 based, separated by a space if two) that scales selected points in relation to the center of their contour. With base points all control points are included in the transformation. But if additional control points are selected (and their base points carefully de-selected), they will also scale. Note the simple logic for accomplishing this.

The other two functions demonstrate that the selected status can also be updated, which can be used to chain functions or to support "selection macros" for easier hand-manipulation.

I updated the documentation to reflect these changes. 

The Spiro system already has a separate interface, which these changes do not affect. I tried to make the `highlighted` flag consistent for quadratic contours as well, but the since the control points are "shared" it appears that there are subtle rules for which on-curve point stores the selected status that I couldn't decipher (in a short time looking). However, the whole system seems like a mess for quadratic curves anyway, so I'm not sure this really hurts anything.

Given that I did change the meaning of `selected` for cubic curve points, this is a change rather than a mere addition. However, given that spent several hours changing no points at all, before learning by Googling widely that the trick is updating the layer, I'm not sure how many scripts out the world might break with this change. In any case, the value is always false-evaluable in the current implementation for cubics.

* The interface is sort of implicit in the instructions for `preserveLayerAsUndo`, but pretty mysterious otherwise. 

With all my pull requests, but especially this one, I'm happy to try to implement alternative designs as suggested. 

### Final checklist
- [ x] My code follows the code style of this project found [here](https://github.com/fontforge/fontforge/blob/master/CONTRIBUTING.md#coding-style).
- [ x] My change requires a change to the documentation.
- [ x] I have updated the documentation accordingly.
- [x ] I have read the [**CONTRIBUTING**](https://github.com/fontforge/fontforge/blob/master/CONTRIBUTING.md) guidelines.

### DIFF
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -935,13 +935,38 @@ fontforge.registerImportExport(importGlyph,exportGlyph,None,"foosball","foo","fo
     </TR>
     <TR>
       <TD><CODE>on_curve</CODE></TD>
-      <TD colspan=2>Whether this is an on curve point or an off curve point (a
-	control point)</TD>
+      <TD colspan=2>True for a point on the curve and false for (in the
+      case of a cubic contour) a control point.</TD>
     </TR>
     <TR>
       <TD><CODE>selected</CODE></TD>
-      <TD colspan=2>Whether this point is selected in the UI. If an off-curve point
-	is selected in means the preceding (interpolated) on-curve point is selected.</TD>
+      <TD colspan=2>For an on-curve point, whether it is selected.
+	For a control point, whether its on-curve point is selected.
+	<P>Most transformations of on-curve points should also include
+	their control points. The <CODE>selected</CODE> flag makes it easier
+	to pick all relevant points using on-curve selection.
+        <P>Setting or clearing the flag will change the selection status of
+	an on-curve point but have no effect for an off-curve point.
+	<P>For quadratic points off-curve points, selected should be true
+	if the preceding <em>interpolated</em> on-curve point is selected. 
+      </TD>
+    </TR>
+    <TR>
+      <TD><CODE>highlighted</CODE></TD>
+      <TD colspan=2>For an on-curve point this will be false. For an off-curve
+	point, whether it is selected.
+	<P>The <CODE>highlighted</CODE> flag can be used when the selection
+	status of a cubic control point is relevant in the python interface.
+	If, for example, a transformation should include all selected on-curve
+	points and their associated off-curve points, and also any additional
+	selected off-curve points, the script can use <CODE>p.selected or
+	p.highlighted</CODE>
+        <P>Setting or clearing the flag will change the selection status of
+	a control point but have no effect for an on-curve point.
+	<P>This interface is only partially implemented for quadratic
+	contours, and some selected off-curve points may not be marked
+	correctly. 
+      </TD>
     </TR>
     <TR>
       <TD><CODE>name</CODE></TD>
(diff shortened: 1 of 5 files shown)


skef:
It's worth noting that the combination of `on_curve`, `selected`, and `highlighted` allow a programmer to make most of the decisions they might need to without ever mapping control point to base point.

skef:
After having used the programmatic contour interface a bit more, I've come to realize that it is more fundamentally incomplete, in that even adding a layer of python on top (whether or not selection state is available) wouldn't let one do all that one needs to do. So I am going to withdraw this for now and think more about what a more complete API would look like.

Issue #3077, opened 2017-06-02
pangoxft not installed makes Travis CI fail on MacOS


gioele:
In addition to the issue #3001, current Travis CI builds on MacOS fail because the package `pangoxft` is not installed in the test VMs. See, for example, <https://travis-ci.org/fontforge/fontforge/jobs/237801652>.

`./configure` fails with the following error:

>     checking for pango >= 1.10 pangoxft... no
>     configure: error: Package requirements (pango >= 1.10 pangoxft) were not met:
>
>     No package 'pangoxft' found
>
>     Consider adjusting the PKG_CONFIG_PATH environment variable if you
>     installed software in a non-standard prefix.

jtanx:
Hombrew being 💩 as usual. Someone decided it was a good idea to remove the x11 option from the pango formula.

frank-trampe:
I don't have time to do it, but I'm tempted to try to hack Travis to implement the MacPorts build that we use for the release packages. I think that the number of people who use FontForge on Macintosh and do not want a GUI is very low, which makes supporting Homebrew in its post-X11 a waste of resources. @DomT4, what do you think?

MikeMcQuaid:
> I think that the number of people who use FontForge on Macintosh and do not want a GUI is very low, which makes supporting Homebrew in its post-X11 a waste of resources.

I don't feel strongly either way but for just for your decision making: 1,772 people installed FontForge on Homebrew in the last 30 days.

jtanx:
https://bintray.com/homebrew/bottles/fontforge#statistics

For comparison, the bundle has been downloaded 28921  times, so brew installs represents ~6%.

DomT4:
> 1,772 people installed FontForge on Homebrew in the last 30 days.

I wouldn't be surprised if most of these are people not reading the `desc` or `caveats` beforehand. As someone who has used FontForge quite a bit I don't think I have ever wanted the CLI. IMO there's a case for deprecating the formula in favour of the cask completely.

> @DomT4, what do you think?

It physically pains me to say this given my history with Homebrew, but yes, I think given your specific needs MacPorts is a better fit for testing FontForge.

jtanx:
Hmm... I seem to recall some third party scripts that depend only on the cli, so that may constitute a small portion of it.

It would be nice to merge in that gdk work at some point since it more or less works, and well, it's hard to get testing without people actually using that. Though probably wouldn't want to do that before #2841.

[...]

Issue #2734, opened 2016-06-25
Menu unification and rally to push mainline.


monkeyiq:
This is to track an update to https://github.com/fontforge/fontforge/issues/2257 for a menu unification for the charview, fontview, and metricsview. 

Interested parties are invited to explore the code in my public branch at https://github.com/monkeyiq/fontforge-fork/tree/2016menus

From a developer perspective this code introduces the new sharedmenu.c which contains the unification of the menu code from all three files: charview.c, fontview.c and metricsview.c. There is a crude type system and a new vtable (contained in sharedmenu_funcs) to allow each type of window to service the menu calls as is best done by the window. Some calls, such as almost all of the entries in the "File" menu can be serviced by a single code path. That code gets around being called with a 'gw' containing the userdata pointer for different window types by using the new type system and calls from classtypeui.h such as tryObtainGDataFontView(). This call will convert a CharView\* to a FontView\* by returning the delegated fv pointer contained in the charview. The calling code doesn't need to care, it just gets a correct fontview pointer from all three window types and can operate from there. All cv, mv, and fv can be converted to a fontview because all either are a fontview or contain a pointer to one.

Window types that allow zero, one, or more characters to be selected can implement the getActiveSplineChar() virtual function to tell the caller if 1+ chars are selected. This way, menu code and other interested callers can check/find the active character without needing to care what window or class type they are dealing with.

The type system, object coercion, and virtual table functions setup the code base to allow single functions to be applied to multiple window types. At the moment, there are some menuitems in some windows which are disabled. Some of these do not really have to be disabled, the code behind that menu item could possibly be updated to work across more window types than it was originally created for. For example, something that used to only be offered on the charview window might be enabled in the menu for the fontview window. Selecting that item in the fontview window would simply apply the same function that the charview menu item does, but for all chars that are selected in the fontview window that the menu item was selected from. This case is exactly what the new visitor mechanism visitAllCharViews() in fontview.c has been created to assist with. Guidance is requested as to which menu items should be enabled in which window for what menu items first. Some things are likely to be more urgent than others.

I have gone through the menus on the three window types checking for crashes but might have missed some, especially when the menu item is contingent on previous setup. I have also tried to unify the code that disables the menu items depending on if they should be made available in that window (ie, the code supports their use and they made sense).
### TODO
- [ ] maybe enable some more menuitems which are currently disabled.
- [ ] update code for some menu items which no longer work for some window+menu combinations
[...]

Issue #3321, opened 2018-07-26
centos7 and python3.6.5 , install fontforge error


fa2002:
centos7 and python3.6.5 , install fontforge error:

make[2]: Leaving directory `/soft/fontforge-master/po'
Making all in pycontrib
make[2]: Entering directory `/soft/fontforge-master/pycontrib'
Making all in gdraw
make[3]: Entering directory `/soft/fontforge-master/pycontrib/gdraw'
make[3]: Leaving directory `/soft/fontforge-master/pycontrib/gdraw'
Making all in simple
make[3]: Entering directory `/soft/fontforge-master/pycontrib/simple'
make[3]: Leaving directory `/soft/fontforge-master/pycontrib/simple'
Making all in collab
make[3]: Entering directory `/soft/fontforge-master/pycontrib/collab'
make[3]: Leaving directory `/soft/fontforge-master/pycontrib/collab'
Making all in graphicore
make[3]: Entering directory `/soft/fontforge-master/pycontrib/graphicore'
make[3]: Leaving directory `/soft/fontforge-master/pycontrib/graphicore'
make[3]: Entering directory `/soft/fontforge-master/pycontrib'
make[3]: Nothing to be done for `all-am'.
make[3]: Leaving directory `/soft/fontforge-master/pycontrib'
make[2]: Leaving directory `/soft/fontforge-master/pycontrib'
Making all in share
make[2]: Entering directory `/soft/fontforge-master/share'
make[2]: Leaving directory `/soft/fontforge-master/share'
Making all in tests
make[2]: Entering directory `/soft/fontforge-master/tests'
make[2]: Leaving directory `/soft/fontforge-master/tests'
fatal: No names found, cannot describe anything.
make[2]: Entering directory `/soft/fontforge-master'
make[2]: Leaving directory `/soft/fontforge-master'
make[1]: Leaving directory `/soft/fontforge-master'

What is "fatal: No names found, cannot describe anything."?

JoesCat:
This is a copy-n-paste from testing file .travis.yml
```sh
packages: [autoconf, automake, autotools-dev, bzip2, gcc-7, libtool,
      libjpeg-dev, libtiff4-dev, libpng12-dev, libfreetype6-dev, libgif-dev,
      libx11-dev, libxml2-dev, libpango1.0-dev, libcairo2-dev, python3.6-dev, lcov]
```
You are likely missing files needed to build
Can you check to see your installation, or add any possibly missing packages listed in this list above?
[...]

PR #3353, opened 2018-10-28
Clarify Xwindows OR GDK when choosing to use FontForge with a GUI

JoesCat:
GDK graphical backend was recently added, and we need to organize the
options as a list to show users two GUI choices.

GDK brings its' own copies of pango and cairo, but X has the choice of
including cairo, so we show cairo as a subset of X (Pango required).
Add some more info for pango (see issues #2734, #2968, where's xft?),
and do extra header and function checks for distros that forget to go
look into /usr/local.
GDK is of better interest on the Mac, so issue #3077 not as important.

Put collab in the GUI section of the list since it is not used in the
command line version of FontForge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ addons:
       - deadsnakes
     packages: [autoconf, automake, autotools-dev, bzip2, gcc-7, libtool,
       libjpeg-dev, libtiff4-dev, libpng12-dev, libfreetype6-dev, libgif-dev,
-      libx11-dev, libxml2-dev, libpango1.0-dev, libcairo2-dev, python3.6-dev, lcov]
+      libx11-dev, libxml2-dev, libpango1.0-dev, libpangoxft-1.0-dev, libcairo2-dev, python3.6-dev, lcov]
   coverity_scan:
     project:
       name: "fontforge/fontforge"
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ addons:
       - deadsnakes
     packages: [autoconf, automake, autotools-dev, bzip2, gcc-7, libtool,
       libjpeg-dev, libtiff4-dev, libpng12-dev, libfreetype6-dev, libgif-dev,
-      libx11-dev, libxml2-dev, libpango1.0-dev, libpangoxft-1.0-dev, libcairo2-dev, python3.6-dev, lcov]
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -32,7 +32,7 @@ addons:
       - deadsnakes
     packages: [autoconf, automake, autotools-dev, bzip2, gcc-7, libtool,
       libjpeg-dev, libtiff4-dev, libpng12-dev, libfreetype6-dev, libgif-dev,
-      libx11-dev, libxml2-dev, libpango1.0-dev, libcairo2-dev, python3.6-dev, lcov]
+      libx11-dev, libxml2-dev, libpango1.0-dev, libpangox-1.0-dev, libcairo2-dev, python3.6-dev, lcov]
```
jtanx: Why did the travis requirements change?

configure.ac
```diff
@@ -419,7 +419,15 @@ if test x"${have_glib}" != xpkg; then
 fi
 
 if test x"${i_do_have_x}" = xyes; then
-   PKG_CHECK_MODULES([PANGO],[pango >= 1.10 pangoxft])
+   # pango depends on glib2. Required by GUI for {cairo or gtk/gtk2 or Mac CoreText}
+   # pangoxft is submodule of pango (used in gxcdraw) (there are other versions of Xft)
+   PKG_CHECK_MODULES([PANGO],[pango >= 1.10 pangoft2 pangoxft])
+   have_libpango=no
+   AC_CHECK_HEADER([pangoxft.h],
```
jtanx: Why do we need this check when we're already checking for pangoxft in PKG_CHECK_MODULES, this is failing travis

JoesCat:
Issue #3321 got my interest, and at the moment, I'm currently limited to command-line compiling because I don't have the GDK or Xwindows dev packages loaded on my computer yet, so it makes it's easier to verify failure.
If I loaded the required dev packages, I couldn't test-out failure.
The first pull request in this list was using a method used in the m4/fontforge_arg_with.m4 of following checks with ac_check_headers which seems to help with older distros and some BSDs, but what it shows instead was an unexpected failure. The pull requests following that was trying-out to see if I could get pangoxft.h to be seen.

Due to the number of additional pull requests on this branch (failing), it's better to avoid merging this, and I'll try some extra stuff to see how travis behaves (which should hopefully help some other distros).

PR #3354, opened 2018-10-28
Upgrade libunicode functions to avoid segfaults or other errors.

JoesCat:
Check values coming into copyn functions before processing data to avoid segfaults and/or other errors.

### DIFF
--- a/Unicode/memory.c
+++ b/Unicode/memory.c
@@ -47,11 +47,10 @@ char *copyn(const char *str,long n) {
      *      https://github.com/fontforge/fontforge/issues/1239
      */
     char *ret;
-    if ( str==NULL )
-    	return( NULL );
-
-    ret = (char *) malloc(n+1);
-    memcpy(ret,str,n);
-    ret[n]='\0';
-    return( ret );
+    if ( str!=NULL && ++n>0 && (ret=(char *)(malloc(n)))!=NULL ) {
+	if ( (--n) ) memcpy(ret,str,n);
+	ret[n]='\0';
+	return( ret );
+    }
+    return( NULL );
 }
--- a/Unicode/ustring.c
+++ b/Unicode/ustring.c
@@ -334,10 +334,13 @@ unichar_t *u_copynallocm(const unichar_t *pt, long n, long m) {
 }
 
 unichar_t *u_copy(const unichar_t *pt) {
-    if(pt)
-return u_copyn(pt,u_strlen(pt));
+/* Copy unichar_t string and if error, return( "" ). */
+    long n;
 
-return((unichar_t *)0);
+    if ( pt && (n=u_strlen(pt))>0 )
+	return( u_copyn(pt,n) );
+
+    return( calloc(1,sizeof(unichar_t)) );
 }
 
 unichar_t *u_concat(const unichar_t *s1, const unichar_t *s2) {
@@ -382,7 +385,6 @@ unichar_t *uc_copyn(const char *pt, int n) {
 unichar_t *uc_copy(const char *pt) {
 /* Copy unsigned char{0..255} string to unichar type */
 /* string. If not printable, return an empty string. */
-    unichar_t *res, *rpt;
     int n;
 
     if ( pt && (n=strlen(pt))>0 )
@@ -391,38 +393,44 @@ unichar_t *uc_copy(const char *pt) {
     return( calloc(1,sizeof(unichar_t)) );
 }
 
-char *cu_copyn(const unichar_t *pt,int len) {
+char *cu_copyn(const unichar_t *pt, int n) {
+/* Copy unichar string to str. Assume values 0..255. */
     char *res, *rpt;
+    unichar_t t;
 
-    if(!pt)
-return(NULL);
-
+    if ( pt && ++n>0 ) {
 #ifdef MEMORY_MASK
-    if ( (len+1)>=MEMORY_MASK )
-	len = MEMORY_MASK-1;
+	if ( n>MEMORY_MASK )
+	    n = MEMORY_MASK;
 #endif
-    res = (char *) malloc(len+1);
-    for ( rpt=res; --len>=0 ; *rpt++ = *pt++ );
-    *rpt = '\0';
-return(res);
+	if ( (res=rpt=malloc(n)) ) {
+	    if ( (--n) ) {
+		while ( --n>=0 ) {
+		    if ( (t=*pt++)>0xff ) {
+			free(res);
+			return( NULL );
+		    }
+		    *rpt++ = t;
+		}
+	    }
+	    *rpt = '\0';
+	    return( res );
+	}
+    }
+    return( NULL );
 }
 
 char *cu_copy(const unichar_t *pt) {
-    char *res, *rpt;
+/* Copy unichar to char string, assume values 0..255 */
+/* and if error, return( "" ) which can be made free */
+    char *res;
     int n;
 
-    if(!pt)
-return((char *)0);
+    if ( pt && (n=u_strlen(pt))>0 )
+	if ( (res=cu_copyn(pt,n))!=NULL )
+	    return( res );
 
-    n = u_strlen(pt);
-#ifdef MEMORY_MASK
-    if ( (n+1)>=MEMORY_MASK )
-	n = MEMORY_MASK/sizeof(unichar_t)-1;
-#endif
-    res = (char *) malloc(n+1);
-    for ( rpt=res; --n>=0 ; *rpt++ = *pt++ );
-    *rpt = '\0';
-return(res);
+    return( calloc(1,sizeof(unichar_t)) );
 }
 
 double u_strtod(const unichar_t *str, unichar_t **ptr) {
--- a/Unicode/ustring.c
+++ b/Unicode/ustring.c
@@ -845,7 +845,7 @@ long utf8_strlen(const char *utf8_str) {
     long len = 0;
 
     while ( utf8_ildb(&utf8_str)>0 && ++len>0 );
-    return( len );
+    return( len < 0 ? -1 : len );
 }
 
 long utf82u_strlen(const char *utf8_str) {
@@ -856,7 +856,7 @@ long utf82u_strlen(const char *utf8_str) {
     while ( (ch = utf8_ildb(&utf8_str))>0 && ++len>0 )
 	if ( ch>=0x10000 )
 	    ++len;
-    return( len );
+    return( len < 0 ? -1 : len );
 }
 
 void utf8_strncpy(register char *to, const char *from, int len) {
--- a/Unicode/ustring.c
+++ b/Unicode/ustring.c
@@ -323,7 +323,7 @@ unichar_t *u_copynallocm(const unichar_t *pt, long n, long m) {
     if ( pt && ++n>0 && ++m>0 && m>=n ) {
 #ifdef MEMORY_MASK
 	if ( n>MEMORY_MASK/sizeof(unichar_t) )
-	    n = MEMORY_MASK/sizeof(unichar_t)-1;
+	    n = MEMORY_MASK/sizeof(unichar_t);
 #endif
 	if ( (res=(unichar_t *)(calloc(m,sizeof(unichar_t)))) ) {
 	    if ( (--n) ) memcpy(res,pt,n*sizeof(unichar_t));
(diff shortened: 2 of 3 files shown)


Unicode/memory.c
```diff
@@ -47,11 +47,10 @@ char *copyn(const char *str,long n) {
...
      */
     char *ret;
-    if ( str==NULL )
-    	return( NULL );
-
-    ret = (char *) malloc(n+1);
-    memcpy(ret,str,n);
-    ret[n]='\0';
-    return( ret );
+    if ( str!=NULL && ++n>0 && (ret=(char *)(malloc(n)))!=NULL ) {
```
jtanx: This makes it quite obtuse in what it's trying to do, especially the stuff around incrementing then decrementing `n`.

Something like

```c
    char *ret;
    if ( str==NULL || n <=0 )   // <-- modify here
    	return( NULL );

    ret = (char *) malloc(n+1);
    if (ret == NULL)             // <-- add here
        return;
    memcpy(ret,str,n);
    ret[n]='\0';
return( ret );
```

Is way more readable. Although I'd question the necessity of the `<0` check on `n`, at that point you're just masking issues with code which is doing something bad.

Unicode/ustring.c
```diff
@@ -301,26 +301,36 @@ return( NULL );
 
 unichar_t *u_copyn(const unichar_t *pt, long n) {
     unichar_t *res;
+
+    if ( pt && ++n>0 ) {
 #ifdef MEMORY_MASK
-    if ( n*sizeof(unichar_t)>=MEMORY_MASK )
-	n = MEMORY_MASK/sizeof(unichar_t)-1;
+	if ( n>MEMORY_MASK/sizeof(unichar_t) )
```
jtanx: can you explain this one a bit more

Unicode/ustring.c
```diff
@@ -301,33 +301,46 @@ return( NULL );
...
+    return( NULL );
 }
 
 unichar_t *u_copy(const unichar_t *pt) {
-    if(pt)
-return u_copyn(pt,u_strlen(pt));
+/* Copy unichar_t string and if error, return( "" ). */
+    long n;
+
+    if ( pt && (n=u_strlen(pt))>0 )
```
jtanx: Why, where did you actually see this be an issue; `u_copyn` with `u_strlen` returning 0 would have been valid in the old case anyway

Unicode/ustring.c
```diff
@@ -846,7 +866,7 @@ long utf82u_strlen(const char *utf8_str) {
     while ( (ch = utf8_ildb(&utf8_str))>0 && ++len>0 )
 	if ( ch>=0x10000 )
 	    ++len;
-    return( len );
+    return( len < 0 ? -1 : len );
```
jtanx: why

Unicode/ustring.c
JoesCat: testing for failing
if ( pt=NULL || n<0 || n+1<0 )
testing for pass
if ( pt!=NULL && n>=0 && n+1>0 )
or testing for pass (n+1)
if ( pt!=NULL && n>0 && n>0 )

Unicode/ustring.c
JoesCat: was thinking about utype.c where we also have -1

Unicode/memory.c
frank-trampe: That is a lot more readable.

Unicode/ustring.c
frank-trampe: And perhaps add a comment indicating what MEMORY_MASK is. The name doesn't seem to pertain to the use.

Unicode/ustring.c
```diff
@@ -301,33 +301,46 @@ return( NULL );
...
+	if ( n>MEMORY_MASK/sizeof(unichar_t) )
+	    n = MEMORY_MASK/sizeof(unichar_t);
 #endif
-    res = (unichar_t *) malloc((n+1)*sizeof(unichar_t));
-    memcpy(res,pt,n*sizeof(unichar_t));
-    res[n]='\0';
-return(res);
+	if ( (res=(unichar_t *)(malloc(n*sizeof(unichar_t)))) ) {
+	    if ( (--n) ) memcpy(res,pt,n*sizeof(unichar_t));
+	    res[n] = 0;
```
frank-trampe: I think that `'\0'` is the right way to do this since it is a character rather than an integer.

Unicode/ustring.c
frank-trampe: @JoesCat, I still don't understand. What value of pt and pointed contents would behave differently between these two cases? The only one I can think of is one in which n gets overflowed.

Unicode/ustring.c
JoesCat: how does it behave with big endian?

Unicode/ustring.c
JoesCat: Looking lower down in the code, we need to build for malloc((n+1)*sizeof(unichar_t)), so we still need room for one more additional sizeof(unichar_t).
Moving the n+1 up to the top avoids several repeated computations....
and moving the sizeof(unichar_t) to the other side of the equation lets the compiler precompute the value once instead of all the users computing it repeatedly during run time.

Unicode/ustring.c
frank-trampe: Oh. I forgot. That is an integer type. Never mind.

Unicode/ustring.c
JoesCat: for n==0 there wasn't a need to go through the exercise of testing variables, malloc and insert 0 as there's already was a calloc(1,sizeof(unichar_t)) below that did the exact same thing.

frank-trampe: Yes, but you added the calloc call, right? Was the ambiguity of sometimes returning a static string and other times a dynamic string the thing you were trying to fix?

JoesCat: Fixing the static string was definitely the main interest.
When you look at calls used everywhere, there is a good potential to cleanup and free memory instead of losing and leaking memory everywhere.
There is a calloc added internally.
Looking at all the feedback here, another commit is needed, the git comment would look pretty ugly on a squash commit, so probably better to create a fresh commit.

frank-trampe: @JoesCat, I still don't understand some of what you've done. Let me finish up comments tomorrow, and then we can finalize changes. The changeset is of sufficiently singular purpose to go all into one giant commit, I think, as long as the functionality is appropriately commented.

Unicode/ustring.c
JoesCat: Reviewing the MEMORY_MASK using git blame (on a copy of 20120731 code), seems this code came in around 2001 and seems to be nowhere else other than within ustring.h
Maybe an abandoned idea?
Take it out completely?

Unicode/ustring.c
JoesCat: We need to pay attention to size when looking at code.
Looking at some issues, what appears to work fine in 32bit is probably failing in 64bit

Unicode/ustring.c
frank-trampe: @JoesCat, n will be in a register in any reasonably compiled binary, and it's not even in the inner loop. So all of this might save two clock cycles per string copy. I don't think that that justifies the significant decrease in readability.

Unicode/ustring.c
```diff
@@ -301,33 +301,46 @@ return( NULL );
...
+	}
+    }
+    return( NULL );
 }
 
 unichar_t *u_copynallocm(const unichar_t *pt, long n, long m) {
+/* Alloc space for m unichar chars and copy n unichar chars */
     unichar_t *res;
+
+    if ( pt && ++n>0 && ++m>0 && m>=n ) {
```
frank-trampe: The adjusted offsets are also sort of confusing here.

Unicode/ustring.c
frank-trampe: I don't think that this actually saves significant execution time in the zero-length case, which is rare, and it increases execution time in the other cases (two function calls).

Unicode/ustring.c
```diff
@@ -301,33 +301,46 @@ return( NULL );
...
-    if(pt)
-return u_copyn(pt,u_strlen(pt));
+/* Copy unichar_t string and if error, return( "" ). */
+    long n;
+
+    if ( pt && (n=u_strlen(pt))>0 )
+	return( u_copyn(pt,n) );
 
-return((unichar_t *)0);
+    return( calloc(1,sizeof(unichar_t)) );
```
frank-trampe: How often does this function get called with a NULL pointer? Will returning something to be freed in that case cause leaks?

Unicode/ustring.c
frank-trampe: If that never happens here, we ought to remove the check since it is bound to confuse people.

Unicode/ustring.c
```diff
@@ -345,72 +358,79 @@ return( u_copy( s1 ));
...
+	    return( res );
+	}
+    }
+    return( NULL );
 }
 
 char *cu_copy(const unichar_t *pt) {
-    char *res, *rpt;
+/* Copy unichar to char string, assume values 0..255 */
+/* and if error, return( "" ) which can be made free */
```
frank-trampe: Yay! A comment.

Unicode/ustring.c
```diff
@@ -345,72 +358,79 @@ return( u_copy( s1 ));
...
     int n;
 
-    if(!pt)
-return((char *)0);
+    if ( pt && (n=u_strlen(pt))>0 )
+	if ( (res=cu_copyn(pt,n))!=NULL )
+	    return( res );
 
-    n = u_strlen(pt);
-#ifdef MEMORY_MASK
```
frank-trampe: We need to take out `MEMORY_MASK` completely or leave it. A partial removal invalidates its purpose but still leaves clutter. I'm always partial to the minimum-change approach, of course.

Unicode/ustring.c
```diff
@@ -345,72 +358,79 @@ return( u_copy( s1 ));
...
-    if ( (n+1)>=MEMORY_MASK )
-	n = MEMORY_MASK/sizeof(unichar_t)-1;
-#endif
-    res = (char *) malloc(n+1);
-    for ( rpt=res; --n>=0 ; *rpt++ = *pt++ );
-    *rpt = '\0';
-return(res);
+    return( calloc(1,sizeof(unichar_t)) );
```
frank-trampe: As elsewhere, we need to be sure that this effective change in behavior (returning dynamically allocated `""` instead of `NULL` when given a NULL pointer) doesn't break anything. @JoesCat knows more about the use cases than I do.

Unicode/ustring.c
```diff
@@ -345,72 +358,79 @@ return( u_copy( s1 ));
...
-char *cu_copyn(const unichar_t *pt,int len) {
+char *cu_copyn(const unichar_t *pt, int n) {
+/* Copy unichar string to str. Assume values 0..255. */
     char *res, *rpt;
+    unichar_t t;
 
-    if(!pt)
-return(NULL);
-
+    if ( pt && ++n>0 ) {
```
frank-trampe: As before, I think that adjusting `n` is unnecessarily confusing.

Unicode/ustring.c
```diff
@@ -345,72 +358,79 @@ return( u_copy( s1 ));
...
-#endif
-    res = (unichar_t *) malloc((n+1)*sizeof(unichar_t));
-    for ( rpt=res; --n>=0 ; *rpt++ = *(unsigned char *) pt++ );
-    *rpt = '\0';
-return(res);
+    return( calloc(1,sizeof(unichar_t)) );
 }
 
-char *cu_copyn(const unichar_t *pt,int len) {
+char *cu_copyn(const unichar_t *pt, int n) {
```
frank-trampe: Renaming the parameter from `len` to a letter makes the calling convention less clear.

Unicode/ustring.c
```diff
@@ -345,72 +358,79 @@ return( u_copy( s1 ));
...
-    if ( (len+1)>=MEMORY_MASK )
-	len = MEMORY_MASK-1;
+	if ( n>MEMORY_MASK )
+	    n = MEMORY_MASK;
 #endif
-    res = (char *) malloc(len+1);
-    for ( rpt=res; --len>=0 ; *rpt++ = *pt++ );
-    *rpt = '\0';
-return(res);
+	if ( (res=rpt=malloc(n)) ) {
```
frank-trampe: This is vastly less readable than the previous implementation. Can we just add the out-of-bounds check to a body for the old for loop?

Unicode/ustring.c
```diff
@@ -345,72 +358,79 @@ return( u_copy( s1 ));
 return( pt );
 }
 
-unichar_t *uc_copyn(const char *pt,int len) {
-    unichar_t *res, *rpt;
-
-    if(!pt)
-return((unichar_t *)0);
+unichar_t *uc_copyn(const char *pt, int n) {
```
frank-trampe: Same concerns (code clarity and parameter names) here as in cu_copyn.

Unicode/ustring.c
JoesCat: two function calls? where?

Unicode/ustring.c
JoesCat: I added comments - will push another patch.

Unicode/ustring.c
JoesCat: The code was already there.... and I think it was in fontforgeexe/displayfonts.c that got my attention in cleaning-up after doing a printout. It's a lot easier to cleanup a bunch of mallocs.
I think it may be a problem attempting to differentiate a constant (unichar_t *)0 which can't be free, vs an alloc(), which can be free().
Running valgrind tells you there's a lot of stuff that can be cleaned up.

Unicode/ustring.c
JoesCat: I'll push a patch with more of these ;-P

Unicode/ustring.c
JoesCat: seems there's more places where it appears. eg github rust, and a couple other places... looks like something to do with intel cpus. Maybe a hard limit. The ustring.c code is 2001, so we're probably looking at an older bit of code.

Unicode/ustring.c
JoesCat: I mentioned return( "" ), but we're talking about calloc(1,sizeof(unichar_t)).
The code of interest is return( (unichar_t)0 )

Unicode/ustring.c
JoesCat: and using
if ( pt && n>=0 && n+1>0 )
says we know exactly what we're doing - hehehe

Unicode/ustring.c
JoesCat: function call uc_... uses n, and both functions are cu_copyn() and uc_copyn()
It seemed to be more consistent to go with n, instead of changing the other function to use len.

JoesCat:
I stopped when I got to the strstr and strmatch functions as I see we have more than one issue here.

For example tolower() only goes 0..65535 right now, results are unsigned shorts, plus we also need to look at unicode above 65535 as well as non-unicode values inbetween.
This may be better having a pull request just for copy, for now, and do another pull request for string matching afterwards

PR #3376, opened 2018-12-02
Clarify Xwindows OR GDK when choosing to use FontForge with a GUI

### LINKED ISSUE #3220: Old bug in FontForge-2017-07-31-Windows-r2: obsolete copywrite
The blob at the start of running `fontforge -lang=ff …` says:

     Based on sources from 08:20 UTC 31-Jul-2017-ML-TtfDb-D.
     Based on source from git with hash: 8b6e7f07a5511e4de9c8a710db6ca8e770de8f4e

It should be 2000-2017.
- [ ] Screenshot

JoesCat:
GDK graphical backend was recently added, and we need to organize options as a list to show users two GUI choices to avoid confusion of thinking you need to include everything to build with a GUI.

This sort of closes #3353 by helping the user identify differences in ./configure
...will have to leave the pango.xft checks of #3353 for resolving later.

This also updates a few values to 2018 (closes #3220), and adds a reminder in configure.ac for when we build next release to check files and update dates accordingly.

### DIFF
--- a/m4/fontforge_arg_enable.m4
+++ b/m4/fontforge_arg_enable.m4
@@ -219,7 +219,7 @@ if test x$use_gdk = xyes ; then
             AC_MSG_NOTICE([building the GUI with the GDK2 backend...])
         ],
         [
-            AC_MSG_ERROR([Cannot build GDK backend without GDK installed. Please install the GDK Developer Package.])
+            AC_MSG_ERROR([Cannot build GDK backend without GDK installed. Please install the GTK+ Developer Package.])
         ])
     else
         PKG_CHECK_MODULES([GDK],[gdk-3.0 >= 3.10],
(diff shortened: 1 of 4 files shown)


m4/fontforge_arg_enable.m4
```diff
@@ -214,7 +219,7 @@ if test x$use_gdk = xyes ; then
             AC_MSG_NOTICE([building the GUI with the GDK2 backend...])
         ],
         [
-            AC_MSG_ERROR([Cannot build GDK backend without GDK installed.])
+            AC_MSG_ERROR([Cannot build GDK backend without GDK installed. Please install the GDK Developer Package.])
```
jtanx: There's no such thing, you'd need to install GTK development libraries to get GDK.

JoesCat: If you want to go ahead and rewrite a correct version, go ahead, and close both pull requests.
For now, I'm still running command line right now (lacking GUI stuff) and haven't followed up on loading a full GUI yet, ...but at the rate I'm going now, it may be next year before I get there.  :-/

JoesCat:
Thanks for adding the edit Jeremy - much appreciated.

Issue #3056, opened 2017-04-15
Provide an AppImage for many Linux distributions


probonopd:
Providing an [AppImage](http://appimage.org/) would have, among others, these advantages:
- Works for most Linux distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional desktop integration with `appimaged`
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using AppImageUpdate
- Can optionally GPG2-sign your AppImages (inside the file)
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions

[Here is an overview](https://github.com/probonopd/AppImageKit/wiki/AppImages) of projects that are already distributing upstream-provided, official AppImages.

I have produced an experimental, proof-of-concept AppImage by converting the binaries from the Ubuntu trusty PPA to AppImage using this file
https://github.com/probonopd/AppImages/blob/master/recipes/meta/FontForge.yml

It can be downloaded from
https://bintray.com/probono/AppImages/FontForge#files

If you have questions, AppImage developers are on #AppImage on irc.freenode.net.

JoesCat:
Adding/Including [Linux Standard Base](https://en.wikipedia.org/wiki/Linux_Standard_Base)

ror6ax:
This would be great. My Fedora 27 offers version of 2016 year, which is very suboptimal ;)

probonopd:
Is there any interest from the FontForge developers? I'd be happy to help.

JoesCat:
Your question is welcome, your timing is pretty good now, and if your strengths include this particular area, you can likely see things we overlooked.
Read the latest pull request conversations concerning flatpak as it is also relevant, 3362, 3368.
I believe one or more distros are also wanting to have this area automated instead of copy-pasting files in these related directories.
...I am a little doubtful one binary will satisfy all the various distros since FontForge carries a lot of dependencies, but it's worth giving a try and see how far we succeed.

probonopd:
[...]

Issue #3053, opened 2017-04-03
128x128 and 64x64 icons is not installed by 'make install'


sl1pkn07:
`make install` not install 128x128 and 64x64 icons

```
---snip---
make[3]: No se hace nada para 'install-exec-am'.
 /usr/bin/mkdir -p '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/applications'
 /usr/bin/mkdir -p '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/icons/hicolor/22x22/apps'
 /usr/bin/mkdir -p '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/icons/hicolor/24x24/apps'
 /usr/bin/mkdir -p '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/icons/hicolor/16x16/apps'
 /usr/bin/mkdir -p '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/icons/hicolor/32x32/apps'
 /usr/bin/mkdir -p '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/icons/hicolor/256x256/apps'
 /usr/bin/mkdir -p '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/icons/hicolor/scalable/apps'
 /usr/bin/mkdir -p '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/icons/hicolor/48x48/apps'
 /usr/bin/mkdir -p '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/mime/packages'
 /usr/bin/mkdir -p '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/icons/hicolor/scalable/apps'
 /usr/bin/install -c -m 644 fontforge.desktop '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/applications'
 /usr/bin/install -c -m 644 icons/22x22/apps/fontforge.png '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/icons/hicolor/22x22/apps'
 /usr/bin/install -c -m 644 icons/32x32/apps/fontforge.png '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/icons/hicolor/32x32/apps'
 /usr/bin/install -c -m 644 icons/24x24/apps/fontforge.png '/tmp/makepkg/fontforge/pkg/fontforge/usr/share/icons/hicolor/24x24/apps'
[...]

Issue #892, opened 2013-11-01
Add AppData.xml 


davelab6:
http://blogs.gnome.org/hughsie/2013/08/29/gnome-software-center-and-you/

divayprakash:
@davelab6 I'd like to take this up.

JoesCat:
Can you check if this applies?
https://github.com/fontforge/fontforge/tree/master/desktop

davelab6:
@divayprakash awesome! please do!! :)

divayprakash:
@JoesCat That definitely applies, however there are a few changes :
- `<application>.....</application>` tags are replaced by `<component type="desktop">.....</component>`
-  `<id type="desktop">` is now simply `<id>`
- `<license>` tag needs to be replaced by `<metadata_license>` and `<project_license>`
- We need to add `<name>` and `<summary>` tags
- Screenshot code needs to be changed to 
```
<screenshot>
    <image>IMAGE_REF</image>
    <caption>IMAGE_CAPTION</caption>
</screenshot>
```
- `<updatecontact>` is changed to `<update_contact>`

divayprakash:
@JoesCat @davelab6 Should I make the changes to that same file or create another?

davelab6:
Same file, with a pull request, I guess?

JoesCat:
On June 14, 2016 03:47:35 PM Dave Crossland wrote:
> Same file, with a pull request, I guess?

Git seems to recognize if you move a file, so if you need to move it, do 
that first, then modify the file. (two git updates). I have not been keeping 
tabs on what's happening, but check the file history for the files in that 
directory. I think Redhat may also be affected, so try to cover all or as 
many distros as you can with your improvements.

You also may want to check out what Debian is doing too so that you can 
merge what they are doing into the improvements (see issue #2732, and 
please give credit where credit is due if you pull in their mods - if you 
have trouble with reassigning-credit, one of us can take care of that, as 
it's not really as easy as click-n-change (I think)).

Thanks,
Looking forward to your help/improvements.

divayprakash:
@JoesCat I've done some more research on this, and the `appdata.xml` file specification is more-or-less common across distros. I'm incorporating the following specs -
- https://wiki.debian.org/AppStream/Guidelines
- https://wiki.gnome.org/Initiatives/GnomeGoals/AppDataGnomeSoftware
- https://fedoraproject.org/wiki/Packaging:AppData
- https://people.freedesktop.org/~hughsient/appdata/

[...]

PR #3387, opened 2018-12-24
theme and desktop changes

JoesCat:
Add Freedesktop desktop support files and also merge-in added
modifications from Tobias Mueller <muelli@cryptobitch.de>
Support starts from older v1.0 upto what appears currently in
/usr/share/{appdata,applications,icons,mime,metainfo,pixmaps}
Theme support also added plus additional images included for
pixmaps and 2012 themes. Renamed directories from icons-old and
icons to 2012 and tango (allowing for further theme expansion).

Closes issue #3372, #3056, #3053, #892, #3368 

@copyninja, @pnemade, @muelli - please check.

FYI, my internet is down, so I won't be able to reply quickly.
Add any patching you think necessary.
Thanks.


frank-trampe:
I don't really understand what's happening, but I'm satisfied that this is safe to merge.

frank-trampe:
Other stakeholders please comment promptly, as I would really like to tag a release before the end of the year.

JoesCat:
Almost ok.... had too many arguments for AM_CONDITIONAL (4 is too many)....got a fix here by putting decision in an if/then decision, but internet problems are causing havoc.... hope to get replacement modem in a couple days.

  AM_CONDITIONAL to if/then
    AM_CONDITIONAL complained of too many arguments.

--------------------------------- configure.ac ---------------------------------
index ace319b..69549f8 100644
@@ -686,9 +686,13 @@ AC_PATH_PROG([DESKTOP_FILE_INSTALL],[desktop-file-install],[:])
 AC_PATH_PROG([UPDATE_MIME_DATABASE],[update-mime-database],[:])
 AC_PATH_PROG([UPDATE_DESKTOP_DATABASE],[update-desktop-database],[:])
 
-AM_CONDITIONAL([XDG_DESKTOP],
-   [test x"${fontforge_desktop}" != xno -a test x"${DESKTOP_FILE_INSTALL}" != "x:"
-    -a test x"${UPDATE_MIME_DATABASE}" != "x:" -a test x"${UPDATE_DESKTOP_DATABASE}" != "x:"])
+i_have_desktop=yes
+if test x"${fontforge_desktop}" = xno || test x"${DESKTOP_FILE_INSTALL}" = "x:"; then
+  i_have_desktop=no
+if test x"${UPDATE_MIME_DATABASE}" = "x:" || test x"${UPDATE_DESKTOP_DATABASE}" = "x:"; then
+  i_have_desktop=no
+fi
+AM_CONDITIONAL([XDG_DESKTOP],[test x"${i_have_desktopUPDATE_MIME_DATABASE}" = xyes])
 
 #--------------------------------------------------------------------------
 # Create and Set FontForge Compiler Flags

desktop/Makefile.am
```diff
@@ -25,32 +25,107 @@
...
+if USE_THEME_TANGO
+ICON_SRC = $(srcdir)/tango
+endif
+
+if USE_THEME_2012
+ICON_SRC = $(srcdir)/2012
+endif
+
+desktopdir = $(PACKAGEROOTDIR)/applications
+desktop_DATA = fontforge.desktop
```
muelli: Have you not changed the appid on purpose?
Otherwise these files, i.e. the desktop, appdata, and icon files, need to be renamed to resemble the appid, i.e. org.fontforge.FontForge.

desktop/fontforge.appdata.xml
```diff
@@ -84,21 +83,8 @@
...
   <screenshots>
-    <screenshot type="default" height="675" width="1200">
-      <image>https://github.com/fontforge/fontforge/raw/master/doc/html/ff-screenshot.png</image>
-      <caption>Fontforge showing a glyph being edited</caption>
-    </screenshot>
+    <screenshot type="default" width="1200" height="675">https://github.com/fontforge/fontforge/raw/master/doc/html/ff-screenshot.png</screenshot>
   </screenshots>
   <url type="homepage">https://github.com/fontforge/fontforge</url>
   <update_contact>fontforge-devel@lists.sourceforge.net</update_contact>
-  <content_rating type="oars-1.1"/>
```
muelli: What's the purpose of removing OARS and release history and so on?

desktop/Makefile.am
JoesCat: I'll need to review this in better detail when I get propper internet again.
This patch attempts to try meet adding a desktop (requested), and also attempt to maintain older ver1 information, but also try and include newer data in the metainfo directory to meet latest version requirements.
Do you think making a duplicate of fontforge.metainfo.xml in /usr{/local}/share/appdata could be helpful? I did see a couple metainfo.xml in the appdata directory too.

JoesCat:
@frank-trampe - This can be squash-merge when you're ready to go, but prefer to also see feedback.

if/when --enable-fontforge-desktop it will add desktop support files in /usr{/local}/share/{appdata,applications,pixmaps,icons,metainfo,mime,mime/packages}.
if/when --enable-fontforge-desktop=PATH, then it will apply files in location PATH instead.
If DESTDIR is empty, then assumption made that this is for user PC and not some other location.

Patch is based on looking at a whole bunch of different applications pushing files into the /usr/share area, plus what we need for FontForge, including tango, 2012, and potentially other themes.

JoesCat:
@muelli - a quick check gives the impression we're looking at two different specs....
1 - desktop: https://specifications.freedesktop.org/desktop-entry-spec/1.2/ar01s02.html
2 - appstream: https://www.freedesktop.org/software/appstream/docs/
...the desktop spec (1) seems ok with maintaining "fontforge.desktop" in ver1.2
...I'll read-thru (2) since existing /usr directory examples I looked at didn't appear to show reverse dns naming except for (mainly) org.kde and one or two org.gnome files.

muelli:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic is relevant

JoesCat:
Based on latest patches, is is best reduced from:
"Add ./configure --enable-fontforge-desktop"
to
"theme selection"

desktop/Makefile.am
```diff
@@ -25,23 +25,109 @@
...
-EXTRA_DIST = icons/src/icon-16x16-apps-fontforge.svg	\
-	icons/src/icon-24x24-apps-fontforge.svg		\
-	icons/src/icon-32x32-apps-fontforge.svg
+PACKAGEROOTDIR = ${datarootdir}
+
+# Where .desktop files go.
+desktopdir = $(PACKAGEROOTDIR)/applications
+desktop_DATA = org.fontforge.FontForge.desktop
+
+appdatadir = $(PACKAGEROOTDIR)/appdata
```
jtanx: What's going on here? In the old makefile, `org.fontforge.FontForge.appdata.xml` got installed to the metainfo folder, but now it gets installed to appdata instead? The AppImage packager expects the former.

desktop/Makefile.am
```diff
@@ -25,23 +25,109 @@
...
+	org.fontforge.FontForge.appdata.xml		\
+	org.fontforge.FontForge.desktop			\
+	fontforge.xml					\
+	org.fontforge.FontForge.metainfo.xml
+
+install-data-hook: touch-icon-dir
+uninstall-hook: touch-icon-dir
+
+touch-icon-dir:
+	if test -z "$(DESTDIR)"; then $(TOUCH) $(DESTDIR)$(HICOLORDIR) || : ; fi
```
jtanx: Why is this needed?

desktop/Makefile.am
muelli: FTR: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html mentions `appdata` being legacy and `metainfo` being the directory du jour.
This may be related to [this change](https://github.com/fontforge/fontforge/pull/3387#discussion_r244451928) I haven't understood.

JoesCat: On older distro installs, everything is in appdata, should continue to keep appdata in there.
I'll clone another copy into metainfo too.
I've looked at a couple of recent distro /usr/share/metainfo and it's sort of a jumble of both types (bit of a mess in there).

Also removing the 'Extra_DIST = desktop' I added in the main Makefile.am as this isn't the right way to 'make dist' anyways.

desktop/Makefile.am
JoesCat: found that tidbit of info while looking at all sorts of desktop installs:
https://github.com/GNOME/gimp/blob/master/desktop/Makefile.am
It was ADDDESKTOP earlier. This seemed a bit more descriptive.
take it out?

desktop/Makefile.am
JoesCat: @muelli - directory du jour is a good description. Some specs seems to shift with time :-)
To avoid having to re-install, you may want to look at a couple of live distros inside those directories, I'd suggest a couple that have been around for quite a while.... look at knoppix 6, puppy linux, then compare it with a couple of recent ones, like mageia 7, ubuntu, etc. This way you avoid installing unnecessary programs but at the same time get a lay-of-the-land.

desktop/Makefile.am
jtanx: Yeah let's keep it simple and not do this

PR #3389, opened 2018-12-26
Fix the direction of the underline position correction for TTF output.

frank-trampe:
TTF uses the up-positive distance from the baseline to the top of the underline. PostScript uses the up-positive distance from the baseline to the center of the underline. FontForge follows the PostScript convention. The old code was partially non-conformant and caused compounding errors on round-tripping TrueType files. This fixes the problem.

@jtanx, please review. @khaledhosny, please test.

### Important

### Motivation and Context

Per issue #3323, TTF underline position does not round-trip currently, since FontForge adjusts negatively both ways.

Closes #3323.

### Types of changes
It is a bug fix. Correct workflows previously broken ought to work.
Workflows hacked to work around the problem by manipulating the value may break.

### Description

### Final checklist

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -4147,7 +4147,7 @@ static void dumppost(struct alltabs *at, SplineFont *sf, enum fontformat format)
 
     putlong(at->post,shorttable?0x00030000:0x00020000);	/* formattype */
     putfixed(at->post,sf->italicangle);
-    putshort(at->post,sf->upos-sf->uwidth/2);		/* 'post' defn says top of rect, while FontInfo def says center of rect */
+    putshort(at->post,sf->upos+sf->uwidth/2);		/* 'post' defn says top of rect, while FontInfo def says center of rect */
     putshort(at->post,sf->uwidth);
     putlong(at->post,at->isfixed);
     putlong(at->post,0);		/* no idea about memory */


khaledhosny:
I believe the right fix is to just restore the old FontForge code as it was, it was working as intended and nothing was wrong with it. This patch will change the semantics of existing SFD files for no good reason.

frank-trampe:
I believe that the intent was for FontForge to use the PostScript convention. Apparently, I broke it in 2015. See commit 211463509c94cb98d5def5315da69560061e8a4b.

khaledhosny:
Then you need to change the `-` below back to `+` as it was before 2114635, otherwise reading and writing OTF files won’’t round-trip: https://github.com/fontforge/fontforge/blob/085c51a930575613f981f7a4cbfc337a1269d244/fontforge/parsettf.c#L5190

frank-trampe:
The operations can't both be additive. One must add, one must subtract.

khaledhosny:
I’m going to stop here as I’m now completely confused. Though I stand by my original advice, restore all related code to where it was when George Williams touched it, it was correct back then and no fixes were needed.

frank-trampe:
@khaledhosny, I believe that this restores originally intended functionality. I would suggest that you test round-tripping a TTF that was causing problems previously to verify.

frank-trampe:
I have written [a technical bulletin](http://fontforge.github.io/en-US/bulletins/underline_position_2018_1/) to help users fix their files after the next release. I would appreciate eyeballs on that from all present here.

frank-trampe:
Tested and working. @JoesCat, take a look.

frank-trampe:
@skef, since you're online, can you look at this one?

PR #3391, opened 2018-12-28
Fix uninitialized variable in metrics dialog.

frank-trampe:
Fix a problem with an uninitialized static variable breaking the display of metrics adjustment dialogs.

### Important

### Motivation and Context

FVCreateWidth, which is responsible for editing metrics values, initializes the relevant views only if the static pointers are null. The pointers were not initialized, blocking initialization and risking bad accesses. This still causes a one-time memory leak.

### Types of changes

### Description
We now initialize the static array to all null pointers.

### Final checklist

### DIFF
--- a/fontforgeexe/fvmetricsdlg.c
+++ b/fontforgeexe/fvmetricsdlg.c
@@ -125,7 +125,7 @@ static void FVCreateWidth( void *_fv,SplineChar* _sc,void (*doit)(CreateWidthDat
     GGadgetCreateData gcd[11], boxes[2], topbox[2], *hvs[17], *varray[8], *buttons[6];
     GTextInfo label[11];
     static CreateWidthDlg cwd;
-    static GWindow winds[5];
+    static GWindow winds[5] = {NULL, NULL, NULL, NULL, NULL};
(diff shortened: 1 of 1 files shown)


skef:
Just sticking my head in here:

This change won't cause any harm, of course. However, it's a bit mysterious why it would fix any existing problem. The C language spec (eg C99 6.7.9 paragraph 10) requires that static and global pointers be initialized to NULL (other types to zero), and that each element of an array be initialized accordingly. So unless there is some nonconforming compiler, the explicit initialization should have no effect.

frank-trampe:
I'm using gcc, but this change fixed the problem that I was having somehow. I wonder whether the way GWindow is defined impacts it somehow. Can you point to the relevant part of the specification?

frank-trampe:
Hmm. It's pretty clear on that. Maybe we have a compiler flag set somewhere that makes gcc work outside the specification. I definitely had the problem before the change and not afterwards, and the Makefile was the same in both cases.

frank-trampe:
I'm merging. We'd need to do a lot of digging to figure out what's happening, and this seems to be a harmless fix.

jtanx:
Oops wrong pr

Issue #3271, opened 2018-03-17
Exporting SVG clips bottom of the glyph


iyoossaev:
While exporting a single glyph (in my case it's uni4E1C and all others from CJK ideographs from the SimSun TTF font) to SVG, the glyph seems to be wronly placed/centered/clipped. This issue only manifests when exporting to SVG; exporting the same glyph to PDF or PNG works fine. Examples in the attached archive.

Issues experienced with the latest FontForge version (31-Jul-2017) on a machine running Windows 10.

To reproduce this behavior, open the easily available SimSun font, double-click on a glyph and use File → Export (as SVG)
[FontForge problems.zip](https://github.com/fontforge/fontforge/files/1821810/FontForge.problems.zip)

shivekkhurana:
Hi @MrVocabulary 

Was this issue resolved for you? I'm facing the same problem. 

```
$ fontforge -v
Copyright (c) 2000-2019. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Version: 20190801
 Based on sources from 12:50 UTC 14-Nov-2019-D.
fontforge 20190801
build date: 12:50 UTC 14-Nov-2019
```

MrVocabulary:
Hi @shivekkhurana,

unfortunately not :(

skef:
The vertical metrics for the viewBox are determined by the ascender and decender metrics of the font and changing that (at least as a default) would screw up a lot of workflows.

Now that we have flags for glyph-level export we could add one for this case but it would be good to know how and why it's important, given that:

1. The glyphs of most fonts fit into those metrics and
2. The problem is easily worked around using other tools. (For example, I just opened your example in Inkscape, went to File -> Document Properties ... and pressed the "Resize page to drawing or selection" button and got the desired result.

shivekkhurana:
My use case needs me to generate svg version of a text in multiple fonts. 

To avoid embedding fon/css in svg, my plan is to convert all font characters to svg and then combine multiple characters together to form the text again.

I need to convert 100s of fonts to svgs for this purpose. Doing it manually isn't very efficient. 

Re 1: I have tried multiple fonts and the letters "g", "y" etc are getting clipped at the bottom
Re 2: It'll not be efficient to export each glyph in a font manually using Inkscape for my particular use case

[...]

PR #3395, opened 2019-01-04
Fix viewBox in exported SVG glyph

skef:
This push request started out as a fix for the vertical alignment of the ViewBox in exported glyph SVGs, and the earlier comments relate to that issue. Request #3428 now fills that role and has a copy of the original description that appeared here. 

This request can now serve as a place to work out what the exported format *should* be, and perhaps could be given sufficient warning. Some desirable features are:

1. All of every glyph, or all of almost every glyph, is visible when opening the SVG in a standard tool. 
2. The left and right side bearings:
    1. are maintained by an export/import cycle through SVG
    2. can be extracted by a script without too much difficulty.
    3. will or could be displayed in sister tools
3. The italic slant and perhaps other metrics are also "available" in the SVG.

@ctrlcctrlv has suggested adding sodipodi:guides as a solution and that does seem like a possibility. This would involve:

1. Adjusting ViewBox parameters 1 and 3 to display the entire glyph
2. Placing guides with reference points at (FontForge space) zero-Y and X points at zero X (corresponding to the start) and `sc->width` (corresponding to the right side bearing). The second X value could then be converted the RSB coordinate with minimal difficulty. 
3. Angle both guides at the italic angle of the font. 
4. On import use the value from 2 to set the RSB (and continue to use the SVG zero-relative placement to determine the LSB). 

The main thing standing in the way of a prototype of this proposal is that sodipodi:guide appears to be a sub-part of sodipodi:namedview, and the latter has a lot of aspects we might not want to specify. (It doesn't help that the DTD is no longer at its documented location and I haven't tracked it down yet -- it may never have existed.)  My initial impression is that these guides may not be very "separable", making them a less attractive solution. 

Vertical display is a separate and more limited problem. I've run into at least one font with accented capital characters that exceed font's vertical height. Personally I think it would be OK, and possibly beneficial, to continue to cut the tops off of such glyphs with the ViewBox as an indication that an important limit has been exceeded.

### DIFF
--- a/fontforge/svg.c
+++ b/fontforge/svg.c
@@ -1049,9 +1049,11 @@ return( 0 );
 return( ret );
 }
 
+#define SVGMINLRPAD 10
+
 int _ExportSVG(FILE *svg,SplineChar *sc,int layer) {
     char *end;
-    int em_size;
+    int em_size, xstart, xend;
     DBounds b;
 
     SplineCharLayerFindBounds(sc,layer,&b);
@@ -1064,8 +1066,19 @@ int _ExportSVG(FILE *svg,SplineChar *sc,int layer) {
     switch_to_c_locale(&tmplocale, &oldlocale); // Switch to the C locale temporarily and cache the old locale.
     fprintf(svg, "<?xml version=\"1.0\" standalone=\"no\"?>\n" );
     fprintf(svg, "<!DOCTYPE svg PUBLIC \"-//W3C//DTD SVG 1.1//EN\" \"http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd\" >\n" );
-    fprintf(svg, "<svg xmlns=\"http://www.w3.org/2000/svg\" xmlns:xlink=\"http://www.w3.org/1999/xlink\" version=\"1.1\" viewBox=\"0 0 %d %d\">\n",
-	    (int) ceil(sc->width), (int) ceil(em_size));
+    // Adjust horizontal ViewBox to display entire glyph 
+    xstart = floor(b.minx);
+    if (xstart > SVGMINLRPAD)
+	xstart = 0; // Start from origin when sufficiently past it
+    else
+	xstart -= SVGMINLRPAD; // Give glyphs starting near or before the origin some extra space
+    xend = ceil(b.maxx);
+    if (xend < ceil(sc->width) - SVGMINLRPAD)
+	xend = ceil(sc->width); // End at the advance width when sufficiently short of it
+    else
+	xend += SVGMINLRPAD; // Give glyphs ending near or past the advance width some extra space
+    fprintf(svg, "<svg xmlns=\"http://www.w3.org/2000/svg\" xmlns:xlink=\"http://www.w3.org/1999/xlink\" version=\"1.1\" viewBox=\"%d 0 %d %d\">\n",
+	    xstart, xend - xstart, (int) ceil(em_size));
     fprintf(svg, "  <g transform=\"matrix(1 0 0 -1 0 %d)\">\n",
 	    sc->parent->ascent );
     if ( sc->parent->multilayer || sc->parent->strokedfont || !svg_sc_any(sc,layer)) {


skef:
`viewBox` is only parsed twice in the code; once in `svg.c:SVGuseTransform()` and once in `svg.c:ParseSVG()`. In neither case are the first two values stored or used. This push doesn't alter the fourth (height) value. So the question is whether the (new version of the push) width changes will make a difference. 

In the case of `SVGuseTransform()` the answer is straightforwardly "no" because the code only uses that value (`swid`) if there is also a `width` attribute and the export code doesn't add a width. The story in `SVGParseSVG()` is similar: because there is no `width` or `height` attribute both will be set to `1`, which means `width>height` will be false, which means `st.transform[0]` and `[3]` will be initialized with `sheight` (the fourth value) rather than `swidth` (the third). 

Speaking more generally, the SVG import code appears to be intended to work with a variety of code styles (within limits) and just uses the viewBox for scaling hints.

ctrlcctrlv:
If we were to add the X value of the advance and Y value of the start of the descender, it would be more logical to use `<sodipodi:guide>` tags instead, which are the tags used by Inkscape, and due to their long history in the format (over 12 years), are something of a de-facto standard among other SVG editors/tools as well.

skef:
Closes #3271

frank-trampe:
@skef, I'm not sure about the propriety of adding padding. It might break certain assumptions in other software. Maybe ask on the mailing list.

skef:
@frank-trampe: If you're worried about other pieces of software, can you say more about the scope of that worry? Or if you're worried about re-importing the files into FontForge can you say how the analysis of ViewBox in the discussion above is not sufficient? Given that only the ViewBox is changed I would think that verifying that the changes won't effect the use of ViewBox on import would be enough as far as FontForge goes.

ctrlcctrlv:
@frank-trampe I was also quite concerned about that but I've found that imports work the same as always with every SVG I tested. This patch is good IMHO&mdash;it only changes the width, never the height, and the coordinates of the SVG are unchanged.

When I make my fonts I import SVG files and do many round-trips and found this patch not to break my workflow. :-)

Now of course some script out there somewhere may use the ViewBox...but let's please remember:
image

This change is good and should be merged.

frank-trampe:
@skef, I don't know of a particular case to test. But, in the millions of scripts in the world, there are certain to be a few that use the view box as the canonical source of bearings for some purpose, and the arbitrary padding could be confusing. These cases are much more common than those in which the aesthetics of the automatic padding would be useful, I think.

frank-trampe:
@ctrlcctrlv, the automatic padding is conditional and may not be triggering in your test cases. We do not have enough developer time to revisit issues, so I try very hard to keep all changes practically non-regressive. The automatic padding is easily one that would have unforeseen consequences, possibly only realized after the release and requiring another follow-up release.

frank-trampe:
Yes. Let's split the change. I want to get the base fix merged.

skef:
This is where I get to regret The Squash Imperative ...

frank-trampe:
@skef, now that GitHub lets you do it at merge time, there's no need to keep code squashed in the development branch.

skef:
I'm working on this, but looking at the previous code I also want to ask whether and now the present concerns are justified. That code was:

```
    SplineCharLayerFindBounds(sc,layer,&b);
...
    fprintf(svg, "<svg xmlns=\"http://www.w3.org/2000/svg\" xmlns:xlink=\"http://www.w3.org/1999/xlink\" version=\"1.1\" viewBox=\"%d %d %d %d\">\n",
	    (int) floor(b.minx), (int) floor(b.miny),
	    (int) ceil(sc->width), (int) ceil(em_size));
```
The present hypothesis, I take it, is that the existing code allows the left and right side bearings to be extracted by some script. It seems like that's true if a user combines SVG x value 0 and ViewBox parameter 3 but that's ... kind of weird. Shouldn't we at least pin ViewBox parameter 1 at 0 if that's supposed to be the idea? These values don't make sense in relation to one another.

frank-trampe:
Yes. That change from before (regarding the basis) was right. I'm only apprehensive about the conditional adjustments.

skef:
Ok, so the proposal at hand for the ViewBox is: `0 0 ceil(sc->width)  ceil(em_size)`rather than `floor(b.minx) 0 ceil(sc->width) ceil(emsize)`, is that correct?

frank-trampe:
Yes.

skef:
OK. My branch is updated. I've tested with 'f` in AmbrosiaItalic.sfd and it worked as expected.

frank-trampe:
The conditional adjustments of xstart and xend are still in there somehow. But don't delete them here, as we ought to keep the code for comment on the prospective change. Let's make a new pull request with your trimmed changeset.

skef:
> The conditional adjustments of xstart and xend are still in there somehow. But don't delete them here, as we ought to keep the code for comment on the prospective change. Let's make a new pull request with your trimmed changeset.

Lines 1074-1086 in my updated push are a block comment. Are you asking that I do a new push with that block comment removed for more total separation of the two issues?

jtanx:
is there anything left in this pr?

skef:
A simplified version of this went in as #3428. The remaining changes are on old until it can be determined whether they will harm some scripted process.

frank-trampe:
Did we ever ask about this on the mailing list?

skef:
I think we either didn't ask or asked in the sense of "any ideas?" rather than for permission for a specific change. 

I was enthusiastic about @ctrlcctrlv's `sodipodi:guide` suggestion and looked into it a bit. I may be mistaken but my sense was that the XML context for that tag is pretty deep and might not be separable from other directives, so marking the boundaries that way isn't very clean. And I didn't find other standard conventions.

ctrlcctrlv:
XML context?

skef:
@ctrlcctrlv the hierarchy of other tags that tag is expected to be "embedded" in. A given tag is only valid in a given context or contexts.

skef:
(See, for example, this reference: https://gist.github.com/xenomachina/559f1eb49a25b487545a?short_path=dc880fe . One question is what Inkscape would "do" on encountering a mostly-empty sodipodi:namedview tag, without all the `pagecolor`, `document-units` stuff. It might be fine, it might be considered invalid, depending on the XTD (if there *is* an XTD -- there isn't always). `sodipodi-guide` wasn't intended for easy independent use because the coders who added it weren't thinking about our problem.)

ctrlcctrlv:
I tried it, it's fine. SVG is a very forgiving format. I'm not home right now or I'd look in my bookmarks for other software that uses it how you would, last time we talked about this I think I found one.

ctrlcctrlv:
@skef I've determined the minimum required for Inkscape to accept Sodipodi guides without complaining:

```xml
<svg
   xmlns:svg="http://www.w3.org/2000/svg"
   xmlns="http://www.w3.org/2000/svg"
   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
   width="210mm"
   height="297mm"
   viewBox="0 0 210 297">
  <sodipodi:namedview>
    <sodipodi:guide
       position="-27.214285,170.84524"
       orientation="0,1"/>
    <sodipodi:guide
       position="105.83333,195.03572"
       orientation="1,0"/>
  </sodipodi:namedview>
</svg>
```
# Notes 
* Without the width and height on `svg` element, the guides still appear, but in the wrong place. Same for if viewBox is missing.
* Without the `xmlns`, Inkscape complains in `stdout`, but everything still works fine.

ctrlcctrlv:
Are you still interested in pursuing this @skef?

skef:
@ctrlcctrlv: Here's an honest assessment with the caveat that I may have forgotten some of the issues at this point. 

Given that fonts have a specified height I'm OK with cutting stuff above that height out of the SVG bounding box. (Usually if there's anything up there it will be the accent on some precombined character.) It doesn't make sense to me to cut off the left and right sides. That's what my later code changes on this push were trying to fix. 

Round-trip concerns led to the suggestion of recording the width and zero-point somehow. And that led to the further suggestion of using `sodipodi:guide` so that the record could be more generally useful. However:

1. There's as yet no clear "customer" who needs those guides, and
2. My analysis of the code didn't turn up any round-trip problems on the FontForge side.

I think, on reflection, that the zero mark question is a moot point, as the glyph isn't being moved in SVG space, the viewbox value is just being made negative sometimes. There's no reason to always add a guide marking x=0 in SVG space. The reason not to "pad" (to use @frank-trampe's term) therefore seems to boil down to the possibility of SVG scripts that rely on the bounding box marking the width. And, I guess, the hope would be that such people could switch to a different convention if there was one. Maybe those people would happily adopt a `sodipodi:guide` solution, even if it's a bit "wordy". Hard to tell. 

Given that FontForge doesn't appear to set the width that way on import I think it would be fine to extend the viewbox past the width in only those cases that require it and wait for someone to file an issue. That's easy, wouldn't affect roundtrips through FontForge, and is what this PR already does.

frank-trampe:
Okay. As discussed, the original behavior is wrong and also essentially precludes proper use of a font with content outside the bearings, and all of the people who care enough to weigh in (including me) prefer the correct behavior. So I think that we ought to make the change.

skef:
In that case I'll "refresh" and test this code later tonight to make sure it's really in sync with the master and we can get it in.

skef:
I force-pushed but the only difference is using a define for the padding amount. 

To summarize for the record: Because this new change does not move the SVG-coordinate-space x=0 line, and that line is interpreted as the FontForge-coordinate-space x=0 line, this change has no effect on glyph placement when importing. It *would* affect the char width if FontForge used the ViewBox to set that parameter, but FontForge does not currently do that. And anyway the ViewBox is only adjusted outward when it would clip the character; normally there is enough room on both sides and the box will start at x=0 and end at the width.

PR #3398, opened 2019-01-07
cvpalettes.c: Recreate cvtools/bvtools each time it needs to be associated with a new window

### LINKED ISSUE #3397: Tool palette does not move with window focus/use
As I've been looking at other problems I have compiled FontForge with debugging turned on and sort of half-assed my build environment. All that time I've been having one strange problem that I attributed to a poor build. Now that things are closer to release I decided to replicate the OS's build as closely as I could and see if the problem remained. 

It did, so I figured I should open an issue about it. 

Environment: I use Arch Linux with its standard X server and KDE5 Plasma. I synced with the current Arch repository before this latest testing. It provides a version of FontForge it calls 20170731-2 that doesn't have the problem.

Method to reproduce and *expected* behavior (based on the 20170731-2 package): 

Open a font file and then a glyph outline window. The tool palette appears in the latter. Now open a second glyph outline window and the tool palette disappears from the first and appears in the second.
Now foreground the first outline window and the tool palette does not shift back immediately, but does shift once you click anywhere in the window. In this way you can move the the palette back and forth between windows. 

(Note: this is actually a bit of a pain if you have a tool other than the arrow selected, because you can accidentally create points trying to get the palette back, but fine if you remember to click in the ruler area.) 

The problem:

When I compile the current git source (plus my viewBox push -- but I've been seeing this problem for a couple months with different fixes) and follow the procedure above, the tool palette always stays in the first window. Here is a screenshot taken after repeatedly clicking in the "C" window:
ow

I can still switch between tools by going back to the first window or with the menu, but obviously it's a pain to use the program this way. 

Note that I also get these warnings that I doubt are related but could be:

      File "/usr/share/fontforge/python/graphicore.py", line 25, in <module>
         import graphicore.shell
      File "/usr/share/fontforge/python/graphicore/shell.py", line 26, in <module>
        import gtk
      ModuleNotFoundError: No module named 'gtk'

I've tried to fix this in the past but I haven't found a combination of gtk and python 3 in Arch that makes FontForge "happy".

jtanx:
Broken since the reparenting stuff was removed. Having a single set of
palette tools seems like a bad idea...

Fixes #3397

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -6442,6 +6442,12 @@ return( GGadgetDispatchEvent(cv->vsb,event));
 	    ActiveCharView = cv;
 	    if ( cv->gic!=NULL )
 		GDrawSetGIC(gw,cv->gic,0,20);
+
+	    // X11 on Windows is broken, non-active windows
+	    // receive this event on mouseover
+#if !defined(_WIN32) || !defined(FONTFORGE_CAN_USE_GDK)
+	    CVPaletteActivate(cv);
+#endif
 	}
       break;
     }
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -3689,7 +3689,7 @@ void CVChangeSC( CharView *cv, SplineChar *sc )
     GDrawSetWindowTitles8(cv->gw,buf,title);
     CVInfoDraw(cv,cv->gw);
     free(title);
-    _CVPaletteActivate(cv,true);
+    _CVPaletteActivate(cv,true,false);
 
     if ( cv->tabs!=NULL ) {
 	for ( i=0; i<cv->former_cnt; ++i )
@@ -4701,7 +4701,7 @@ static void CVSwitchActiveSC( CharView *cv, SplineChar* sc, int idx )
     GDrawSetWindowTitles8(cv->gw,buf,title);
     CVInfoDraw(cv,cv->gw);
     free(title);
-    _CVPaletteActivate(cv,true);
+    _CVPaletteActivate(cv,true,false);
     
     TRACE("CVSwitchActiveSC() idx:%d\n", idx );
 
@@ -6445,7 +6445,7 @@ return( GGadgetDispatchEvent(cv->vsb,event));
 
 	    // X11 on Windows is broken, non-active windows
 	    // receive this event on mouseover
-#if !defined(_WIN32) || !defined(FONTFORGE_CAN_USE_GDK)
+#if !defined(_WIN32) || defined(FONTFORGE_CAN_USE_GDK)
 	    CVPaletteActivate(cv);
 #endif
 	}
--- a/fontforgeexe/cvpalettes.c
+++ b/fontforgeexe/cvpalettes.c
@@ -3765,7 +3765,7 @@ void CVPalettesRaise(CharView *cv) {
 	GDrawRaise(cvlayers2);
 }
 
-void _CVPaletteActivate(CharView *cv,int force) {
+void _CVPaletteActivate(CharView *cv, int force, int docking_changed) {
     CharView *old;
 
     CVPaletteCheck(cv);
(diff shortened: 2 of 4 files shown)


fontforgeexe/charview.c
```diff
@@ -6442,6 +6442,12 @@ return( GGadgetDispatchEvent(cv->vsb,event));
 	    ActiveCharView = cv;
 	    if ( cv->gic!=NULL )
 		GDrawSetGIC(gw,cv->gic,0,20);
+
+	    // X11 on Windows is broken, non-active windows
+	    // receive this event on mouseover
+#if !defined(_WIN32) || !defined(FONTFORGE_CAN_USE_GDK)
```
skef: Can't vouch for this, but the code matches the comment.

jtanx:
I'll need to make one more change, I don't want to recreate the tools if it's not docked, because it's not required in that case

fontforgeexe/cvpalettes.c
```diff
@@ -3790,7 +3790,19 @@ void _CVPaletteActivate(CharView *cv,int force) {
 	    else
 		SaveOffsets(old->gw,cvlayers,&cvlayersoff);
 	}
-	GDrawSetUserData(cvtools,cv);
+
```
skef: I'm not thrilled with the two calls to `CVPalletteCheck()` (given that the one at the start of the function remains) and the potential for creating, destroying, and creating the palette windows when `force` is true. 

But beyond that pickiness, is the logic right? I presume that `GDrawDestroyWindow()` will throw away the visibility statuses set starting on line 3772. Is that OK? If it is, can't we move the `if ( (old = GDrawGetUserData(cvtools))!=cv || force) {` to the top and put that stuff in an else, eliminating the double `CVPaletteCheck()`? And if it isn't OK, doesn't the code need a broader reorganization?  

(Also, the indentation is wonky in this version.)

fontforgeexe/cvpalettes.c
```diff
@@ -4556,19 +4568,28 @@ void BVPaletteSetVisible(BitmapView *bv,int which,int visible) {
...
 
     BVPaletteCheck(bv);
-    if ( (old = GDrawGetUserData(bvtools))!=bv ) {
+    if ( ((old = GDrawGetUserData(bvtools)) != bv) || force ) {
 	if ( old!=NULL ) {
 	    SaveOffsets(old->gw,bvtools,&bvtoolsoff);
 	    SaveOffsets(old->gw,bvlayers,&bvlayersoff);
 	    SaveOffsets(old->gw,bvshades,&bvshadesoff);
 	}
+
```
skef: The double `BVPaletteCheck()` seems even more superfluous here. `GDrawGetUserData()` is NULL-tolerant so just move the initial call to an `else`.

fontforgeexe/cvpalettes.c
```diff
@@ -4694,32 +4719,13 @@ void BVPaletteChangedChar(BitmapView *bv) {
...
-	    BitmapView *bv = GDrawGetUserData(bvtools);
-	    if ( bv!=NULL ) {
-            GDrawDestroyWindow(bvtools);
-            GDrawDestroyWindow(bvlayers);
-            GDrawDestroyWindow(bvshades);
-            bvtools = bvlayers = bvshades = NULL;
-            BVPaletteActivate(bv);
-	    }
-	}
+
```
skef: Nice

fontforgeexe/cvpalettes.c
```diff
@@ -3791,16 +3791,20 @@ void _CVPaletteActivate(CharView *cv,int force) {
 		SaveOffsets(old->gw,cvlayers,&cvlayersoff);
 	}
 
-    // Now recreate cvtools
-    GDrawDestroyWindow(cvtools);
+    if (palettes_docked || docking_changed) {
```
skef: Not sure this makes sense in the execution context -- does your indentation here reflect your not thinking you are inside the `if ( (old = GDrawGetUserData(cvtools))!=cv || force) {` context? And therefore aren't you already protected from superfluous destruction/creation cycles?

fontforgeexe/cvpalettes.c
```diff
@@ -4579,12 +4583,14 @@ static void _BVPaletteActivate(BitmapView *bv, int force) {
 	    SaveOffsets(old->gw,bvshades,&bvshadesoff);
 	}
 
-    // Recreate the bvtools
-    GDrawDestroyWindow(bvtools);
-    GDrawDestroyWindow(bvlayers);
-    GDrawDestroyWindow(bvshades);
-    bvtools = bvlayers = bvshades = NULL;
-    BVPaletteCheck(bv);
+    if (palettes_docked || docking_changed) {
```
skef: Same "superfluous" question here.

fontforgeexe/cvpalettes.c
jtanx: Nah it's not superfluous, the conundrum is that you only want to recreate the windows if you're in docked mode, because if you recreate it while it's not docked, it will flicker, which is annoying.

So you need to restrict it to recreating only if you're going from docked to another docked window - but if someone has just toggled from not docked to docked, then `palettes_docked` is true - and if that was your only point of reference, you would never recreate the window, which you must do when toggling the docking.

As for the indentation, that's because I'm just using spaces only, vs. tabs + spaces in the old code. Generally in new code I just use spaces, because the old indentation rules are stupid. (there's been discussions in the past of mass reformatting but there's some arguments about it causing issues with git blame - my current thought is just doing it once and for all fwiw).

fontforgeexe/cvpalettes.c
jtanx: Discarding current visibility state is probably ok, since the rest of the activation routines determine what should be visible. As for removing the topmost `CVPaletteCheck`, yeah it's probably possible, but I'd have to study this more to find out the consequences - and that's a problem for another day (or patches welcome).

fontforgeexe/cvpalettes.c
jtanx: At a cursory glance that seems so

fontforgeexe/cvpalettes.c
skef: I don't understand why `GDrawGetUserData(cvtools))!=cv` wouldn't be a good test in those cases too, but the guard shouldn't foul anything up so I guess it doesn't matter much.

frank-trampe:
No objections here.

PR #3401, opened 2019-01-10
Fix for #3347 SIGSEGV on trying to Merge CID Font

### LINKED ISSUE #3347: SIGSEGV on trying to Merge CID Font (Noto Sans Japanese) into other font
When merging Noto Sans Japanese into other font (Noto Sans Latin for example), I get a crash in fvfonts.c:1049 (in FVMergeRefigureMapSel). The issue seems to be that for a lot of characters, `o_sf->glyphs[i]` is NULL. 

I'm not really capable of fixing the error. Instead a working workaround seems to be to flatten the font before merging. Confirmed on current git master (9e9278c66ecc47290accb4e9e4aff82e8578ce1e) at the time, on both Linux and OSX. Also confirmed on version "19:05 UTC 30-Jul-2017" on OSX.

Reproduce, in i.e. Python:

    font = fontforge.open('NotoSans-Bold.ttf')
    font.mergeFonts('Noto-Sans-Japanese-Bold.ttf')

[fonts.zip](https://github.com/fontforge/fontforge/files/2477654/fonts.zip)

- [ ] Screenshot

skef:
Closes #3347 . See that issue for analysis. 

This should definitely be reviewed by someone more familiar with CID Fonts and their implications than I am. For convenience I'm including my test script, an `.sfd` saved after the merge and a `.ttf` generated after the merge in this zip: [bug3347.zip](https://github.com/fontforge/fontforge/files/2743972/bug3347.zip)

### Types of changes

### Description

### Final checklist

.

### DIFF
--- a/fontforge/fvfonts.c
+++ b/fontforge/fvfonts.c
@@ -1087,8 +1087,7 @@ static void _MergeFont(SplineFont *into,SplineFont *other,struct sfmergecontext 
     emptypos = into->glyphcnt;
 
     mapping = malloc(other->glyphcnt*sizeof(int));
-    for (i = 0; i<other->glyphcnt; ++i)
-	mapping[i] = -1;
+    memset(mapping,-1,other->glyphcnt*sizeof(int));
 
     bitmap_into = into->cidmaster!=NULL? into->cidmaster : into;
 
--- a/fontforge/fvfonts.c
+++ b/fontforge/fvfonts.c
@@ -1079,62 +1079,60 @@ static void FVMergeRefigureMapSel(FontViewBase *fv,SplineFont *into,SplineFont *
 
 static void _MergeFont(SplineFont *into,SplineFont *other,struct sfmergecontext *mc) {
     int i, cnt, doit, emptypos, index, k;
-    SplineFont *o_sf, *bitmap_into;
+    SplineFont *bitmap_into;
     BDFFont *bdf;
     FontViewBase *fvs;
     int *mapping;
 
     emptypos = into->glyphcnt;
+
     mapping = malloc(other->glyphcnt*sizeof(int));
-    memset(mapping,-1,other->glyphcnt*sizeof(int));
+    for (i = 0; i<other->glyphcnt; ++i)
+	mapping[i] = -1;
 
     bitmap_into = into->cidmaster!=NULL? into->cidmaster : into;
 
+    /* If other starts as CID it should be flattened before calling */
+    cnt = 0;
     for ( doit=0; doit<2; ++doit ) {
-	cnt = 0;
-	k = 0;
-	do {
-	    o_sf = ( other->subfonts==NULL ) ? other : other->subfonts[k];
-	    for ( i=0; i<o_sf->glyphcnt; ++i ) if ( o_sf->glyphs[i]!=NULL ) {
-		if ( doit && (index = mapping[i])!=-1 ) {
-		    /* Bug here. Suppose someone has a reference to our empty */
-		    /*  char */
-		    SplineCharFree(into->glyphs[index]);
-		    into->glyphs[index] = SplineCharCopy(o_sf->glyphs[i],into,mc);
-		    into->glyphs[index]->orig_pos = index;
-		    if ( into->bitmaps!=NULL && other->bitmaps!=NULL )
-			BitmapsCopy(bitmap_into,other,index,i);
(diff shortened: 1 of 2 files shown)


frank-trampe:
@skef, there are some odd whitespace changes that make it difficult to see what actually changed. Could you perhaps use GitHub comments to annotate and to explain actual changes? Otherwise, I would need to perform a zero-based review on the code, which would be very tedious.

fontforge/fvfonts.c
```diff
@@ -1079,62 +1079,60 @@ static void FVMergeRefigureMapSel(FontViewBase *fv,SplineFont *into,SplineFont *
...
     mapping = malloc(other->glyphcnt*sizeof(int));
-    memset(mapping,-1,other->glyphcnt*sizeof(int));
+    for (i = 0; i<other->glyphcnt; ++i)
+	mapping[i] = -1;
 
     bitmap_into = into->cidmaster!=NULL? into->cidmaster : into;
 
+    /* If other starts as CID it should be flattened before calling */
+    cnt = 0;
     for ( doit=0; doit<2; ++doit ) {
```
skef: Everything below to (green) line 1135 is the same except that the `do ... while()` loop level is removed. The code used to "loop once" with `other` if there were no subfonts and over the subfonts if there were. The new code just uses other as the font. `o_sf` is accordingly eliminated and that string replaced with `other`.

fontforge/fvfonts.c
```diff
@@ -1256,8 +1254,11 @@ return;
     /*  If both are CID then merge each subfont separately */
     if ( fv->sf->cidmaster!=NULL && other->subfonts!=NULL )
 	CIDMergeFont(fv->sf->cidmaster,other,preserveCrossFontKerning);
-    else
+    else {
+	if (other->subfonts!=NULL)
```
skef: Here we flatten if there are subfonts as the comment on (old) line 1254 suggests. 

I take it that the loop level I removed was intended do be equivalent to what `SFFlatten()` does, and perhaps faster or better in some other way. I discuss in the comments on #3347 why the code doesn't work and would be difficult to make work.

fontforgeexe/fontview.c
```diff
@@ -7120,7 +7120,7 @@ static void FontView_ReformatAll(SplineFont *sf) {
     MetricsView *mvs;
     extern int use_freetype_to_rasterize_fv;
 
-    if ( ((FontView *) (sf->fv))->v==NULL || ((FontView *) (sf->fv))->colcnt==0 )			/* Can happen in scripts */
+    if ( sf->fv==NULL || ((FontView *) (sf->fv))->v==NULL || ((FontView *) (sf->fv))->colcnt==0 )			/* Can happen in scripts */
```
skef: This function was called as a side-effect of `SFFlatten()` and wasn't quite set up to be run without a UI.

skef:
> Wait, so does this drop the [broken] code path for merging a CID font into a CID font without flattening?

If I understand your question correctly the answer is no. This change only affects the merge of a CID font with a non-CID font. The case of merging two CID fonts is caught and addressed (how well I don't know) by lines 1257-8 of the existing code and 1255-6 of the new code, which are unchanged.

frank-trampe:
Ah, I see. So `_MergeFont` drops flattening, and we just flattening the incoming font when in that code path?

skef:
> Ah, I see. So `_MergeFont` drops flattening, and we just flattening the incoming font when in that code path?

Yes, that is the essence of the fix.

frank-trampe:
The changes look fine to me in that context. Given the complexity, I'd like @jtanx to double-check before we merge.

fontforge/fvfonts.c
```diff
@@ -1079,62 +1079,60 @@ static void FVMergeRefigureMapSel(FontViewBase *fv,SplineFont *into,SplineFont *
...
+    SplineFont *bitmap_into;
     BDFFont *bdf;
     FontViewBase *fvs;
     int *mapping;
 
     emptypos = into->glyphcnt;
+
     mapping = malloc(other->glyphcnt*sizeof(int));
-    memset(mapping,-1,other->glyphcnt*sizeof(int));
+    for (i = 0; i<other->glyphcnt; ++i)
```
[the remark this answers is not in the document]
jtanx: As mentioned in #3347, memsetting to `-1` should work just fine.

PR #3409, opened 2019-01-13
Miscellaneous python fixes (mostly string conversion)

### LINKED ISSUE #3278: Glyph name is sometimes not in the font, though the glyph was just created
After creating a glyph using Python, the glyph’s name is sometimes not `in` the font. I would expect it to always be `in` the font.

When I run this script:
```python
#!/usr/bin/env python
import fontforge
font = fontforge.font()
for cp in range(0x00, 0xFF + 1):
    name = hex(cp)
    glyph = font.createChar(cp, name)
    glyph.width = 0
    assert name in font, name
font.generate('out.ttf')
```
I get this error:
```pytb
  File "./build.py", line 8, in <module>
    assert name in font, name
AssertionError: 0x1
```

That means it failed on the second iteration of the loop, meaning there is no glyph named “0x1”, though there is a glyph named “0x0”. However, when I comment out the assertion and inspect the resulting font, all 256 glyphs have their expected names.

I am running OS X El Capitan and I am using these versions of Python and FontForge:
```console
$ python --version
Python 2.7.13
$ fontforge --version
 Based on sources from 09:30 UTC 21-Mar-2018-D.
 Based on source from git with hash: 
fontforge 09:30 UTC 21-Mar-2018
libfontforge 20180321
```

### LINKED ISSUE #3300: python: typecheck errors on subtable-name when calling addPosSub
### The Problem
When attempting to call `addPosSub` in order to add an entry to a subtable, you get one of two type errors on the first argument, making it completely unusable at the moment (fully tested with Single Substitution and Ligature lookups)

#### Examples
* If you call it with a `str` you get `TypeError: expected bytes, str found`. For example:
```python
def sample():
    font = ff.activeFont()
    lname = '\'=>\' Single Substitution'
    font.addLookup(lname, 'gsub_single', 0, ())
    font.addLookupSubtable(lname, lname)
    glyph = font['equal']
    # glyph.addPosSub(bytes(lname,'utf-8'),'LIG')
    glyph.addPosSub(lname,'LIG')

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-6f614e1d4e6c> in <module>()
----> 1 sample()

/G/src/Inconsoliga/CreateLigLookup.py in sample()
     48     glyph = font['equal']
     49     # glyph.addPosSub(bytes(lname,'utf-8'),'LIG')
---> 50     glyph.addPosSub(lname,'LIG')
     51 
     52 def createLookup(ligname, parts=None, chars=None):

TypeError: expected bytes, str found
```

* But if you provide the argument as `bytes` you get `argument 1 must be str, not bytes`
```python
def sample():
    font = ff.activeFont()
    lname = '\'=>\' Single Substitution'
    font.addLookup(lname, 'gsub_single', 0, ())
    font.addLookupSubtable(lname, lname)
    glyph = font['equal']
    glyph.addPosSub(bytes(lname,'utf-8'),'LIG')
    # glyph.addPosSub(lname,'LIG')

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-6f614e1d4e6c> in <module>()
----> 1 sample()

/G/src/Inconsoliga/CreateLigLookup.py in sample()
     47     font.addLookupSubtable(lname, lname)
     48     glyph = font['equal']
---> 49     glyph.addPosSub(bytes(lname,'utf-8'),'LIG')
     50     # glyph.addPosSub(lname,'LIG')
     51 

TypeError: argument 1 must be str, not bytes
```

This happens both when called directly from python with the extension installed, and when called as a script with fontforge. Same thing happens if instead of creating a Single Substitution lookup I create a Ligature one for example. In no case does the function call get past the parsing of the subtable name: If for example I provide a `bytes` value that doesn't match any existing subtable it does report so with an error saying `Unknown lookup subtable`

### Environment
Arch Linux 4.8.14-1 
python-3.6.5
fontforge 1:20170731.r230.ged2aac4d8-1 configured with ```--with-x \
        --prefix=/usr \
        --enable-shared \
        --enable-gtk2-use \
        --mandir=/usr/share/man \
        --with-freetype-source="$srcdir"/freetype-2.6.1
        --enable-pyextension```

skef:
Glyph name race condition
Typecheck errors on addPosSub
Fix for horizontalBaseline and verticalBaseline (including doc)
Fix ParseLookupFlagsItem
Fix ParseLookupFlags

Closes #3278 
Closes #3300 

I have only tested these changes with the python on my system. However, the new code conforms to that used in other functions that would have frequently reported bugs if they were not working.

### Types of changes

### Final checklist

### DIFF
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -3326,8 +3326,7 @@ pen = None;				# Finalize the pen. This tells FontForge
 	extent</CODE> tuples.
 	<P>
 	A <CODE>feature extent</CODE> tuple is<BR>
-	<CODE>(feature-tag,min-extent,max-extent, (tuple of feature
-	extents))</CODE><BR>
+	<CODE>(feature-tag,min-extent,max-extent)</CODE><BR>
 	<CODE>feature-tag</CODE> is a 4 letter string specifying an opentype (GPOS
 	or GSUB) feature tag,<CODE>min</CODE>/<CODE>max-extent</CODE> are numbers
 	specifying how far above and below the baseline characters go in this
@@ -3339,14 +3338,18 @@ pen = None;				# Finalize the pen. This tells FontForge
   (("cyrl","romn",(1405,-288,0),()),
    ("grek","romn",(1405,-288,0),()),
    ("latn","romn",(1405,-288,0),
-    ("dflt",-576,1913,
-     ("NoAc",-576,1482))
-    ("ENG ",-576,1482,())
+    (("dflt",-576,1913,
+      (("NoAc",-576,1482),
+       ("ENG ",-576,1482))
+     ),
+    )
    )
   )
 )
 </PRE>
 	</BLOCKQUOTE>
+	(Note: The comma after the <CODE>dflt</CODE> tuple puts it into a
+	one-element tuple.)
       </TD>
     </TR>
     <TR>
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -515,7 +515,7 @@ static uint32 StrToTag(char *tag_name, int *was_mac) {
     int feat, set;
 
     if ( tag_name==NULL ) {
-	PyErr_Format(PyExc_TypeError, "OpenType tags must be represented strings" );
+	PyErr_Format(PyExc_TypeError, "OpenType tags must be represented as strings" );
 return( BAD_TAG );
     }
 
@@ -547,6 +547,16 @@ return( BAD_TAG );
 	}
     }
 return( (foo[0]<<24) | (foo[1]<<16) | (foo[2]<<8) | foo[3] );
+}
+
+static uint32 StrObjToTag(PyObject *obj, int *was_mac) {
+    char *str;
+    uint32 r;
+
+    PYGETSTR(obj, str, BAD_TAG);
+    r = StrToTag(str, was_mac);
+    ENDPYGETSTR();
+    return r;
 }
 
 static PyObject *TagToPythonString(uint32 tag,int ismac) {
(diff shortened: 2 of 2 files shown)


skef:
Hmm. The test script failed on OS X so someone should look at that. 

I don't have access to a Mac right now, but with remote access of some kind I could probably pin down the problem.

doc/html/python.html
```diff
@@ -3339,14 +3338,18 @@ <H3>
   (("cyrl","romn",(1405,-288,0),()),
    ("grek","romn",(1405,-288,0),()),
    ("latn","romn",(1405,-288,0),
-    ("dflt",-576,1913,
```
frank-trampe: What's happening here?

frank-trampe:
This looks good to me, but I'd like to get a second opinion from @jtanx before we merge.

doc/html/python.html
skef: Basically: 

1. I looked through `python.c` for code with suspicious string conversions and this function had some.
2. I looked for documentation on the interface so that I could test all the conversions and found this description.
3. I got the code so that I thought it should work but it couldn't parse the documented example. 
4. After trying and failing to get the interface to work many times I used the GUI to enter identical parameters into a font and then read it out using the same interface. It spit out something close to what I have documented. 
5. After looking back at the GUI and the working case I decided it made sense and the documentation didn't, so I edited the documentation to match the implementation.

skef:
> This looks good to me, but I'd like to get a second opinion from @jtanx before we merge.

If this code is right then the test I added in #3410 should work, but they seem to fail on Mac OS X. I don't currently have access to the relevant hardware so I don't know why.

fontforge/python.c
```diff
@@ -12290,6 +12305,8 @@ return( -1 );
 	    BaseFree(base);
 return( -1 );
 	}
+	bs = chunkalloc(sizeof(struct basescript));
+	bs->next = NULL;
```
jtanx: this shouldn't be needed since you just chunkalloc'ed it (which uses calloc)

fontforge/python.c
```diff
@@ -12290,6 +12305,8 @@ return( -1 );
 	    BaseFree(base);
 return( -1 );
 	}
+	bs = chunkalloc(sizeof(struct basescript));
```
jtanx: wow, I guess this never worked before?

fontforge/python.c
frank-trampe: I somehow missed all the chunkalloc calls. I guess I ought to have looked more closely.

fontforge/python.c
skef: Correct. If I'm remembering right there are paths through the function that worked but anything that reached this line would have failed.

skef: Hmm, actually maybe only error-producing paths worked. 

Anyway a number of the bugs I've fixed had implementations that could never have worked. #3355 was another example in the same file.

PR #3410, opened 2019-01-13
Python string conversion tests

skef:
These are some rudimentary tests I wrote while preparing the #3409 fixes, integrated into the "make check" hook. Not all correspond to problems -- the comment checks just verified the existing code. 

I did not include this change with the fix push because:

1. I'm not really familiar with the test infrastructure or FontForge testing more generally.
2. I can imagine additional tests creating problems even when the fixes are correct. (Obscure versions of python could be one cause of such problems.) 

So I'll leave it up to more experienced hands what to do with these. If nothing else they should help clarify what #3409 addresses and how to test that code.

### DIFF
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -3326,8 +3326,7 @@ pen = None;				# Finalize the pen. This tells FontForge
 	extent</CODE> tuples.
 	<P>
 	A <CODE>feature extent</CODE> tuple is<BR>
-	<CODE>(feature-tag,min-extent,max-extent, (tuple of feature
-	extents))</CODE><BR>
+	<CODE>(feature-tag,min-extent,max-extent)</CODE><BR>
 	<CODE>feature-tag</CODE> is a 4 letter string specifying an opentype (GPOS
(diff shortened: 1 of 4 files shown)


jtanx:
So the reason it was passing on linux but not mac/windows is that the linux ci build runs against python 3, but mac/windows run against python2.

skef:
> So the reason it was passing on linux but not mac/windows is that the linux ci build runs against python 3, but mac/windows run against python2.

That's better for coverage purposes. Almost all of my python experience is with 3 so I suppose I should have guessed the problem would be something like what it was. 

If the updated test does work on 2 then we're likely in good shape, right?

jtanx:
Yep, no complaints from me

skef:
@jtanx : As a separate question, does our building the Mac OS X version against 2 have any relation to how it is likely to be built and/or used by end-users? It's getting a bit late in the day to be putting out new releases against 2. 

(I guess Windows raises the same question, but I assume the issues are more complex and fraught in that context.)

PR #3413, opened 2019-01-14
Re-enable tooltips (aka "popup windows") in GXdraw

skef:
I mentioned in #3397 that I wasn't seeing any tooltips in the more recent builds. So I figured I should look into that too. 

The function responsible for displaying a tooltip after the timer goes off is `GGadgetPopupTest()`, which starts like this:
```
static int GGadgetPopupTest(GEvent *e) {
    unichar_t *msg;
    int lines, temp, width;
    GWindow root = GDrawGetRoot(GDrawGetDisplayOfWindow(popup));
    GRect pos, size;
    unichar_t *pt, *ept;
    int as, ds, ld, img_height=0;
    GEvent where;

    if ( e->type!=et_timer || e->u.timer.timer!=popup_timer || popup==NULL )
return( false );
    popup_timer = NULL;

    /* Is the cursor still in the original window? */
    if (GDrawGetPointerWindow(root) != popup_info.gw)
        return true;
```
The last two lines are a recent change/addition. 

`GDrawGetPointerWindow()` and its innards are:
```
static Window _GXDrawGetPointerWindow(GWindow w) {
    GXWindow gw = (GXWindow) w;
    Display *display = gw->display->display;
    int junk;
    Window parent, child, wjunk;
    int x, y; unsigned int state;

    parent = gw->display->groot->w;
    for (;;) {
        child = None;
        if ( !XQueryPointer(display,parent,&wjunk,&child,&junk,&junk,&x,&y,&state))
    break;
        if ( child==None )
    break;
        parent = child;
    }
return( parent );
}

static GWindow GXDrawGetPointerWindow(GWindow w) {
    GXWindow gw = (GXWindow) w;
    Display *display = gw->display->display;
    void *ret;
    Window parent;

    parent = _GXDrawGetPointerWindow(w);
    if ( (gw->w&0xfff00000) == (parent&0xfff00000)) {
        /* It is one of our windows, so it is safe to look for it */
        if ( XFindContext(display,parent,gw->display->mycontext,(void *) &ret)==0 )
return( (GWindow) ret );
    }
return( NULL );
}
```
Note that `_GXDrawGetPointerWindow()` pulls the display and then the root window from the
argument, and `GXDrawGetPointerWindow()` has that bitmask-based test for whether the window ID maps to the same application. Basically `GXDrawGetPointerWindow()` is designed to be called with some application window rather than the root window. 

This change calls it with `popup_info.gw` (which given that the code pulls the root window from the argument is as good as anythng), which reenables tooltips. 

### Types of changes

### Final checklist

### DIFF
--- a/gdraw/ggadgets.c
+++ b/gdraw/ggadgets.c
@@ -554,7 +554,7 @@ return( false );
     popup_timer = NULL;
 
     /* Is the cursor still in the original window? */
-    if (GDrawGetPointerWindow(root) != popup_info.gw)
+    if (GDrawGetPointerWindow(popup_info.gw) != popup_info.gw)
         return true;
 
     lines = 0; width = 1;


skef:
Hmm. I guess I should test this with gdk, huh?

jtanx:
Nice, thanks! Seems to work on GDK going off the appveyor windows build.

PR #3417, opened 2019-01-17
Import/Export dialog fixes

### LINKED ISSUE #551: Import Dialog: Format is not Remembered
If I have previously selected _SVG_ as format in the import dialog, then on the next import, although _SVG_ is still selected, I can't see any SVG files unless I reselect _SVG_ in the drop down format menu.

skef:
# Dialog sizing 

The CharView (and bitmap view) -> Export dialog always comes up with a strange size that is much too short vertically. It's easy enough to resize but that shouldn't be necessary. The import dialog doesn't have the same problem.

The code in `_Import()` and `_Export()` is quite similar, so for the most part this change just adds some lines from the former that are missing in the latter. It also slightly increases the default width so that the right margin matches the left. 

I've tested this with X and GDK backends on Arch Linux. 

Does this work well with a modified `ScreenWidthInches`? Nope. But the Export dialog situation in that case is no better or worse than the Import dialog situation.

# Format logic

The logic to record import format was using the offset in the pulldown rather than recording the format enum itself, and this wasn't reliable in some cases (such as when the spiro format was disabled). With this push the format and position are tracked separately. That also removes the need for a separate `wildchr` table. 

The export dialog has no wildcards so its logic is simpler. However, some of the existing logic dated to when there were only two bitmap formats. Since the bitmap formats are all available in the outline dialog I updated the logic to just use the single table and format identifier, subtracting the table offset as needed. 

I tested these changes on Arch Linux in many combinations to verify that the format is remembered separately, the import dialog remembers and updates the file type and list correctly, and the export format is remembered (separately for charview and bitmap view) and filename is updated correctly. 

Closes #551 

### Types of changes

### Final checklist

### DIFF
--- a/fontforgeexe/cvexportdlg.c
+++ b/fontforgeexe/cvexportdlg.c
@@ -233,19 +233,6 @@ struct gfc_data {
     int layer;
 };
 
-
-static GTextInfo bcformats[] = {
-/* 0=*.xbm, 1=*.bmp, 2=*.png, 3=*.xpm, 4=*.c(fontforge-internal) */
-    { (unichar_t *) N_("X Bitmap"), NULL, 0, 0, (void *) 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, '\0' },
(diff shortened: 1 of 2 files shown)


frank-trampe:
I can't necessarily verify all of the logic, but I don't see anything that would pass tests and still be wrong.

PR #3418, opened 2019-01-18
Edit discontinued utility

Marr11317:
Autotrace is discontinued since 2004: https://alternativeto.net/software/autotrace/


ctrlcctrlv:
It still works so...why?

Marr11317:
So people know they won't get help with it.
But actually, I think you're right: discontinued it a bit rough, no one will use it anymore if we add it.

PR #3424, opened 2019-01-24
Fix for #3133 Can't open subfont with name containing parentheses

### LINKED ISSUE #3133: can't open font that has parentheses in its name
In Python

    font = fontforge.open(file_name + '(' + font_name + ')')

where `file_name == 'Automatica -BRK-xy.ttf'` and `font_name == 'Automatica (BRK)xy'` throws `EnvironmentError: Open failed`. `fontforge.open(file_name)` works fine, and this is the only font I have that has parentheses in its name which is why I think the parentheses are the problem.

[Automatica -BRK-xy.zip](https://github.com/fontforge/fontforge/files/1216914/Automatica.-BRK-xy.zip)

skef:
I replumbed the TTF path a bit so that the suffix convention isn't so intrusive. Unfortunately that was just too convoluted for `macbinary.c` so I replaced the "parsing" code where it was. 

There is a `.ttc` in [NotoSans.zip](https://github.com/fontforge/fontforge/files/2790233/NotoSans.zip) that can be tested with this:

```
#!/usr/bin/env python

import sys
import fontforge

for i, s in enumerate(('Noto (Sans)', 'Noto (Sans) UI')):
    for k in (str(i), s):
        f = fontforge.open(sys.argv[1] + '(' + k + ')')
        print(f.fullname)
        if f.fullname != s:
            raise ValueError('Expected fontname "' + s + '" instead of "' + f.fontname + '" for key "' + k + '"')
        f.close()
```
None of this will work if the embedded parentheses are not balanced, but it won't (or shouldn't) crash. 

I added the test because although the code path is tested in the internal scripting language that test

1. Doesn't try opening by offset and
2. Doesn't verify that the right subfont was opened. 

Closes #3133

### Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

### Final checklist

### DIFF
--- a/fontforge/macbinary.c
+++ b/fontforge/macbinary.c
@@ -2080,8 +2080,8 @@ static SplineFont *SearchTtfResources(FILE *f,long rlistpos,int subcnt,long rdat
     SplineFont *sf;
     int which = 0;
     char **names;
-    char *pt,*lparen, *rparen;
-    char *chosenname=NULL;
+    char *pt,*lparen;
+    char *find=NULL,*chosenname=NULL;
 
     fseek(f,rlistpos,SEEK_SET);
     if ( subcnt>1 || (flags&ttf_onlynames) ) {
@@ -2106,15 +2106,9 @@ static SplineFont *SearchTtfResources(FILE *f,long rlistpos,int subcnt,long rdat
 return( (SplineFont *) names );
 	}
 	if ((pt = strrchr(filename,'/'))==NULL ) pt = filename;
-	/* Someone gave me a font "Nafees Nastaleeq(Updated).ttf" and complained */
-	/*  that ff wouldn't open it */
-	/* Now someone will complain about "Nafees(Updated).ttc(fo(ob)ar)" */
-	if ( (lparen = strrchr(pt,'('))!=NULL &&
-		(rparen = strrchr(lparen,')'))!=NULL &&
-		rparen[1]=='\0' ) {
-	    char *find = copy(lparen+1);
-	    pt = strchr(find,')');
-	    if ( pt!=NULL ) *pt='\0';
+	if ( lparen = SFSubfontnameStart(pt) ) {
+	    find = copy(lparen+1);
+	    find[strlen(find)-1] = '\0';
 	    for ( which=subcnt-1; which>=0; --which )
 		if ( strcmp(names[which],find)==0 )
 	    break;
@@ -2180,7 +2174,7 @@ return( (SplineFont *) names );
 	    len += temp;
 	}
 	rewind(ttf);
-	sf = _SFReadTTF(ttf,flags,openflags,NULL,NULL);
+	sf = _SFReadTTF(ttf,flags,openflags,NULL,NULL,NULL);
 	fclose(ttf);
 	if ( sf!=NULL ) {
 	    free(buffer);
@@ -2577,10 +2571,10 @@ static FOND *PickFOND(FOND *fondlist,char *filename,char **name, int *style) {
     char *find = NULL;
 
     if ((pt = strrchr(filename,'/'))!=NULL ) pt = filename;
-    if ( (lparen = strchr(filename,'('))!=NULL && strchr(lparen,')')!=NULL ) {
+    if ( lparen = SFSubfontnameStart(pt) ) {
 	find = copy(lparen+1);
-	pt = strchr(find,')');
-	if ( pt!=NULL ) *pt='\0';
+	find[strlen(find)-1] = '\0';
+	find = copy(lparen+1);
 	for ( test=fondlist; test!=NULL; test=test->next ) {
 	    for ( i=0; i<48; ++i )
 		if ( test->psnames[i]!=NULL && strcmp(find,test->psnames[i])==0 ) {
@@ -2810,7 +2804,7 @@ return( (SplineFont *) ret );
 	dlen -= len;
     }
     rewind(temp);
-    sf = _SFReadTTF(temp,flags,openflags,NULL,NULL);
+    sf = _SFReadTTF(temp,flags,openflags,NULL,NULL,NULL);
     fclose(temp);
     free(buffer);
 return( sf );
--- a/fontforge/macbinary.c
+++ b/fontforge/macbinary.c
@@ -2574,7 +2574,6 @@ static FOND *PickFOND(FOND *fondlist,char *filename,char **name, int *style) {
     if ( lparen = SFSubfontnameStart(pt) ) {
 	find = copy(lparen+1);
 	find[strlen(find)-1] = '\0';
-	find = copy(lparen+1);
 	for ( test=fondlist; test!=NULL; test=test->next ) {
 	    for ( i=0; i<48; ++i )
 		if ( test->psnames[i]!=NULL && strcmp(find,test->psnames[i])==0 ) {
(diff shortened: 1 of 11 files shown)


fontforge/splinefont.c
```diff
@@ -931,14 +931,37 @@ return(copy(tmpfilename));			/* The filename does not exist */
...
 
+/* The following won't work if the fontname has unbalanced
+ * parentheses, but short of requiring those to be escaped
+ * somehow that case is tricky. (One could guess the cutoff
+ * by looking for a file extension.)
+ * */
+char *SFSubfontnameStart(char *fname) {
+    char *rparen, *paren;
+    int cnt = 1;
+    if ( fname==NULL || (rparen = strrchr(fname,')'))==NULL || rparen[1]!='\0' )
```
frank-trampe: So fname is invalid if the closing parenthesis occurs before the last character? Why is that?

fontforge/macbinary.c
```diff
@@ -2106,15 +2106,9 @@ static SplineFont *SearchTtfResources(FILE *f,long rlistpos,int subcnt,long rdat
...
-	/* Now someone will complain about "Nafees(Updated).ttc(fo(ob)ar)" */
-	if ( (lparen = strrchr(pt,'('))!=NULL &&
-		(rparen = strrchr(lparen,')'))!=NULL &&
-		rparen[1]=='\0' ) {
-	    char *find = copy(lparen+1);
-	    pt = strchr(find,')');
-	    if ( pt!=NULL ) *pt='\0';
+	if ( lparen = SFSubfontnameStart(pt) ) {
+	    find = copy(lparen+1);
+	    find[strlen(find)-1] = '\0';
```
frank-trampe: It looks like SFSubfontnameStart always returns `NULL` or a string of length greater than zero. Is that the basis of this assumption? If so, I would put that in a comment so that nobody worries about a segfault on `""`.

fontforge/macbinary.c
```diff
@@ -2577,10 +2571,10 @@ static FOND *PickFOND(FOND *fondlist,char *filename,char **name, int *style) {
     char *find = NULL;
 
     if ((pt = strrchr(filename,'/'))!=NULL ) pt = filename;
-    if ( (lparen = strchr(filename,'('))!=NULL && strchr(lparen,')')!=NULL ) {
+    if ( lparen = SFSubfontnameStart(pt) ) {
+	find = copy(lparen+1);
+	find[strlen(find)-1] = '\0';
 	find = copy(lparen+1);
```
frank-trampe: This line shows up twice and might cause a memory leak.

frank-trampe:
@skef, would you explain a bit more about the origin and use of `info-chosenname` value?

fontforge/splinefont.c
skef: I added a comment that explains this, which was a preexisting convention. The code had been:
```
    if ( (paren = strrchr(pt,'('))!=NULL &&
	    (rparen = strrchr(paren,')'))!=NULL &&
	    rparen[1]=='\0' ) {
```

The restriction that the subfontname come at the end of the string likely corresponds to the possibility of a font filename containing parentheses, given that they are valid characters in most filesystems.

fontforge/macbinary.c
skef: Added comment about this for `SFSubfontnameStart`

fontforge/macbinary.c
skef: Removed

skef:
> @skef, would you explain a bit more about the origin and use of `info-chosenname` value?

All of these changes ultimately have to do with the paragraph starting "When opening a TTC file, or a mac dfont" in the [File -> Open](http://fontforge.github.io/en-US/documentation/interface/filemenu/#open) documentation. `ttfinfo.chosenname` is a preexisting field that stores the name of the subfont chosen by the user in one of the ways indicated in the (also preexisting) comment. 

What I've done in the reorganization of the TTF code is parse the filename once at the top and then pass in the "chosen name" (if any) via that same argument. If it was really chosen by name, the value will be freed and replaced but should be identical. If the convention was by number it will be replaced with the name corresponding to that offset in the subfont array. 

I would have made the same changes in `macbinary.c` except that the code paths were too convoluted.

frank-trampe:
I'm normally averse to changing the arguments for a function (in order to preserve patch-sharing with FontForge forks), but this improves readability and maintainability in a way that more than justifies a little hassle for downstream projects, in my mind. @jtanx, @JoesCat, are you okay with this?

skef:
In this case I felt that what is really a UI convention was dipping *way* too far into the guts of TTF processing and it seemed worth trying to separate. (The fact that the old code reparsed the name two or sometimes three times wasn't great either.) That I didn't manage to do the same in `macbinary.c` still bothers me a little.

frank-trampe:
Uh oh, @skef. We have a merge conflict as a result of the Python test changes.

jtanx:
> I'm normally averse to changing the arguments for a function (in order to preserve patch-sharing with FontForge forks), 

That rarely, if ever happens, so I'm completely fine with that.

skef:
> Uh oh, @skef. We have a merge conflict as a result of the Python test changes.

This was just from both fixes adding a test (adjacent lines)

frank-trampe:
@jtanx, I'm glad we're on the same page about the indentation style.

fontforge/splinefont.c
JoesCat: > The restriction that the subfontname come at the end of the string likely corresponds to the possibility of a font filename containing parentheses, given that they are valid characters in most filesystems.

Filenames should never contain parentheses because the file operating systems sees this as a "list".
The fontforge open command looks at this differently as mentioned in #3133 
```code
fontforge.open(file_name + '(' + font_name + ')')
```
file_name == 'Automatica -BRK-xy.ttf' and font_name == 'Automatica (BRK)xy'

skef:
@jtanx it seems better than just using a random style I prefer. Maybe a style for new code could be documented in CONTRIBUTING.md?

fontforge/splinefont.c
jtanx: > Filenames should never contain parentheses because the file operating systems sees this as a "list".

What, that doesn't make sense, the file system doesn't care, it's your bash/shell interpreter that's doing that. Escape properly and you can use braces all you like.

fontforge/macbinary.c
JoesCat: When you create code, take a look at what is compiled, and not simply looking for an X or a checkmark saying it passes.... for example, if you follow the github checkmark and look at what was compiled, you'll see this should be avoided:
```code
  if ( lparen = SFSubfontnameStart(pt) ) {
       ^~~~~~
2574:10: warning: suggest parentheses around assignment used as truth value [-Wparentheses]
```

JoesCat:
> I'm normally averse to changing the arguments for a function (in order to preserve patch-sharing with FontForge forks), but this improves readability and maintainability in a way that more than justifies a little hassle for downstream projects, in my mind. @jtanx, @JoesCat, are you okay with this?

I was reading some of the conversation, and should point-out...
As long as the parentheses is part of the "subfont", and not part of the "filename" because the operating system sees parentheses as a list: [posix](https://unix.stackexchange.com/questions/26063/how-are-parentheses-interpreted-at-the-command-line), [windows](https://www.google.com/search?q=windows+command+line+parentheses)
If the subfont is going to be saved/exported as a file, the parentheses need to be converted to another character, like maybe "-" to avoid having the part within parentheses seen as a list.

PR #3426, opened 2019-01-26
Fix for #3425 SVG causes NaN value in spline creation

### LINKED ISSUE #3425: SVG icon causes FontForge to crash NaN value in spline creation
FontForge version: `fontforge 12:31 UTC  3-Oct-2018, libfontforge 20181003`

### To reproduce:

1. Open FontForge
1. Open the "A" glyph by double-clicking
1. Choose "File" "Import"
1. Choose the `hand-middle-finger.svg` [attached here and yes, we see the irony too](https://github.com/fontforge/fontforge/files/2798081/hand-middle-finger.svg.zip)

### Result:

FontForge crashes (segfault I believe)

We also see this in our Python-based script:

```
Internal Error: NaN value in spline creation
Internal Error: No Solution
```

This SVG file has gone through some cleaning procedures as we prepare it for our build step as part of creating a new release of Font Awesome. We know it has to be something related to that but the root cause is hard for us to determine.

We can get around this issue by using the un-altered / un-cleaned up version. FontForge has no problem with that one.

- [ ] Screenshot

skef:
In the comments for #3425 I note that the guts of `SVGTraceArc()` are based on section [F.6.5](https://www.w3.org/TR/SVG11/implnote.html#ArcConversionEndpointToCenter) of the SVG 1.1 spec, which corresponds to section [B.2.4](https://www.w3.org/TR/SVG2/implnote.html#ArcConversionEndpointToCenter) of the SVG 2 spec. There are no substantial differences between the two. 

The current master has this code:
```
    if ( rx!=0 && ry!=0 ) {
        /* Page 647 in the SVG 1.1 spec describes how to do this */
        /* This is Appendix F (Implementation notes) section 6.5 */
        cosr = cos(axisrot); sinr = sin(axisrot);
        x1p = cosr*(current->x-x)/2 + sinr*(current->y-y)/2;
        y1p =-sinr*(current->x-x)/2 + cosr*(current->y-y)/2;
        /* Correct for bad radii */
        lambda = x1p*x1p/(rx*rx) + y1p*y1p/(ry*ry);
        if ( lambda>1 ) {
           lambda = sqrt(lambda);
           rx *= lambda;
           ry *= lambda;
        }
        factor = rx*rx*ry*ry - rx*rx*y1p*y1p - ry*ry*x1p*x1p;
        if ( RealNear(factor,0))
            factor = 0;         /* Avoid rounding errors that lead to small negative values */
        else
            factor = sqrt(factor/(rx*rx*y1p*y1p+ry*ry*x1p*x1p));
        if ( large_arc==sweep )
            factor = -factor;
        cxp = factor*(rx*y1p)/ry;
        cyp =-factor*(ry*x1p)/rx;
        cx = cosr*cxp - sinr*cyp + (current->x+x)/2;
        cy = sinr*cxp + cosr*cyp + (current->y+y)/2;
```
The `RealNear()` check seems to be Williams' addition, and due to not fully interpreting F.6.6/B.2.5 step 4:
> Proceed with the remaining elliptical arc computations, such as those in the Conversion from endpoint to center parameterization algorithm. Note: As a consequence of the radii corrections in this section, the equation (5.2) for the center of the ellipse always has at least one solution (i.e. the radicand is never negative). In the case that the radii are scaled up using equation (eq. 6.3), the radicand of (eq. 5.2) is zero and there is exactly one solution for the center of the ellipse.

In other words, `lambda>1` is the intended test for whether it is necessary to calculate `factor`. 

I have successfully given myself the bird with the file from #3425 and tested with some other SVGs I have lying around, and things seem fine on that basis. @ctrlcctrlv : you reported in another bug that SVG round-trips are a key part of your workflow. If you have some time, could you give this change some exercise and report back? 

Closes #3425

### Types of changes

### Final checklist

### DIFF
--- a/fontforge/svg.c
+++ b/fontforge/svg.c
@@ -1264,16 +1264,15 @@ static void SVGTraceArc(SplineSet *cur,BasePoint *current,
 	   lambda = sqrt(lambda);
 	   rx *= lambda;
 	   ry *= lambda;
-	}
-	factor = rx*rx*ry*ry - rx*rx*y1p*y1p - ry*ry*x1p*x1p;
-	if ( RealNear(factor,0))
-	    factor = 0;		/* Avoid rounding errors that lead to small negative values */
-	else
+	   cxp = cyp = 0;
+	} else {
+	    factor = rx*rx*ry*ry - rx*rx*y1p*y1p - ry*ry*x1p*x1p;
 	    factor = sqrt(factor/(rx*rx*y1p*y1p+ry*ry*x1p*x1p));
-	if ( large_arc==sweep )
-	    factor = -factor;
-	cxp = factor*(rx*y1p)/ry;
-	cyp =-factor*(ry*x1p)/rx;
+	    if ( large_arc==sweep )
+		factor = -factor;
+	    cxp = factor*(rx*y1p)/ry;
+	    cyp =-factor*(ry*x1p)/rx;
+	}
 	cx = cosr*cxp - sinr*cyp + (current->x+x)/2;
 	cy = sinr*cxp + cosr*cyp + (current->y+y)/2;
 


frank-trampe:
@skef, could you perhaps add some of the knowledge you've acquired here in the form of comments in the code? It's rather laborious for me to review this, and I can only imagine how difficult it would be for somebody to understand the function without the context of this pull request.

skef:
I'll confess: My *understanding* of any of this is dim at best. Williams' understanding may not have been much better. The preexisting code comment already refers to Appendix F in the SVG 1.1 spec. The idea of the first half of this function is to implement what is described there. If you put that documentation next to this code the correspondence is (in my opinion) quite straightforward, down to the variable names. The adjustments immediately before the section I quote above correspond to [F.6.2](https://www.w3.org/TR/SVG11/implnote.html#ArcOutOfRangeParameters). The main thrust of the algorithm corresponds to [F.6.5](https://www.w3.org/TR/SVG11/implnote.html#ArcConversionEndpointToCenter). That algorithm is then "adjusted" corresponding to the advice of [F.6.6](https://www.w3.org/TR/SVG11/implnote.html#ArcCorrectionOutOfRangeRadii). This pull request is based on what I take to be a more accurate reading of the spec, nothing more. 

Given my level of understanding I'm not sure I could do much more in the comments beyond recording this correspondence, potentially down to equation numbers. Would that be beneficial? Should I just, or also, add a comment that emphasizes the value of the spec in assessing the code?

frank-trampe:
@skef, I don't understand the rationale for making the use of `factor` conditional on `lambda`. It seems to me from section 6.6 of the 1.1 specification that only the scaling of the radii by `sqrt(lambda)` is subject to the value of `lambda`. Am I missing something?

skef:
> @skef, I don't understand the rationale for making the use of `factor` conditional on `lambda`. It seems to me from section 6.6 of the 1.1 specification that only the scaling of the radii by `sqrt(lambda)` is subject to the value of `lambda`. Am I missing something?

The spec says "In the case that the radii are scaled up using equation (F.6.6.3), the radicand of (F.6.5.2) is zero and there is exactly one solution for the center of the ellipse." -- a "radicand" being whatever is under the squareroot sign. I interpret this as a *stipulation* that *factor* (in Williams' code) is zero in that case.

frank-trampe:
I thought that it meant that the radicand evaluated to zero in that case, not that we were responsible for suppressing it. I'm going to run that through Maxima quickly.

frank-trampe:
It definitely doesn't collapse to zero. So I guess that that is an imperative.

frank-trampe:
The specification is not very clear, but I think that you've interpreted it correctly.

frank-trampe:
@ctrlcctrlv, please test.

skef:
> It definitely doesn't collapse to zero. So I guess that that is an imperative.

Would the opposite conclusion ensured the `double` value would collapse to zero anyway?

frank-trampe:
@skef, I don't understand that.

skef:
Suppose that the algebraic form did collapse to zero after all. Would that imply that the C language expression of the calculation would also do so? It seems like the potential need for protection from wonky close-to-zero results applies in either case.

frank-trampe:
Ah, no. But the structure of the denominator of the "factor" and the translation ensuring non-zero x and y (at least for a non-zero-length segment) ensures that there are no discontinuities that would amplify any aberrations. Other conditions would result in similar erroneous small non-zero values for `cxp` and `cyp`, and I assume that the code is robust to those.

frank-trampe:
I've attached my mathematical model for future reference by the curious.

[SVG_Elliptical_Arc_Exploration_20190126_1.wxmx.gz](https://github.com/fontforge/fontforge/files/2798808/SVG_Elliptical_Arc_Exploration_20190126_1.wxmx.gz)

skef:
That makes sense. As it happens in this case the issue isn't with amplification but with the square root, and the fact that (apparently) at values close enough to zero the double calculation of the radicand can go negative.

frank-trampe:
Ah. I forgot about that. I still don't understand why the radicand drops abruptly to zero when Lambda crosses 1, but if it tests right, that's about all we can ask.

ctrlcctrlv:
I compiled it and found no problems with _my_ SVGs.

skef:
> I compiled it and found no problems with _my_ SVGs.

Is this the kind of thing you could test with a font-level diff of the old and new, either with fontforge itself or a version generated from files imported with the old and with this code?

ctrlcctrlv:
@skef I've got a massive font project underway right now, even bigger than my KJV1611 project, but nothing on my Github yet. It's totally scripted due to how much work it would be to do manually, so I can test importing hundreds of SVGs at a time with different FontForge variants...granted they're all made by me with InkScape, so not much variety, but still.

skef:
@ctrlcctrlv I'm just wondering how/hoping that we can verify the result of the import process with each code base is either identical or any differences are inconsequential. Just based on its "mechanics" the changes are less likely to introduce a crash than a subtle inaccuracy in the import.

ctrlcctrlv:
All glyphs imported fine, I noticed no breakage or differences between the two versions of FontForge

skef:
Good! That's probably a more comprehensive test than we could have hoped for (short of round-tripping non-SVG fonts through svg with a script, I guess).

ctrlcctrlv:
Right without a test suite of many SVG I'm not sure how to do better than this. I also made a font a long time ago called "Some Time Later" ,the SVGs from that also import fine.

frank-trampe:
@skef, for future reference, diffing sfds is actually a fairly good check. Timestamps and a few other things may vary, but the number of spurious changes is small. The other option is to export to TrueType or OpenType, to dump to XML with ttx, and to diff the results.

frank-trampe:
With testing successful, I think that the time has come to merge.

robmadole:
That was fast! What's involved in getting this released so we can do an easy `brew install`? I noticed that it's no longer possible to install with [Homebrew's `--HEAD` option](https://github.com/Homebrew/homebrew-core/pull/16226/files).

PR #3441, opened 2019-02-05
Increase buffer size to prevent stack smashing

keidax:
### Motivation and Context
This fixes a crash with `*** stack smashing detected ***` when opening an SFD file.

### Types of changes

### Description
This prevents a crash when opening an SFD file with embedded BDF properties longer than 200 chars. In my case, for example, it was a very long copyright field.

### Final checklist

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -6042,7 +6042,7 @@ static void SFDFixupBitmapRefs( BDFFont *bdf ) {
 
 static int SFDGetBitmapFont(FILE *sfd,SplineFont *sf,int fromdir,char *dirname) {
     BDFFont *bdf, *prev;
-    char tok[200];
+    char tok[2000];
     int pixelsize, ascent, descent, depth=1;
     int ch, enccount;
 


keidax:
@skef Thanks for the write-up. I should have mentioned earlier, but that's exactly how I arrived at the value of 2000.

I'm also confused by the compilation errors.

skef:
Attempting to rebuild with close/open ...

jtanx:
Ahh... these arbitrary buffer sizes and code that assumes that size just feels so wrong...

frank-trampe:
Is there a reason we couldn't have at least a macro for buffer size with limiting code in the parsing functions?

skef:
> Is there a reason we couldn't have at least a macro for buffer size with limiting code in the parsing functions?

I'm not sure I understand this suggestion -- the `get...` functions already limit internally; the problem is coordinating the internal limit with the buffer passed as an argument. 

It would take a bit of work but the system could be replumbed with a buffer struct that includes its length. That way you could either:

1. Eliminate the `get...` limits and just use the length of the passed-in buffer structure, or
2. Keep both types of length, use the shorter one and log a warning if the length of the passed-in buffer is smaller than expected.

frank-trampe:
I'm not suggesting that it be handed dynamically, just that we have one macro TOKMAX (or something like that) that defines the size of tok and how it is used.

skef:
> I'm not suggesting that it be handed dynamically, just that we have one macro TOKMAX (or something like that) that defines the size of tok and how it is used.

Oh, I was assuming there was some reason for the different buffer sizes -- most likely that some are allocated in functions called recursively and 2k buffers everywhere could exceed the stack depth. If that's not a live issue then everything becomes very easy.

PR #3442, opened 2019-02-06
Fix includes to appease gnulib

jtanx:
Broken since https://github.com/coreutils/gnulib/commit/b39cc72608e7ae56e959e2f12f7c8b355e63a35a


jtanx:
Yeah, so gnulib replaces parts of stdlib with its own implementation, and so something probably changed there. 

Generally you have to include the config header first (`fontforge-config.h`) for it to work, so the reordering I've done is to make sure that happens where gnulib was complaining

skef:
Looks good to me!

jtanx:
Thanks!

PR #3449, opened 2019-02-07
Revert "Upgrade copy functions in libgunicode to avoid segfaults and …

### LINKED ISSUE #3439: Images in buttons are aligned incorrectly
### Important

- [X] Screenshot
screen shot 2019-02-01 at 4 57 01 pm

c96d3eb78197aeb1447a8957b37b687398a1b529 - Mac OS X 10.14.3
Expected to have the images in the buttons centered correctly instead of off-center.
Build under Mac OS X with gdk enabled.

### Additional Info
Running on a MacBook Pro with a Retina display.

jtanx:
…other errors."

This reverts commit d45ecfe875acd4bf731f111985ac37342802c79c.

Fixes #3439.

Likely reason why it broke:
* Guess, but `u_copy` now always returning non null instead of returning null if input was null -> gbuttons.c -> label now not null - try to calc length -> now icons are off

I'm reverting this without any further alterations because:

a. They worked just fine before, with clear failure conditions
b. The chance of anyone trying to use these functions outside of FontForge is effectively 0
c. The changes are needlessly complex with very little benefit - I mentioned it before, but the whole `++n`/`--n` routine is just not good practice

### DIFF
--- a/Unicode/ustring.c
+++ b/Unicode/ustring.c
@@ -300,63 +300,34 @@ return( NULL );
 }
 
 unichar_t *u_copyn(const unichar_t *pt, long n) {
-/* Copy n chars of input string to a new res string. */
     unichar_t *res;
-
-    /* verify input valid for pt, and n>=0 and n+1>0 */
-    if ( pt==NULL || ++n<=0 )
-	return( NULL );
-
 #ifdef MEMORY_MASK
-    /* ensure n+1 unichar_t fits inside MEMORY_MASK. */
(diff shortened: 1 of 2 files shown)


skef:
It seems like the main worry with revisions like this is the chance that some aspect of a subsequent merge depends on the new behavior. So I just grabbed a diff with the child commit and the current master and looked for the altered functions. There's only one occurrence of one function and it's the context right before the altered code. That's no guarantee but the risk of this reversal is limited.

PR #3458, opened 2019-02-08
Python point, contour, and layer interface bug fixes and improvements

skef:
When I sent the message fontforge-dev about removing points from contours and contours from layers I had assumed that the functionality might just not be obvious. However, after playing around with some possible solutions and then more generally with the documented interface I realized much of it was just broken, in part because the slice implementation dated to the early days of Python 2. 

For the most part these changes are an attempt to slightly expand the documented interface and then make that interface work. The doc changes and test code are the place to look for how things should be now. Almost all the change is upwardly compatible, but there are some combinations of parameters (for initializing a point, for example) that would have worked before and won't now. However, given how non-functional things were I would be surprised if there are many substantial working scripts out there. (You couldn't even assign a point to a contour because of a missing reference count increment.) 

Rather than going further here I'll give my code an initial review to point out the reasoning where it may not be clear.   

### Types of changes

### Description

### Final checklist

### DIFF
--- a/fontforge/ffpython.h
+++ b/fontforge/ffpython.h
@@ -109,7 +109,7 @@ extern void PyFF_Glyph_Set_Layer(SplineChar *sc,int layer);
 typedef struct ff_point {
     PyObject_HEAD
     /* Type-specific fields go here. */
-    float x,y;
+    double x,y;
     uint8 on_curve;
     uint8 selected;
     char *name;
--- a/fontforge/glyphcomp.c
+++ b/fontforge/glyphcomp.c
@@ -197,8 +197,8 @@ return( true );
 			 ((dy=s->to->me.y-here->y)<=3 || dy<=3*err ) && (dy>=-3 || dy>=-3*err)) ) {
 		    s = s->to->next;
 		    t = 0;
-		    if ( (adx = (3*s->splines[0].a* t + 2*s->splines[0].b)* t + s->splines[0].c)<0 ) adx = -adx;
-		    if ( (ady = (3*s->splines[1].a* t + 2*s->splines[1].b)* t + s->splines[1].c)<0 ) ady = -ady;
+		    if ( s!=NULL && (adx = (3*s->splines[0].a* t + 2*s->splines[0].b)* t + s->splines[0].c)<0 ) adx = -adx;
+		    if ( s!=NULL && (ady = (3*s->splines[1].a* t + 2*s->splines[1].b)* t + s->splines[1].c)<0 ) ady = -ady;
 		} else
 	    break;
 	    }
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -44,6 +44,17 @@
 #define ENDPYGETSTR()
 #endif
 
+#if PY_VERSION_HEX > 0x03000000
+#define PYCMPF(a) NULL
+#define PYRCMPF(b) b
+#elif PY_VERSION_HEX > 0x02020000
+#define PYCMPF(a) a
+#define PYRCMPF(b) b
+#else
+#define PYCMPF(a) a
+#define PYRCMPF(b) NULL
+#endif
+
 extern int old_sfnt_flags;
 
 #include "autohint.h"
@@ -345,7 +356,7 @@ static void FreeStringArray( int cnt, char **names ) {
     free(names);
 }
 
-#if PY_MAJOR_VERSION >= 3
+#if PY_VERSION_HEX > 0x02020000
 
 typedef int (*cmpfunc)(PyObject*, PyObject*);
 
@@ -398,7 +409,7 @@ static PyObject *enrichened_compare(cmpfunc compare, PyObject *a, PyObject *b, i
     return result;
 }
 
-#endif /* PY_MAJOR_VERSION >= 3 */
+#endif /* PY_VERSION_HEX > 0x02020000 */
 
 static int FlagsFromString(const char *str, struct flaglist *flags, const char *flagkind) {
     int i;
@@ -1739,16 +1750,65 @@ return( reto );
 /* ************************************************************************** */
 /* Points */
 /* ************************************************************************** */
-static PyObject *PyFFPoint_dup(PyFF_Point *self) {
-    PyFF_Point *ret = (PyFF_Point *)PyFF_PointType.tp_alloc(&PyFF_PointType, 0);
-    if ( ret==NULL )
+
+static PyFF_Point *PyFFPoint_CNew(double x, double y, int on_curve, int sel, char *name) {
+    /* Convenience routine for creating a new point from C */
+    PyFF_Point *self = (PyFF_Point *)PyFF_PointType.tp_alloc(&PyFF_PointType, 0);
+    if ( self==NULL )
 	return( NULL );
-    ret->x = self->x;
-    ret->y = self->y;
-    ret->on_curve = self->on_curve;
-    ret->selected = self->selected;
-    ret->name = copy(self->name);
-return( (PyObject *) ret );
+    self->x = x;
+    self->y = y;
+    self->on_curve = on_curve;
+    self->selected = sel;
+    self->name = copy(name);
+    return( self );
+}
+
+static PyObject *PyFFPoint_dup(PyFF_Point *self) {
+    PyFF_Point *ret = PyFFPoint_CNew(self->x, self->y, self->on_curve,
+                                     self->selected, self->name);
+    return( (PyObject *) ret );
+}
+
+static PyFF_Point *PyFFPoint_Parse(PyObject *args, bool dup, bool always) {
+    double x,y;
+    PyFF_Point *p=NULL;
+    int i, on, sel;
+
+    if ( args==NULL && !always )
+	return NULL;
+
+    x = y = 0.0;
+    on = true;
+    sel = false;
+    if ( !PyArg_ParseTuple( args, "(ddii)", &x, &y, &on, &sel )) {
+	PyErr_Clear();
+	if ( !PyArg_ParseTuple( args, "(ddi)|i", &x, &y, &on, &sel )) {
+	    PyErr_Clear();
+	    if ( !PyArg_ParseTuple( args, "(dd)|ii", &x, &y, &on, &sel )) {
+		PyErr_Clear();
+		if ( !PyArg_ParseTuple( args, "dd|ii", &x, &y, &on, &sel )) {
+		    PyErr_Clear();
+		    if ( PyType_IsSubtype(&PyFF_PointType, Py_TYPE(args)) ) { 
+			p = (PyFF_Point *) args;
+		    } else if ( !always ) {
+			return( NULL );
+		    }
+		}
+	    }
+	}
+    }
+
+    if ( p==NULL ) {
+	p = PyFFPoint_CNew(x,y,on,sel,NULL);
+	if ( p==NULL )
+	    return( NULL );
+    } else if ( dup ) {
+	p = (PyFF_Point *) PyFFPoint_dup(p);
+    } else {
+	Py_INCREF( p );
+    }
+    return p;
 }
 
 static void PyFF_TransformPoint(PyFF_Point *self, double transform[6]) {
@@ -1790,17 +1850,17 @@ return( reductionTuple );
 static int PyFFPoint_compare(PyFF_Point *self,PyObject *other) {
     double x, y;
 
-    /* I'd like to accept general sequences but there is no PyArg_ParseSequence*/
-    if ( PyTuple_Check(other) && PyTuple_Size(other)==2 ) {
-	if ( !PyArg_ParseTuple(other,"dd", &x, &y ))
-return( -1 );
-    } else if ( PyType_IsSubtype(&PyFF_PointType, Py_TYPE(other)) ) {
-	x = ((PyFF_Point *) other)->x;
-	y = ((PyFF_Point *) other)->y;
-    } else {
-	PyErr_Format(PyExc_TypeError, "Unexpected type");
-return( -1 );
+    if ( !PyArg_ParseTuple(other,"dd", &x, &y ) ) {
+	PyErr_Clear();
+	if ( PyType_IsSubtype(&PyFF_PointType, Py_TYPE(other)) ) {
+	    x = ((PyFF_Point *) other)->x;
+	    y = ((PyFF_Point *) other)->y;
+	} else {
+	    PyErr_Format(PyExc_TypeError, "Unexpected type");
+	    return( -1 );
+	}
     }
+
     if ( RealNear(self->x,x) ) {
 	if ( RealNear(self->y,y))
 return( 0 );
@@ -1830,11 +1890,11 @@ static int PyFFPoint_set_name(PyFF_Point *self,PyObject *value, void *UNUSED(clo
 return( 0 );
 }
 
-#if PY_MAJOR_VERSION >= 3
+#if PY_VERSION_HEX > 0x02020000
 static PyObject *PyFFPoint_richcompare(PyObject *a, PyObject *b, int op) {
     return enrichened_compare((cmpfunc) PyFFPoint_compare, a, b, op);
 }
-#endif /* PY_MAJOR_VERSION >= 3 */
+#endif /* PY_VERSION_HEX > 0x02020000 */
 
 static PyObject *PyFFPoint_Repr(PyFF_Point *self) {
     char buffer[200];
@@ -1852,9 +1912,9 @@ return( STRING_TO_PY(buffer));
 }
 
 static PyMemberDef FFPoint_members[] = {
-    {(char *)"x", T_FLOAT, offsetof(PyFF_Point, x), 0,
+    {(char *)"x", T_DOUBLE, offsetof(PyFF_Point, x), 0,
      (char *)"x coordinate"},
-    {(char *)"y", T_FLOAT, offsetof(PyFF_Point, y), 0,
+    {(char *)"y", T_DOUBLE, offsetof(PyFF_Point, y), 0,
      (char *)"y coordinate"},
     {(char *)"on_curve", T_UBYTE, offsetof(PyFF_Point, on_curve), 0,
      (char *)"whether this point lies on the curve or is a control point"},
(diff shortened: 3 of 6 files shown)


fontforge/python.c
```diff
@@ -44,6 +44,17 @@
 #define ENDPYGETSTR()
 #endif
 
+#if PY_VERSION_HEX > 0x03000000
+#define PYCMPF(a) NULL
+#define PYRCMPF(b) b
+#elif PY_VERSION_HEX > 0x02020000
```
skef: Support for "Rich comparison" was apparently added in this release and support for normal comparison dropped in 3, hence the three combinations.

fontforge/python.c
```diff
@@ -1739,16 +1750,65 @@ return( reto );
...
+    PyFF_Point *p=NULL;
+    int i, on, sel;
+
+    if ( args==NULL && !always )
+	return NULL;
+
+    x = y = 0.0;
+    on = true;
+    sel = false;
+    if ( !PyArg_ParseTuple( args, "(ddii)", &x, &y, &on, &sel )) {
```
skef: Some of these embedded-tuple combos are probably pointless but they were supported by insertPoint()

fontforge/ffpython.h
```diff
@@ -109,7 +109,7 @@ extern void PyFF_Glyph_Set_Layer(SplineChar *sc,int layer);
 typedef struct ff_point {
     PyObject_HEAD
     /* Type-specific fields go here. */
-    float x,y;
+    double x,y;
```
skef: float was causing conversion problems. I don't believe there is a dependency between this representation and the one chosen for "internal" splines.

fontforge/glyphcomp.c
```diff
@@ -197,8 +197,8 @@ return( true );
 			 ((dy=s->to->me.y-here->y)<=3 || dy<=3*err ) && (dy>=-3 || dy>=-3*err)) ) {
 		    s = s->to->next;
 		    t = 0;
-		    if ( (adx = (3*s->splines[0].a* t + 2*s->splines[0].b)* t + s->splines[0].c)<0 ) adx = -adx;
-		    if ( (ady = (3*s->splines[1].a* t + 2*s->splines[1].b)* t + s->splines[1].c)<0 ) ady = -ady;
+		    if ( s!=NULL && (adx = (3*s->splines[0].a* t + 2*s->splines[0].b)* t + s->splines[0].c)<0 ) adx = -adx;
```
skef: I think this code may have been assuming that only closed contours would be compared. In any case s->to->next can wind up being NULL in practice and this prevents a segfault.

fontforge/python.c
```diff
@@ -1880,22 +1940,8 @@ static PyGetSetDef FFPoint_getset[] = {
...
-    self = (PyFF_Point *)type->tp_alloc(type, 0);
-    if ( self!=NULL ) {
-	self->x = x; self->y = y;
-	self->on_curve = on;
-	self->selected = sel;
-    }
-
+static PyObject *PyFFPoint_New(PyTypeObject *UNUSED(type), PyObject *args, PyObject *UNUSED(kwds)) {
+    PyFF_Point *self = PyFFPoint_Parse(args, true, true);
     return( (PyObject *)self );
```
skef: This is probably a double-reference increment. I'll check that and fix it.

fontforge/python.c
```diff
@@ -2082,7 +2117,6 @@ static void PyFFContour_ClearSpiros(PyFF_Contour *self) {
 	free(self->spiros);
     self->spiros = NULL;
     self->spiro_cnt = 0;
-    free(self->name);
```
skef: ???

fontforge/python.c
```diff
@@ -2391,33 +2425,31 @@ static Py_ssize_t PyFFContour_Length( PyObject *self ) {
 return( ((PyFF_Contour *) self)->pt_cnt );
 }
 
+static PyFF_Contour *PyFFPointList_Parse(PyObject *args);
```
skef: These functions call each other, but it's safe because the _Concat functions never call PointList_Parse when called with a second contour.

fontforge/python.c
```diff
@@ -2428,10 +2460,13 @@ return( NULL );
 	self->points[i] = c1->points[i];
     }
     for ( i=0; i<c2->pt_cnt; ++i ) {
-	Py_INCREF(c2->points[i]);
+	if ( c2!=&dummy )
+	    Py_INCREF(c2->points[i]);
 	self->points[c1->pt_cnt+i] = c2->points[i];
     }
-Py_RETURN( (PyObject *) self );
+    if ( ((PyObject *)c2)!=_c2 && c2!=&dummy )
```
skef: PyFFPoint_Parse grabs a reference while PyFFPointList_Parse doesn't grab references for the copied points, So we need to increment when copying from the latter or a passed contour but not from the "dummy" point. 
(The contour returned from PyFFPointList_Parse is really just a handy container for a point sequence.)

fontforge/python.c
```diff
@@ -2441,169 +2476,98 @@ static PyObject *PyFFContour_InPlaceConcat( PyObject *_self, PyObject *_c2 ) {
...
-Py_RETURN( self );
+    Py_RETURN( self );
 }
 
 static PyObject *PyFFContour_Index( PyObject *self, Py_ssize_t pos ) {
     PyFF_Contour *cont = (PyFF_Contour *) self;
     PyObject *ret;
 
-    if ( pos<0 || pos>=cont->pt_cnt ) {
+    if ( pos<-cont->pt_cnt || pos>=cont->pt_cnt ) {
```
skef: Add support for negative indexing.

fontforge/python.c
```diff
@@ -2625,14 +2589,172 @@ static PySequenceMethods PyFFContour_Sequence = {
     PyFFContour_Concat,		/* concat */
     NULL,			/* repeat */
     PyFFContour_Index,		/* subscript */
-    PyFFContour_Slice,		/* slice */
+    NULL,			/* slice */
```
skef: These are the obsolete slice interfaces that correspond to the functions removed above.

fontforge/python.c
```diff
@@ -2625,14 +2589,172 @@ static PySequenceMethods PyFFContour_Sequence = {
...
     PyFFContour_IndexAssign,	/* subscript assign */
-    PyFFContour_SliceAssign,	/* slice assign */
+    NULL,			/* slice assign */
     PyFFContour_Contains,	/* contains */
     PyFFContour_InPlaceConcat,	/* inplace_concat */
     NULL			/* inplace repeat */
 };
 
+static PyObject *PyFFContour_Sub( PyObject *self, PyObject *key ) {
```
skef: Slice indexing is now supported via the mapping interface.

fontforge/python.c
```diff
@@ -2625,14 +2589,172 @@ static PySequenceMethods PyFFContour_Sequence = {
...
+    int i;
+
+    if ( PyInt_Check(key)) {
+        return PyFFContour_Index(self, PyNumber_AsSsize_t(key, PyExc_IndexError));
+    }
+    if ( !PySlice_Check(key) ) { 
+	PyErr_Format(PyExc_IndexError, "Contour indexed by integer only");
+	return( NULL );
+    }
+#if PY_VERSION_HEX > 0x03060100
```
skef: The docs explain the chain of support. As it turns out any recent version of python 2 or 3 will do pretty much the same thing underneath these functions because the new implementations were back-ported.

fontforge/python.c
```diff
@@ -2625,14 +2589,172 @@ static PySequenceMethods PyFFContour_Sequence = {
...
+    }
+
+    ret = (PyFF_Contour *)PyFF_ContourType.tp_alloc(&PyFF_ContourType, 0);
+    ret->is_quadratic = cont->is_quadratic;
+    ret->closed = false;
+    ret->pt_max = ret->pt_cnt = len;
+    ret->points = PyMem_New(PyFF_Point *,ret->pt_max);
+
+    for ( i = 0; i < len; i++ ) {
+	ret->points[i] = cont->points[i*step + start];
```
skef: The tricky step-supporting assignment semantic.

fontforge/python.c
```diff
@@ -2807,42 +2929,35 @@ Py_RETURN( self );
...
 
-    x = y = 0.0; pos = -1; on = true;
-    if ( !PyArg_ParseTuple( args, "(ddi)|i", &x, &y, &on, &pos )) {
+    x = y = 0.0;
+    pos = -1;
+    on = true;
+    sel = false;
+    if ( !PyArg_ParseTuple( args, "(ddii)|i", &x, &y, &on, &sel, &pos )) {
 	PyErr_Clear();
-	if ( !PyArg_ParseTuple( args, "(dd)|ii", &x, &y, &on, &pos )) {
```
skef: This is the most obvious aspect of the old interface that I broke, and I broke it because it was crazy. Did the user of the previous interface mean to specify on-curve or pos? Who knows? Just leave the point spec in a tuple and the pos outside.

skef:
Oh: The python 2 stuff is not entirely guesswork. I broke my vow of never thinking about python 2 for a moment and compiled and tested against 2.7 (easy on Arch). However, I didn't try anything other than that and (very recent) Python 3.7.2

skef:
Maybe a high-level description of the changes to `python.c` would be helpful:

# Comparison functions

The old code supported standard and "rich" comparison functions, but used the former with python 2 and the latter with 3. Support for rich comparison was later back-ported to python 2.2, and subsequent version of python 2 could use either. The rich function probably overrides the poor when both are defined, but for completeness sake both are now added in the versions that support either. (The important thing is that the rich is added for later versions of 2.)

# Point creation

I added support for defining whether a point is selected at creation time. I also restructured the creation code a bit, moving the argument parsing to a new function `PyFFPoint_Parse`. That also involved some reordering of the code. 

# Contour creation and assignment

Contour slicing was broken because support for the old `_Slice` and `_SliceAssign` function types was long ago dropped. The new means of slice support is via the mapping interface. So I removed those old functions and added `_Sub` and `_SubAssign`. I also added support for "point lists", which are sequences of either actual points or point initialization tuples (noted above). These can be used as one argument in a concatenation, or in slice assignment. This also means one can directly initialize a contour 
like this:
```
c[:] = [(0,0,True,True),(1,1,False,),(2,2,False),(3,3),(4,4),(5,5,False),(6,6,False),(7,7),(8,8,False),(9,0,False)]
```
The diff for these changes is more confusing than the changes themselves, because of the way the code lines matched up. I recommend just comparing the old and new `PyFFContour_IndexAssign` by other means. 

The code for "unpacking" slice objects was changed twice, once in 3.2 and once in 3.6. I haven't tested with the former, but my ifdefs match the python documentation.

# "Sequence" deletion

At the C-API level, `del c[i]` corresponds to calling `_IndexAssign` on c with `i` as the index and a null argument. This is now supported for contours and layers, providing a straightforward way of removing elements from those sequences.

fontforge/python.c
```diff
@@ -1790,17 +1850,17 @@ return( reductionTuple );
...
-	if ( !PyArg_ParseTuple(other,"dd", &x, &y ))
-return( -1 );
-    } else if ( PyType_IsSubtype(&PyFF_PointType, Py_TYPE(other)) ) {
-	x = ((PyFF_Point *) other)->x;
-	y = ((PyFF_Point *) other)->y;
-    } else {
-	PyErr_Format(PyExc_TypeError, "Unexpected type");
-return( -1 );
+    if ( !PyArg_ParseTuple(other,"dd", &x, &y ) ) {
+	PyErr_Clear();
```
frank-trampe: So, just to be clear, the idea is to parse even if it generates an error and then to depend on the subtype check to determine validity?

fontforge/python.c
frank-trampe: Good catch.

fontforge/python.c
```diff
@@ -2391,33 +2425,31 @@ static Py_ssize_t PyFFContour_Length( PyObject *self ) {
...
+    PyFF_Contour *c1 = (PyFF_Contour *) _c1, *c2 = (PyFF_Contour *)_c2;
     PyFF_Contour *self;
     int i;
     PyFF_Contour dummy;
     PyFF_Point *dummies[1];
     double x,y;
 
-    if ( PyType_IsSubtype(&PyFF_PointType, Py_TYPE(c2)) ) {
+    if ( PyType_IsSubtype(&PyFF_ContourType, Py_TYPE(c1)) &&
+         PyType_IsSubtype(&PyFF_ContourType, Py_TYPE(c2)) ) {
```
frank-trampe: I don't immediately understand what is going on here. Does ContourType inherit from PointType?

frank-trampe: Oh. I see. You reordered it.

frank-trampe: Any particular reason?

fontforge/python.c
skef: The existing code is basically redundant, except perhaps in the difference in error messages generated by `PyErr_Format()` and `ParseTuple()` (if the latter doesn't find elements it can convert to numbers.) The new code first tries to parse the tuple and if that fails checks if the argument is a point, and if that fails produces the error. This is a better match for most of the rest of the tuple parsing code in the file.

fontforge/python.c
skef: I wanted to make sure I catch the double-contour case first because this function now calls `PyFFPointList_Parse()` and vice versa. That's OK because `PyFFContour_Concat()` never calls  `PyFFPointList_Parse()` when passed two contours, and the latter only calls the former with two contours. I wanted to lessen the chance that some future update leads to an infinite call chain.

skef: Oh, slightly correction: `PyFFPointList_Parse()` calls `PyFFContour_InPlaceConcat()` and vice versa. I reordered `PyFFContour_Concat()` to match `PyFFContour_InPlaceConcat()`.

fontforge/python.c
jtanx: Just a note, but I believe the build system enforces Python >= 2.7

fontforge/python.c
```diff
@@ -2441,169 +2476,98 @@ static PyObject *PyFFContour_InPlaceConcat( PyObject *_self, PyObject *_c2 ) {
...
-    ret->pt_max = ret->pt_cnt = len;
-    ret->points = PyMem_New(PyFF_Point *,ret->pt_max);
-
-    if ( end<start ) {
-	for ( i=start; i<cont->pt_cnt; ++i )
-	    ret->points[i-start] = cont->points[i];
-	for ( i=0; i<end; ++i )
-	    ret->points[(cont->pt_cnt-start)+i] = cont->points[i];
+    if ( val==NULL ) {
+	for ( i=pos; i<=cont->pt_cnt-1; ++i )
```
jtanx: This check doesn't look right; should this condition be `i < cont->pt_cnt-1`? Otherwise wouldn't it overflow

fontforge/python.c
```diff
@@ -2625,14 +2589,172 @@ static PySequenceMethods PyFFContour_Sequence = {
...
+    if (PySlice_GetIndicesEx(key, cont->pt_cnt, &start, &end, &step, &len) < 0) {
+	return( NULL );
+    }
+#else
+    if (PySlice_GetIndicesEx((PySliceObject *) key, cont->pt_cnt, &start, &end, &step, &len) < 0) {
+	return( NULL );
+    }
+#endif
+    if ( !(step == 1 || step == -1) ) {
+	PyErr_Format(PyExc_IndexError, "Only supported steps are 1 and -1");
```
jtanx: I'm not overly familiar with the data structure at hand, but why not support arbitrary steps?

fontforge/python.c
```diff
@@ -2625,14 +2589,172 @@ static PySequenceMethods PyFFContour_Sequence = {
...
+	Py_INCREF(ret->points[i]);
+    }
+    return( (PyObject *) ret );
+}
+
+// Caller responsible for incrementing reference count
+static void PyFFContour_CInsertPoint(PyFF_Contour *self, PyFF_Point *p, int pos) {
+	int i;
+
+	if ( pos<0 || pos>=self->pt_cnt-1 )
```
jtanx: nit, but I'd prefer `pos>self->pt_cnt-1` over redundantly assigning if it's equal to the limit

fontforge/python.c
```diff
@@ -2625,14 +2589,172 @@ static PySequenceMethods PyFFContour_Sequence = {
...
+		pos = self->pt_cnt-1;
+	if ( self->pt_cnt >= self->pt_max ) {
+		/* Messes with self->points */
+		PyMem_Resize(self->points,PyFF_Point *,self->pt_max += 10);
+	}
+	for ( i=self->pt_cnt-1; i>pos; --i )
+	self->points[i+1] = self->points[i];
+	self->points[pos+1] = p;
```
jtanx: This seems a bit surprising that it would insert the point at `pos` instead of `pos+1`

fontforge/python.c
```diff
@@ -2625,14 +2589,172 @@ static PySequenceMethods PyFFContour_Sequence = {
...
+    rpl = PyFFPointList_Parse(val);
+    if ( rpl==NULL ) { 
+	PyErr_Format(PyExc_TypeError, "Replacement must encode a point list");
+	return( -1 );
+    }
+    if ( !PySlice_Check(key) ) { 
+	PyErr_Format(PyExc_IndexError, "Contour indexed by integer only");
+	return( -1 );
+    }
+#if PY_VERSION_HEX > 0x03060100
```
jtanx: It feels like this slice unpacking code should be either macro'ed or made into a function

fontforge/python.c
skef: Contour points come in on-curve/off-curve sequences that, because there are lines without off-curve points in between and (for quadratic curves) interpolated points, will not generally fall on regular intervals. Supporting steps would be both extra work and likely lead to confusion for users rather than benefit.

fontforge/python.c
```diff
@@ -3867,27 +3982,36 @@ return( ret );
...
-return( -1 );
+	return( -1 );
     }
 
     old = layer->contours[pos];
-    layer->contours[pos] = contour;
+
+    if ( val==NULL) {
+	for ( i=pos; i<=layer->cntr_cnt-1; ++i )
+	    layer->contours[i] = layer->contours[i+1];
```
jtanx: Same comment here re oob acess on the contours

fontforge/python.c
skef: Don't you mean that the other way around? 

I agree that `pos` would be better but the interface has been implemented and documented the other way.

fontforge/python.c
skef: I think that's right. I'll check and test.

fontforge/python.c
skef: I think this is just from the old `Contour_IndexAssign`, and therefore retained so as not to change the behavior.

fontforge/python.c
skef: It's only two instances ...

fontforge/python.c
skef: Hmm, you don't mean it the other way around. Yeah, this seems wrong. Which means my test must be wrong ...

skef: The existing documentation for `insertPont` says "Adds point to the contour. If the optional third argument is give, the line will be added after the pos'th point, otherwise it will be at the end of the contour." So (weirdly) if you specify 0 the point should end up in position 1, and if you specify 1 it should end up in position 2.

Say there are three points and the user specifies `1`. Then you want to move existing point 2 to position 3 and existing point 1 to position 2, and then assign to 2. So I think that `i>pos` is the right test for the loop, and `pos+1` the right location, after all.

fontforge/python.c
skef: This and the other one are now fixed in 60b59d575049e968294299a06838939b792918c5.

fontforge/python.c
frank-trampe: The two lines are consistent, at least. The last iteration of the loop runs with i = pos + 1, which vacates that spot for an insertion at pos + 1. The calling convention is still sort of absurd, but what can we do?

fontforge/python.c
frank-trampe: It looks okay now. I assume that this was resolved?

PR #3459, opened 2019-02-09
Fix compilation when compiled without python support

jtanx:
Update travis NOX target to also compile without python support

Regression from ca2f4bb97794bf983bbc0b25145248c21c9030d1


skef:
Oof, sorry about that ...

PR #3467, opened 2019-02-11
Expose selection of off-curve cubic points in python interface

skef:
When writing python scripts to "interactively" modify splines it can be useful to know which off-curve points are selected -- a status FontForge already tracks separately from that of on-curve points. At present the python interface just sets `selected` to false for all off-curve points. 

This change bidirectional `selected` status tracking for *cubic* off-curve points. (The status was already bidirectional for on-curve points.) 

Should this change also be made for quadratic off-curve points? One would think so! Unfortunately the original implementer applied an alternate interpretation to the field:

> If an off-curve point is selected in means the preceding (interpolated) on-curve point is selected.

So it's unclear what to do about that. 

This change is borderline-incompatible in that a script that assumes only on-curve points will be selected will no longer work as intended. 

### Types of changes

### DIFF
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -942,8 +942,15 @@ fontforge.registerImportExport(importGlyph,exportGlyph,None,"foosball","foo","fo
     </TR>
     <TR>
       <TD><CODE>selected</CODE></TD>
-      <TD colspan=2>Whether this point is selected in the UI. If an off-curve point
-	is selected in means the preceding (interpolated) on-curve point is selected.</TD>
+      <TD colspan=2>Whether this point is selected in the UI. In the case of
+        an off-curve point of a quadratic spline, selected is true when the
+	preceding (interpolated) on-curve point is selected. Otherwise (as with
+	a cubic curve) selected for an off-curve point also corresponds to 
+	UI selection.
+	<P>The value can also be set to change the selected status in the UI.
+        When not running with a UI FontForge will retain point selection status
+        between most calls and will usually mark the status in saved files as
+        it would when a UI is present.</TD>
     </TR>
     <TR>
       <TD><CODE>name</CODE></TD>
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -4954,6 +4954,7 @@ return( ss );
 	    if ( !c->points[previ]->on_curve ) {
 		sp->prevcp.x = c->points[previ]->x;
 		sp->prevcp.y = c->points[previ]->y;
+		sp->prevcpselected = c->points[previ]->selected;
 		// if ( sp->prevcp.x!=sp->me.x || sp->prevcp.y!=sp->me.y ) // This is unnecessary since the other converter only makes a control point if there is supposed to be one.
 		    sp->noprevcp = false;
 	    }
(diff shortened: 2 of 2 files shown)


skef:
I guess I should mention #3352 here, which was my original thought about this problem. Separating out the state as "selected" (for whether the base point is selected) and "highlighted" (for whether a control point is, well, selected) has some nice properties, in that there are times you want to know whether the base point is selected, and this approach provides that info without the user having to "find" it. (The best example of this in FontForge is the transformation UI interface, which, when run on "selected points" actually transforms the selected on-curve points together with their control points. Any amount of playing around with selected `transform` makes it clear this is the semantic of choice, in that it comes closest to "doing what you expect".) 

I eventually decided that providing the two states was likely to be rejected as too "baroque" for what is intended to be a low-level interface. It's not exactly hard to find a control point -- this is my quickie code for the cubic case:
```
def controlPt(c, i):
    if c[i].on_curve:
        return c[i]
    if c[i-1].on_curve:
        return c[i-1]
    ip = (i+1)%len(c)
    if c[ip].on_curve:
        return c[ip]
    return None
```
However, I mention it here because it does happen to address the quadratic curve issue, and is also more backward-compatible: by adding a state to track off-curve point selection status you can retain the current semantics of `selected`.

frank-trampe:
@skef, it would be rather helpful for review purposes and also for future reference it you would annotate your changes in python.c.

frank-trampe:
Oh, wait, some of those are in #3458. I'll work on that first, and then we work on the incremental changes.

skef:
Yeah, I anticipated these changes depending on those and arranged my own branches accordingly. I don't think I wound up with any actual dependencies, though.

skef:
Now that the other changes are merged I've updated this request to the relevant change set alone. The question of what to do for quadratic splines remains.

skef:
After thinking about this for a while, the most sensible options for dealing with quadratic beziers are:

1. Do nothing (compromising the python/UI interactions).
2. Add an "interpolated" bool to `FontForge.point` and add interpolated points into the point list. 

On merging a cubic contour back in the flag would be ignored entirely. On merging a quadratic contour we could verify that any interpolated point is centered between its control points, setting the internal flag if it is and throwing an error otherwise. 

2 is probably the better option. Whether they should nor not, interpolated fontforge points can have (temporary?) types and (temporary?) names. If we don't throw them into the point list the names are lost in the current code and the types are lost after the merge of #3481. 

I'll work on some code to do this to have something more concrete to evaluate. It will be a breaking interface change for some scripts, but any script that breaks would probably be doing something questionable anyway.

frank-trampe:
@skef, I don't quite understand what you're saying. What's wrong with the current patch in the context of quadratics?

skef:
> @skef, I don't quite understand what you're saying. What's wrong with the current patch in the context of quadratics?

Selected off-curve quadratic points are not marked as such in python. So a round-trip through a PyFF_Layer will remove that status, and it isn't available to a script. This isn't trivial to fix because the `selected` flag for quadratics is "overloaded" to track selection of interpolated on-curve points. 

I'm working on a fix for this. With luck it will be done in a few hours.

jtanx:
I'll hold off on merging this then

skef:
Closing in preference to #3513

Issue #3437, opened 2019-02-01
Spiro segmentation fault


ctrlcctrlv:
* Enable Spiro mode
* Draw a circle (create four curve points) (it seems like only one is strictly necessary)

![c](https://user-images.githubusercontent.com/838783/52124528-77341d80-2664-11e9-97c2-c131bb214845.png)

* Right click one of the points, select _Get Info&hellip;_
* Close the dialog via your window manager. It won't close. Try again, you'll get a segfault.

Tested on Arch Linux, commit c96d3eb78197aeb1447a8957b37b687398a1b529

```
Program received signal SIGSEGV, Segmentation fault.
0x0000155554a665e0 in CVRemoveTopUndo () from /usr/lib/libfontforge.so.2
(gdb) bt
#0  0x0000155554a665e0 in CVRemoveTopUndo () at /usr/lib/libfontforge.so.2
#1  0x000015555523c505 in  () at /usr/lib/libfontforgeexe.so.2
#2  0x000015555523c9e9 in  () at /usr/lib/libfontforgeexe.so.2
#3  0x00001555550cb958 in  () at /usr/lib/libgdraw.so.5
#4  0x00001555550cbf18 in  () at /usr/lib/libgdraw.so.5
#5  0x0000155555116bb6 in  () at /usr/lib/libgdraw.so.5
#6  0x0000155555119141 in  () at /usr/lib/libgdraw.so.5
#7  0x000015555533da3f in fontforge_main () at /usr/lib/libfontforgeexe.so.2
#8  0x0000555555555159 in main (argc=1, argv=0x7fffffffe188) at main.c:39
```

ctrlcctrlv:
I think I figured out why this happens, `CVRemoveTopUndo` attempts to use memory freed by the previous invocation of `CVRemoveTopUndo`. The solution would be to do the same action as the "Close" button on the dialog to the "X" provided by window manager.

ctrlcctrlv:
Just figured out how to get `gdb` to produce an even better dump :-)

```
Program received signal SIGSEGV, Segmentation fault.
0x000015555486a90e in CVRemoveTopUndo (cv=0x55555580cc10) at cvundoes.c:1138
1138        cv->layerheads[cv->drawmode]->undoes = undo->next;
(gdb) bt
#0  0x000015555486a90e in CVRemoveTopUndo (cv=0x55555580cc10) at cvundoes.c:1138
[...]

PR #3468, opened 2019-02-12
Update CONTRIBUTING.md

ctrlcctrlv:
Clarify the correct way to debug FontForge to prevent a recurrence of https://github.com/fontforge/fontforge/issues/3437#issuecomment-462346232, a very subtle bug.


ctrlcctrlv:
This is just a documentation upgrade, so the failed build is meaningless.

Issue #2829, opened 2016-09-13
Deleting a point while its point info dialogue is open results in segfault


jtanx:
* Open the point info dialogue for a point (Right click->get info)
* Back in charview, delete the point, leave the point info dialogue open
* Try adding more points --> crash (may take more than one try)

PR #3469, opened 2019-02-12
Close #3437

### LINKED ISSUE #3437: Spiro segmentation fault

ctrlcctrlv:
This patch is courtesy @jtanx, I'm just making the PR to save him some time.

### Important

### Motivation and Context

### Types of changes

### Description

Small change to `pi_e_h` function which prevents crash when Spiro Info dialog closed.

### Final checklist

### DIFF
--- a/fontforgeexe/cvgetinfo.c
+++ b/fontforgeexe/cvgetinfo.c
@@ -61,7 +61,6 @@ typedef struct gidata {
     int interp_start, interp_end;
     GGadgetCreateData* gcd;
     GGadget *group1ret, *group2ret;
-    int nonmodal;
 } GIData;
 
 #define CID_BaseX	2001
@@ -1753,27 +1752,6 @@ static void PI_DoCancel(GIData *ci) {
     SCUpdateAll(cv->b.sc);
 }
 
-static int pi_e_h(GWindow gw, GEvent *event) {
-    if ( event->type==et_close ) {
-	GIData  *d = GDrawGetUserData(gw);
-	if( d->nonmodal ) {
-	    PI_Destroy((struct dlistnode *)d);
-	} else {
-	    PI_DoCancel( GDrawGetUserData(gw));
-	}
-    } else if ( event->type==et_char ) {
-	if ( event->u.chr.keysym == GK_F1 || event->u.chr.keysym == GK_Help ) {
-	    help("getinfo.html");
-return( true );
-	}
-return( false );
-    } else if ( event->type == et_map ) {
-	/* Above palettes */
-	GDrawRaise(gw);
-    }
-return( true );
-}
-
 static void PIFillup(GIData *ci, int except_cid);
 
 static void PI_FigureNext(GIData *ci) {
@@ -1869,32 +1847,49 @@ void PI_Destroy(struct dlistnode *node) {
     free(d);
 }
 
-static void PI_Close(GGadget *g) {
-    GWindow gw = GGadgetGetWindow(g);
-    GIData  *d = GDrawGetUserData(gw);
+static void PI_Close(GIData *d) {
     PI_Destroy((struct dlistnode *)d);
 }
 
 static int PI_Cancel(GGadget *g, GEvent *e) {
     if ( e->type==et_controlevent && e->u.control.subtype == et_buttonactivate ) {
-	PI_DoCancel( GDrawGetUserData(GGadgetGetWindow(g)));
-	PI_Close(g);
+	GIData *d = GDrawGetUserData(GGadgetGetWindow(g));
+	PI_DoCancel(d);
+	PI_Close(d);
     }
 return( true );
 }
 
-static int PI_Ok(GGadget *g, GEvent *e) {
-    if ( e->type==et_controlevent && e->u.control.subtype == et_buttonactivate ) {
-	GIData *ci = GDrawGetUserData(GGadgetGetWindow(g));
-
+static void PI_DoOk(GIData *ci) {
 	PI_FixStuff(ci);
 	_PI_ShowHints(ci,false);
 
 	ci->done = true;
 	/* All the work has been done as we've gone along */
-	PI_Close(g);
-    }
+	PI_Close(ci);
+}
 
+static int PI_Ok(GGadget *g, GEvent *e) {
+    if ( e->type==et_controlevent && e->u.control.subtype == et_buttonactivate ) {
+	GIData *ci = GDrawGetUserData(GGadgetGetWindow(g));
+	PI_DoOk(ci);
+    }
+    return( true );
+}
+
+static int pi_e_h(GWindow gw, GEvent *event) {
+    if ( event->type==et_close ) {
+	PI_DoOk(GDrawGetUserData(gw));
+    } else if ( event->type==et_char ) {
+	if ( event->u.chr.keysym == GK_F1 || event->u.chr.keysym == GK_Help ) {
+	    help("getinfo.html");
+return( true );
+	}
+return( false );
+    } else if ( event->type == et_map ) {
+	/* Above palettes */
+	GDrawRaise(gw);
+    }
 return( true );
 }
 
@@ -3338,7 +3333,6 @@ static void PointGetInfo(CharView *cv, SplinePoint *sp, SplinePointList *spl) {
 
 	GHVBoxFitWindow(mb[0].ret);
 
-	gi->nonmodal = 1;
 	dlist_pushfront( &cv->pointInfoDialogs, (struct dlistnode *)gi );
 	GWidgetHidePalettes();
 	GDrawResize(gi->gw,
@@ -3462,13 +3456,32 @@ static int PI_SpiroChanged(GGadget *g, GEvent *e) {
 return( true );
 }
 
+static void PI_SpiroDoOk(GIData *ci) {
+	ci->done = true;
+	/* All the work has been done as we've gone along */
+	PI_Close(ci);
+}
+
 static int PI_SpiroOk(GGadget *g, GEvent *e) {
     if ( e->type==et_controlevent && e->u.control.subtype == et_buttonactivate ) {
 	GIData *ci = GDrawGetUserData(GGadgetGetWindow(g));
+	PI_SpiroDoOk(ci);
+    }
+    return( true );
+}
 
-	ci->done = true;
-	/* All the work has been done as we've gone along */
-	PI_Close(g);
+static int spi_e_h(GWindow gw, GEvent *event) {
+    if ( event->type==et_close ) {
+	PI_SpiroDoOk(GDrawGetUserData(gw));
+    } else if ( event->type==et_char ) {
+	if ( event->u.chr.keysym == GK_F1 || event->u.chr.keysym == GK_Help ) {
+	    help("getinfo.html");
+return( true );
+	}
+return( false );
+    } else if ( event->type == et_map ) {
+	/* Above palettes */
+	GDrawRaise(gw);
     }
 return( true );
 }
@@ -3519,7 +3532,7 @@ static void SpiroPointGetInfo(CharView *cv, spiro_cp *scp, SplinePointList *spl)
 	if ( pos.y+pos.height+20 > screensize.height )
 	    pos.y = screensize.height - pos.height - 20;
 	if ( pos.y<0 ) pos.y = 0;
-	gip->gw = GDrawCreateTopWindow(NULL,&pos,pi_e_h,gip,&wattrs);
+	gip->gw = GDrawCreateTopWindow(NULL,&pos,spi_e_h,gip,&wattrs);
 
 	memset(&gcd,0,sizeof(gcd));
 	memset(&label,0,sizeof(label));


jtanx:
Thanks. @ctrlcctrlv do you still want that behaviour of keeping the current state when closing the dialogue? Because this will revert state.

skef:
> Thanks. @ctrlcctrlv do you still want that behaviour of keeping the current state when closing the dialogue? Because this will revert state.

My own two cents, because I thought about this issue a bit while looking at #2829 (which is the first time I've encountered a segfault condition that made me think "huh, maybe this is best the way it is.")

Short answer: there is no good answer because the implementation model is wrong, but:

1. *If* an immediate undo consistently brings you back to the state that "cancel" would, mapping close to OK is probably the better option, because it leaves both outcomes available to the user. 
2. Otherwise, mapping close to cancel is the best option. 

I'm far from a Microsoft person, but [this document](https://docs.microsoft.com/en-us/windows/desktop/uxguide/win-dialog-box#modeless-dialog-boxes) matches my understanding of the usual (and most intuitive) model: Modeless dialogs should introduce their changes immediately into the "stream" just like other modeless tools (such as a palette tool). In that case the way to move out of a modeless-window-induced change would be to undo, just like with anything else. 

In contrast, these modeless windows save a context when they are opened, and cancel switches back to that context. This can be weird and counter-intuitive. Do three things in the dialog context and seven things outside of it, and then hit cancel, and all might go away. 

So the proper fix is to switch to an immediate commit model and use undo/redo to move back and forth in the stream, and thereby replace "OK" and "Cancel" with "Close", leaving no ambiguity in the meaning of the window decoration. Short of that, either choice is just making the best of bad semantics.

skef:
Added: Oh, I'm mostly thinking of the non-Spiro getinfo in that analysis. Now I believe I recall that the Spiro-specific dialog is actually modal (and that caused the problem). 

If that's right the better fix would be to remove or disable the decoration. Modal dialogs should only be decorated at all if they can be moved (which is appropriate when one might want to look at the window "below" for reference), but generally it's better to force the user to interact with the in-window controls than support OS-level controls when the latter would be ambiguous.

ctrlcctrlv:
When I made the original issue, it seemed like opinion was against me that the window decorator's "X" should mean OK. Luckily the tide seems to have turned in that regard, so I just made it so that this PR will make the Spiro info dialog act exactly like the Bézier info dialog. I tested it and it works on my system. :-)

As to whether there should be window decorations at all, this would affect all three dialogs in the file, the anchor, Bézier, and Spiro dialogs. I feel unqualified to comment, but I'm not sure how the dialogs would be moved around without their decorations - perhaps the decorator can be instructed to put them in a disabled state.

skef:
>  Luckily the tide seems to have turned in that regard, so I just made it so that this PR will make the Spiro info dialog act exactly like the Bézier info dialog. 

I don't think this is accurate, as the Spiro dialog is still modal while the Bezier dialog isn't. 

> As to whether there should be window decorations at all, this would affect all three dialogs in the file, the anchor, Bézier, and Spiro dialogs.

And given that the modality is different this doesn't follow. (But it's true that the Spiro and Anchor windows should have corresponding decorations.) 

These issues have more impact than you might think. @ctrlcctrlv: Since you basically have the same environment, bring up the Spiro Get Info or Anchor Get Info dialog and then use the decoration to minimize it. Then try to get out of that situation to use the program again. Is there any clear way?

ctrlcctrlv:
I really think that the modality has nothing to do with making a friendly UI but rather everything to do with the underlying weakness of the Spiro implementation. If points are moved around and changed while the dialog is open, crashes could result&mdash;just a hunch, I can test it tomorrow.

The modality (i.e. making the Character View inactive while it's open) has no bearing at all on the active state of the window decorations as far as I can tell. I can still minimize the Spiro Info dialog, just the CharView comes with it, while in the Bézier Info dialog, it minimizes independently of its associated CharView.

skef:
> I can still minimize the Spiro Info dialog, just the CharView comes with it
Huh, what happens on my system is that minimizing the Spiro dialog makes it disappear and have no representation in the window list, while the CharView doesn't minimize. I can get out of the situation by minimizing and then un-minimizing the CharView, but those steps are fare from obvious. 

> The modality (i.e. making the Character View inactive while it's open) has no bearing at all on the active state of the window decorations as far as I can tell.

Right, but it *should* have a bearing. Or if the program allows modal dialogs to be minimized it should at least give them OS-level window representation.

ctrlcctrlv:
I've tried to get the Spiro dialog to lose its modality, but so far nothing has resulted in a truly satisfactory result. However, at least it doesn't crash where I thought it might, so the fix might be trivial.

First, the patch; I just applied some of the elements of `PointGetInfo` to `SpiroPointGetInfo`:

```diff
diff --git a/fontforgeexe/cvgetinfo.c b/fontforgeexe/cvgetinfo.c
index 38df9aa08..3f8be1061 100644
--- a/fontforgeexe/cvgetinfo.c
+++ b/fontforgeexe/cvgetinfo.c
@@ -3496,7 +3496,7 @@ static void SpiroPointGetInfo(CharView *cv, spiro_cp *scp, SplinePointList *spl)
     GDrawGetSize(root,&screensize);
 
        memset(&wattrs,0,sizeof(wattrs));
-       wattrs.mask = wam_events|wam_cursor|wam_utf8_wtitle|wam_positioned|wam_isdlg|wam_restrict;
+       wattrs.mask = wam_events|wam_cursor|wam_utf8_wtitle|wam_positioned|wam_isdlg;
        wattrs.event_masks = ~(1<<et_charup);
        wattrs.restrict_input_to_me = 1;
        wattrs.positioned = 1;
@@ -3745,6 +3745,7 @@ static void SpiroPointGetInfo(CharView *cv, spiro_cp *scp, SplinePointList *spl)
 
        GHVBoxFitWindow(pb[0].ret);
 
+       gip->nonmodal = 1;
     GWidgetHidePalettes();
     GDrawSetVisible(gip->gw,true);
     while ( !gip->done )
```

This successfully makes the dialog nonmodal, but it has a nasty side effect: unlike the PointGetInfo dialog, the SpiroPointGetInfo dialog no longer follows its master CharView when closed. If you attempt to use the dialog after the CharView close, you get the following predictable crash (SIGSEGV):

```gdb
#0  0x0000155554a83622 in SCTickValidationState (sc=0x5555558395c0, layer=-1) at splinechar.c:1861
#1  0x000015555517bac5 in CVSetCharChanged (cv=0x5555558395d0, changed=1) at charview.c:5132
#2  0x000015555517c292 in _CV_CharChangedUpdate (cv=0x5555558395d0, changed=1) at charview.c:5243
#3  0x000015555517c45d in CV_CharChangedUpdate (cv=0x5555558395d0) at charview.c:5271
#4  0x00001555551e56cf in PI_SpiroChanged (g=0x555555963fc0, e=0x7fffffffa270) at cvgetinfo.c:3457
#5  0x0000155555050d6e in GTextFieldChanged (gt=0x555555963fc0, src=-1) at gtextfield.c:209
#6  0x0000155555054974 in GTextFieldIncrement (gt=0x555555963fc0, amount=1) at gtextfield.c:1076
#7  0x00001555550573f6 in gnumericfield_mouse (gt=0x555555963fc0, event=0x7fffffffa660) at gtextfield.c:1571
#8  0x0000155555057d80 in gtextfield_mouse (g=0x555555963fc0, event=0x7fffffffa660) at gtextfield.c:1674
#9  0x0000155554ff81fc in _GWidget_Container_eh (gw=0x555555963650, event=0x7fffffffa660) at gcontainer.c:315
#10 0x0000155554ff9f1e in _GWidget_TopLevel_eh (gw=0x555555963650, event=0x7fffffffa660) at gcontainer.c:745
#11 0x0000155555074d01 in _GGDKDraw_CallEHChecked (gw=0x555555963650, event=0x7fffffffa660, eh=0x155554ff9590 <_GWidget_TopLevel_eh>)
    at ggdkdraw.c:335
#12 0x000015555507746d in _GGDKDraw_DispatchEvent (event=0x5555555b65c0, data=0x555555645cf0) at ggdkdraw.c:1203
#13 0x000015555417cae6 in  () at /usr/lib/libgdk-3.so.0
#14 0x00001555541ae985 in  () at /usr/lib/libgdk-3.so.0
#15 0x0000155552fa2a2f in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#16 0x0000155552fa45e9 in  () at /usr/lib/libglib-2.0.so.0
#17 0x0000155552fa462e in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#18 0x0000155555079eee in GGDKDrawProcessOneEvent (gdisp=0x555555645cf0) at ggdkdraw.c:2157
#19 0x0000155554ffd3a3 in GDrawProcessOneEvent (gdisp=0x555555645cf0) at gdraw.c:767
#20 0x00001555551e7d2a in SpiroPointGetInfo (cv=0x5555558395d0, scp=0x555555791118, spl=0x555555931b20) at cvgetinfo.c:3752
#21 0x00001555551e7f4b in CVPGetInfo (cv=0x5555558395d0) at cvgetinfo.c:3787
#22 0x0000155555202c3b in CVPopupSelectInvoked (v=0x55555591c570, mi=0x55555595fa00, e=0x7fffffffbd50) at cvpalettes.c:3472
#23 0x000015555501fe75 in gmenu_mouse (m=0x5555559605b0, event=0x7fffffffbd50) at gmenu.c:943
#24 0x0000155555021924 in gmenu_eh (w=0x55555595fb90, ge=0x7fffffffbd50) at gmenu.c:1381
#25 0x0000155555074d01 in _GGDKDraw_CallEHChecked (gw=0x55555595fb90, event=0x7fffffffbd50, eh=0x15555502184b <gmenu_eh>) at ggdkdraw.c:335
#26 0x000015555507746d in _GGDKDraw_DispatchEvent (event=0x555555945820, data=0x555555645cf0) at ggdkdraw.c:1203
#27 0x000015555417cae6 in  () at /usr/lib/libgdk-3.so.0
#28 0x00001555541ae985 in  () at /usr/lib/libgdk-3.so.0
#29 0x0000155552fa2a2f in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#30 0x0000155552fa45e9 in  () at /usr/lib/libglib-2.0.so.0
#31 0x0000155552fa462e in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#32 0x0000155555079f6a in GGDKDrawEventLoop (gdisp=0x555555645cf0) at ggdkdraw.c:2168
#33 0x0000155554ffd486 in GDrawEventLoop (gdisp=0x555555645cf0) at gdraw.c:787
#34 0x000015555539b7c0 in fontforge_main (argc=1, argv=0x7fffffffe178) at startui.c:1389
#35 0x0000555555555159 in main (argc=1, argv=0x7fffffffe178) at main.c:39
```

Fixing this should be trivial, but I'm out of energy for the night :-)

ctrlcctrlv:
PS. The last two lines of `static void SpiroPointGetInfo` seemingly do nothing. Why are they there?

```diff
@@ -3745,10 +3745,9 @@ static void SpiroPointGetInfo(CharView *cv, spiro_cp *scp, SplinePointList *spl)
 
        GHVBoxFitWindow(pb[0].ret);
 
+       gip->nonmodal = 1;
     GWidgetHidePalettes();
     GDrawSetVisible(gip->gw,true);
-    while ( !gip->done )
-       GDrawProcessOneEvent(NULL);
 }
```

skef:
(Partly based on supposition:)

Many infrastructures allow the lifecycle of a modal dialog to exist in a stack context -- basically to correspond to the lifecycle of a function call. That semantic allows (for example) a function that creates a dialog to return a value derived from the dialog. But for that to work (in a single-thread model) there needs to be code in the function to actually do the UI drawing and event handling. That's what this code does -- it's basically a "replica" of the event loop that is normally implemented higher up in the stack.

frank-trampe:
Is there a reason we couldn't just have `pi_e_h` call the okay or cancel handler routine directly on a close event? I'd like to cut the number of code paths as much as possible.

ctrlcctrlv:
I figured out how to make it non-modal at last, check https://github.com/fontforge/fontforge/pull/3469/commits/d46a11172344e96017747286687fbf28499d7628 :-)

ctrlcctrlv:
@frank-trampe I'm pretty sure my commit already does what you're asking :)

ctrlcctrlv:
@skef When you get a chance, please test it out and let me know what you think

frank-trampe:
@ctrlcctrlv, I mean to bind the code paths together by calling, for example PI_Ok or PI_Cancel if there are no bad side effects from that.

ctrlcctrlv:
@frank-trampe I'm not sure how to do that due to the mismatched types. `PI_Ok` and `PI_SpiroOk` expect a `GGadget`, while `PI_Destroy` expects a type only derivable from a `GWindow`, `GIData`&hellip;it's possible to get a `GWindow` from a `GGadget`, but I don't think it's possible to do the opposite in a sane way, and `pi_e_h` itself takes a `GWindow` 😵

I'm fixing a serious crash here, and the code path existed before my changes; I actually simplified matters a lot from its previous state. Someone with more talent than me will need to simplify things further, I'm afraid.

frank-trampe:
@ctrlcctrlv, then can we try the syntax from the patch from @jtanx? (With the DoCancel call conditional and the Destroy call following either way.) As this problem suggests, code gets copied and pasted a lot, and I want to be sure that we account for the modal and non-modal cases here. Likewise, if pushing to the point info dialogs list is conditional on modality, let's either note that or insert a conditional.

ctrlcctrlv:
@frank-trampe DoCancel does a revert while we decided in this thread that one isn't appropriate, especially if the dialog is no longer modal, so I can't restore that I don't think. I assume that it doesn't hurt to push a modal dialog onto `cv->pointInfoDialogs`, but the reason that it isn't done is because it's pointless as `dlist_foreach` will never be called on it.

ctrlcctrlv:
Is there anything I need to do to get this merged?

skef:
I just realized there's an aspect of the progression of these changes that might be unclear ...

If the dialog were still modal I would probably have done a review already and "signed off" if I didn't find any problems. Because the dialog is now non-modal, however, the scope of a proper review is much wider than just the code that's been changed. One now has to worry about all potential interactions between this window and others. And also how to judge the problems, given that there will likely be crashes that mirror those of the non-Spiro "Get Info" dialog, like #2829. Given that I'm not very familiar with the relevant code paths, it would take me a while just to ramp up to a review.

ctrlcctrlv:
Do you want me to revert back to the modal dialog?

The current crash is so egregious that even if there are minor problems or reactions with other windows/dialogs, I still think it'd be better to merge it. I've been using the `issue3437` branch merged into master of fontforge/fontforge for my work the past few days (except when reporting #3482 to be sure it happens even without my patch) and haven't gotten any crash.

skef:
It's not my place to stipulate what features should or should not be accepted. However, if you split the two changes into separate issues I would likely review the crash fix soon after.

ctrlcctrlv:
@skef Done. This PR now only fixes the crash, while PR#3484 will make the dialog non-modal (and also fix the crash since it relies on this being merged first).

skef:
As of this review point, the substantive change in the code is that instead of calling `PI_DoCancel` when `pi_e_h` is called for a modal dialog with event `et_close`, `PI_Destroy` is called. 

For reasons foreshadowed by @frank-trampe, this doesn't quite work. Run fontforge and open a new font. Then open a cv and create some Spiro splines. Then open the Get Info dialog for one point and press the close decoration -- the dialog will close. Now close the character window and then the font window. Unfortunately, FontForge is still running. Why? Because it's still hanging out in the loop at the end of `SpiroPointGetInfo()` waiting for `done` to be set. 

What we want is the equivalent of pressing OK in the dialog. But to get that right we need to know what kind of dialog we have, because `PI_Ok` and `PI_SpiroOk` don't quite do the same things. 

I'm contemplating this ...

skef:
How about something like this (off the current master)?
```
diff --git a/fontforgeexe/cvgetinfo.c b/fontforgeexe/cvgetinfo.c
index 200be05af..3759557d5 100644
--- a/fontforgeexe/cvgetinfo.c
+++ b/fontforgeexe/cvgetinfo.c
@@ -62,6 +62,7 @@ typedef struct gidata {
     GGadgetCreateData* gcd;
     GGadget *group1ret, *group2ret;
     int nonmodal;
+    void (*closefunc)(struct gidata *);
 } GIData;
 
 #define CID_BaseX      2001
@@ -1756,11 +1757,7 @@ static void PI_DoCancel(GIData *ci) {
 static int pi_e_h(GWindow gw, GEvent *event) {
     if ( event->type==et_close ) {
        GIData  *d = GDrawGetUserData(gw);
-       if( d->nonmodal ) {
-           PI_Destroy((struct dlistnode *)d);
-       } else {
-           PI_DoCancel( GDrawGetUserData(gw));
-       }
+       (*d->closefunc)(d);
     } else if ( event->type==et_char ) {
        if ( event->u.chr.keysym == GK_F1 || event->u.chr.keysym == GK_Help ) {
            help("getinfo.html");
@@ -1869,33 +1866,34 @@ void PI_Destroy(struct dlistnode *node) {
     free(d);
 }
 
-static void PI_Close(GGadget *g) {
-    GWindow gw = GGadgetGetWindow(g);
-    GIData  *d = GDrawGetUserData(gw);
+static void PI_Close(GIData *d) {
     PI_Destroy((struct dlistnode *)d);
 }
 
 static int PI_Cancel(GGadget *g, GEvent *e) {
     if ( e->type==et_controlevent && e->u.control.subtype == et_buttonactivate ) {
-       PI_DoCancel( GDrawGetUserData(GGadgetGetWindow(g)));
-       PI_Close(g);
+       GIData *d = GDrawGetUserData(GGadgetGetWindow(g));
+       PI_DoCancel(d);
+       PI_Close(d);
     }
 return( true );
 }
 
-static int PI_Ok(GGadget *g, GEvent *e) {
-    if ( e->type==et_controlevent && e->u.control.subtype == et_buttonactivate ) {
-       GIData *ci = GDrawGetUserData(GGadgetGetWindow(g));
-
+static void PI_DoOk(GIData *ci) {
        PI_FixStuff(ci);
        _PI_ShowHints(ci,false);
 
        ci->done = true;
        /* All the work has been done as we've gone along */
-       PI_Close(g);
-    }
+       PI_Close(ci);
+}
 
-return( true );
+static int PI_Ok(GGadget *g, GEvent *e) {
+    if ( e->type==et_controlevent && e->u.control.subtype == et_buttonactivate ) {
+       GIData *ci = GDrawGetUserData(GGadgetGetWindow(g));
+       PI_DoOk(ci);
+    }
+    return( true );
 }
 
 static void mysprintf( char *buffer, char *format, real v) {
@@ -2694,6 +2692,7 @@ static void PointGetInfo(CharView *cv, SplinePoint *sp, SplinePointList *spl) {
     gi->curspl = spl;
     gi->oldstate = SplinePointListCopy(cv->b.layerheads[cv->b.drawmode]->splines);
     gi->done = false;
+    gi->closefunc = PI_DoOk;
     CVPreserveState(&cv->b);
 
     root = GDrawGetRoot(NULL);
@@ -3462,15 +3461,18 @@ static int PI_SpiroChanged(GGadget *g, GEvent *e) {
 return( true );
 }
 
+static void PI_SpiroDoOk(GIData *ci) {
+       ci->done = true;
+       /* All the work has been done as we've gone along */
+       PI_Close(ci);
+}
+
 static int PI_SpiroOk(GGadget *g, GEvent *e) {
     if ( e->type==et_controlevent && e->u.control.subtype == et_buttonactivate ) {
        GIData *ci = GDrawGetUserData(GGadgetGetWindow(g));
-
-       ci->done = true;
-       /* All the work has been done as we've gone along */
-       PI_Close(g);
+       PI_SpiroDoOk(ci);
     }
-return( true );
+    return( true );
 }
 
 static void SpiroPointGetInfo(CharView *cv, spiro_cp *scp, SplinePointList *spl) {
@@ -3493,6 +3495,7 @@ static void SpiroPointGetInfo(CharView *cv, spiro_cp *scp, SplinePointList *spl)
     gip->curspl = spl;
     gip->oldstate = SplinePointListCopy(cv->b.layerheads[cv->b.drawmode]->splines);
     gip->done = false;
+    gip->closefunc = PI_SpiroDoOk;
     CVPreserveState(&cv->b);
 
     root = GDrawGetRoot(NULL);
```

ctrlcctrlv:
@skef Sure ! :smile: Do you want me to close this and you can open your own? Or you can just push to issue3437 branch...

Sorry I did so poorly you had to redo all my work :disappointed:

skef:
> Sorry I did so poorly you had to redo all my work

Naw, it's not a straightforward problem. And here I am using a dasterdly function pointer. 

Unfortunately, having written this code I can no longer review it.

ctrlcctrlv:
>Unfortunately, having written this code I can no longer review it.

Paging @frank-trampe :+1:

ctrlcctrlv:
Bump. (Is bumping allowed?)

fontforgeexe/cvgetinfo.c
```diff
@@ -62,6 +62,7 @@ typedef struct gidata {
     GGadgetCreateData* gcd;
     GGadget *group1ret, *group2ret;
     int nonmodal;
+    void (*closefunc)(struct gidata *);
```
jtanx: Not a fan of this, adding a function pointer just for this introduces too much indirection.

Since it's not used any more, rename the `nonmodal` flag to be `is_spiro` or something like that, and just do a conditional on that in the close event

skef:
Even simpler! The help event should probably display a separate page for Spiro's dialog anyway :-P

PR #3476, opened 2019-02-14
Drop outdated warning about extension lookups

khaledhosny:
Extension lookups used to be proplemetaic in the past (some old versions of Microsoft office), but this is not an issue any more and many tools (e.g. FontTools) will generate such lookups without issuing warnings.

### Types of changes


khaledhosny:
I removed the commit that was making `-quiet` suppress `LogError()` as I think it was heavey handed. I think I’ll see if I can introduced `LogWraning()` and reserve `LogError()` for hard errors.

PR #3478, opened 2019-02-15
Handle reused mark filtering set in feature files

khaledhosny:
The subsequent used of a mark filtering set will cause parsing errors due to incorrectly set flag mask.

### Types of changes


frank-trampe:
@khaledhosny, let's leave the zombie comma code. It is helpful to have such explanation of non-adherence to published specifications for future developers. The other commits look fine to me.

khaledhosny:
The [current spec](https://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#4.d) says the flags are space separated, so these comments are outdated.

PR #3479, opened 2019-02-15
Add an option to not generate duplicated Mac names

khaledhosny:
Modern macOS applications will use name entries from MS platform if the Mac platform is missing. They also use 8-bit encodings that do not cover the whole Unicode, which can be problematic for non-ASCII names.

This adds an option to not generate the legacy Mac names (which FontForge does by default).

### Types of changes

### DIFF
--- a/fontforgeexe/savefontdlg.c
+++ b/fontforgeexe/savefontdlg.c
@@ -87,6 +87,7 @@ static int nfnt_warned = false, post_warned = false;
 #define CID_TTF_OldKern		1109
 #define CID_TTF_GlyphMap	1110
 #define CID_TTF_OFM		1111
+#define CID_TTF_NoMacNames	1112
 #define CID_TTF_PfEdLookups	1113
 #define CID_TTF_PfEdGuides	1114
 #define CID_TTF_PfEdLayers	1115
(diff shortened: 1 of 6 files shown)


fontforgeexe/savefontdlg.c
```diff
@@ -87,6 +87,7 @@ static int nfnt_warned = false, post_warned = false;
 #define CID_TTF_OldKern		1109
 #define CID_TTF_GlyphMap	1110
 #define CID_TTF_OFM		1111
+#define CID_TTF_NoMacNames	1112
```
frank-trampe: Are you sure we're not stepping on any phantom toes here?

frank-trampe:
Are you sure you want to pose the option as a negative?

khaledhosny:
> Are you sure you want to pose the option as a negative?

I don’t want to change the default, making it `macnames` would require that.

fontforgeexe/savefontdlg.c
khaledhosny: Of course not, but I tested it and it worked.

frank-trampe: Okay. I guess since the value doesn't end up in SFD, we can always renumber it if we find a problem later.

PR #3484, opened 2019-02-21
Close #3483

### LINKED ISSUE #3483: Spiro dialog should not be modal
Per https://github.com/fontforge/fontforge/pull/3469#issuecomment-464050310

ctrlcctrlv:
This makes the dialog non-modal, splitting PR#3469 in two; #3469 now only fixes the crash.


ctrlcctrlv:
Closing voluntarily due to #3469 being totally reworked - I'll have to base this off of https://github.com/fontforge/fontforge/commit/9e271c125ac100760333d36b47a4715e1c51dbdb now, not https://github.com/fontforge/fontforge/pull/3484/commits/bed0688bf75271f496d09b1febc3adb3615540b1

skef:
It seems that if you have multiple tabs open in the same CharView window, open a "normal" Get Info dialog on a point in one tab, and then shift to another tab, the point in the dialog doesn't change. But then if you use the "next point" button you wind up on a point in some contour in the current tab. But "next on contour" (when pressed first) just cycles around the points in the contour on the invisible tab. 

That seems ... bad.

ctrlcctrlv:
This has got to be refactored to your new force-pushed branch (#3469) but I'll keep that in mind while doing so :smile: (if and when I get around to doing so, and if you haven't already done so yourself)

Issue #3482, opened 2019-02-20
Attempting to add an anchor results in infinite "Anchor Class Name" windows


ctrlcctrlv:
### When reporting a bug/issue:
- [x] Screenshot
![bug](https://user-images.githubusercontent.com/838783/53088574-1ef68a00-3545-11e9-8e38-5c3b8e879839.jpg)

- [x] The FontForge version and the operating system you're using
Arch Linux x64, https://github.com/fontforge/fontforge/commit/6abdba08e52d9431fcc49936c4e1127ac542d225
- [x] The behavior you expect to see, and the actual behavior
I expect a mark class to be added after I enter the name of it and press OK / enter key on keyboard.
- [x] Steps to reproduce the behavior 
* Open [N~.sfd](https://github.com/fontforge/fontforge/files/2884400/N.zip) ; or make your own new font without a `mark` OT feature
* Open tilde character view
* Right click, add anchor
* Attempt to enter anchor name
* Bug occurs

- [x] \(optional) Possible solution/fix/workaround
Create a `mark` feature in GPOS and add some classes before trying to add an anchor. Very non-obvious for users that you need to do this, or that it will even work.

khaledhosny:
The bug happens when one presses Enter, pressing the OK button will work fine. Similar thing happens with the Anchor Point Info dialog, pressing enter there will open a new dialog, attempting to close them will crash FontForge.

But, I think you need to create a mark or mkmk lookup first, add anchor classes, then add anchors to the glyphs. Adding anchors without lookups will not end up in the compiled font.

skef:
This fixes the problem and is at least roughly right but it's hard to evaluate the scope of potential side effects. (At worst (I *think*) menu acceleration for popup menus could be missing in some contexts, as that's what `GMenuPopupCheckKey(event)` is responsible for.)

```
diff --git a/gdraw/gcontainer.c b/gdraw/gcontainer.c
index da1c88ba2..9f6f3d012 100644
--- a/gdraw/gcontainer.c
+++ b/gdraw/gcontainer.c
@@ -496,7 +496,8 @@ static int _GWidget_TopLevel_Key(GWindow top, GWindow ew, GEvent *event) {
     }
     /* Check for mnemonics and shortcuts */
     if ( event->type == et_char && !GKeysymIsModifier(event->u.chr.keysym) ) {
-       handled = GMenuPopupCheckKey(event);
+       if ( topd->popupowner != NULL )
[...]

PR #3494, opened 2019-02-26
Fix for #3482 infinite "Anchor Class Name" windows with return char

skef:
I'm making this into a push request instead of just a diff in the issue because

1. I'm afraid it will get lost otherwise and
2. even if it's not perfect it's probably better than the current code. 

Here is the problem in summary: 

`GMenuPopupCheckKey()` is responsible for calling `gmenu_key()` on the current popup menu (if there is one) to convert relevant key events into popup menu operations. A return key event is "relevant" in that it gets passed to `GMenuSpecialKeys()`. This code seems to exist to support keyboard selection of popup-menu items, but I haven't found any way of getting that to work by actually using the keyboard. 

Anyway, what winds up happening is that even though the old popup menu from which "Anchor Point" was selected is no longer visible, it hasn't been cleaned out of (static) `most_recent_popup_menu`. "Return" maps, roughly speaking, to activating the currently selected item which is "Anchor Point" -- because that's what was selected to activate the dialog in the first place. Hence the new dialog. 

With this change `GMenuPopupCheckKey()` is called only if the current window is the "`popupowner`". There is a similar clause shortly below. I *think* this is fine, and anyway `GMenuPopupCheckKey()` may serve no constructive purpose now. But I'm not certain.  

### Types of changes

### Final checklist

### DIFF
--- a/gdraw/gmenu.c
+++ b/gdraw/gmenu.c
@@ -1313,6 +1313,9 @@ static int gmenu_key(struct gmenu *m, GEvent *event) {
     GMenu *top;
     unichar_t keysym = event->u.chr.keysym;
 
+    if ( m->dying )
+	return( false );
+
     if ( islower(keysym)) keysym = toupper(keysym);
     if ( event->u.chr.state&ksm_meta && !(event->u.chr.state&(menumask&~(ksm_meta|ksm_shift)))) {
 	/* Only look for mneumonics in the child */


ctrlcctrlv:
Just wanted to say thanks for this&mdash;issue's not really a big deal but things like this make FontForge much more usable, users don't have to remember as many work arounds.

jtanx:
I think this stops the hotkeys on popup menus from working. The way you activate them is via `Alt+letter` where `letter` is the letter that is underscored in the name of the option. (Really though you should be able to select it just by pressing the letter, but this is just another gdraw quirk)

skef:
Since you've been in the gdraw code a lot, can you help me understand why this doesn't work? Why isn't the target for key events in a window with a popup the popup owner? Or more specifically why is `topd->popupowner` NULL in the normal case? What is `popupowner` for if not for this?

jtanx:
The answer to that is, I don't know :P

Debugging gdraw code involves a lot of stepping through in gdb and a ton more of hair pulling...

skef:
You recently said that one transient is gone before another is displayed. But this ordering implies the opposite -- the static pointing the menu is still populated when the dialog comes up. The other route to fixing this is to get that static NULL again. Is there some further cleanup step beyond "destroy" that clears that pointer?

jtanx:
Alright, I think I see what's happening.

When the menu event handler is called, the request to destroy it has already been sent, and the window (as far as gdraw is concerned) is dead/dying. However, it's not cleaned up until that menu window receives the destroy notification.

So picture this (although I'm fairly sure this applies to gxdraw, just implicitly in how windows are handled): the reference count on the menu window when the event is received is 2; 1 for when the window was created, and another for when the event was dispatched.

The menu event is processed, and GDrawDestroyWindow is called. and it does this, decementing the reference count. It's now at 1, still above 0, because it's still in the event dispatch from that window.

The anchor info window is created, and then *it* pumps the message loop - while still being in the event dispatch of that menu, so it can never be destroyed, *until* the anchor window is closed - only then is it finally able to be removed, because the event dispatch finally finishes, the reference count drops to 0, and the et_destroy message is published.

I've run out of time today to look into this, but perhaps menu events can be ignored (there's an is_dying flag in gmenu.c). The other option is to post the creation of the anchor info window, so the menu window can be cleaned up (i.e. some method of asynchronously creating the window, instead of pumping the loop off the menu's event dispatch)

skef:
Thank you -- that's all very helpful. I might be able to modify the popup event code to check for dying.

skef:
This new version preserves the menu alt shortcuts and arrow key behavior while still fixing the problem.

Issue #2882, opened 2016-09-30
OpenType 1.8 support yet?


aaron-elkins:
Does FF support to export as OpenType 1.8 (Open­Type Variable Font) right now?

frank-trampe:
Ohh no. We weren't invited to participate in the design of the specification, so everything is new to us. It will probably be a long time. It is a very large set of additional functionality, and, before adding it, we would need to figure out a user interface model. And we would probably need to get outside funding for such a big project.

KrasnayaPloshchad:
FontLab and Glyphs has already got support OT 1.8, and HarfBuzz also supported OT Variable font.

http://www.typeisbeautiful.com/2016/09/10968/zh-hant/

KrasnayaPloshchad:
This is very useful to you if you want to develop a CJK or Tangut font with FontForge.

KrasnayaPloshchad:
On [this article](http://www.typeisbeautiful.com/2016/09/10968/zh-hant/) I found some informations about the support for this:

> 2016 年 9 月在華沙召開的第 60 屆 ATypI 年會上，OpenType 1.8 正式發布。它基於的不是 Adobe 的 MM，而是蘋果貢獻的 TrueType GX 技術。谷歌開發了一個可以在 Win­dows 上實時預覽 1.8 版字體的小工具叫 FreeView。微軟在 Windows 10 的周年慶典版 (Anniversary Edition) 的 Edge 瀏覽器中，已經率先支持了新版的規範。而蘋果則走得更遠：macOS Sierra 中的系統字體 San Francisco UI，已經變成了 OpenType 可變字體。在會議召開的同時，相關人士也着手與 W3C 的 CSS 工作組洽談，以求讓新規範更快獲得廣泛的瀏覽器支持。

My translation: _On the 60th Annual Meeting of ATypeI, OpenType 1.8 is released. It is not based on Adobe MM, instead is TrueType GX techonology contributed by Apple. Google have developed a small tool named FreeView, which can preview OT 1.8 font lively on Windows. Microsoft have already supported new specification firstly in Edge on Windows 10 Anniversary Edition. But Apple go much further: Mac OS Sierra’s system font San Francisco UI is become OpenType Variable Font. During the meeting, related persons are also take actions to contact CSS Working Group of W3C, to let the new spec being supported by more web browsers soon._

This paragraph mentiond some softwares and fonts supported this spec, I guess we can do some hackings on them to know how this technology works, then everything would be clear.

[...]

PR #3505, opened 2019-03-07
Add more OpenType table names

MaddTheSane:
This adds table names for some OpenType tables.

### Motivation and Context
- [X] Why is this change required? What problem does it solve?
This adds more OpenType table names.
Somewhat related to #2882.

### Types of changes
- [X] New feature (non-breaking change which adds functionality)

### Description
- [X] Describe your changes in detail.
This adds descriptions for the following tables: CBDT, CBLC, CFF2, HVAR, MERG, meta, MVAR, six, STAT, 'SVG ', and VVAR.

### Final checklist
- [X] My code follows the code style of this project found [here](https://github.com/fontforge/fontforge/blob/master/CONTRIBUTING.md#coding-style).


ctrlcctrlv:
Hi - why add these table names when FontForge provides no UI for editing them?

MaddTheSane:
For future work. And for reference.

Also, there are already tables in the list that aren't currently supported by FontForge (the justification tables, `Zaph`, for starters).

ctrlcctrlv:
So if FontForge doesn't know how to edit a certain table will it just appear greyed out, or ... ?

MaddTheSane:
It pops up in the warnings window. With this patch, it says what the unreadable table is.

skef:
> So if FontForge doesn't know how to edit a certain table will it just appear greyed out, or ... ?

These aren't the lookup names, as in the Lookups dialog, but a level above that. The risk would be FontForge's logic thinking it knew more about one of these tables than it does. 

@MaddTheSane: for potential reviewers, can you add some references to code points (filename/line number or just quoted code) verifying the safety of these changes?

MaddTheSane:
This specific table is only referenced four times in the source code, all in reference to printing info about unknown tables (look for `stdtables`):
https://github.com/fontforge/fontforge/blob/b561e1aa3a1ca6b971f6dcc21e06d2cea7104547/fontforge/parsettf.c#L1232-L1240

ctrlcctrlv:
Referring to the tag that way (`%c%c%c%c`, then bitshifting for some reason to get the table name) can't be a good programming practice, can it?

MaddTheSane:
It's probably the fastest way to print an ASCII FourChar instead of byte-swapping (on Little Endian machines), copying the resulting integer into a temporary buffer, and print the resulting string out.

Note that I did not directly touch the code that references `stdtables`.

PR #3506, opened 2019-03-07
Cross-application modal input restriction

### LINKED ISSUE #3498: GDK modality changes
This is (probably?) not a bug report, but a question. I guess the dev list would be a better place for this officially, but "coverage" there seems spotty.

gxdraw implements the familiar sort of application-wide modal window. The "trick" is more or less i the `inputRedirect` stuff in `GXDrawSetVisible()` and the other functions that deal with `gdisp->input`. 

At present ggdkdraw implements a "lighter" version of modality by way of `_GGDKDraw_FilterByModal()`. This more or less checks for windows in the chain from that receiving the event up to the root, and throws away events if any of those have a modal dialog (which, being a separate window, isn't in the chain). 

This is "lighter" because you can use other windows at the same time. For example, in gxdraw if you have a modal "Add Lookup" dialog open on top of the Font Info -> Lookups window, you have to close it before doing anything else with the application. In ggdkdraw you can use the Font and CharViews at the same time -- only the Font Info window is blocked. 

Considered in isolation as a UI semantic, I'm not really "bothered" by this or that worried that users may find it confusing. (So this is different from the Point Info dialog concerns I raised in another recent issue.) But I am worried in a program-correctness sense. Dialogs are often modal for a reason, with the code relying on nothing else changing for its lifetime. 

So: do we *know* this *isn't* a problem, or did the implementation just "wind up" this way? 

I take it @jtanx is the expert on this subject.

skef:
This is my variant of @jtanx's https://github.com/fontforge/fontforge/tree/gdk-transients/gdraw branch. It maintains a count of the restricted windows, partly to speed up `_GGDKDraw_FilterByModal` but mostly to make it easy to a) check whether there is currently an input restriction and b) trigger a change on a 0->1 or 1->0 transition. (Hence the currently one-line `GGDK_ChangeRestrictCount`.) 

I've been using this for the past few days and it seems to work fine, but on the inspection level there's a potential problem:
```
static void GGDKDrawSetVisible(GWindow w, int show) {
    Log(LOGDEBUG, "0x%p %d", w, show);
    GGDKWindow gw = (GGDKWindow)w;
    _GGDKDraw_CleanupAutoPaint(gw->display);
    if (show) {
        if (gdk_window_is_visible(gw->w)) {
            Log(LOGDEBUG, "0x%p %d DISCARDED", w, show);
            return;
        }
#ifdef GDK_WINDOWING_QUARTZ
        // Quartz backend fails to send a configure event after showing a window
        // But FF expects one.
        _GGDKDraw_OnFakedConfigure(gw);
#endif
        gdk_window_show(gw->w);
    } else {
        GGDKDrawSetTransientFor((GWindow)gw, NULL);
        gdk_window_hide(gw->w);
    }
}
```
This works fine for new windows because they've already had Transients set when relevant. And it works fine on a visible->invisible transition because of the `GGDKDrawSetTransientFor()` call. But on its face it's not likely to work when making a window that has been hidden visible again. 

Do we know that either:
1. We never make hidden transient windows visible again, or
2. the code that does so already calls `GDrawSetTransientFor()` to handle the case?  

Closes #3498

### Types of changes

### Final checklist

### DIFF
--- a/gdraw/ggdkdraw.c
+++ b/gdraw/ggdkdraw.c
@@ -791,25 +791,25 @@ static bool _GGDKDraw_FilterByModal(GdkEvent *event, GGDKWindow gw) {
     GPtrArray *stack = gw->display->transient_stack;
 
     if ( gww && gw->display->restrict_count == 0 )
-	return false;
+        return false;
 
     while (gww != NULL) {
-		GGDKWindow last_modal = NULL;
+        GGDKWindow last_modal = NULL;
 
-		for (int i = ((int)stack->len)-1; i >= 0; --i) {
-			GGDKWindow ow = (GGDKWindow)stack->pdata[i];
-			if (ow == gww || ow->restrict_input_to_me) { // fudged
-				last_modal = ow;
-				break;
-			}
-		}
+        for (int i = ((int)stack->len)-1; i >= 0; --i) {
+            GGDKWindow ow = (GGDKWindow)stack->pdata[i];
+            if (ow == gww || ow->restrict_input_to_me) { // fudged
+                last_modal = ow;
+                break;
+            }
+        }
 
-		if (last_modal == NULL || last_modal == gww) {
-			return false;
-		}
+        if (last_modal == NULL || last_modal == gww) {
+            return false;
+        }
 
-		gww = gww->parent;
-	}
+        gww = gww->parent;
+    }
 
     if (event->type != GDK_MOTION_NOTIFY && event->type != GDK_BUTTON_RELEASE) {
         gdk_window_beep(gw->w);
@@ -998,7 +998,7 @@ static void _GGDKDraw_DispatchEvent(GdkEvent *event, gpointer data) {
                         gw == gdisp->bs.release_w &&
                         gevent.u.mouse.button == gdisp->bs.release_button &&
                         (int32_t)(gevent.u.mouse.time - gdisp->bs.last_press_time) < gdisp->bs.double_time &&
-                        gevent.u.mouse.time >= gdisp->bs.last_press_time) {	// Time can wrap
+                        gevent.u.mouse.time >= gdisp->bs.last_press_time) {  // Time can wrap
 
                     gdisp->bs.cur_click++;
                 } else {
@@ -1576,19 +1576,19 @@ static void GGDKDrawSetTransientFor(GWindow transient, GWindow owner) {
         gdk_window_set_transient_for(gw->w, ow->w);
         gdk_window_set_modal_hint(gw->w, true);
         gw->istransient = true;
-	g_ptr_array_add(gdisp->transient_stack, gw);
-	if (gw->restrict_input_to_me)
-	    _GGDKDraw_ChangeRestrictCount(gdisp, 1);
+        g_ptr_array_add(gdisp->transient_stack, gw);
+        if (gw->restrict_input_to_me)
+            _GGDKDraw_ChangeRestrictCount(gdisp, 1);
     } else {
         gdk_window_set_modal_hint(gw->w, false);
         gdk_window_set_transient_for(gw->w, gw->display->groot->w);
         gw->istransient = false;
 
-	// Be extra careful and ensure the window was on the list, in
-	// case there are spurious remove calls.
+        // Be extra careful and ensure the window was on the list, in
+        // case there are spurious remove calls.
         // TODO: optimise; search from the back; it is meant to be a stack...
         if ( g_ptr_array_remove(gdisp->transient_stack, gw) && gw->restrict_input_to_me )
-	    _GGDKDraw_ChangeRestrictCount(gdisp, -1);
+            _GGDKDraw_ChangeRestrictCount(gdisp, -1);
     }
 
     gw->transient_owner = ow;
--- a/gdraw/ggdkdraw.c
+++ b/gdraw/ggdkdraw.c
@@ -205,7 +205,7 @@ static gboolean _GGDKDraw_OnWindowDestroyed(gpointer data) {
                     tw->istransient = false;
                     if (tw->restrict_input_to_me)
                         _GGDKDraw_ChangeRestrictCount(gw->display, -1);
-                    }
+                }
             }
             if (!gdk_window_is_destroyed(gw->w)) {
                 gdk_window_destroy(gw->w);
--- a/gdraw/ggdkdraw.c
+++ b/gdraw/ggdkdraw.c
@@ -172,7 +172,7 @@ static bool _GGDKDraw_TransmitSelection(GGDKDisplay *gdisp, GdkEventSelection *e
     return true;
 }
 
-static void GGDK_ChangeRestrictCount(GGDKDisplay *gdisp, int c) {
+static void _GGDKDraw_ChangeRestrictCount(GGDKDisplay *gdisp, int c) {
     gdisp->restrict_count += c;
 }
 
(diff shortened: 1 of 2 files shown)


gdraw/ggdkdraw.c
```diff
@@ -785,30 +789,44 @@ static bool _GGDKDraw_FilterByModal(GdkEvent *event, GGDKWindow gw) {
...
+    if ( gtkb_Grabbed(gww->display->gtkb_state) ) {
+	// If gtk has an active grab, push the (already filtered)
+	// event up to it.
+	gtkb_ProcessEvent(gww->display->gtkb_state, event);
+	// Keep the "beep" (XXX?)
+	gww = NULL;
     }
+#endif
+
+    if ( gww && gw->display->restrict_count == 0 )
```
jtanx: I'm not 100% on the need for `restrict_count`, vs. just checking the length of the transient stack/list.

gdraw/ggdkdraw.c
```diff
@@ -192,12 +196,17 @@ static gboolean _GGDKDraw_OnWindowDestroyed(gpointer data) {
...
-                tw->istransient = false;
-            }
+			for (int i = ((int)gw->display->transient_stack->len) - 1; i >= 0; --i) {
+				GGDKWindow tw = (GGDKWindow)gw->display->transient_stack->pdata[i];
+				if (tw->transient_owner == gw) {
+					g_ptr_array_remove_index(gw->display->transient_stack, i);
+					gdk_window_set_transient_for(tw->w, gw->display->groot->w);
+					tw->transient_owner = NULL;
+					tw->istransient = false;
+					if ( tw->restrict_input_to_me )
```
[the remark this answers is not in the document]
skef: Half of that is yours!

But will be fixed anyway.

gdraw/ggdkdraw.c
```diff
@@ -785,30 +789,44 @@ static bool _GGDKDraw_FilterByModal(GdkEvent *event, GGDKWindow gw) {
...
-            }
-        }
-        gww = gww->parent;
-    }
+    GGDKWindow gww = gw;
+    GPtrArray *stack = gw->display->transient_stack;
 
-    if (gww == gw->display->groot) {
-        return false;
+#ifdef FONTFORGE_CAN_USE_GTK_BRIDGE
```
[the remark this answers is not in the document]
skef: That shouldn't be there -- leaked in from my backport. Will remove. 

Do you have interest in seeing my GTK compatibility branch? I don't have any custom signals in yet, but it basically works.

gdraw/ggdkdraw.c
```diff
@@ -172,6 +172,10 @@ static bool _GGDKDraw_TransmitSelection(GGDKDisplay *gdisp, GdkEventSelection *e
     return true;
 }
 
+static void GGDK_ChangeRestrictCount(GGDKDisplay *gdisp, int c) {
+    gdisp->restrict_count += c;
```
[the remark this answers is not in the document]
skef: Indulge me this one.

gdraw/ggdkdraw.c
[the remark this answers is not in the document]
jtanx: That's a case of do as I say, not as I do :P

That commit was also fairly wip, so I wouldn't have left it as-is

gdraw/ggdkdraw.c
skef: Clearly not everything in the transient list is meant to have `restrict_input_to_me` set, given that the code checks the flag. And it doesn't seem like the bottom would be guaranteed to have it.

skef:
Re-pushed.

gdraw/ggdkdraw.c
```diff
@@ -1566,20 +1572,23 @@ static void GGDKDrawSetTransientFor(GWindow transient, GWindow owner) {
...
+	    _GGDKDraw_ChangeRestrictCount(gdisp, 1);
     } else {
         gdk_window_set_modal_hint(gw->w, false);
         gdk_window_set_transient_for(gw->w, gw->display->groot->w);
         gw->istransient = false;
+
+	// Be extra careful and ensure the window was on the list, in
+	// case there are spurious remove calls.
+        // TODO: optimise; search from the back; it is meant to be a stack...
+        if ( g_ptr_array_remove(gdisp->transient_stack, gw) && gw->restrict_input_to_me )
```
jtanx: I'd recommend you run the autoformatter if you can, because I still see mixed tabs/spaces here. (although I know it indents some macros which I'd prefer it didn't, not sure how to get around that...)

gdraw/ggdkdraw.c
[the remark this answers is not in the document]
jtanx: Yeah, that'd be interesting to see I guess. Was interested in moving off GDK, given that what we do is basically impossible with the next iteration of GDK

gdraw/ggdkdraw.c
skef: It's one thing to have different styles in different files. It's another to divide the project up into domains with independent tool chains. The project can't agree on a new style, so everything that gets added has a style determined by whoever added it?

jtanx: I still think we should format the whole code base (there was a long discussion on this somewhere), and adhere to it. The biggest drawback that was raised was losing meaningful git blame, although there are some workarounds there.

The common theme I generally see for new code is:
* Use spaces for indentation, don't mix tabs/spaces, don't dedent returns
* Egyptian braces

But yes, generally whoever wrote it will have a different flavour on that. Bugfix code usually (but not always) also follows the indentation style of the surrounding code.

As for this case in particular, all the GDK code conforms to the astyle formatter code I mentioned (barring the indentation of the ignore deprecated warning macros). You don't have to use the formatter if you don't want to, but it should generally conform to what it does.

skef: This should be done now.

gdraw/ggdkdraw.c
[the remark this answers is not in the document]
skef: Let's talk over at jtanx/fontforge#2

gdraw/ggdkdraw.c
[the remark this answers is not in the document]
khaledhosny: I second @jtanx here, it is just one line that does not do anything complex. Also the name is a bit misleading, it increments the count not just changing it.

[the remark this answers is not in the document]
skef: c can be negative. 

This function becomes
```
static void _GGDKDraw_ChangeRestrictCount(GGDKDisplay *gdisp, int c) {
    gdisp->restrict_count += c;

    if ( c > 0 && (gdisp->restrict_count - c) == 0 ) {
        gtkb_Grab(gdisp->gtkb_state, true);
    } else if ( c < 0 && gdisp->restrict_count == 0 ) {
        gtkb_Grab(gdisp->gtkb_state, false);
    }
}
```
once we track 0->1 and 1->0 transitions.

skef:
The "fun test case" reproduces the same way with this code.

gdraw/ggdkdraw.c
skef: My first response was a little terse. 

Suppose you right-click on a CharView to bring up the popup menu. I believe that the menu "window" is a transient for the CharView but doesn't set `restrict_input_to_me` (you can use other windows while that menu is "popped"). Now suppose you use that menu to set an anchor point. GDraw may not show the menu when the subsequent dialog is displayed, but it's "there" and will only get cleaned up later. I *believe* it's in the stack still as well, but I'm not sure. Anyway, there are other plausible scenarios in which you can have: a) a base window, b) a transient non-restricting window, and then c) a "restricting" or modal transient window. 

So what GTK would call a "grab" -- where you can't use other windows -- corresponds to whether there is a displayed window that sets `restrict_count_to_me`. Once you have one of these, any further transients *setting that flag* should be transients for that window (unless some other event, like a python call, brings up another modal window somehow) , but the number is as arbitrary as the number of modals that can bring up further modals (in practice, probably 3?). 

So to determine whether there is presently a modal grab somewhere, you can either go through the full transient list or track the number of `restrict_input_to_me` flags among those transients. And if you want a *signal* when the status changes, you need to do one of those two things whenever the list contents change. Since there was already optimization around adding and removing elements from the list, I went with tracking the value rather than iterating every time.

jtanx: Ok, I've spent way too long looking at this. In part, a transient window *implies* modality (or at least modality for the window it was transient for).

Wrt. the case you mention, the menu is actually destroyed before the new window is created, and it is also marked with the restrict_input_to_me flag.

Anyway, I've kept it, but I've moved everything into the GGDKDrawSetTransientFor call, so if you need to track that transition, do it in there. (there was also a potential bug of the count being incremented too much if say GGDKDrawSetTransientFor was called multiple times on the same window (where ow!=null), although I guess this isn't really a valid thing to do.

I've also added a check that setting transients can only be done on toplevel (and not child) windows.

skef: That works. The new review looks good to me.

jtanx:
That's a good point, really if there were transients on a window, you wouldn't expect them to live for longer than the window they were transient for. 

So if everything was working as expected, this code should never run.

I guess this was aimed more at not leaving around invalid pointers, but perhaps it makes more sense to destroy them. Hmmm...

skef:
> That's a good point, really if there were transients on a window, you wouldn't expect them to live for longer than the window they were transient for.
> 
> So if everything was working as expected, this code should never run.
> 
> I guess this was aimed more at not leaving around invalid pointers, but perhaps it makes more sense to destroy them. Hmmm...

That was an idle thought, but I just came up with a test case:

1. Bring up the font info window
2. Go to lookups and open the "add lookup" dialog
3. Close the font info window with the window manager decoration. It closes!
4. Press "cancel" in the still open dialog -> core dump

So I guess this is a genuine, if corner case, problem.

skef:
This is likely also a gxdraw problem, and similar to the #3437 thing. I think it's fine to file a brief issue and deal with it separately and later.

jtanx:
That's odd, which window manager are you using again? I can't get this to occur with fluxbox (I can't close the font info dlg while the add lookups one is open)

skef:
I guess the window manager for KDE Plasma is "KWin". 

With KDE, if the window decoration doesn't close it, what about the right-click menu for the window in the task manager?

jtanx:
It doesn't close via that either

input

image

skef:
Weird -- not sure what to say. I suppose my environment could be confused; it's been a while since I rebooted.

jtanx:
I'll merge it for now, but if you keep running into it, I guess file another issue, maybe with a stack trace if you can

PR #3513, opened 2019-03-09
Expose interpolated quadratic points in python contours

skef:
This extends #3467 to property set `selected` on off-curve points in quadratic contours. It includes the commit from that one, so if this is merged that should be closed. 

To do this properly and also expose the selection status of interpolated on-curve points, this also adds an `interpolated` flag and includes those in the python contour. Most code should be fine with interpolated points, especially given that points can be marked as never-interpolated in FontForge. Users who don't want them, perhaps because an old script doesn't expect them, can strip them out with the documented code fragment. 

The `interpolated` flag is ignored when setting a layer. Setting aside `never_interpolated`, FontForge generally treats any point midway between its control points as interpolated. Therefore there has never been much difference between including and omitting an interpolatable point. (Future code could add an error on encountering a non-midway-point marked `interpolated`, but ignoring it is easier and better in almost all cases.) 

Note that the section status of off-curve points is tracked using `nextcpselected` on the previous off-curve point except for the last off-curve point in the list, in which case it is tracked with `prevcpselected` on the first point. I suspect this is a *de facto* standard based on the code in `CheckPoint()`. It might not make sense, but it's hard to judge how much code depends on it so I'm just following it in this new code. 

### Types of changes

### Final checklist

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -1774,7 +1774,7 @@ static PyObject *PyFFPoint_dup(PyFF_Point *self) {
 static PyFF_Point *PyFFPoint_Parse(PyObject *args, bool dup, bool always) {
     double x,y;
     PyFF_Point *p=NULL;
-    int i, on, sel;
+    int i, on, sel, interp;
 
     if ( args==NULL && !always )
 	return NULL;
@@ -1782,13 +1782,14 @@ static PyFF_Point *PyFFPoint_Parse(PyObject *args, bool dup, bool always) {
     x = y = 0.0;
     on = true;
     sel = false;
+    interp = false;
     if ( !PyArg_ParseTuple( args, "(ddii)", &x, &y, &on, &sel )) {
 	PyErr_Clear();
 	if ( !PyArg_ParseTuple( args, "(ddi)|i", &x, &y, &on, &sel )) {
 	    PyErr_Clear();
 	    if ( !PyArg_ParseTuple( args, "(dd)|ii", &x, &y, &on, &sel )) {
 		PyErr_Clear();
-		if ( !PyArg_ParseTuple( args, "dd|ii", &x, &y, &on, &sel )) {
+		if ( !PyArg_ParseTuple( args, "dd|iii", &x, &y, &on, &sel, &interp )) {
 		    PyErr_Clear();
 		    if ( PyType_IsSubtype(&PyFF_PointType, Py_TYPE(args)) ) { 
 			p = (PyFF_Point *) args;
@@ -1804,6 +1805,8 @@ static PyFF_Point *PyFFPoint_Parse(PyObject *args, bool dup, bool always) {
 	p = PyFFPoint_CNew(x,y,on,sel,NULL);
 	if ( p==NULL )
 	    return( NULL );
+	if ( interp )
+	    p->interpolated = true;
     } else if ( dup ) {
 	p = (PyFF_Point *) PyFFPoint_dup(p);
     } else {
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -5009,7 +5009,7 @@ return( NULL );
     if ( c->closed ) {
 
 	SplineMake(ss->last,ss->first,c->is_quadratic);
-	if ( c->is_quadratic && ss->last->nextcpselected || ss->first->prevcpselected ) {
+	if ( c->is_quadratic && (ss->last->nextcpselected || ss->first->prevcpselected) ) {
             // The convention for tracking selection of quadratic control
 	    // points is to use nextcpselected except at the tail of the
 	    // list, where it's prevcpselected on the first point.
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -4956,6 +4956,7 @@ return( ss );
 	    if ( !c->points[previ]->on_curve ) {
 		sp->prevcp.x = c->points[previ]->x;
 		sp->prevcp.y = c->points[previ]->y;
+		sp->prevcpselected = c->points[previ]->selected;
 		// if ( sp->prevcp.x!=sp->me.x || sp->prevcp.y!=sp->me.y ) // This is unnecessary since the other converter only makes a control point if there is supposed to be one.
 		    sp->noprevcp = false;
 	    }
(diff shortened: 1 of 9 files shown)


jtanx:
Is it possible to add a test on this?

skef:
Yeah, I should add some round-trip tests. *sigh*.

skef:
I'm going to get some rest soon and after that I'll write some tests. So that I can better stage my efforts later, if and when this is confirmed to work is it the one likely to be integrated? I want to get #3481 in shape relative to the best candidate.

jtanx:
Yeah, I'm happy to merge this once some tests are added. What are your thoughts on including this pre/post the upcoming release?

skef:
I've worked through a preliminary design for a spline manipulation library I'll probably pursue pretty soon. Before any system like that is widely adopted by FontForge users, a python "round trip" needs to preserve the more common spline stuff (at least). Not "never-interpolate", but definitely point types and selection status. (And selection is needed as a UI interface, because there's not much else available.) 

So in my view this change and the type stuff are requirements for a robust tool set, and I'd like to see them in if they can fit the schedule. However, I recognize other factors that could weigh against and I'm not going to throw a tantrum either way.

I do think it's fairly low risk in this sense: #3458 fixed several *very* basic problems, as well as a lot of less basic ones. My attempts to do anything very complex around December generally ended in frustration. "Break" implies a change in something not already broken, and doubt there is much of that out there in script-land.

skef:
As with other aspects of quadratic spline construction, a lot of the work in this change is accomplished by making `SplineRefigure2()` impose the convention (based on whether the point is selected "from either side"). 

The change to `SplineSetReverse()` is unrelated to the other changes, but does the appropriate thing. (There are probably other broken things here, like `dontinterpolate` moving around.)

fontforge/python.c
```diff
@@ -4893,11 +4899,10 @@ return( ss );
 	    } else {
 		if ( !c->points[i-1]->on_curve ) {
 		    sp = SplinePointCreate((c->points[i]->x+c->points[i-1]->x)/2,(c->points[i]->y+c->points[i-1]->y)/2);
-		    sp->selected = c->points[i]->selected;
-		    sp->name = copy(c->points[i]->name);
```
skef: These two lines were part of the convention of putting the name and selection status of interpolated points on the following off-curve point. If a user wants to set these in the modified interface they should just add an interpolated point to their contour and set them there.

fontforge/python.c
```diff
@@ -4999,7 +5007,15 @@ return( NULL );
...
     if ( c->closed ) {
+
 	SplineMake(ss->last,ss->first,c->is_quadratic);
+	if ( c->is_quadratic && (ss->last->nextcpselected || ss->first->prevcpselected) ) {
+            // The convention for tracking selection of quadratic control
+	    // points is to use nextcpselected except at the tail of the
+	    // list, where it's prevcpselected on the first point.
+	     ss->last->nextcpselected = false;
+	     ss->first->prevcpselected = true;
+	}
```
skef: The other part of the convention is handled by `RefigureSpline2()`. This code is needed because the last off-curve point is handled differently.

fontforge/python.c
```diff
@@ -1838,12 +1839,13 @@ static PyObject *PyFFPoint_pickleReducer(PyFF_Point *self, PyObject *UNUSED(args
...
     Py_INCREF(_new_point);
     PyTuple_SetItem(reductionTuple,0,_new_point);
-    argTuple = PyTuple_New(4);
+    argTuple = PyTuple_New(5);
     PyTuple_SetItem(reductionTuple,1,argTuple);
     PyTuple_SetItem(argTuple,0,Py_BuildValue("d", (double)self->x));
     PyTuple_SetItem(argTuple,1,Py_BuildValue("d", (double)self->y));
     PyTuple_SetItem(argTuple,2,Py_BuildValue("i", self->on_curve));
     PyTuple_SetItem(argTuple,3,Py_BuildValue("i", self->selected));
+    PyTuple_SetItem(argTuple,4,Py_BuildValue("i", self->interpolated));
```
jtanx: `PyFFi_newPoint` should be updated, or else this will be lost on unpickling.

skef:
Both done

fontforge/python.c
```diff
@@ -1767,27 +1767,29 @@ static PyFF_Point *PyFFPoint_CNew(double x, double y, int on_curve, int sel, cha
...
     sel = false;
+    interp = false;
     if ( !PyArg_ParseTuple( args, "(ddii)", &x, &y, &on, &sel )) {
 	PyErr_Clear();
 	if ( !PyArg_ParseTuple( args, "(ddi)|i", &x, &y, &on, &sel )) {
 	    PyErr_Clear();
 	    if ( !PyArg_ParseTuple( args, "(dd)|ii", &x, &y, &on, &sel )) {
 		PyErr_Clear();
-		if ( !PyArg_ParseTuple( args, "dd|ii", &x, &y, &on, &sel )) {
+		if ( !PyArg_ParseTuple( args, "dd|iii", &x, &y, &on, &sel, &interp )) {
```
jtanx: Not really related to this PR, but looking at this again, why are there so many combinations of this?

From the docs:
> The C variables corresponding to optional arguments should be initialized to their default value — *when an optional argument is not specified, PyArg_ParseTuple() does not touch the contents of the corresponding C variable(s)*.

jtanx: Actually perhaps that's answer by: what do the braces in the format string mean?

skef: The short answer (based on experimentation) is that optional arguments only work at the "top level", not inside sub-tuple (which is what the braces mean). 

The braces are to support things like `p = fontforge.point((4,5), True)`. That was supported "before" with the more limited argument set, so I've preserved the semantic. 

I don't think it's necessary to have a documented interface for `interpolated` on point creation, especially given that it's ignored when setting a layer. This is just an undocumented means of setting it to make the pickler happy. It will move further down the list in #3481.

skef: Maybe it would be better to only support setting `on_curve` inside the sub-tuple and leave the rest as optional outside. I suggest we leave this the way it is now and revisit the question in the type-support issue. (I'll likely change it around there and you can tell me if you want it changed back.)

jtanx: I'm happy to leave as is for now, but just something to keep in mind, wouldn't changing the order in the tuple cause incompatibility when unpickling between older/newer versions?

skef: The May 2017 interface only included setting `on_curve`, and that would remain the same. I would just be altering stuff that extrapolates from that semantic but has never released. (Basically stuff from the past month or two.)

skef:
@frank-trampe: do you want to review this or should it be considered approved?

fontforge/python.c
jtanx: If we merge the two in the same release then yeah all good. The other thing to keep in mind is that if we have to add more params in the future, then we have to be mindful of the ordering to not break when reading older pickles

PR #3520, opened 2019-03-12
Fix loading of internal python module when used from pyhook

### LINKED ISSUE #3518: Pickling broken in non-embedded contexts
All of the below applies to python 3. The code works in python2. 

Running `python tests/test932.py` on the (updated) #3513 code results in:
```
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named '__FontForge_Internals___'
Segmentation fault (core dumped)
```
This is because `PyFFPoint_pickleReducer()` tries to load the module `__FontForge_Internals___` if `_new_point` (a static) is NULL. Unfortunately in python 3 `__FontForge_Internals___` is not enough of a real module entity outside of the embedded context to load. 

This is all ultimately a side-effect of the partial fix in #2891. That fix works in general (otherwise `psMat` would still be broken), but it relies on each python 3 module having its own python-level identity. So if that is going to remain the way we do module plumbing, I'm pretty sure we need to add the files for `__FontForge_Internals___` to `pyhook`. 

@jtanx what do you think?

jtanx:
Fixes #3518

This is equivalent to doing something like

```py
import sys
sys.modules['__FontForge_Internals___'] = the_module
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -141,7 +141,7 @@ script:
   - make check
   - if [ ! -z "$LINUX_FULL" ]; then make distcheck; fi
   - fontforge -version
-  - if [ -z "$LINUX_NOX" ]; then $PYTHON -c "import fontforge; import psMat; print(fontforge.__version__, fontforge.version()); fontforge.font();"; fi
+  - if [ -z "$LINUX_NOX" ]; then $PYTHON travis-scripts/pyhook_smoketest.py; fi
   - if [ "$TRAVIS_OS_NAME" == "osx" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_BRANCH" == "master" ]; then travis-scripts/ffosxbuild.sh $PREFIX $TRAVIS_COMMIT; fi
   - if [ ! -z "$LINUX_RELEASE" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then travis-scripts/ffappimagebuild.sh; fi
 
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -18818,6 +18818,8 @@ static void RegisterAllPyModules(void) {
      * It allows an 'import some_module' to always work, as python
      * knows where to find the module in already-loaded code without
      * having to search the filesystem for module files/libraries.
+     * NOTE: This should only be called from the embedded python,
+     *       when used from the pyhook, we can't (or shouldn't) do this
      */
     fixup_module_definitions();
     for ( unsigned i=0; i<NUM_MODULES; i++ ) {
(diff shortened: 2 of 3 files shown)


skef:
This doesn't yet quite meet the spirit of `pickle`, for the reason I suspected in #3518. 

Generate a file in this way:
```
#!/usr/bin/env python

import fontforge
import pickle

# Point Pickling 

p = fontforge.point(4.4, 5.5)
p.on_curve = False
p.selected = True
fl = open('/tmp/pkl', 'bw')
pickle.dump(p, fl)
fl.close()
```
And look at the binary you find (among the junk):
```
$ cat /tmp/pkl
�c__FontForge_Internals___
newPoint
q(G@������G@KKtqRq.
```
That reference is so that the object can be unpickled ("encucumbered"?) regardless of what modules are loaded at the time. So run `load` in a context without `fontforge` imported:
```
#!/usr/bin/env python

import pickle

fl = open('/tmp/pkl', 'br')
p = pickle.load(fl)
```
and you get:
```
  File "/tmp/ss", line 6, in <module>
    p = pickle.load(fl)
ModuleNotFoundError: No module named '__FontForge_Internals___
```
We can say that everyone needs to have `fontforge` imported. But in that case it seems like we're just drawing a pointless circle. Isn't the whole reason `ff_internals` is separated out into a module so that it can load quickly under these circumstances? Why else do that? It's like we're following some example of the appropriate way to plumb unpickling but loosing the thread along the way.

jtanx:
I doubt it was for speed reasons, because whenever you import one module, all the other modules effectively get initialised. To be honest, the whole existence of that internal fontforge module is quite dubious, so perhaps you can do away with it (at the cost of breaking existing pickles).

Anyway, doing what you've suggested from outside of the embedded python would never have been possible in any of the previous iterations, since that internal fontforge module is not exposed directly via pyhook, so I'm ok with having to import the fontforge module beforehand for it to work.

skef:
My theory was more that the original developer was following some example designed for speed and didn't fully understand that. Otherwise the methods should have just been in `module_def_fontforge`. 

I'm just going to say this and then drop it: Given that adding the real plumbing for `__FontForge_Internals___` in `pyhook` would actually make everything work and have the intended performance advantage (in python 3, at least), why not just do that? Embarrassment? Kruft concerns? We're sitting on that point in the namespace anyway ...

jtanx:
What performance advantage are you thinking of? Because as mentioned before, the way the initialisation function works is that it will initialise *all* the modules, regardless of which one was imported.

I think the docstring gives a pretty good idea of why that module exists:

> I use this to get access to certain python objects I need, and to hide some internal python functions. I don't expect users ever to care about it.

Having said that, it probably should have just all been on the fontforge module.

I'm reluctant to add it to pyhook because it's an internal module, so it feels wrong to expose it like that. In any case, why would you ever not import the fontforge module but work with the pickled data? It's not like you can do much with just the internal module.

By not exposing it, we're also definitely not breaking anything that currently uses pyhook since it wouldn't have worked in the past either.

skef:
> why would you ever not import the fontforge module but work with the pickled data?
Again, I'm not especially energized about this issue. However, the same question could be raised about any serialization system. 

On the conceptual level: The reasons such systems work this way is that they allow whatever object to be embedded within or next to whatever other object. Having to already know what is inside an opaque packet of binary data in order to open it partly defeats the point. 

On the implementation level: The pickler embeds the name of a module it expects to import in order to allow the object to be unpacked. By making that module un-importable we violate the expectations of the implementation.

jtanx:
Fair enough. but in that case, I'd argue the answer there would not be to expose the internal module, it would be to get rid of the internal module and have it on the fontforge module.

If there's no other objections, I'll merge this as is, we can look at changing how that internal module works at a later date.

PR #3521, opened 2019-03-12
Update PULL_REQUEST_TEMPLATE.md

jtanx:
The PR template is too bloated, there's no need for a wall of text

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,29 +1,9 @@
-Provide a general summary of your changes in the **Title** above.
+<!-- Provide a description of the change here. -->
+<!-- See also: https://github.com/fontforge/fontforge/blob/master/CONTRIBUTING.md -->
 
-### Important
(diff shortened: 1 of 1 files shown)


JoesCat:
May be better to ask the author to simply erase everything that doesn't apply after reading.
I recall the ping-pong that led to this in the first place.

PR created...
Why are you doing this?
because...
ping...
pong...
ping...
pong...

skef:
Couldn't there be a web hook on pull_request -> opened to remove unwanted boilerplate? It might not take effect before the email, but that doesn't seem like a huge deal. It seems that mere edits don't trigger another email.

jtanx:
Hardly anyone fills out the details anyway, so it wasn't of much benefit. The other thing is that by making the descriptions as html comments, they don't show up, even if they're not explicitly deleted by the author.

I've never seen a web hook to do as suggested, seems a bit much instead of just slimming down the template

jtanx:
Any further comments or strong objections? I'll merge this in a day or so otherwise.

Issue #3510, opened 2019-03-09
Slow rendering of CharView when glyph contains "many" points (≳500)


ctrlcctrlv:
Arch Linux x64, https://github.com/fontforge/fontforge/commit/95e67e50cb1df6bae0c79c654af05372c96f848c

Reproduction steps:
1. Create a new font
1. Import an SVG with many points
1. Window will freeze for a time depending on speed of your CPU

Alternative reproduction steps:
1. Open [Issue3510.zip](https://github.com/fontforge/fontforge/files/2947993/Issue3510.zip)
1. Open "zero" charview
1. Window will freeze for a time depending on speed of your CPU

I debugged this with GDB by killing the program ten times and seeing where it was stuck. It was always stuck in cairo_stroke called by CVDrawSplineSetOutlineOnly in charview.c:1601

Here is an example full traceback with debugging symbols and compiled with `-g -O0`:

```
#0  0x000015555394d3ee in active_edges () at /usr/lib/libcairo.so.2
#1  0x000015555394d49d in intersection_sweep () at /usr/lib/libcairo.so.2
#2  0x000015555394e126 in _cairo_polygon_intersect () at /usr/lib/libcairo.so.2
#3  0x00001555538f10a3 in _cairo_clip_get_polygon () at /usr/lib/libcairo.so.2
#4  0x00001555539834ea in combine_clip_as_traps () at /usr/lib/libcairo.so.2
#5  0x00001555539842c7 in create_composite_mask () at /usr/lib/libcairo.so.2
#6  0x0000155553984484 in clip_and_composite_with_mask () at /usr/lib/libcairo.so.2
#7  0x00001555539856e6 in clip_and_composite () at /usr/lib/libcairo.so.2
#8  0x00001555539875ee in clip_and_composite_traps () at /usr/lib/libcairo.so.2
#9  0x000015555398828f in _cairo_traps_compositor_stroke () at /usr/lib/libcairo.so.2
#10 0x00001555538f9104 in _cairo_compositor_stroke () at /usr/lib/libcairo.so.2
#11 0x00001555539c19d4 in _cairo_xlib_surface_stroke () at /usr/lib/libcairo.so.2
#12 0x000015555397aec1 in _cairo_surface_stroke () at /usr/lib/libcairo.so.2
#13 0x0000155553906b6b in _cairo_gstate_stroke () at /usr/lib/libcairo.so.2
[...]

PR #3522, opened 2019-03-12
Greatly improve CharView performance, see #3510

ctrlcctrlv:
This probably shouldn't be merged, but for some reason greatly increases
the performance of the character view window. I'm mostly opening this to
try to start a discussion on why it's so slow in the first place.

"Rough" looking glyphs are quite common in fonts, so there's no reason
FontForge shouldn't be able to manipulate them.

See #3510 for how this affects the UI. I'm currently using this patch in 
working on my own fonts. CC: @monkeyiq

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -1597,14 +1597,12 @@ void CVDrawSplineSetOutlineOnly(CharView *cv, GWindow pixmap, SplinePointList *s
 		GDrawPathClose(pixmap);
 
 	    switch( strokeFillMode ) {
-            case sfm_stroke_trans:
-                GDrawPathStroke( pixmap, fc );
-                break;
             case sfm_stroke:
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
I've found this patch to greatly improve performance, this combined with other stability patches made it so I only got one crash while tracing this "M", as significant improvement.

https://twitter.com/HW_BEAT_THAT/status/1108991165527789568

jtanx:
For the most part, this is equivalent to setting the `Outline Thickness` preference to 1 (or 0).

For reasons I don't quite understand yet, when the outline thickness is more than 1, it first creates a clip from the path, then strokes it again. It's the complexity of that clip that's causing the slow performance; what you're doing in this case is creating the clip, but then doing nothing with it (and in which case, you'd be better off not creating the clip at all, by setting that preference)

PR #3524, opened 2019-03-14
Chomp comments. Minor speed increases (sequence of if tests).

JoesCat:
Added some comments around some code as help for developers.
Moved least likely tests to back/last tests to give a minor speed improvement.


JoesCat:
If you're ok with it. squash merge is fine.

PR #3527, opened 2019-03-15
Run (almost all) python tests both embedded and through python

### LINKED ISSUE #3525: Add Makefile target to run python tests via modules/pyhook instead of FontForge
I only caught the Pickling problem (after adding a test) because I tried to run it using python manually. Even if it's not part of travis (it probably should be), there should be a way of running the python tests through python. 

I'm making this an issue instead of a pull request because the whole testsuite m4 situation makes this addition non-obvious. I'm guessing generating a second file is the best way to go, but I'm not sure.

skef:
Closes #3525 

### Types of changes

### Final checklist

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -48,9 +48,13 @@ noinst_FILES = * fonts/* results/* results
 TESTSUITE = $(srcdir)/testsuite
 EXTRA_DIST = testsuite.at $(srcdir)/package.m4 $(TESTSUITE) atlocal.in
 
+TESTSUITEFLAGS =
 if !PYTHON_SCRIPTING
-TESTSUITEFLAGS = FONTFORGE_NO_PYTHON=1
+TESTSUITEFLAGS += FONTFORGE_NO_PYTHON=1
 endif !PYTHON_SCRIPTING
+if !PYTHON_EXTENSION
+TESTSUITEFLAGS += FONTFORGE_NO_PYTHON_EXTENSION=1
+endif !PYTHON_EXTENSION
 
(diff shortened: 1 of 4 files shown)


skef:
Lord knows what this will do on Windows.

skef:
I don't have a windows environment handy. I presume the way to work through this is to get to the point after a "make", open a shell, and play around with running a particular test using variants of the command. 

The most recent variant has:
```
m4_defun([run_python_py],[(PYTHONPATH="${abs_top_builddir}/pyhook/.libs:$PYTHONPATH" ${python_exec} "${abs_srcdir}/$1" "$2" || /bin/false)])
```
Some earlier variants passed `-Ss` to python. The primary goal is to make the modules in `pyhook/.libs` have precedence. The secondary goal is to make it less likely that some other set of modules will be picked up if those are missing. 

The latter is *probably* only of minor importance. I've tested with all four combinations of `--disable-python-scripting` and `--disable-python-extension`, and the test suite doesn't try to run tests it hasn't been built for. Therefore any problem that prevents populating `pyhook/.libs` with modules *should* be caught earlier in the build process. Anyway, it's fine to preserve the existing `PYTHONPATH` but the build directory should come first. `-Ss` was kind of a stab in the dark, but marginally contributes to the secondary goal.

jtanx:
The problem was dll resolution, fixed by altering PATH. We should just statically link all our own code and get rid of the dlls...

skef:
Looks good to me! 

I have no idea how much this will help, but by only running "Various stroke tests" in one mode it doesn't add much time to `make check`. I only caught the pickling problem by trying to run that test by hand and this seems like better practice.

PR #3529, opened 2019-03-15
Add ./configure --enable-fontforge-desktop=DIR

JoesCat:
Add method to disable making desktop to avoid overwriting an existing
setup, or provide a path to another desktop, such as on a virtual PC.

Default is to create freedesktop.org desktop in default @datarootdir@


JoesCat:
copy-n-paste from earlier code, works as expected for no programs, but did not check with gui yet.
...needs further checking...
...almost Saturday... leave as for another day?

khaledhosny:
What is the point of having an option? either one is building GUI and gets .desktop files or one is not, adding options just for the sake of it does not make any sense.

JoesCat:
99% of us don't care about this option because the default is good enough for most of us, however...
for the 1% that have customizations in mind, this option exposes this variable:
```code
PACKAGEROOTDIR = ${datarootdir}
```
hidden away in a file: [mk/layout.am](https://github.com/fontforge/fontforge/blob/20170731/mk/layout.am) where very few of us would think of looking in, or even know about.
If you can answer a problem before it's asked, that's good (we avoid adding to the extra 1007+ issues).

Also, up to now, FontForge has never had an automated desktop install and has relied on distro maintainers to do this task themselves. If you read some of the issues, there are requests for help in this area. The best we've "had up to now" was [dist files](https://github.com/fontforge/fontforge/blob/20170731/desktop/Makefile.am) which they then had to figure-out themselves.

Although we all are reasonably expecting everything to go smoothly without problems for this new release, we should also allow for the possibility of problems and corner-cases, so this option allows users to turn this feature off. There are some issues on the web about putting the desktop elsewhere too, so let's take care of this now while we're all looking at this new feature now/anyways.

Right now is also the best time to "go wild" about the desktop directory because we also have @probonopd (interested in appdata), @muelli (interested in flatpak), and we've been looking through a bunch of different desktop installs for problems, features, issues, etc. If there's going to be a problem, hopefully they spot it now while we're busy here now. ...in summary, better to deal with this topic now while it's hot and of interest, vs cold and not the topic du Jour.

khaledhosny:
Customization for sake of customization is also pointless. Unless someone can come up with a real world case of needing (not just wanting) to build the GUI without the desktop files, and don’t see why we need to complicate the already unnecessarily complex build system. Extra build option is an extra axis in the build matrix that can potentially conflict with other options, lead to nonworking combinations and otherwise bit-rot. If anything, we should drop some of the current build options not add more.

JoesCat:
If I knew you were going to ask for real world situations, I would have kept more notes. LOL :-)
...let's see....time for yet another R&D dive....
https://gitlab.gnome.org/haarek/gimp/blob/master/configure.ac
https://gitlab.gnome.org/GNOME/evince/blob/master/configure.ac
... is 2 enough? or do you need more examples?

...unfortunately, your reference to bitrot can be avoided if more attention is given to the code.
Real world examples, FontForge readline was added, and it doesn't bother anybody. Several other options are based on this initial PR and seem to be functioning fine. Much of the gutils has had a good look-through for potential flaws, and in terms of issues, it is relatively quiet. I do recall argument about "nobody does bitmaps" anymore, but there are at least two issues referencing drag-n-drop, so that does show there are users still using bitmaps. The Unicode directory is seeing some settling down in reference to issues, where we've had some issues that caused security call warnings like tolower() buffer overflows. The cure to bitrot is not simply cutting-out code because it's not functioning the way you expected (example sortsmill cutting out code and then having a bunch of orphaned functions that serve no purpose afterwards), but to look at the code, figure-out the problem(s), fix it/them! ...and not simply fix the one issue, but if you see others, take care of them as you see them too, regardless of if there is an issue associated with the potential problem or not.

There is no Issue related to this PR, but there are real world examples where it seemed worth making this effort to add this option.

khaledhosny:
Did you actually read the configure.ac files you are citing? Neither of them is doing anything even remotely related to what you want to do here. But that is my last comment on this issue, I have no write access and I don’t accept or reject PRs.

frank-trampe:
@khaledhosny, your absence from the developer group was an oversight on my part. Sorry about that. I just sent an invitation.

jtanx:
I'm inclined to agree with the point that I don't see much value in this, and I'd rather not add yet another option to an already complicated build system.

JoesCat:
@khaledhosny, please check again. This is a realworld example:
https://gitlab.gnome.org/haarek/gimp/blob/master/configure.ac
```code
##########################################################
# Determine where to install the desktop & mime info files
##########################################################

AC_ARG_WITH(desktop-dir, [  --with-desktop-dir=PATH where to install desktop files (default=PREFIX/share)])

if test "x$with_desktop_dir" != x; then
  DESKTOP_DATADIR=$with_desktop_dir
else
  DESKTOP_DATADIR="$datadir"
fi

AM_CONDITIONAL(DESKTOP_DATADIR, test "x$with_desktop_dir" != xno)
AC_SUBST(DESKTOP_DATADIR)
```
You'll see more in the Makefile.am

and here is the other example:
```code
AC_ARG_ENABLE([libgnome-desktop],
        [AS_HELP_STRING([--disable-libgnome-desktop], [Disable GNOME Desktop (Thumbnail cache)])],
        [enable_gnome_desktop=$enableval],
        [enable_gnome_desktop=auto])
```

BTW = Happy Birthday.

@jtanx - will go with that. Good enough. That's why it was never pulled in earlier.

khaledhosny:
> @khaledhosny, your absence from the developer group was an oversight on my part. Sorry about that. I just sent an invitation.

That is no ones fault, I had right access then decided I don’t need or want it and removed myself. I haven’t changed my mind about it, but thanks for the invitation anyway.

khaledhosny:
> @khaledhosny, please check again. This is a realworld example:
> https://gitlab.gnome.org/haarek/gimp/blob/master/configure.ac
> 
> ```
> ##########################################################
> # Determine where to install the desktop & mime info files
> ##########################################################
> 
> AC_ARG_WITH(desktop-dir, [  --with-desktop-dir=PATH where to install desktop files (default=PREFIX/share)])
> 
> if test "x$with_desktop_dir" != x; then
>   DESKTOP_DATADIR=$with_desktop_dir
> else
>   DESKTOP_DATADIR="$datadir"
> fi
> 
> AM_CONDITIONAL(DESKTOP_DATADIR, test "x$with_desktop_dir" != xno)
> AC_SUBST(DESKTOP_DATADIR)
> ```

This is setting where to install the files not whether to install them or not.

> You'll see more in the Makefile.am
> 
> and here is the other example:
> 
> ```
> AC_ARG_ENABLE([libgnome-desktop],
>         [AS_HELP_STRING([--disable-libgnome-desktop], [Disable GNOME Desktop (Thumbnail cache)])],
>         [enable_gnome_desktop=$enableval],
>         [enable_gnome_desktop=auto])
> ```

“Disable GNOME Desktop (***Thumbnail cache***)”.

JoesCat:
> This is setting where to install the files not whether to install them or not.

There is a bug in that code - if you look carefully.
AC_ARG_WITH and AC_ARG_WITHOUT calls the same code.
...but someone can call --without-desktop-dir
...which seems to make sense to a user, but there is no code to deal with it.

Similarly, we have --enable-desktop, and --disable-desktop
...which makes sense to add the necessary code to deal with 'disable'
I was mostly interested in the DIR=PACKAGEROOTDIR, but you should also take into consideration yes, no branches.

ctrlcctrlv:
I hate to close @JoesCat's work, but as both Khaled and Jeremy see no point in this option, and for the record I don't either, especially with a new build system coming&hellip;I'm closing this on the basis that it was rejected by the membership.

PR #3530, opened 2019-03-15
Synchronise old docs with fontforge.github.io

jtanx:
- Fixup all docs links (where available)
- Reduce the number of false positives when scanning for missing assets
- Standardise body ids
- LF Line endings
- Encode all docs in utf-8
- Add syncing script to sync docs to fontforge.github.io

See also
https://github.com/fontforge/fontforge.github.io/pull/124


skef:
My initial reaction to this push was strongly negative, but I've decided I was mentally overstating it's impact. So let me confirm that thinking. After the release was finalized I was planning to:

1. diff the docs directory using the two release tags,
2. make a list of all the docs changes between the two releases
3. work those changes back into the website docs. 

At first glance these changes struck me as making that procedure hopeless/useless. But looking more closely most of them are to files in the language subdirectories, and maybe those are less of an issue. 

However, either way it seems that I could instead do two diffs, one from the 2017 release to before these commits, and another from just after these commits to the 2019 release. That should isolate all the doc updates merged in that time.

Does that sound right?

jtanx:
The bulk of the diff is just re-encoding the Japanese/German docs into utf-8; the Japanese docs on the website were already utf-8 encoded, so that was bringing it in line with that. There aren't very many changes to the English docs, just one sync back from changes made on the website to macformats.html.

When you say:
> work those changes back into the website docs.

Do you mean working them into the *new* docs, or the *old* docs? If it's the latter, then this already does that, if it's the former, then perhaps it's time to discuss what we're going to do about this long term.

This PR was really only about keeping the old docs between this repo and the website in line.

skef:
> Do you mean working them into the new docs, or the old docs? If it's the latter, then this already does that, if it's the former, then perhaps it's time to discuss what we're going to do about this long term.

I mean updating the markdown format documents in the website repository to include any changes made since 2017 to the html documents in the source code repository.

skef:
> This PR was really only about keeping the old docs between this repo and the website in line.

OK, but there are (at least) two separate issues: One is what to do going forward, and one is how to round up the 2017-2019 changes made when there was no system in place. I'm currently fretting over the second.

jtanx:
A `git diff 20170731...HEAD  *.html` works pretty well in the docs folder

skef:
Pretty well, but less well with changes like the one under review.

jtanx:
Alright, how about this, I'll hold off on this for now, we cut the release, you then update the markdown, then we work out what to do with the docs long term. 

Let's consolidate to writing the docs once, say in markdown, because it's untenable to continue updating the old docs, while then also trying to sync it back to the website markdown

skef:
That sounds good. A "last-minute" synchronized switch would also work. After doing my doc update on the types code I'll take a look and start rounding up the changes. (That work is needed regardless of what path we take anyway.) 

One tricky question long-term question is whether the website should have a) only the docs for the current version, b) separate docs trees for separate versions, c) an integrated doc tree that mentions changes made in particular releases. The answer might change what makes the most sense to do with the docs in the code repo.

skef:
Update on the more general project: I've now fixed up the differences between the code repo HTML docs and the website repo HTML docs. It was a big pain. I have reason to believe that there is a lot of incompleteness, so the next step is to do a pass over the python and native scripting docs to fill in the gaps. Then I'll port the changes over to the Markdown docs. 

Unfortunately a lot of the Markdown docs look a bit older than the HTML docs in the repo so I'm worried information will be lost anyway. It's a big mess.

ctrlcctrlv:
Since we've been accepting commits that just change the old HTML files up to now, we should also put in CONTRIBUTING.md how developers should edit the documentation.

skef:
The formatting of the python documentation pretty much requires a more sophisticated markup language than Kramdown. Kramdown has tables but no effective way to include much structure inside them. 

I'm going to make the website version of the python document html again, and I guess any argument about that can go here or in the eventual pull request.

ctrlcctrlv:
(in re: Python docs)
How about LaTeX? :-D It seems fitting as this is a font software to have its documentation well typeset, and we can use `htlatex` or `tex4ht` to get HTML back. I use XeLaTeX to generate my font's documentation.

jtanx:
> How about LaTeX?

I'd have to say to that, yeah no :)

There's probably some extensions that we can add to enrich the markdown, depending on how what exactly is too hard to do otherwise

skef:
For tables within tables? Or description lists within tables? 

One you get past a certain level of complexity the name of the extension is "embedded HTML".

skef:
Let me add: I don't see the issue as whether it's possible to do what is needed in some variant of Markdown. I see it as whether the variant in question is really Markdown in *spirit*. If it's not straightforwardly legible in the source it's not functioning like Markdown is supposed to. 

I've been doing my personal authoring in Pandoc Markdown for a few years now, usually with a LaTeX back-end. It's great for relatively normal writing needs. It's kind of wonky for normal tables because of the compromises you generally have to make in the source (line wraps become an issue). More complicated stuff becomes impractical.

ctrlcctrlv:
I would convert the Python page to LaTeX if we'd actually use it.

skef:
With LaTeX the only reasonable options for screen use would be PDF-like or back-translation to HTML via something like pandoc. And it's only a mild improvement over Kramdown markup-wise. 

I think there may be some mental overstatement about my proposed route. I'm going to set `parse_block_html` (if it isn't set already) and organize the python member and method documentation into tables. I'll continue to use Kramdown span markup for italic and code and such. The overall file organization -- everything outside of the tables -- will still be in Kramdown. 

Maybe that's not ideal, but the file is almost unusable as it was left, and this gets it back into shape while still fitting into the overall system.

ctrlcctrlv:
Uh, no, htlatex for translation to HTML is far more powerful than pandoc as it's an actual TeX engine.

ctrlcctrlv:
But if you have your own plan I don't object to it. :-) My offer remains open though.

skef:
I think the main thing is that I'm not suggesting much change, and don't see a need for much change. I want to do a different thing *only in those documents that need it* and only in the *parts* that need it. I think the current CSS for the website is questionable, but that has nothing to do with the markup language. 

I rewrote the `psMat` section using a Kramdown table and it worked fine, although it looks pretty crappy in the file, which isn't really in the Markdown spirit. But what about "hooks" in [fontforge](http://dmtr.org/ff.php#fontforge)? Or the many key-value options lists? The [current state](http://fontforge.github.io/en-US/documentation/scripting/python/) of that stuff is borderline unusable.

khaledhosny:
Or rewrite the Python API documentation to not use a table, it is a very terrible layout to begin width.

Something like:
```
# psMat

## Methods

###  `identity ()`
Returns an identity matrix as a 6 element tuple.

### `compose (mat1, mat2)`
Returns a matrix which is the composition of the two input transformations.

### `inverse  (mat)` 
Returns a matrix which is the inverse of the input transformation.
(Note: There will not always be an inverse).

### `rotate  (theta)` 
Returns a matrix which will rotate by theta. Theta is expressed in radians.

### `scale (x[, y])`
Returns a matrix which will scale by `x` in the horizontal direction
and`y` in the vertical direction. If y is omitted, it will scale by the
same amount (`x`) in both directions.

### `skew (theta)` 
Returns a matrix which will skew by theta (to produce a oblique font).
Theta is expressed in radians.

### `translate (x,y)`
Returns a matrix which will translate by `x` in the horizontal direction
and `y` in the vertical direction.

# Types
None
```
# psMat

## Methods

###  `identity ()`
Returns an identity matrix as a 6 element tuple.

### `compose (mat1, mat2)`
Returns a matrix which is the composition of the two input transformations.

### `inverse  (mat)` 
Returns a matrix which is the inverse of the input transformation.
(Note: There will not always be an inverse).

### `rotate  (theta)` 
Returns a matrix which will rotate by theta. Theta is expressed in radians.

### `scale (x[, y])`
Returns a matrix which will scale by `x` in the horizontal direction
and`y` in the vertical direction. If y is omitted, it will scale by the
same amount (`x`) in both directions.

### `skew (theta)` 
Returns a matrix which will skew by theta (to produce a oblique font).
Theta is expressed in radians.

### `translate (x,y)`
Returns a matrix which will translate by `x` in the horizontal direction
and `y` in the vertical direction.

# Types
None

skef:
The time to have this discussion was five years ago, not when a volunteer is 8-10 hours into an update project. 

Going to a sub-section model is viable but will turn the higher-level structure into mush. Any sense of whether one is in the `font` or `glyph` section will only be evident by hunting for a sliver of slightly larger type. To do that switch properly you would need to switch the top-level sectioning to a separate file model.

skef:
That suggestion isn't all that different from what's there now, just with larger headers, and the file is almost unusable: http://fontforge.github.io/en-US/documentation/scripting/python/ .

skef:
Well, fine, screw-it. I guess the docs matter less than I thought they did.  I'll just do a push request that updates the HTML and the markdown on the website but skips over the python markdown.

khaledhosny:
I’m sorry you feel that way about a simple and well meaning suggestion. That is the last you hear from me.

skef:
It's the combination of the long-term neglect and the indecision that's bothering me.

1. Someone did this translation many years ago now with no plan for how to move the documentation forward.
2. The markup language chosen was inadequate and not designed for complex tasks (as is evident all over the Kramdown issues tracker). 
3. Sitting at the top of this document for all that time is a note pointing to http://dmtr.org/ff.php#Font *as a reference* for how things should wind up. 

So I've been trying to use the tool chosen to make the document look pretty much the reference example, which is possible but involves a largely illegible and more difficult to maintain source file. How much does that matter? It's hard to say. After many hours I'm down to `font` Methods, and this is before integrating the actual changes and the link targets that were just stripped out of the file: [py.zip](https://github.com/fontforge/fontforge/files/2998373/py.zip)

I understand what you're saying, but I'm sure you understand how it feels to spend significant effort trying to implement something *labeled as a reference* that someone then comes along and calls "*very terrible*". And this is after *five years* of things just sitting that way. How is someone supposed to help with this area of the project? Are contributors just supposed to bang their heads against the wall until the blood eventually spatters in an aesthetically pleasing pattern? 

Either *have an up-front decision-making mechanism* or *accept incremental changes that don't make things worse*.

jtanx:
Just my 2c.

From the git commit history, it's pretty telling that the migration to the new layout is unfinished (and somewhat abandoned), so I would take anything that says 'use this as a guide' with a large grain of salt.

What you're doing in terms of trying to bring the docs up to spec is great, since as you've rightly noted, it's been sorely neglected.

As you've spent the effort to try to update what's already there, I'd happily accept whatever you did, which is almost certainly an improvement over what's there now.

But if you're finding that's too hard, and you want to take on the challenge, there's nothing tying us to using Jekyll/Kramdown for docs. Gut it, use [sphinx](http://www.sphinx-doc.org/en/master/), rst, latex, however readthedocs works, or whatever you fancy, and I'm sure we can find a way to integrate it. Of late, you've spent the most time in this area, so imo, you have the most say on what to do here.

This project does unfortunately suffer from a lot of indecision, and at times a fear of changing things because it *might* break something, but (I'd hope) changing up how we do documentation should not be a contentious issue.

skef:
Aside question: Why isn't HTML on that list? 

The glaring problem blocking my taking on any significant tool change is the top-level organization problem. The HTML documents in the fontforge repo don't have a great organization, but its better than the current website. The top-level "Documentation" page is weird: https://fontforge.github.io/en-US/documentation/ . The "Design with Fontforge" hierarchy isn't bad, but the section of the translated old docs that tell you how to use the program is labeled "User Interface", and the top level is a list: http://fontforge.github.io/en-US/documentation/interface/ 

It seems like whoever did the website was going down a path of *rewriting the documentation* and then petered out after a surface-level reorganization. An actual rewrite is beyond the scope of my present interests. The existing documentation is written to fit into the `overview.html` model, which has a "through-line". Unless the answer is "put that organization back, with its own URL hierarchy on the website", I'm not sure what to do.

skef:
The easiest path to something like comprehensive documentation on the website -- *by a long shot* -- is to update and normalize the existing HTML and present it pretty much like it is now, with overview.html as the "guide" page, and with tabs similar to those at the top, except probably with (at least) two levels instead of one.

frank-trampe:
The bad organization does not bother me terribly, and I don't know that it's a good use of anybody's time to fix it. Certainly, single-streaming the documentation is a great starting point. If @skef is interested in mucking through all this, I'm perfectly happy to take the path that he has proposed.

skef:
Well, wait. My baseline proposal is:

1. Shove all the Kramdown files corresponding to repo documents into an unused sub-folder (or just delete them).
2. Copy the HTML repo hierarchy as-is into the website, including the images. 
3. Invite people to improve the docs in the repo, while keeping them self-contained.

The website also has copies of all the HTML documents, but with a different heading and images in different places. Those changes make it a huge pain to integrate changes made in the repo versions, because you have to sort through all the changed image links when diffing.

skef:
The general problem is that there are currently three copies of all the documents, and no one is updating two of them as updates are made to the repo copies. I went through and did this (no push request yet), but got hung up on `python/index.md` because it's basically broken -- both difficult to read and almost all of the link targets are missing. 

All I want is a way for people to find out about the improvements to the spline interface. What am I, as a code contributor, supposed to do in this case? How are things supposed to work?

jtanx:
When you say copy the HTML in, do you mean the HTML generated by Jekyll from the markdown, or the old HTML files back in?

skef:
I mean the old HTML files.

skef:
Right now, if you download the repo and point your browser at the `doc/html` directory, you get a documentation experience that is:

1. More hokey,
2. Up to date
3. Vastly better organized and more informative.

jtanx:
The website already hosts all the old documentation too (i.e. the point of this pr). Are you saying to just remove the markdown, so there's just one source?

skef:
The website hosts a *version* of the old documentation with images moved to different locations (making re-synchronizing difficult) and no pointer that I could find from the website. Removing the markdown would be a good first step. Replacing the modified files with the repo versions, or repo versions with *only* a modified header, and not modified image links, would be better.

skef:
I did do the synchronizing already, but it involved sorting through all the modified image links and porting the changes over by hand. It was a big pain; it shouldn't be necessary.

jtanx:
There's a synchronisation script in this PR that handles rewriting the links. Personally, I think it's for a good reason, having all the assets in one folder is better practice than having them interspersed.

skef:
> There's a synchronisation script in this PR that handles rewriting the links. Personally, I think it's for a good reason, having all the assets in one folder is better practice than having them interspersed.

Isn't that for the other direction?

skef:
I mean, this PR is moving docs from the website to the code repo, which means the answer to "which doc set is the master copy" is "both", right?

jtanx:
No. That script's purpose is to copy whatever's in the `doc` folder here, and to dump it into the website repo, rewriting links into the assets folder as necessary. It also adds that banner in, fwiw.

The only thing I synchronised back into this repo was one html file, and that was by hand.

(https://github.com/fontforge/fontforge/pull/3530/commits/d4c492ca31849268d077b26e7e1d1ebbf1f1c4ef)

Anyway, I'm fine with removing the markdown versions for now.

My opinion, but the reason I didn't suggest HTML earlier is because (assuming we were starting from scratch), it's generally easier to write/grok it in markdown/rst. It also allows you to change styling way easier (i.e. what Jekyll's currently doing) - if we're going down this path as is, the styling is pretty glaringly different between the main page and the docs. 

Anyway, that's a problem for another time, there's tools that can be used to transpose into the format of choice (probably with some effort/manual fixups).

skef:
In that case, I will focus my efforts in this direction:

1. Push the markdown (of the documents in the code repo, not any other documents) aside for now.
2. Work on updating the styling of the repo documents to better match the modern styling of the website.
3. Ensure that your script can do the whole job of synchronization. 
4. Alter the top-level organization of the website to best point to `overview.html`.

jtanx:
Sounds good to me. Feel free to push into this PR whatever you do (or create your own PR, whichever is easier).

BTW the only reason this PR is so huge is because I converted all the docs to UTF-8 encoding (affecting the Japanese/German docs). The website already has the Japanese docs UTF-8 encoded, so all the stuff here was to just minimise the diff when running the sync script on the website repo.

skef:
I've already done the work to track down the relevant changes, which if we're pushing aside the markdown isn't necessarily needed anyway. So why don't you merge this and I'll work from there.

jtanx:
Sure. I'll merge fontforge/fontforge.github.io#124 too.

davelab6:
Very happy to see all this activity!

PR #3532, opened 2019-03-16
Fix ttf_flag_nomacnames flag

khaledhosny:
It conflicted with ttf_flag_symbol (probably a bad merge on my part), which meant it could never be disabled from the GUI).


skef:
The value of everything from `ttf_flag_pfed_lookupnames` is changed -- are you *sure* it's OK to remap these?

khaledhosny:
Good point (and I missed it, which is telling), may be we should leave the gap as it was just in case (these flags end up saved in the perfs file).

skef:
That seems better. I *think* these are all "internal" (as opposed to the `fm_flag` group), but the existing gap is fine and safer than doing a full inspection.

PR #3535, opened 2019-03-16
Update version to 20190316

jtanx:


frank-trampe:
@jtanx, @skef, could you perhaps rattle out lists of the changes you've made in the last year from your perspective? I can write up the stuff from other contributors. I'll then make a grand changelist (for use on the tag) and a nice announcement for the mailing list once we have packages ready.

jtanx:
The two biggest ones from me are:
* GDK backend support, which I'll use for the Windows and Mac releases
* Basic woff2 support (via the woff2 library)

skef:
Most of my work has been individual quality fixes that don't merit itemizing. I think it would be fair to credit "Improved feature file support" to myself and @khaledhosny. There's also "Python fixes (especially 3.x) + spline interface improvements".

JoesCat:
We need to do a bunch of major bumps.
I just did a diff of 2017 tagpoint vs HEAD, and here's a summary of what I see:
```code
m4_define([fontforge_major_version], [3]) dnl 20120731-b was 1.0
m4_define([fontforge_minor_version], [0])
m4_define([libgdraw_major_version], [6]) dnl 20120731-b was 4.10
m4_define([libgdraw_minor_version], [0])
m4_define([libgioftp_major_version], [2]) dnl 20120731-b was 1.0
m4_define([libgioftp_minor_version], [1])
m4_define([libgunicode_major_version], [5]) dnl 20120731-b was 3.2
m4_define([libgunicode_minor_version], [0])
m4_define([libgutils_major_version], [3]) dnl 20120731-b was 1.3
m4_define([libgutils_minor_version], [1])
m4_define([libzmqcollab_major_version], [2]) dnl n/a
m4_define([libzmqcollab_minor_version], [1])
```
Reasons:
* libgraw requires to upgrade from 5.0 to 6.0
 - because the functionality of several functions in inc/gdraw.h have changed:
example (diff inc/gdraw.h ../f/inc/gdraw.h)
```code
428a425
> extern int GDrawKeyState(GWindow w, int keysym);
587d583
< extern int GDrawKeyState(int keysym);
```
* libgutils requires to upgrade from 2.0 to 3.0
example, we removed stuff (diff inc/gutils.h ../f/inc/gutils.h):
```code
30c30,31
< extern int16 div_tables[257][2];
```
* libgunicode requires to upgrade from 4.0 to 5.0
example, we changed #definitions like tolower into actual functions tolower()

Doing a diff of collab vs collab, it's ok to say from 2.0 to 2.1 since only one file changed for:
```code
55c55
< #include <glib.h>
---
> #include <ffglib.h>
```

* libgioftp did not have much activity, and nothing flags to say major bump, we can be ok with 2.1

* fontforge_major_version requires a major bump from 2.0 to 3.0 because the lower libraries have one or more that did a major (non-backwards compatible) bump-up

There's a couple of PR that look worth doing (before the final tag point done), but I'll be hands-off and leave this to you to take care of (as there's going to be a lot of churn right now) - good luck!

jtanx:
I don't think it really matters too much, given that we're the only thing that actually uses those dlls; we haven't bumped them in a very long time, and I'm sure there have been breaking changes in those.

JoesCat:
It probably doesn't matters much for dlls, but we've had a couple of past issues with (linux) users that had already installed from distro, and then tried to install these newer versions.
If we can avoid issues with a little prevention - it's good.
Thanks for bumping the version numbers.

PR #3537, opened 2019-03-16
GDK: Don't set last_nontransient_window if the window is a transient

### LINKED ISSUE #3536: GDK: Further modal issues
Annoying issue:
* Open a font, or make a font that causes validation issues (e.g. 
[Inconsolatazi4-Regular.zip](https://github.com/fontforge/fontforge/files/2973763/Inconsolatazi4-Regular.zip))
* Attempt to generate the font, ensure that the validation option is selected
* Validation warning shows, but you can't interact with it, so we're stuck.

This is a side-effect of #3506, which imposes stricter requirements on input restriction. The validation window is not marked as a transient, but it *does* have the restrict_input_to_me flag set. The other compounding problem is that `last_nontransient_window` is NULL, so `GGDKDrawSetTransientFor` never gets called on it, so we can never interact with it, while we can with the generate font dialogue - but we can't close it (for some other reason, probably waiting for the validation window to close first).

jtanx:
In preference to the last nontransient window.

This is the only thing that I can see that makes sense in this situation.

Fixes #3536

### DIFF
--- a/gdraw/ggdkdraw.c
+++ b/gdraw/ggdkdraw.c
@@ -592,8 +592,13 @@ static GWindow _GGDKDraw_CreateWindow(GGDKDisplay *gdisp, GGDKWindow gw, GRect *
             nw->is_dlg = true;
         } else if (!nw->is_dlg) {
             ++gdisp->top_window_count;
-        } else if (nw->restrict_input_to_me && gdisp->last_nontransient_window != NULL) {
-            GGDKDrawSetTransientFor((GWindow)nw, (GWindow) - 1);
+        } else if (nw->restrict_input_to_me) {
+            if (gdisp->last_nontransient_window != NULL) {
+                GGDKDrawSetTransientFor((GWindow)nw, (GWindow) - 1);
+            } else {
+                Log(LOGWARN, "Could not call GGDKDrawSetTransientFor on a restricted window (%p %s), restrict count %d, transient window count %d",
+                    nw, nw->window_title, gdisp->restrict_count, (int)gdisp->transient_stack->len);
+            }
         }
         nw->isverytransient = (wattrs->mask & wam_verytransient) ? 1 : 0;
         nw->is_toplevel = true;
@@ -882,7 +887,7 @@ static void _GGDKDraw_DispatchEvent(GdkEvent *event, gpointer data) {
     if (event->type == GDK_KEY_PRESS || event->type == GDK_BUTTON_PRESS || event->type == GDK_BUTTON_RELEASE) {
         if (gw->transient_owner != NULL && gw->isverytransient && !gw->transient_owner->is_dying) {
             gdisp->last_nontransient_window = gw->transient_owner;
-        } else {
+        } else if (!gw->istransient && !gw->is_dying) {
             gdisp->last_nontransient_window = gw;
         }
     }
--- a/gdraw/ggdkdraw.c
+++ b/gdraw/ggdkdraw.c
@@ -595,7 +595,7 @@ static GWindow _GGDKDraw_CreateWindow(GGDKDisplay *gdisp, GGDKWindow gw, GRect *
         } else if (nw->restrict_input_to_me) {
             if (gdisp->last_nontransient_window != NULL) {
                 GGDKDrawSetTransientFor((GWindow)nw, (GWindow) - 1);
-            } else {
+            } else if (gdisp->top_window_count > 0) {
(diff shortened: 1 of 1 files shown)


gdraw/ggdkdraw.c
```diff
@@ -592,7 +592,7 @@ static GWindow _GGDKDraw_CreateWindow(GGDKDisplay *gdisp, GGDKWindow gw, GRect *
             nw->is_dlg = true;
         } else if (!nw->is_dlg) {
             ++gdisp->top_window_count;
-        } else if (nw->restrict_input_to_me && gdisp->last_nontransient_window != NULL) {
+        } else if (nw->restrict_input_to_me && (gdisp->last_nontransient_window != NULL || gdisp->restrict_count > 0)) {
```
skef: I don't doubt the report that this fixes the associated issue. However, from the standpoint of "inspection" I'm concerned about the generality of the fix. We now apparently know there are windows that set `restrict_input_to_me` and aren't transient. What if one such window is created at a time when `last_nontransient_window` happens to be NULL (for whatever other strange reason)? In that case we'll have "grabbing" window 1 that doesn't wind up on the transient_stack. Then if another `restrict_input_to_me` window opens under these circumstances this guard could evaluate to false (because `last_nontransient_window` could still be null, and we wouldn't have incremented `restrict_count`. 

Doesn't it seem safter to have this guard just be `} else if (nw->restrict_input_to_me) { ` and then have `GGDKDrawSetTransientFor` get it into the `transient_stack` and adjust `restrict_count` in whatever way it needs to to avoid problems? If some of those windows aren't strictly transients, we could always rename the stack ...

skef: Alternative: Since `GGDKDrawSetTransientFor()` is a gdraw function pointer method, we could factor out the stack manipulation code into a separate static function and call that here and from `GGDKDrawSetTransientFor`. That would allow the naming to better match whatever we call this transient-or-restricted status. (I don't care all that much about names matching up in these complex legacy situations, but I won't discount the concern.)

jtanx:
> What if one such window is created at a time when last_nontransient_window happens to be NULL (for whatever other strange reason)?

So having `last_nontransient_window` being NULL while there are other windows around is a bug in of itself. 

Note that a restricted input window *implies* that it is a transient window, by virtue of GGDKDrawSetTransientFor setting istransient, and that we always call that function on restricted input windows.

However, I still think that the previous soln was also valid, it doesn't make sense to have multiple windows at the same time be marked as restricted input and not have them being as transients to each other (in the order they were created).

skef:
Note: merging this will bump the release number. Not sure how this relates to #3535.

jtanx:
This latest version doesn't work on windows

skef:
Can you explain the latest change to me? I thought the idea was that the `restrict_input_to_me && last_transient_window == NULL` combination was supposed to be rare. There's something I (as a reviewer) is still not getting about this stuff.

jtanx:
```
09:43:23.623 WARNING: _GGDKDraw_CreateWindow (ggdkdraw.c:600) - Could not call GGDKDrawSetTransientFor on a restricted window (02f64ee0 Error), restrict count 0, transient window count 0
09:43:23.926 WARNING: _GGDKDraw_CreateWindow (ggdkdraw.c:600) - Could not call GGDKDrawSetTransientFor on a restricted window (02fc8bc0 Open Font), restrict count 0, transient window count 0
```

What's happening is that on startup, FontForge preallocates the error window (but never shows it). It also creates that open font dialogue, marked as restricted input, but obviously there's nothing else around to make it transient to, as it's the only window. So basically this case is valid if there's no other windows around. Once the transient stack has *something* in it, that implies that there is at least one other nontransient window present, which means that you would always expect to be able to make any new restricted input windows transient at least to that.

skef:
Again, checking my understanding: The `transient_stack` is *just* for transients, not "base" windows, right? So this new guard will cut out a portion of genuine warnings, and is in effect kludging around a startup condition?

jtanx:
It's unlikely that it will filter out  genuine warnings, but I've found a better check, against the top window count.

skef:
> It's unlikely that it will filter out genuine warnings, but I've found a better check, against the top window count.

I'm less confident about what state combinations are ruled out than you seem to be, but the new guard addresses those concerns handily. Looks good to me.

skef:
Would it be easy to remove commit aa447e6517c844df77828669522561d0b39c9a8e from this, for clarity's sake?

jtanx:
done

Issue #3343, opened 2018-10-04
[Mac] Crash when scrolling fast


MaddTheSane:
Whenever I scroll quickly in FontForge, it crashes. It does not matter if I use the trackpad or the scroll bars. And yes, the bounds error happens with the same index and bounds.

````
2018-10-04 11:38:28.847 fontforge[77733:12035210] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSArrayI objectAtIndex:]: index 34 beyond bounds [0 .. 33]'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff495da43d __exceptionPreprocess + 256
	1   libobjc.A.dylib                     0x00007fff754e7720 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff49619a38 -[CFPrefsConfigurationFileSource initWithConfigurationPropertyList:containingPreferences:] + 0
	3   CoreFoundation                      0x00007fff4952edf9 -[__NSSingleObjectSetI member:] + 0
	4   libpangocairo-1.0.0.dylib           0x000000010ca8444b pango_core_text_fontset_get_font_at + 62
	5   libpangocairo-1.0.0.dylib           0x000000010ca843e3 pango_core_text_fontset_foreach + 43
	6   libpango-1.0.0.dylib                0x000000010cac4148 get_shaper_and_font + 208
	7   libpango-1.0.0.dylib                0x000000010cac3486 itemize_state_process_run + 963
	8   libpango-1.0.0.dylib                0x000000010cac2d3f pango_itemize_with_base_dir + 110
	9   libpango-1.0.0.dylib                0x000000010cac8522 pango_layout_check_lines + 617
	10  libpango-1.0.0.dylib                0x000000010cacbed6 pango_layout_get_extents_internal + 47
	11  libpango-1.0.0.dylib                0x000000010cacc247 pango_layout_get_pixel_extents + 71
	12  libgdraw.5.dylib                    0x000000010bf37cfd _GXPDraw_DoText8 + 115
	13  libgdraw.5.dylib                    0x000000010bef38a4 GDrawGetText8Bounds + 70
	14  libfontforgeexe.2.dylib             0x000000010bcdc8d3 v_e_h + 3963
	15  libgdraw.5.dylib                    0x000000010bef0488 _GWidget_Container_eh + 239
	16  libgdraw.5.dylib                    0x000000010bf33c31 GXDrawSendExpose + 171
[...]

PR #3538, opened 2019-03-16
Last minute MacOS build fixes

jtanx:
If all goes well, I'll be using the travis mac build output for the release. The main things are:

* Disable listening to mac events, this is broken, and causes weird interaction with FontForge (e.g. doesn't show any windows on start, and only shows something when you click on the dock, it also doesn't quit when all the windows are closed). 
* Pin Pango to 1.42.2, to work around #3343

### DIFF
--- a/travis-scripts/_FontForge
+++ b/travis-scripts/_FontForge
@@ -132,7 +132,7 @@ export PATH="$PATH:$bundle_bin"
 echo $WRAPPER $bundle_bin/fontforge > /tmp/oo
 
 if [ $# -eq 0 ] ; then
-    ( exec $WRAPPER $bundle_bin/fontforge -new & )
+    ( exec $WRAPPER $bundle_bin/fontforge -new  )
 else
-    ( exec $WRAPPER $bundle_bin/fontforge "$@" & )
+    ( exec $WRAPPER $bundle_bin/fontforge "$@"  )
 fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -123,6 +123,9 @@ install:
       sed -i.bak '/fc-cache/d' "$(brew --prefix)/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/fontconfig.rb"
       brew unlink python || true
       brew install ./travis-scripts/fontforge.rb --only-dependencies || true
+      export HOMEBREW_NO_AUTO_UPDATE=1
+      # Pin pango to this version, because of issues, see #3343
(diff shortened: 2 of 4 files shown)


jtanx:
So this does work, the pango pinning was successful. Just playing around with re-enabling listening to events, to see if I an get it to work with file open events

skef:
I'm not sure we should have any dependencies on a project that has not only a *bug*, but a bug that causes a *crash*.  People might get confused and think our project's code could be responsible for that sort of thing.

jtanx:
Pango is a pretty core requirement to FontForge, but it's also hard to get anyone to fix things on the quartz backend. fwiw, I did try to debug it, but I was struggling to even get a version of pango with debugging symbols working with everything else to see what the problem was.

MaddTheSane:
> it also doesn't quit when all the windows are closed[...]

That is standard behavior for Mac apps. But because it doesn't use the Mac menu bar, it does make using the app without a window impossible.

travis-scripts/_FontForge
```diff
@@ -132,7 +132,7 @@ export PATH="$PATH:$bundle_bin"
 echo $WRAPPER $bundle_bin/fontforge > /tmp/oo
 
 if [ $# -eq 0 ] ; then
-    ( exec $WRAPPER $bundle_bin/fontforge -new & )
+    ( exec $WRAPPER $bundle_bin/fontforge -new  )
```
jtanx: This was so annoying, with this it actually uses the FontForge icon now, and it semi-works with opening files from Finder.

jtanx:
The reason things are semi broken is that by the time FF installs the carbon (!!!!) event handlers, it's a bit too late, it's already received the open event (and if it wasn't yet started, the open file event). I believe this is because GDK pumps the quartz(?)/cocoa(?) event loop (couple with us calling GDrawProcessPendingEvents a few times in the super complicated startup sequence), whereas before, it wouldn't have been pumped until the event handlers were installed. 

Anyway, I'm leaving this as is, it's good enough for now.

travis-scripts/_FontForge
skef: Kinda makes sense. What would it mean to background a shell `exec` anyway ?

jtanx: It was probably a workaround for some other issue back when it was an X application, I vaguely recall mention of the dock icon continuously bouncing (this doesn't happen any more from what I can tell).

PR #3548, opened 2019-03-17
Potential further release changes 

skef:
1. Add tests to dist in tests/Makefile.am

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -177,7 +177,7 @@ EXTRA_DIST += findoverlapbugs.py test0001.py test0101.py	\
 	test1002.py test1003.py test1004.py test1005.py		\
 	test1006.py test1007.py test1008.py test1009.py         \
 	test1010.py test926.py test927.py test928.py test929.py \
-	test930.py test932.py 
+	test0002.py test930.py test931.py test932.py 
 
 EXTRA_DIST += link_test.c randomtest.c build-aux/missing
 


skef:
I'm not the right judge of how important the two missing tests are. I have doubts that `testsuite` will function well outside of the `make check` "environment" but I could easily be mistaken.

JoesCat:
low/no priority, don't think they are needed,
but mentioning in case you're modifying these Makefile.am:
2 more EXTRA_DIST, each makefile.am
diff -r fontforge/ ../ft/fontforge-20190317/fontforge/
Only in fontforge/: libstamp.c
Only in fontforge/: stamper.c

jtanx:
pretty sure libstamper and stamper.c can both be nixed, we don't even use that any more.

JoesCat:
> I'm not the right judge of how important the two missing tests are. I have doubts that testsuite will function well outside of the make check "environment" but I could easily be mistaken.

just focus on the github files that don't appear in the dist tarball.
testsuite is built during make dist process, so you don't add it in the EXTRA_DIST (leave it out)

skef:
Ha! From stamper.c:
```
/*
 * This program is used to create a 'release date' time stamp when it is time for the
 * next fontforge release build. See fontforge/GNUmakefile.in
 */
```
Old build infrastructure ...

JoesCat:
3 files missed in pycontrib Makefile.am
diff -r pycontrib/ ../ft/fontforge-20190317/pycontrib/
Only in pycontrib/: FontCompare
Only in pycontrib/: svg2glyph
Only in pycontrib/: svg2sfd.py

frank-trampe:
There is nothing more George Williams than writing a timestamp generator in C.

JoesCat:
3 more missed files in Packaging
diff -r Packaging/ ../ft/fontforge-20190317/Packaging/
Only in Packaging/: AppDir
Only in Packaging/: debian
Only in Packaging/: redhat

skef:
I'm pretty sure those aren't in `pycontrib/Makefile.am` because they're not appropriate to be installed. The usual way of "retiring" stuff there is to just remove it from the Makefile.

frank-trampe:
Yes. It's less of a thing now, but, back when we had multiple active forks going, it was very important to keep deviations as small as possible, and that was a good way to do it.

JoesCat:
This is where I leave-off, got other things happening.
Good luck with this. Seems it may be "just in time" for St Patrick's & green beer - LOL

frank-trampe:
@JoesCat, it's a good thing that we don't theme our releases since I'd probably be forced to do some sort of doodle with Saint Patrick and letters made of snakes.

jtanx:
Let's merge this in before this gets stale

frank-trampe:
Fine with me. I got distracted by the discussion.

PR #3549, opened 2019-03-17
Fix coverity reports:

serval2412:
cid#1399463
cid#1399494
cid#1399499
cid#1399519
cid#1399587
cid#1399594

### Important

### Motivation and Context

### Types of changes

### Description

### Final checklist

### DIFF
--- a/contrib/fonttools/showttf.c
+++ b/contrib/fonttools/showttf.c
@@ -5855,8 +5855,16 @@ static int readcff(FILE *ttf,FILE *util, struct ttfinfo *info) {
 	    }
 	}
     }
+    for ( i = 0; strings[i] != NULL; ++i)
+        free(strings[i]);
     free(strings);
+
+    for ( i = 0; dicts[i] != NULL; ++i)
+        free(dicts[i]);
     free(dicts);
+
+    for ( i = 0; fontnames[i] != NULL; ++i)
+        free(fontnames[i]);
     free(fontnames);
 return( 1 );
 }
--- a/contrib/fonttools/showttf.c
+++ b/contrib/fonttools/showttf.c
@@ -2489,8 +2489,7 @@ static void gposPairSubTable(FILE *ttf, int which, int stoffset, struct ttfinfo 
     } else {
 	printf( "\t   !!! unknown sub-table format !!!!\n" );
     }
-    if (glyphs)
-        free(glyphs);
+    free(glyphs);
 }
 
 static void ShowAttach(FILE *ttf) {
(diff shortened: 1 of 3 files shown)


serval2412:
Just to be sure, any problems with this patch?

contrib/fonttools/showttf.c
```diff
@@ -5860,8 +5855,9 @@ static int readcff(FILE *ttf,FILE *util, struct ttfinfo *info) {
 	    }
 	}
     }
-    if (fontnames)
-        free(fontnames);
+    free(strings);
+    free(dicts);
```
skef: These are string arrays. Looking at line 6428, this will free the array structure of `strings` but not the strings. (Line 5574 analogous.)

skef: See line 3957 for an example of freeing these. Could be worth pairing `readcfffontnames` with `freecfffontnames` (and so forth) containing that code rather than repeating it -- more modular that way.

serval2412:
@skef: I read your first comment, made the changed and pushed a new commit.
I've just read your last comment.
I don't see what you mean with line 3957 of this file.
Indeed, I got:
   3957     if ( flags&0x4000 )
   3958         printf( "Don't Advance Glyph " );
from show_indicflags
About freecfffontnames, I suppose you meant to create a new method just containing a loop to free inner elements of the array then free the array, didn't you?

skef:
Sorry, that line number was referring to `parsettf.c` -- I got mixed up with my files. 

About the separate function: I don't care that much either way. The nice thing about pairing the allocation and the freeing, beyond not repeating the (rather small amount of code), is that the implementation can change. (One can *imagine* allocating a struct starting with a `char [][]` followed by a `char[]`, allocating the whole mess into that struct, casting it to `char **`, and then just freeing it at the end, leaving the memory structure different but the interface the same.) But people have different opinions on this and it's not that important in the scheme of things anyway.

jtanx:
'm not too fussed about contrib code, I'd be more interested in coverity reports on the main source repo (anything in fontforge/... fontforgeexe/..., gdraw/... gutils/...)


coverity is broken for me at the moment though :(

serval2412:
@jtanx: I understand it's better to fix contrib upstream but :
- not so easy to find repo (eg: where can I find cidmap or fonttools repos ? I only found Python code for Fonttools )
- not sure they'll accept quickly the patch
- if it's accepted, how much time for Fontforge to retrieve the fix
So I don't think it's a waste of time to fix bugs from contrib parts.

Now I agree with you, fixing the non contrib code should be more interesting but as I say, I'm just giving a try to low hanging fruits.

BTW, no problem here to browse Fontforge bugs in Coverity.

PR #3554, opened 2019-03-18
Fix for #3552 Up/down arrow keys broken in Execute Script

### LINKED ISSUE #3552: Up/down arrow keys are broken in Execute Script
In the “Execute Script” dialog box, pressing the up or down arrow key moves the cursor to the first line of the buffer. The left and right arrow keys work fine. I am using Fontforge 20190317 on macOS Mojave 10.14.3.

skef:
Closes #3552


jtanx:
I take it this was a regression of some sort, what was the cause of that?

jtanx:
Actually no, I retract that, it was in the 2017 release too. Just another classic gdraw bug :P

skef:
> I take it this was a regression of some sort, what was the cause of that?
Not really. @khaledhosny was the last to touch this code in b6a454a27bf9da34f5440d1184271d2df16b2c22, but that was to remove support for the pango alternative. 

It's two more amazing blocks of GW code with exactly one problem that could never have worked.

PR #3556, opened 2019-03-18
Be graceful about negative unicode codepoints

### LINKED ISSUE #3555: nameFromUnicode does not handle negative integers well
`fontforge.nameFromUnicode` returns `'uniFFFFFFFF'` when passed -1 and segfaults when passed any other negative integer. It should instead fail gracefully as it does for other invalid inputs. The stack trace shows that the problem is in `StdGlyphName`.

skef:
I'm not sure it's necessary to pick a validity cutoff with this function; if it is a reviewer can choose one. This eliminates the reported crash. 

Closes #3555

### DIFF
--- a/fontforge/namelist.c
+++ b/fontforge/namelist.c
@@ -184,7 +184,7 @@ const char *StdGlyphName(char *buffer, int uni,enum uni_interp interp,NameList *
     if ( (uni>=0 && uni<' ') ||
 	    (uni>=0x7f && uni<0xa0) )
 	/* standard controls */;
-    else if ( uni!=-1  ) {
+    else if ( uni>0 && uni <= 0x10ffff ) {
 	if ( uni>=0xe000 && uni<=0xf8ff &&
 		(interp==ui_trad_chinese || for_this_font==&ams)) {
 	    const int *pua = interp==ui_trad_chinese ? cns14pua : amspua;
@@ -200,15 +200,23 @@ const char *StdGlyphName(char *buffer, int uni,enum uni_interp interp,NameList *
 			(name = nl->unicode[up][ub][uc])!=NULL )
 	    break;
 	}
+    } else {
+	LogError( _("Warning: StdGlyphName returning name for value %d outside of Unicode range\n"), uni );
     }
     if ( name==NULL ) {
-	if ( uni>=0x10000 )
+	if ( uni>=0x10000 || uni < 0 )
 	    sprintf( buffer, "u%04X", uni);
 	else
 	    sprintf( buffer, "uni%04X", uni);
 	name = buffer;
     }
 return( name );
+}
+
+const char *StdGlyphNameBoundsCheck(char *buffer, int uni,enum uni_interp interp,NameList *for_this_font) {
+    if ( uni<0 || uni > 0x10ffff )
+	return NULL;
+    return StdGlyphName(buffer, uni, interp, for_this_font);
 }
 
 #define RefMax	40
--- a/fontforge/namelist.h
+++ b/fontforge/namelist.h
@@ -8,6 +8,7 @@ extern char **AllNamelistNames(void);
 extern char **SFTemporaryRenameGlyphsToNamelist(SplineFont *sf, NameList *new);
 extern const char *RenameGlyphToNamelist(char *buffer, SplineChar *sc, NameList *old, NameList *new, char **sofar);
 extern const char *StdGlyphName(char *buffer, int uni, enum uni_interp interp, NameList *for_this_font);
+extern const char *StdGlyphNameBoundsCheck(char *buffer, int uni, enum uni_interp interp, NameList *for_this_font);
 extern int UniFromName(const char *name, enum uni_interp interp, Encoding *encname);
 extern NameList *DefaultNameListForNewFonts(void);
 extern NameList *LoadNamelist(char *filename);
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -763,9 +763,10 @@ return( ret );
 
 static PyObject *PyFF_NameFromUnicode(PyObject *UNUSED(self), PyObject *args) {
     char buffer[400], *nlist = NULL;
+    const char *name;
     int uniinterp, uni;
     NameList *for_new_glyphs;
-    PyObject *ret;
+    PyObject *ret = NULL;
 
     if ( !PyArg_ParseTuple(args, "i|s", &uni, &nlist) )
 return( NULL );
@@ -785,8 +786,12 @@ return( NULL );
 	for_new_glyphs = fv_active_in_ui->sf->for_new_glyphs;
     }
 
-    ret = Py_BuildValue("s", StdGlyphName(buffer,uni,uniinterp,for_new_glyphs));
-return( ret );
+    name = StdGlyphNameBoundsCheck(buffer,uni,uniinterp,for_new_glyphs);
+    if ( name!=NULL )
+	ret = Py_BuildValue("s", name);
+    else
+	PyErr_Format(PyExc_ValueError, "Value %d has no Unicode name.", uni);
+    return( ret );
 }
 
 static PyObject *PyFF_UnicodeAnnotationFromLib(PyObject *UNUSED(self), PyObject *args) {


fontforge/namelist.c
```diff
@@ -202,7 +202,7 @@ const char *StdGlyphName(char *buffer, int uni,enum uni_interp interp,NameList *
 	}
     }
     if ( name==NULL ) {
-	if ( uni>=0x10000 )
+	if ( uni>=0x10000 || uni < 0 )
 	    sprintf( buffer, "u%04X", uni);
```
khaledhosny: If `uni < 0` then neither of the two branches here make any sense (they are checking whether to use `u` or `uni` prefix, but both expect +ve number). I think the code should return either NULL or if caller don’t check for that, some nonsensical name as `.none` (not `.null` because there is standard glyph with that name).

skef: Would that be really much more sensible than what the code does for MAXINT? Is there any justification for FontForge having made this a signed value in the first place (other than convenience)? 

I am happy to make U+10FFFF the upper bound if my reviewers are. I'm less happy to implement an arbitrary bound at MAXINT+1.

khaledhosny: Making the upper bound the maximum valid Unicode characters makes sense as well.

skef: Hmm, actually, adding checks at all of those points means adding error paths in some places that currently don't have them. That just seems like an intrusive change likely to create bugs. 

My current inclination is to add `StdGlyphNameBoundsCheck()` with the proposed behavior to centralize the logic and to leave StdGlyphName as I have altered it. I'll shift any calls made from the Python interface to use the new function.

khaledhosny: Make `StdGlyphName()` call the new `StdGlyphNameBoundsCheck()` and check its result and return a dummy glyph name if it returned `NULL`. Thinking again, I think the new function should return `NULL` for -ve numbers only to be a less disruptive change.

skef: I'm not comfortable altering the code to start assigning the same name to multiple glyphs when called in questionable ways. ...BoundsCheck can return NULL for anything outside of 0..0x10FFFF. StdGlyphName can log an error but return a string unique to every int it's called with. 

Your route may be better in absolute terms but it requires more code inspection than I'm interested in doing to fix this problem.

JoesCat: > Would that be really much more sensible than what the code does for MAXINT? Is there any justification for FontForge having made this a signed value in the first place (other than convenience)?

The reason for the signed value is to work with .nodef. in some Unicode/utype.c tables.
If you're going to work on this code, you'll want to consider unicode goes to 0..10ffff, but an older ISO has an even larger range, I added comments here: https://github.com/fontforge/fontforge/blob/master/Unicode/ustring.c:699, however, do what you think improves it versus doing nothing at all, since an incremental improvement is preferred over nothing.
Additionally, it's also worth thinking of possible scenarios between int32 and int64 mayhem, so if you can avoid plain int, it may help prevent mystery magic later on.

skef: I just did a new push. The integer size shouldn't affect the behavior of `StdGlyphNameBoundsCheck()` as all relevant values are comfortably within the range of either. `StdGlyphName()` will print an error for larger or negative values, but should be fine as long as the buffer passed is big enough.

skef:
@frank-trampe this problem involves a balance between extent of change and best practices, but my current compromise has the virtue that it's short and (I think) easy to review. Can you weigh in?

fontforge/namelist.c
```diff
@@ -184,7 +184,7 @@ const char *StdGlyphName(char *buffer, int uni,enum uni_interp interp,NameList *
     if ( (uni>=0 && uni<' ') ||
 	    (uni>=0x7f && uni<0xa0) )
 	/* standard controls */;
-    else if ( uni!=-1  ) {
+    else if ( uni>0 && uni <= 0x10ffff ) {
```
skef: *Could* preferentially check the NameList if it comes in non-NULL for larger values, but only if we reliably know how long it is.

skef:
Can we kick off discussion of this again? @JoesCat or @frank-trampe? The current code adds a function with a bounds-check (for python use). It then alters the existing function to always return a name without crashing, but logs a warning if the name is requested for a number outside the unicode range. 

This seems like a reasonable way to fix the crash but avoid rewriting every caller of the function to handle an error if it happens to pass a weird value. (Many of those callers may already "know" the value they're passing is valid.)

fontforge/namelist.c
JoesCat: ...okay, I took a more careful look here. We have several balls up in the air here, so trying to juggle this is a bit interesting... to begin with, the code comes from a time when ints were maybe 16 bits long. Since this code can end up on commandline with old/ancient PCs, and to play nice with those older PCs, we should probably use long instead of int for these variables, use Unicode/ustring.c:char *utf8_idpb() as a reference. I would suggest changing int up, ub, uc; to long up, ub, uc; that ensures all PCs are using 32bit int, and then the sprintf at 206..208 should be printing long, not int, you can see an example of the 'l' here... libuninameslist/tests/call-test.c:42..44, this also shows we need to consider values larger than ffffh that then require more than 4 hex digits ...there's probably a better example that uses 5,6,7,8 digits elsewhere but I can't recall where ATM.
My error for mentioning 10ffff earlier, the unicode range is 0...11ffffh (looking at makeutype.c:135). Another max point is MAXC seen at line 126, or if you want to use it within FontForge, it is FF_UTYPE_MAXC which is seen at inc/utype.h:43   FF_UTYPE_MAXC is where NameList stops, so everything at this point and higher will only have uni values or hex codes.

skef: I take it that if I change the prototype of `uni` to long I don't need to worry about the callers, as an int can always be converted to a long? Or are you suggesting a more systemic change up the calling hierarchy?

fontforge/namelist.c
```diff
@@ -200,9 +200,11 @@ const char *StdGlyphName(char *buffer, int uni,enum uni_interp interp,NameList *
 			(name = nl->unicode[up][ub][uc])!=NULL )
 	    break;
 	}
+    } else {
+	LogError( _("Warning: StdGlyphName returning name for value %d outside of Unicode range\n"), uni );
     }
```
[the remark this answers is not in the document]
JoesCat: One of the problems with python/native scripting is this Logerror. It stops abruptly in commandline mode, which is ok, but in GUI mode it gives you a message and keeps on trucking as if nothing wrong.
Ideally, you want it to stop too. So you will want to tackle this with two plans. For nameslist, the ideal situation is keep your house clean, and if you have an error, return an error when you fail. You want to push the problem upstream and because LogError has a dual personality, you may want to push this function into python.c and scripting.c instead. Short term, commandline will exit on LogError(), long term strategy, we fix LogError() to not exit, and we do the proper stop at the root level where you process instructions, not deep in the python commands.
If you do a diff between 20120731 scripting.c and current scripting.c you'll probably see a lot of editing in trying to push this exit() out into the  command processing routine in anticipation of fixing LogError(). This should fix the different behaviour between commandline scripting stopping like it should, vs GUI scripting doing let's carry-on as if nothing really happened.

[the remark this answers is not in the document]
skef: I actually don't want it to stop in either case. I just want the error to be evident to a careful observer.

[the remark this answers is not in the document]
skef: I think some of my grousing may have been lost in the shuffle. I've added a new function `StdGlyphNameBoundsCheck()`. That's the one with the error-handling conventions. This one doesn't have error handling *conventions*, but I do want it to print a warning (and not stop) when it's called with a weird value. 

My goal with this change is to 1) add a function that has error handling conventions and is correct, and use that when calling from python, 2) fix the crash and add the warning printing in the existing function. 

If someone else (e.g. @JoesCat) wants to come in after and remove all the calls to the existing function by adapting the callers to have error handling, that would be great.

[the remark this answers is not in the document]
skef: I don't see in `nouiutil.c` where `LogError` would exit, but I guess I can test that easily enough.

skef:
@JoesCat let's concentrate on the logic of the two functions first and then work out what needs to change. You mentioned the real range is "0...11ffffh (looking at makeutype.c:135)." Should I replace the 10ffff guards with 11ffffh or something similar? What is valid and invalid to be looking for the name of?

skef:
Will someone please take this over to fix the crash itself in some way?

skef:
Despite the concerns, this fixes a crash and is not likely to be a source of significant harm. Someone else can always step in and add error handling upstream. With @jtanx's approval I'm merging this.

PR #3560, opened 2019-03-18
Fix crash opening generate dialog for bitmap fonts

khaledhosny:
Opening generate font dialog for a bitmap font currently crashes when FontForge is build without WOFF2 support. This is because existing code assumes the last element in `savefont_extensions` array equals `ff_none` and depends on it, but when built without WOFF2 support that array will be one element short and bad things happen.

Fixed by making the calling code more robust, and also removing the `#ifdefs` from the array; it should have the same number of elements regardless of the build options (yay for having more optional features).

### DIFF
--- a/fontforge/savefont.c
+++ b/fontforge/savefont.c
@@ -68,9 +68,7 @@ const char *savefont_extensions[] = { ".pfa", ".pfb", ".res", "%s.pfb", ".pfa", 
 	".t42", ".t11",
 	".ttf", ".ttf", ".suit", ".ttc", ".dfont", ".otf", ".otf.dfont", ".otf",
 	".otf.dfont", ".svg", ".ufo", ".ufo2", ".ufo3", ".woff",
-#ifdef FONTFORGE_CAN_USE_WOFF2
 	".woff2",
-#endif
 	NULL };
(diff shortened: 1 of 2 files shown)


fontforge/savefont.c
```diff
@@ -84,9 +82,7 @@ const char *savefont_extensions[] = { ".pfa", ".pfb", ".bin", "%s.pfb", ".pfa",
 	".ufo2",
 	".ufo3",
 	".woff",
-#ifdef FONTFORGE_CAN_USE_WOFF2
 	".woff2",
-#endif
```
skef: I'm not thrilled about this. Seems better to
```
#ifdef FONTFORGE_CAN_USE_WOFF2	
	".woff2",
#else
	NULL,
#endif
```
No harm in an extra NULL to preserve the length.

khaledhosny: Not sure if it gains us much. With my change `font.generate("font.woff2")` does nothing, with your proposed change it generates a *Type 1* font with the `font.woff2` extension

khaledhosny: In the GUI, the WOFF2 entry is greyed either way.

PR #3561, opened 2019-03-19
Fix for #3551 altuni is unreliable

### LINKED ISSUE #3551: altuni is unreliable
Setting the field `altuni` of a `fontforge.font` in Python does not take immediate effect. Resetting the encoding fixes it. I am using Fontforge 20190317 on macOS Mojave 10.14.3 with Python 2.7.15.

The following script creates one glyph and assigns it to two code points.
```python
import os

import fontforge

def draw(cp):
    glyph_name = 'zero'
    if glyph_name in font:
        glyph = font[glyph_name]
        new_altuni = ((cp, -1, 0),)
        if glyph.altuni is None:
            glyph.altuni = new_altuni
        else:
            glyph.altuni += new_altuni
        return glyph
    glyph = font.createChar(cp)
    glyph.right_side_bearing = 200

font = fontforge.font()
font.encoding = 'UnicodeFull'
draw(0x30)
draw(0x31)
#font.encoding = font.encoding
for glyph in font.glyphs():
    print glyph.glyphname, glyph.unicode, glyph.altuni
font.generate('out.otf')
```
And indeed, the output shows that FontForge does associate multiple code points with the glyph.
```
zero 48 ((49, -1, 0),)
```
However, only one of the code points gets the glyph.
```console
$ hb-shape out.otf 01
[zero=0+200|.notdef=1+200]
```
Only one of the font’s 'cmap' subtables is correct.
```console
$ ttx -o /dev/stdout -t cmap out.otf 2>/dev/null
<ttFont sfntVersion="OTTO" ttLibVersion="3.26">

  <cmap>
    <tableVersion version="0"/>
    <cmap_format_4 platformID="0" platEncID="3" language="0">
      <map code="0x30" name="zero"/><!-- DIGIT ZERO -->
    </cmap_format_4>
    <cmap_format_0 platformID="1" platEncID="0" language="0">
      <map code="0x0" name="zero"/>
      <map code="0x30" name="zero"/>
      <map code="0x31" name="zero"/>
    </cmap_format_0>
    <cmap_format_4 platformID="3" platEncID="1" language="0">
      <map code="0x30" name="zero"/><!-- DIGIT ZERO -->
    </cmap_format_4>
  </cmap>

</ttFont>
```
Uncommenting the commented line of the script, to reset the encoding, works around the problem.
```console
$ hb-shape out.otf 01
[zero=0+200|zero=1+200]
```

skef:
It works. That's something. (Worth the hours it took to nail down? Hmm.)

Closes #3551

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -6677,6 +6677,7 @@ static int PyFF_Glyph_set_altuni(PyFF_Glyph *self,PyObject *value, void *UNUSED(
     struct altuni *head, *last=NULL, *cur;
     int uni, vs, fid;
     PyObject *obj;
+    FontViewBase *fvs;
 
     if ( value == Py_None )
 	head = NULL;
@@ -6705,7 +6706,13 @@ return( -1 );
 
     AltUniFree(self->sc->altuni);
     self->sc->altuni = head;
-return( 0 );
+
+    for ( fvs=self->sc->parent->fv; fvs!=NULL; fvs=fvs->nextsame ) {
+	fvs->map->enc = &custom;
+	FVSetTitle(fvs);
+    }
+
+    return( 0 );
 }
 
 static PyObject *PyFF_Glyph_get_changed(PyFF_Glyph *self, void *UNUSED(closure)) {


dscorbett:
Thanks!

PR #3563, opened 2019-03-19
Fix Coverity reports:

serval2412:
- cid#1399752
- cid#1399753
- cid#1399754
- cid#1399755
- cid#1399756

### Important

### Motivation and Context

### Types of changes

### Description

### Final checklist

### DIFF
--- a/contrib/fonttools/showttf.c
+++ b/contrib/fonttools/showttf.c
@@ -5542,7 +5542,7 @@ static void readcffprivate(FILE *ttf, struct topdicts *td, char **strings, int s
 	fseek(ttf,td->cff_start+td->private_offset+td->subrsoff,SEEK_SET);
 	readcffsubrs(ttf,td,&td->local_subrs, 1, name );
     }
-    if (name != "<Nameless>") free(name);
+    if (name != NULL && *name != "<Nameless>") free(name);
 }
 
 static struct topdicts **readcfftopdicts(FILE *ttf, char **fontnames, int cff_start) {
--- a/contrib/fonttools/showttf.c
+++ b/contrib/fonttools/showttf.c
@@ -5410,6 +5410,8 @@ static void readcffprivate(FILE *ttf, struct topdicts *td, char **strings, int s
     int32 end = td->cff_start+td->private_offset+td->private_size;
     char *name = NULL;
 
+    const char *nameless_str = "<Nameless>";
+
     fseek(ttf,td->cff_start+td->private_offset,SEEK_SET);
 
     td->subrsoff = -1;
@@ -5542,7 +5544,7 @@ static void readcffprivate(FILE *ttf, struct topdicts *td, char **strings, int s
 	fseek(ttf,td->cff_start+td->private_offset+td->subrsoff,SEEK_SET);
 	readcffsubrs(ttf,td,&td->local_subrs, 1, name );
     }
-    if (name != NULL && *name != "<Nameless>") free(name);
+    if (name != NULL && *name != *nameless_str) free(name);
 }
 
 static struct topdicts **readcfftopdicts(FILE *ttf, char **fontnames, int cff_start) {
--- a/contrib/fonttools/showttf.c
+++ b/contrib/fonttools/showttf.c
@@ -5410,7 +5410,7 @@ static void readcffprivate(FILE *ttf, struct topdicts *td, char **strings, int s
     int32 end = td->cff_start+td->private_offset+td->private_size;
     char *name = NULL;
 
-    const char *nameless_str = "<Nameless>";
+    char *nameless_str = "<Nameless>";
 
     fseek(ttf,td->cff_start+td->private_offset,SEEK_SET);
 
@@ -5421,7 +5421,7 @@ static void readcffprivate(FILE *ttf, struct topdicts *td, char **strings, int s
     td->bluescale = .039625;
 
     name = td->fontname?td->fontname:
-		td->sid_fontname?getsid(td->sid_fontname,strings,smax):"<Nameless>";
+		td->sid_fontname?getsid(td->sid_fontname,strings,smax):nameless_str;
 
     printf( "\n Private Dict for %s\n", name );
     while ( ftell(ttf)<end ) {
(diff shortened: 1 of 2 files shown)


contrib/fonttools/showttf.c
```diff
@@ -5542,7 +5542,7 @@ static void readcffprivate(FILE *ttf, struct topdicts *td, char **strings, int s
 	fseek(ttf,td->cff_start+td->private_offset+td->subrsoff,SEEK_SET);
 	readcffsubrs(ttf,td,&td->local_subrs, 1, name );
     }
-    if (name != "<Nameless>") free(name);
+    if (name != NULL && *name != "<Nameless>") free(name);
```
jtanx: this makes no sense, `name` is a `char*`, and you're dereferencing it to compare against a const string?

serval2412: Thank you for your feedback.
I saw this:
   5421     name = td->fontname?td->fontname:
   5422                 td->sid_fontname?getsid(td->sid_fontname,strings,smax):"<Nameless>";
Re reading these lines, I'm not sure know what to do.

What should I put here?

serval2412: Sorry, I meant "I'm not sure now what to do"

skef: I should have caught this last time -- sorry. Too much scripting work in my recent history. 

Try putting `const char *nameless_str = "<Nameless>"` at the top and using `nameless_str` on both lines 5422 and 5545. That gives you proper pointer equality for the lifetime of the function.

skef:
Close, but for this to work you also have to use it on (new) line 5424. The idea is to have the memory location of the string "<Nameless>" be what you're comparing against. You assign that pointer on 5424 and then check and avoid freeing in that case.

serval2412:
@skef: you mean replacing these:
    name = td->fontname?td->fontname:
		td->sid_fontname?getsid(td->sid_fontname,strings,smax):"<Nameless>";
by these:
name = td->fontname?td->fontname:
		td->sid_fontname?getsid(td->sid_fontname,strings,smax):nameless_str;

and finally do:
if (name != nameless_str) free(name);
?

skef:
That would be the idea. Although thinking about it the "const" is only going to cause assignment problems. Just make `nameless_str` a `char *` instead. It's fine.

serval2412:
Ok, I'll do it when I get back home after my day time job.
Thank you!

jtanx:
An easier fix is to just `copy("Nameless")` and always free it.

contrib/fonttools/showttf.c
khaledhosny: That should use `strcmp()`, that is how strings are compared in C, not by checking pointers for equality.

skef: That's true in general but would be a weird choice in this case. The "question of interest" is whether `name` points to memory that should be `free`d. A pointer comparison is usually the right kind of test for that. The problem in the initial fix is that the location of the "\<Nameless\>" string is unknown due to how it is originally assigned. 

@jtanx is of course right that just dynamically allocating a "\<Nameless\>" string is a fine solution.  I'm giving advice in the direction of @serval2412's chosen approach because it's not misguided and I figure the folks making these smaller tool-driven improvements move on to larger changes by learning this way.

khaledhosny: I see. What I have seen elsewhere in this code base is to set a boolean variable and only free when it is true, but this requires rewriting the `name` assignment at the top (which would be a good thing instead of the terrible multilevel use of ?: operator)

serval2412:
@jtanx: I applied what skef advised because I don't know what method to use to copy a string, strcpy, memcpy, other. ; there are always subtleties in C that I miss or don't know.

fontforge/python.c
```diff
@@ -5155,6 +5155,7 @@ return( NULL );
 		if ( !c->points[nexti]->on_curve ) {
 		    SplinePointListsFree(ss);
 		    PyErr_Format(PyExc_TypeError, "In cubic splines there must be exactly 2 control points between on curve points");
+                    free(sp);
```
skef: This is "new", and seems to be OK, but if it's added here it makes little sense not to also add it after (new) line 5148.

skef: Also, move this up before the error, so that we're freeing things "together".

fontforge/python.c
```diff
@@ -5832,6 +5833,7 @@ static int PyFF_Glyph_CSetLayer(PyFF_Glyph *self, PyObject *value, int layeri, i
     }
 
     if ( PyErr_Occurred() != NULL ) {
+        free(ss);
```
skef: I'm not happy about this one. `ss` is only set via a call to `_SSFromLayer` or `_SSFromContour`. Those should either set an error and return NULL or return a proper `SplineSet`. If this is from a Coverity report rather than by inspection, we should track down the path that returns a non-NULL `SplineSet` pointer while setting an error. 

And anyway, SplineSets have sub-units and are freed with `SplinePointListsFree()` rather than just free.

fontforge/python.c
```diff
@@ -9203,6 +9205,7 @@ return( NULL );
 
     excludes = SSFromLayer((PyFF_Layer *) obj);
     if ( PyErr_Occurred() != NULL ) {
+        free(excludes);
```
skef: Same with this. How/why are we getting non-NULL with an error?

fontforge/python.c
skef: Oh, maybe it's not new. These fixes tend to blend together after a while.

serval2412: In my last commit, I moved up free(sp) + added a new one just some lines before

fontforge/python.c
serval2412: I removed the free instruction since it was only based on Coverity reports

fontforge/python.c
serval2412: Idem

serval2412:
I don't know what's the use of "Resolve the conversation" but thought it was related with "Changes requested", it doesn't seem so.

skef:
If a reviewer requests changes and the submitter says "no" and then does "resolve the conversation" that's not good. If instead the submitter does an update push with the requested changes, or backing out parts of the change, and does "resolve the conversation", that's fine. Or I think it's fine, at least.

fontforge/python.c
```diff
@@ -13294,6 +13296,7 @@ return( -1 );
 return( -1 );
     }
     if ( PyErr_Occurred() != NULL ) {
+        free(ss);
```
skef: This may not be clear from the code, but a SplineSet should never be "deallocated" with just `free()`. Use `SplinePointListsFree()`. It's NULL-tolerant, so it can be called as `free()` is, and I guess I'm fine calling it here.

PR #3567, opened 2019-03-22
 Use GLib for GetAuthor

jtanx:
I've had this patch in the windows releases for a long time. Might as well add it in to master.


jtanx:
Ah, realised that I should probably keep the behaviour from #3176.

gutils/gutils.c
```diff
@@ -28,70 +28,22 @@
...
-	author[sizeof(author)-1] = '\0';
-	ret = author;
+    if (author[0]) {
+        return author;
+    } else if (getenv("SOURCE_DATE_EPOCH")) {
+        const char *username = getenv("USER");
+        if (username) {
+            snprintf(author, sizeof(author), "%s", username);
+            return author;
+        }
```
skef: I take it it doesn't matter that this will be enabled and have priority under `__MINGW32_`?

jtanx: Yeah that's fine, from what I understand, the SOURCE_DATE_EPOCH+USER env variables were to always have higher priority for reproducibility (and there wasn't a mandate that this only applies to linux)

skef: lgtm

Issue #3570, opened 2019-03-23
HiDPI question - how can I make the CharView's ruler, toolbox larger?


ctrlcctrlv:
GDK has massively helped this project in re: HiDPI, but it seems like no matter what I set `ScreenWidthInches` or `GDK_SCALE` to, the ruler and toolbox remain a constant size.

How can I make them larger so I don't have to use a magnifier anymore to do my work? The massive size imbalance is easily seen in the image below.

![Screenshot_20190323_130416](https://user-images.githubusercontent.com/838783/54861852-330ff000-4d6c-11e9-8ce5-3cd348db0d20.png)

jtanx:
Does something like this work?

```patch
diff --git a/fontforgeexe/charview.c b/fontforgeexe/charview.c
index 67dbedc23..1bcd940b3 100644
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -12598,11 +12598,8 @@ static void _CharViewCreate(CharView *cv, SplineChar *sc, FontView *fv,int enc,i
 
     cv->showdebugchanges = CVShows.showdebugchanges;
 
-    cv->infoh = 13;
-#if defined(__MINGW32__)||defined(__CYGWIN__)
-    cv->infoh = 26;
-#endif
-    cv->rulerh = 16;
+    cv->infoh = GDrawPointsToPixels(NULL, 13);
+    cv->rulerh = GDrawPointsToPixels(NULL, 16);
 
     GDrawGetSize(cv->gw,&pos);
     memset(&gd,0,sizeof(gd));
@@ -12909,11 +12906,8 @@ CharView *CharViewCreateExtended(SplineChar *sc, FontView *fv,int enc, int show
     cv->enc = enc;
     cv->map_of_enc = fv->b.map;		/* I know this is done again in _CharViewCreate, but it needs to be done before creating the title */
 
-    cv->infoh = 13;
-#if defined(__MINGW32__)||defined(__CYGWIN__)
-    cv->infoh = 26;
-#endif
-    cv->rulerh = 16;
+    cv->infoh = GDrawPointsToPixels(NULL, 13);
+    cv->rulerh = GDrawPointsToPixels(NULL, 16);
 
 
     SCLigCaretCheck(sc,false);
```

jtanx:
Fixing the palette is harder, the size is tied to the image size of the icons...

ctrlcctrlv:
[...]

PR #3572, opened 2019-03-23
Make it clear which GUI backend is in use

ctrlcctrlv:
This will help prevent confusion among users about which backend they
are using. This will make conversations like the one in #3570 less
likely to recur.

Just so everyone knows, I indeed tested this PR with `--with-x`, `--without-x`, `--enable-gdk=gdk3` and `--enable-gdk=gdk2` and saw the expected results.

gdk3

### DIFF
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -281,6 +281,15 @@ static void SplashLayout() {
 #ifdef FONTFORGE_CONFIG_USE_DOUBLE
     uc_strcat(pt,"-D");
 #endif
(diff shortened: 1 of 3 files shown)


JoesCat:
Nice touch. Anything that reduces confusion is a good direction.

jtanx:
It was actually possible to do this without modifying the build system; there's `FONTFORGE_CAN_USE_GDK` to distinguish between X11 and GDK, and you can check `GDK_MAJOR_VERSION` to see if it's GDK2 or 3.

PR #3573, opened 2019-03-23
Rework FONTFORGE_ARG_ENABLE_GDK

floppym:
This allows --disable-gdk or --enable-gdk=no to work properly.
Previously, either of these would result in GDK3 being enabled.


floppym:
For an easier code review, I suggest hiding whitespace changes.

https://github.blog/2018-05-01-ignore-white-space-in-code-review/

PR #3562, opened 2019-03-19
Remove breakpad

jtanx:
It's been broken for years, and when it did work, it only worked on Mac. Even then, I don't think it was ever fully functional. It's taking up build time (although the makefile structure was also just inefficient) , so let's just remove it

PR #3576, opened 2019-03-24
Fix Mac bundle creation

jtanx:
This is a follow-up to #3562. With breakpad gone, the `Frameworks` folder didn't exist, so instead of copying `Python.framework` into `Frameworks`, it copied `Python.framework`, renaming it to `Frameworks`. So this ensures that the framework directory is created before copying in that folder.

### DIFF
--- a/travis-scripts/ffosxbuild.sh
+++ b/travis-scripts/ffosxbuild.sh
@@ -51,8 +51,10 @@ echo "Bundling Python libraries..."
 otool -L $workdir/bin/fontforge
 
 pylib=$(otool -L $workdir/bin/fontforge | grep -i python | sed -e 's/ \(.*\)//')
-pycruft=$(dirname $pylib)/../../..
-cp -a $pycruft/Python.framework $outdir/Contents/Frameworks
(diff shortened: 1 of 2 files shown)


travis-scripts/ffosxbuild.sh
```diff
@@ -51,8 +51,10 @@ echo "Bundling Python libraries..."
 otool -L $workdir/bin/fontforge
 
 pylib=$(otool -L $workdir/bin/fontforge | grep -i python | sed -e 's/ \(.*\)//')
-pycruft=$(dirname $pylib)/../../..
-cp -a $pycruft/Python.framework $outdir/Contents/Frameworks
+pycruft=$(realpath $(dirname $pylib)/../../..)
+echo "pycruft: $pycruft"
+mkdir -p $outdir/Contents/Frameworks
+cp -av $pycruft/Python.framework $outdir/Contents/Frameworks
```
frank-trampe: Not worth changing, but I find that this is clearer if the destination parent has a `/` at the end.

frank-trampe:
The changes look fine. If Travis likes it, let's merge.

PR #3578, opened 2019-03-24
Simplify GDK/X11 identification

jtanx:
Uses pre-existing defines, instead of creating new ones.

Follow-up to #3572

### DIFF
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -281,14 +281,14 @@ static void SplashLayout() {
 #ifdef FONTFORGE_CONFIG_USE_DOUBLE
     uc_strcat(pt,"-D");
 #endif
-#ifdef BUILT_WITH_GDK2
-    uc_strcat(pt,"-GDK2");
-#endif
-#ifdef BUILT_WITH_GDK3
+#ifdef FONTFORGE_CAN_USE_GDK
+#  if GDK_MAJOR_VERSION >= 3
     uc_strcat(pt,"-GDK3");
-#endif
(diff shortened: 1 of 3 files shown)


ctrlcctrlv:
Very sorry I created extra work for you sir, I wasn't aware of `GDK_MAJOR_VERSION` :frowning_face:

jtanx:
nah that's fine, I forgot to comment on that pr earlier before it got merged

ctrlcctrlv:
Yes, also if I can say something unrelated, I'm sorry for flooding the issue tracker with backtraces, I don't know if what I'm doing is helping, for the most part I'm just trying to document all the crashes and weird behavior I experience on a day-to-day basis using FontForge for hours at a time to make my own fonts.

I hope it's not seen as spam

skef:
I jump into spiro stuff slower than the standard spline code, but I'm planning to look at those crashes when I'm through with this doc work if someone doesn't first.

ctrlcctrlv:
Sure, I understand that, I personally use Spiro for all my fonts, I find it a joy to use and that it generates splines of much higher quality than I can do with the standard Bézier tools. I understand however that there is a steep learning curve w.r.t. Spiro, not the least part of which is due to its various bugs and crashes and workarounds, and that few users find it in any case as delightsome as me. I appreciate all work done on fixing Spiro crashes, and as long as they aren't seen as spam, I'll continue reporting them if you don't mind.

skef:
I'm just a guy, but I would think any crash is more than fair game to report. 

I'm slow on spiro not because I think it's unused but because of the library dependency.

PR #3580, opened 2019-03-24
basic python.html fix-up 

skef:
Fix UnicodeNames2... table markup
Move top-level names and descriptions outside of tables
Correct faulty use of "id"
Minor changes

Not great, but not wrong and could do for a while.

### DIFF
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -55,10 +55,9 @@
 <BLOCKQUOTE>
 <PRE>fontforge -c 'open(argv[1]).generate(argv[2])'</PRE>
 </BLOCKQUOTE>
-  <TABLE id="module" border=1>
-    <CAPTION>
-      <A NAME="psMat">psMat</A>
-    </CAPTION>
+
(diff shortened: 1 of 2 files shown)


doc/html/python.html
```diff
@@ -3161,21 +3107,16 @@ <H3>
...
-	<P>
-	This type may not be pickled.</TD>
-    </TR>
+
+<H2 id="Font">font</H2>
+
+The font type refers to a fontforge Font object. It generally
+contains a list of glyphs, an encoding to order those glyphs, a fontname,
+a list of GPOS/GSUB lookups and many other things.
+<P>
```
jtanx: What's with these `<p>` tags? I don't see the corresponding `</p`> and I don't think you usually nest tables inside of paragraph blocks.

jtanx: Oh, you don't actually have to close p tags. Weird

skef: You do in xhtml but not normal old html. Same with dt and dd.

PR #3581, opened 2019-03-24
Close #2261

### LINKED ISSUE #2261: Glyph Window right click menu with Spiro
I expect to see 

```
G4 Curve
G2 Curve
Corner
Left Constraints
Right Constraints
```

on Right Click menu when Spiro is turned on. 
Instead it is showing the same `Curve`, `HV Curve`, `Corner`, `Tangent` that applies for beziers.

screen shot 2015-05-06 at 10 47 59 pm

But as you can see the Point menu is changing accordingly.

screen shot 2015-05-06 at 10 48 09 pm

ctrlcctrlv:
This was *much* easier to fix than I anticipated - w00t!

### Types of changes


fontforgeexe/cvpalettes.c
```diff
@@ -3518,34 +3518,72 @@ void CVToolsPopup(CharView *cv, GEvent *event) {
...
+            mi[i].invoke = CVMenuPointType;
+            i++;
+            mi[i].ti.text = (unichar_t *) _("Left Constraint ([)");
+            mi[i].ti.text_is_1byte = true;
+            mi[i].ti.fg = COLOR_DEFAULT;
+            mi[i].ti.bg = COLOR_DEFAULT;
+            mi[i].mid = MID_SpiroLeft;
+            mi[i].invoke = CVMenuPointType;
+            i++;
+            mi[i].ti.text = (unichar_t *) _("Right Constraint (])");
```
skef: What's the bracket about? (Not implying it's wrong.)

ctrlcctrlv: I debated adding it, but I think it makes it more obvious what point type is being talked about, since the icon looks like a bracket. I think it's a UI enhancement to throw it in, but can take it out if you don't like it.

skef: I think it's best to leave it out -- it might look to some users like some unknown form of keyboard acceleration.

skef: Would have also affected i18n, now that I think of it. But anyway ...

skef:
This looks good but I have one question: I don't know that much about Spiro and you've used the GetInfo box a bunch. Is there any circumstance in which a type isn't valid? (Like, greyed out in the box?) I expect not but should ask anyway.

fontforgeexe/cvpalettes.c
ctrlcctrlv: Right, I've taken it out already. :-)

ctrlcctrlv:
There are indeed situations where a certain point type is invalid, but FontForge does not solve this by graying out items, it solves this by just drawing straight lines between all points to show the user their selection is invalid and a Spiro spline cannot be drawn between their chosen points. So, no, there are never grayed out items in the Get Info dialog nor the toolbox.

skef:
Ah, OK. I was playing around and saw that behavior but didn't know what I was seeing. 

This looks good to me.

Issue #3582, opened 2019-03-24
fontforge tries to initialize GDK when no UI will be displayed


floppym:
Downstream bug: https://bugs.gentoo.org/681650

On Gentoo Linux, we run `make check` as part of the package build process. The test suite involves calling the built fontforge binary with options to convert several font files without any user interaction.

In the past, running these tests did not require an active X display.

After upgrading to 20190317, we get the following output from the test suite.

```
../fontforgeexe/fontforge -lang=ff -c 'Open($1); Generate($2)' fonts/Ambrosia.sfd fonts/Ambrosia.pt3
No protocol specified

(fontforge:326): Gdk-WARNING **: 13:12:52.415: cannot open display: 
make[3]: *** [Makefile:1823: fonts/Ambrosia.pt3] Error 1
make[3]: Leaving directory '/var/tmp/portage/media-gfx/fontforge-20190317/work/fontforge-20190317/tests'
make[2]: *** [Makefile:1662: check-am] Error 2
make[2]: Leaving directory '/var/tmp/portage/media-gfx/fontforge-20190317/work/fontforge-20190317/tests'
make[1]: *** [Makefile:1916: check-recursive] Error 1
make[1]: Leaving directory '/var/tmp/portage/media-gfx/fontforge-20190317/work/fontforge-20190317'
make: *** [Makefile:2205: check] Error 2
```
I believe this happens because fontforge always calls the `gdk_init()` function, regardles of whether the user interface will ever be displayed.

jtanx:
Ah, that makes sense. It might be enough to replace `gdk_init` with `gdk_init_check`, as there are checks further down the line to determine if it can get a display or not.

Otherwise we'd need to check the result after `CheckIsScript` is called in `startui.c`, or otherwise move `gdk_init` further down.

floppym:
> otherwise move gdk_init further down.

This seemed like the most obvious solution to me. I tested it and it seems to work ok for both scripts and GUI usage.

Issue #1277, opened 2014-03-07
Command-line processing and "main program"


ghost:
The top-level processing of command-line arguments and start/run/terminate control flow in FontForge is currently several plates of spaghetti spread throughout the system:
- main.c:main() - trivial wrapper that passes argc and argv to fontforge_main(), which has two different definitions in two different files selected by conditionals at the Makefile level.
- startnoui.c:fontforge_main() - does some initialization, prints a banner, passes argc and argv to CheckIsScript which _may or may not return_, then if it returns, parses the command line, then _if Python is not enabled_ passes argc and argv to ProcessNativeScript(), else if Python is enabled (but even if native is also enabled) passes to PyFF_Stdin(), then if that returns, does cleanup and exits.
- startui.c:fontforge_main() - does some command-line parsing _first_, without the initialization done by the other version of this function.  Then runs a bunch of Macintosh- and Windows-specific code, depending on the platform.  Then runs CheckIsScript(), passing it argc and argv; it may or may not return.  Then (if it returns) parses the command line again, differently.  Then does Python initialization (which would NOT have been done in the other version).  Shows the splash screen, runs the GUI loop, does collab stuff.  Does cleanup and exits.
- scripting.c:CheckIsScript() - trivial wrapper that passes argc and argv to _CheckIsScript()
- scripting.c:_CheckIsScript() - parses the command line.  Also attempts to determine whether it was executed via a shebang, using hardcoded assumptions about the name of the FontForge executable.  Passes argc and argv to ProcessNativeScript() or to PyFF_Main(), or invokes PyFF_Stdin() without argc and argv, or just returns, depending on complicated conditions and #ifdefs.
- scripting.c:ProcessNativeScript() - parses the command line.  Obtains native-language script code from a location determined by the command-line parse, runs it, does not do cleanup, and terminates.
- python.c:PyFF_Main() - does some Python-specific initialization.  Parses the command line.  Creates a new, filtered version of argc and argv, passes those to Py_Main(), and then terminates.
- python.c:PyFF_Stdin() - does Python-specific initialization.  Then depending on whether standard input is a TTY, invokes either PyRun_InteractiveLoop() or PyRun_SimpleFile(), and terminates.
[...]

PR #3583, opened 2019-03-24
Defer gdk_init() until after CheckIsScript()

floppym:
This allows fontforge to run without an X server available.

Fixes: https://github.com/fontforge/fontforge/issues/3582

### DIFF
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -900,9 +900,6 @@ int fontforge_main( int argc, char **argv ) {
 #if !(GLIB_CHECK_VERSION(2, 35, 0))
     g_type_init();
 #endif
-#ifdef FONTFORGE_CAN_USE_GDK
-    gdk_init(&argc, &argv);
-#endif
 
     /* Must be done before we cache the current directory */
     /* Change to HOME dir if specified on the commandline */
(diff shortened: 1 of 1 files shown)


skef:
This is not a review (it looks OK to me but others know more about this path and can review with higher confidence). However, while I think this *will* work with the current `PythonUI_Init()`, either:

1. Some "script" may want to do graphical stuff of/on its own, in which case this change removes that possibility, or
2. We don't provide for that opportunity, in which case `PythonUI_Init()` can and probably should be moved down after `gdk_init()`. 

Anyone making use of `PythonUI_Init()` in the future could wind up confused about the graphics layer "not being there" when the function is called.

jtanx:
This is fine, when running via `-script`, it's implicit that it doesn't bring up a gui, gdraw is only initialised around where `GDrawCreateDisplays(display,argv[0]);` is called.

jtanx:
Oh, right I see your point. FontForge's startup procedure is so ridiculously overcomplicated...

skef:
To be clear, this is arguably speculative generality, because `PythonUI_Init` just does something that should work fine, probably can't be used now, and is relevant to the GDK path anyway. But in terms of comprehension it's probably best to move it down.

jtanx:
For now, I think it's fine to leave in it's current place, since it won't change existing behaviour, and I can kind of see why it was grouped there with the other init calls. Longer term, I'm contemplating rewriting the whole startup procedure, because it's just bonkers

skef:
Bootstrapping is intrinsically bonkers. I have yet to encounter a bootstrapping sequence that wasn't.

jtanx:
Let's just say it doesn't need to be as complex as it currently is, #1277 gives a good overview

PR #3585, opened 2019-03-25
Install html docs in htmldir

floppym:
Recent versions of autoconf support  the `--htmldir` option. Let's use it
instead of a custom make variable.


JoesCat:
Apparently autoconf [ver2.63](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Installation-Directory-Variables.html) (we currently have configure.ac minimum set for AC_PREREQ([2.68]), so this seems like a good patch).

My preference would be to put your solution in configure.ac and to eventually delete mk/layout.mk altogether, but otherwise my vote is - I'm ok with this patch.

PR #3587, opened 2019-03-25
Fix the use of deprecated FSRef usage in OpenDocumentsAE.

MaddTheSane:
OpenDocumentsAE uses the deprecated FSRef types for getting the document paths. This changes it to get the future-proof `file://` URLs and get the file paths from that.

### DIFF
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -406,7 +406,7 @@ static FILE *logfile;
 
 /* These are the four apple events to which we currently respond */
 static pascal OSErr OpenApplicationAE( const AppleEvent * theAppleEvent,
-	AppleEvent * reply, SInt32 handlerRefcon) {
+	AppleEvent * reply, SRefCon handlerRefcon) {
  fprintf( logfile, "OPENAPP event received.\n" ); fflush( logfile );
     if ( localsplash )
 	start_splash_screen();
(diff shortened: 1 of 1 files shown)


fontforgeexe/startui.c
```diff
@@ -463,10 +462,16 @@ static pascal OSErr OpenDocumentsAE( const AppleEvent * theAppleEvent,
...
     for(index = 1; index <= count; index++) {
-        err = AEGetNthPtr(&docList, index, typeFSRef,
-                        NULL, NULL, &theFSRef,
-                        sizeof(theFSRef), NULL);// 4
-	err = FSRefMakePath(&theFSRef,(unsigned char *) buffer,sizeof(buffer));
+        AEDesc aDoc;
+        err = AEGetNthDesc(&docList, index, typeFileURL, NULL, &aDoc);
+        size_t bytecount = AEGetDescDataSize(&aDoc);
+        void *pathPtr = malloc(bytecount);
+        err = AEGetDescData(&aDoc, pathPtr, bytecount);
```
jtanx: seems weird that we store the error but do nothing with it

MaddTheSane: That's how the code was originally written. That said, using the `continue` directive when there's an error should prevent errors.

Issue #3431, opened 2019-01-29
Web server in collab mode?


ctrlcctrlv:
Since we're discussing removing possibly deprecated elements (#3399), I'd like to ask about this feature.

What does it do? Has it ever worked? Who came up with this? Does anyone actually use it? 

For me it doesn't work: (Arch Linux, x64)

```
Error: Cannot find module 'socket.io'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:611:15)
    at Function.Module._load (internal/modules/cjs/loader.js:537:25)
    at Module.require (internal/modules/cjs/loader.js:665:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/usr/share/fontforge/nodejs/collabwebview/server.js:12:10)
    at Module._compile (internal/modules/cjs/loader.js:736:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:747:10)
    at Module.load (internal/modules/cjs/loader.js:628:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:568:12)
    at Function.Module._load (internal/modules/cjs/loader.js:560:3)
```

Even if it did I'm dubious about its utility or applicability to the current code base. Most probably it should be removed.

ctrlcctrlv:
@skef Thank you, I was having trouble finding that. 

I think it should be removed from the default installation, but it could be a cool plugin. It doesn't seem to be maintained and is currently broken (for me, most probably for you too).

skef:
It's a little tricky because the code isn't limited to added files. There is a compilation flag and a bunch of hooks in the main code. I've been leaning towards:

1. Ensuring that the contrib and nodejs directories don't install unless the support is compiled in (if this isn't already true) and
2. Warning that it has not been updated and maintained and may not work on many systems. 

Longer term it would be best if more of the implementation were pushed one level out. Currently the transport and addressing and such intrude into the code. It would be better if there were python hooks for sending and receiving "collaboration messages" as changes occur with the transport logic living on the python side. That way:

1. There would be no compilation flags or added libraries for the executable itself.
2. Interested parties (if any) could experiment with different collaboration arrangements without being tied to a single messaging system. 

One could also potentially "collaborate" with a script, opening up new potentially interesting scripting possibilities, such as help functions that demo operations for you. 

However, this would be a bit of work.

skef:
This was removed.

PR #3596, opened 2019-03-28
Close #3594

### LINKED ISSUE #3594: Feature request: Auto-load feature file on startup
My workflow as of late has been putting my kerns into a feature file, loading them into FontForge, closing without saving, and repeating ad nauseum as I adjust little things here and there.

It's rather tiresome to keep having to press Alt+Ctrl+Shift+K, select my feature file, then finally Ctrl+K to test my changes, which are extensive...

```fea
feature kern {
    lookup kern_G {
        pos @lowerA @lowerA -30; # (A)
        pos @lowerA [c] 10;
        pos @lowerA [s] 5;
        pos @upper_groupC @lowerA -30;
    } kern_G;

    lookup kern_S {
        pos N @f_lowers -80;
        pos T @lowerA 10;
        pos @f [@f_lowers r] -110;
        pos F @f_lowers -110;
        pos F @r -90;
        pos [@lower_no_f @upper_no_f_N] @a 10; # (B) This partially cancels out (A)
        pos [@r] @a -65;
        pos [t] @a -45;
        pos @a [n m] -45;
        pos [h n m] @a -20;
        pos [b o] [q d] 30;
        pos o [c e] 30;
        pos [b o] @lowerA 10;
        pos e @a -15;
        pos [s] @lowerA 15;
    } kern_S;

    # This is mostly for handling many repeating `f`'s, as in, "Oofffff!"
    lookup kern_L {
        pos [f_f] [f f_f f_l f_f_l f_i f_f_i] -185;
        pos [f]   [f f_f f_l f_f_l f_i f_f_i] -175;
    } kern_L;
} kern;
```

I already run FontForge with `-nosplash -quiet`. Is there a possibility of adding a `-loadfeaturefile` or similar as well?

ctrlcctrlv:
While working on this I saw https://github.com/fontforge/fontforge/issues/3594#issuecomment-477536109 &hellip;so&hellip;this might be rejected, but I'm going to try anyway.

This feature is extremely useful for anyone who uses the SFD / FEA separation mode of FontForge development.

Even if rejected, I learned a lot writing it, (who knew `ViewPostScriptFont` is also for SFD files? Wtf! :laughing: )  and will likely keep this patch in my own personal builds.

### DIFF
--- a/doc/html/cliargs.html
+++ b/doc/html/cliargs.html
@@ -2,7 +2,7 @@
 <HEAD>
   <!-- Created with AOLpress/2.0 -->
   <!-- AP: Created on: 14-Jan-2002 -->
-  <!-- AP: Last modified: 22-Aug-2008 -->
+  <!-- AP: Last modified: 29-Mar-2019 -->
   <TITLE>Command Line Arguments</TITLE>
   <LINK REL="icon" href="fftype16.png">
   <LINK REL="stylesheet" TYPE="text/css" HREF="FontForge.css">
@@ -70,6 +70,16 @@ $ fontforge -c script-string [arguments]
       -display name
     <DD>
       Specifies the name of the display on which FontForge will open its windows
+    <DT>
+      -featurefile filename
+    <DD>
+      Automatically loads the specified feature file on start up. (This performs
+      the same action as would manually clicking <a
+      href="filemenu.html#Merge-feature">Merge Feature Info&hellip;</a> in the
+      UI.) <BR/> This is useful if you primarily add OpenType features to your fonts
+      via feature (<tt>fea</tt>) files. You could use this, for example, to tweak the
+      kerning in your feature file and test it conveniently in FontForge's <a
+      href="metricsview.html">Metrics window</a>.
     <DT>
       -dontopenxdevices
     <DD>
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -35,6 +35,7 @@
 #include "encoding.h"
 #include "fontforgeui.h"
 #include "lookups.h"
+#include "splinesaveafm.h"
 #include "start.h"
 
 #ifndef _NO_LIBUNICODENAMES
@@ -134,6 +135,7 @@ static void _dousage(void) {
     printf( "\t-quiet\t\t\t (don't print non-essential information to stderr)\n" );
     printf( "\t-unique\t\t\t (if a fontforge is already running open\n\t\t\t all arguments in it and have this process exit)\n" );
     printf( "\t-display display-name\t (sets the X display)\n" );
+    printf( "\t-featurefile filename\t (automatically load feature file)\n" );
     printf( "\t-depth val\t\t (sets the display depth if possible)\n" );
     printf( "\t-vc val\t\t\t (sets the visual class if possible)\n" );
     printf( "\t-cmap current|copy|private\t (sets the type of colormap)\n" );
@@ -902,10 +904,12 @@ int fontforge_main( int argc, char **argv ) {
     int i;
     int recover=2;
     int any;
+    FontViewBase *openedfont;
     int next_recent=0;
     GRect pos;
     GWindowAttrs wattrs;
     char *display = NULL;
+    char *featurefile = NULL;
     FontRequest rq;
     int ds, ld;
     int openflags=0;
@@ -1131,6 +1135,8 @@ int fontforge_main( int argc, char **argv ) {
 	    AddR(argv[0],"Gdraw.Keyboard", argv[++i]);
 	else if ( strcmp(pt,"-display")==0 && i<argc-1 )
 	    display = argv[++i];
+	else if ( strcmp(pt,"-featurefile")==0 && i<argc-1 )
+	    featurefile = argv[++i];
 # if MyMemory
 	else if ( strcmp(pt,"-memory")==0 )
 	    __malloc_debug(5);
@@ -1339,7 +1345,7 @@ exit( 0 );
 	    // already handled above.
 	} else if ( strcmp(pt,"-last")==0 ) {
 	    if ( next_recent<RECENT_MAX && RecentFiles[next_recent]!=NULL )
-		if ( ViewPostScriptFont(RecentFiles[next_recent++],openflags))
+		if ( openedfont = ViewPostScriptFont(RecentFiles[next_recent++],openflags))
 		    any = 1;
 	} else if ( strcmp(pt,"-sync")==0 || strcmp(pt,"-memory")==0 ||
 		    strcmp(pt,"-nosplash")==0 || strcmp(pt,"-recover=none")==0 ||
@@ -1357,6 +1363,9 @@ exit( 0 );
 		    strcmp(pt,"-display")==0 || strcmp(pt,"-recover")==0 ) &&
 		i<argc-1 )
 	    ++i; /* Already done, needed to be before display opened */
+    else if ( (strcmp(pt,"-featurefile")==0) && i<argc-1 ) {
+        ++i;// added to prevent argument being treated as a filename.
+    }
 	else if ( strcmp(pt,"-allglyphs")==0 )
 	    openflags |= of_all_glyphs_in_ttc;
 	else if ( strcmp(pt,"-open")==0 )
@@ -1375,14 +1384,14 @@ exit( 0 );
 		if ( GFileExists(fname)) {
 		    /* It's probably a Unified Font Object directory */
 		    free(fname);
-		    if ( ViewPostScriptFont(buffer,openflags) )
+		    if ( openedfont = ViewPostScriptFont(buffer,openflags) )
 			any = 1;
 		} else {
 		    strcpy(fname,buffer); strcat(fname,"/font.props");
 		    if ( GFileExists(fname)) {
 			/* It's probably a sf dir collection */
 			free(fname);
-			if ( ViewPostScriptFont(buffer,openflags) )
+			if ( openedfont = ViewPostScriptFont(buffer,openflags) )
 			    any = 1;
 		    } else {
 			free(fname);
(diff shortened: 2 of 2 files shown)


skef:
I don't have strong feelings about this. I disagreed with the approach because it doesn't seem to match the past development patterns, which focus on flexibility through scripting rather than a large number of command line arguments.

skef:
Also @ctrlcctrlv - that part of the boilerplate that talks about documentation changes applies to this push. However, it does seem reasonable to wait on that until others have weighed in on whether this is the right direction.

ctrlcctrlv:
I've added it to the --help message, but will of course look for other appropriate places in the documentation to add it to. 👍

ctrlcctrlv:
I've added a brief explanation to `cliargs.html`. I believe that this change will benefit users a lot, it's already benefited my work in just the short hours I've been using it. I'm willing to explain to any developer why this feature should be in the core program; as there is seldom crossover between font authors and FontForge developers, it may not be obvious why this feature is useful.

I'd like to ping my friend @davelab6 to see what he thinks about this feature, given his extensive use of FontForge to create absolutely stunning fonts that in no way compare to mine.

ctrlcctrlv:
Sure! If I read your comment correctly, when supplying `-featurefile`, I should clear the GPOS and GSUB already in the file? I think I can figure out how, but can you confirm that's what you mean before I start?

skef:
Well:

1. *If* there is to be a single command, I *do* think clearing the existing entries before loading the file is likely to be useful for more users. 
2. But one can also imagine having two commands, one to empty and one to load, which can be used separately or together. (And if we do this it might be good to add a python command to do the same, rather than making the user write a loop.)
3. I am not personally convinced this route is the way to go.

However, now that I think about it it's harder to make an argument against option 2. So I guess that amounts to my supporting having those two flags.

skef:
In general the trick with command line flags is the fewer the better, but having enough to avoid forcing an unwanted model on this or that user.

If we add an `-emtpyfeatures` flag (or whatever) someone will inevitably request a flag to get rid of only *some* features. Doing that on a lookup basis is too baroque -- those folks should write scripts. But I can see some sense in being able to choose between GSUB, GPOS, or both -- maybe with an optional parameter to the flag, if that's straightforward to implement. (Libertinus is now arranged so that kerning data is kept in the SFDs while GSUB data is kept in feature files.)

ctrlcctrlv:
I would say that yes if you only want to empty some features it's time for a script. 

But I also insist that keeping your SFD and FEA separate is so _natural_ that FontForge should acknowledge its validity by merging my commit. I've struggled with FontForge for years my friend, making so far three fonts and close to releasing my fourth, with many others in varying states of completeness, and at the end I came to the same conclusion as my dear friend @khaledhosny and many others: keep the files separate.

skef:
There are a lot of good arguments for having a separate feature file be *canonical* for a font. That idea and its advantages seem to have morphed in many people's minds to the idea that the sfd should have no feature data in it. But a) those are two separate things and b) you don't personally believe it because you're asking for an easier interface to load the data. (@khaledhosny doesn't trust the metrics implementation and therefore doesn't care to load the file except before generating, and only then because there isn't yet solid support for an alternative.) 

I don't think keeping the sfd empty of features is natural at all. It seems much easier to make the file canonical and reload it only when its changed or you think the current data might be out of date for another reason. So I don't really "get" how the issues have become so connected in some people's minds.

skef:
(Incidentally, therefore, the interface you yourself would most benefit from is a tools menu command added with an init script that empties the features with a loop like
```
    for gsublname in ff.gsub_lookups:
        ff.removeLookup(gsublname)
```
and then loads your feature file, together with a keyboard accelerator for that command. It shouldn't be necessary for you to quit and restart just to reload the file. That was going to be my original suggestion but I have yet to get a keyboard accelerator working, and haven't yet dug down to figure out if that's my ignorance or if there's some implementation problem.)

ctrlcctrlv:
> It shouldn't be necessary for you to quit and restart just to reload the file.

Hmm :thinking: :thinking: ...this is very true, I didn't think of that, I just figured reloading my feature files over and over again would reveal FontForge bugs, but you are definitely right...

skef:
Do you have a naming convention for your feature files relative to font pathname, font/family name, a fixed name in the same directory, or something else? I can whip up a command for you.

ctrlcctrlv:
Yes I just follow the UFO convention and have one feature file called `features.fea` in the same directory as my SFD.

skef:
So the script in this archive is almost what you really want: [featuremunge.zip](https://github.com/fontforge/fontforge/files/3021999/featuremunge.zip)
put it in `~/.config/fontforge/python` and run the version of your choice from the `Tools` menu. 

It would be precisely what you want if it had a keyboard accelerator. Let me look into that.

skef:
Oh good grief **of course** `fontforgeexe/pythonui.c` has completely different string handing conventions than `fontforge/python.c`.

jtanx:
Not really related to the change, but is it possible to make the commit messages more descriptive? It's fine to reference the issue, but it helps a lot when trying to view the context of changes without having to refer back to github

jtanx:
>Oh good grief of course fontforgeexe/pythonui.c has completely different string handing conventions than fontforge/python.c.

ahah, the dreaded `unichar_t`, that's another legacy where we would have been better off with utf-8 representations throughout

ctrlcctrlv:
> Not really related to the change, but is it possible to make the commit messages more descriptive?

Oh, I am so sorry about that. I just opened a pull request with much the same lack of descriptiveness as this one, sorry, I hadn't seen this request.

I will in future up my descriptiveness to the benefit of all contributors. :)

skef:
Ah, so the documentation is broken as of 2012 and the `pythonui.c` is a red herring. 

@ctrlcctrlv : Say you wanted to add a keyboard shortcut for "Delete and Reload Features" in `featuremunge.py`. to do that you could add a line like this into `~/.config/fontforge/hotkeys`:
```
FontView.Menu.Tools.Delete and Reload Features:Shift+F7
```
However, make sure fontforge isn't running at the time, because that file seems to be zeroed out under some circumstances.

ctrlcctrlv:
`featuremunge.py`?

frank-trampe:
Oh boy. I've missed a lot here. I know that the UFO way is to have the feature file as an independent artwork, but the FontForge philosophy (not to be dogmatic) is to make SFD all-encompassing. When there is a problem there, the solution is usually to extend SFD rather than to add options for working around it. What are the specific limitations that motivated this work?

skef:
@ctrlcctrlv it's in the zip archive I linked above: https://github.com/fontforge/fontforge/pull/3596#issuecomment-477935205

skef:
@frank-trampe: There's no reason FontForge can't support multiple models. 

In terms of specific inherent limitations, features are often common across families, so it makes more sense to maintain a single GSUB source (less so for GPOS in most cases). There's a command to load features from one file into another, but that amounts to the same workflow implemented a different way. 

In terms of specific interface limitations, you still currently can't reorder lookups. Control over ordering in general is very primitive.

ctrlcctrlv:
@frank-trampe :

The use of feature files works around certain UI elements, which on their surface seem to be a sane way of adding OpenType features, but which after repeated use of and experience with, reveal themselves to be either counter-intuitive, click-heavy, incomprehensible, or all three. 

The most easily derided OpenType UI are the substitution dialogs, which have their own domain specific language, and seem to be in many ways an abstraction over OpenType...a simple `sub f 'i by i.alt` would take much more time to add through the UI.

Once the feature files become the canonical source for your font's substitutions, gradually you start experimenting with adding anchors via feature files, kerns via feature files, and of course named stylistic sets.

So&hellip;because for many of our fonts the feature files are a canonical source of truth, we don't apply them and then save the SFD with them applied. For one thing, that could cause corruption due to unforeseen bugs in complex OpenType features; for another, if we want to modify the feature, the vehicle for that will be the feature file, so why save it baked in?

Thus, and I speak here not only for myself, but many other FontForge users, we apply an SFD / FEA separation. The feature file is only applied (a) when it's needed for testing in e.g. the metrics view, and (b) right before generating the font. In both cases it's not saved in.

My commit simply brings this _de facto_ state of affairs for many fonts into the source code&hellip;allowing you to automatically load a feature file when opening a font for either purpose (a) or purpose (b).

@skef:

Wow, thank you so much! I'm sorry I missed this very helpful contribution, I've tried it and it works quite well I must say. It's 9PM here so I won't yet decide whether or not I want to close this issue, some part of me still believes that there should be some acknowledgement of this way of doing things in the actual source, whether it's my C code changes or your `featuremunge`. If we can have a collab thing no one uses and the workings of which are completely mysterious ( #3431 ), I don't see why we can't acknowledge the validity of SFD / FEA separation.

skef:
@ctrlcctrlv Regardless of what happens with this push request, you and I should cooperate on a (modest) tool set for dealing with externally defined features via the Tools menu. Now that the python interface is more stable it's time for some helper "extensions", and this can be one of them. (Spline helper tools is another.)

frank-trampe:
Okay. That makes sense. If the feature file is the most natural way to enter the information, and SFD is at least preserving it correctly, that's fine.

It's probably worth stewing on how best to make user scripts like featuremunge.py accessible to as many users as possible without causing unexpected behavior. One thought that occurs to me is creating a linked list for "arbitrary" preferences, loaded from the configuration file and accessible via GetPrefs, such that user scripts can have preferences. We could then ship featuremunge by default and just have the user enable automatic fea loading in his configuration file, for example.

skef:
Something like that could work (once the script evolves a bit -- there need to be better search conventions for the feature file name, for example). 

However, there should be an effective way for FontForge users to share tools without going through the organization, and therefore an effective way to add them easily. Most python packages are shared with PyPI (and such). If our init system were enhanced to "discover" scripts that follow a naming convention [along these lines](https://packaging.python.org/guides/creating-and-discovering-plugins/), users could install a module the normal way and it would work with FontForge without any centralized management. (It could also be possible for a module to (in some way) "register" for the preferences it wants.)

ctrlcctrlv:
Something like [WeeChat's (predominately Python) script system](https://weechat.org/scripts/)?

ctrlcctrlv:
I'm considering closing this and opening a new PR which would put `featuremunge.py` in `$HOME/.config/fontforge/python/` by default. What do you think @skef?

skef:
First, nothing should install in a user's home directory.I believe the system-wide equivalent (Arch) linux is `/usr/share/fontforge/python`. 

Second, this script isn't ready yet because we haven't generalized the feature file name to more cases. There should be a set of names searched for a given font from the most specific to the most general. 

In a different, still not-quite-there tool I wrote to organize the display un-unicoded glyphs, I used the equivalent this heuristic:

* The Font's filename, with `.fea` instead of `.sfd`. 
* The Font Name, as specified in the `FontName` header, followed by `.fea`.
* The Font Name with spaces replaced by hyphens, by underscores, or omitted, followed by `.fea`
* The same as above but with the Full Name, as specified in the `FullName` header.
* The same but with the Family name, as specified in the `FamilyName` header.
* `features.fea`

It would also be nice to be able to configure a name somehow. 

As for whether this should go into contrib, to me that seems like sending the wrong message. Reading through the more recent history of FontForge there a fairly consistent mix of advice to build one's own features using scripts, and worries over the programs's extensiblity. Why not make "@ctrlcctrlv's and @skef's bundle of useful FontForge stuff", and invest whatever FF development time on making sure it's easy to install and use. Then anyone else who wants to make a bundle of useful FontForge stuff can benefit as well.

davelab6:
>  There's no reason FontForge can't support multiple models.

On that basis this PR seems fine to me. 

I think python script development makes more sense, so glad to see things end up going in that direction :)

ctrlcctrlv:
Voluntarily closing this in favor of using Python scripts and @skef's "hook" philosophy.

PR #3599, opened 2019-03-29
Don't recreate the charview palettes when the charview hasn't changed

### LINKED ISSUE #3598: Losing focus when typing in glyph editor input field
One spectacular feature is being able to see glyphs next to the one you're editing in the Outline Glyph Editor window, which I've been able to easily type into in the latest public release. Such a good feature!

However, in a recent dev build (which I'll show specifically below), after I mouse-click into the input field above the glyph outline editor pane, it immediately loses focus and I can't type anything at all. The only thing I've gotten to work is by pressing + holding the mouse button and typing. However, this only allows me to input one character at a time. Arduous, but still (kind of) usable. The problem always occurs, regardless of the glyph & font being edited.

OS: Windows 8 & 10.

Here's the input field I'm talking about:
glyph-outline-editor-input-field-problem

Here's the good version of FF:
ff-good-version

Here's the bad version of FF that doesn't allow typing:
ff-bad-version

jtanx:
Fixes #3598, introduced from #3398

### DIFF
--- a/fontforgeexe/cvpalettes.c
+++ b/fontforgeexe/cvpalettes.c
@@ -3829,7 +3829,7 @@ void _CVPaletteActivate(CharView *cv, int force, int docking_changed) {
 		SaveOffsets(old->gw,cvlayers,&cvlayersoff);
 	}
 
-    if (palettes_docked || docking_changed) {
+    if ((palettes_docked && old != cv) || docking_changed) {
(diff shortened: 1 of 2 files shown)


ctrlcctrlv:
Wow, awesome! I've confirmed this works! Another papercut fixed this month.

PR #3605, opened 2019-03-30
Bring hotkey strings in line with visible menu strings

### LINKED ISSUE #3604: FontForge deletes hotkeys if no font opened; otherwise ignores them
I have the following in `/home/fred/.config/fontforge/hotkeys`:
```XResources
fontforge.CharView.Hotkey.Tool.SpiroG2:1
fontforge.CharView.Hotkey.Tool.SpiroG4:2
fontforge.CharView.Hotkey.Tool.SpiroCorner:3
fontforge.CharView.Hotkey.Tool.SpiroLeft:4
fontforge.CharView.Hotkey.Tool.SpiroRight:5
```

If I run `fontforge -open` and immediately close it, it deletes the file!

Even if I do open a font, these hotkeys don't work. I've tried as well `fontforge.CharView.Menu.Point.Tools.G2` and `CharView.Menu.Point.Tools.G2`, mentioned in `fontforge/share/default-n`, and those don't work either.

skef:
This update brings the hotkey strings in line with the menu strings. Some of the more important entries in the `default` hotkey file already had these values (and therefore didn't work before). Others have been changed. 

Another option is to eliminate the `defaults` edits and just bring the strings in line with it. However, I think it is preferable to be able to "read off" of the needed string from the menu itself. The vast majority of the strings are already like this. 

This is "breaking" in the sense that people might have files with custom hotkeys that have now changed. However, the edited keys are (for the most part) not very likely to have been customized. It also changes the interface in the sense that it enables some hotkeys that didn't work before, mostly in the Point -> Tools menu. 

I'm also disabling saving of the user's hotkey file, because that process is prone to zeroing out the file in some cases, and there doesn't seem to be a good enough reason for it; it just saves what is marked as `isUserDefined` -- a value set when loading entries from that file. (I'm happy to remove that commit on request/explanation.) 

Closes #3604 

### Types of changes

### Final checklist

### DIFF
--- a/share/default
+++ b/share/default
@@ -89,7 +89,7 @@ CharView.Menu.Point.Tools.Corner:3
 CharView.Menu.Point.Tools.Tangent:4
 CharView.Menu.Point.Tools.HVCurve:5
 
-CharView.Menu.Point.Tools.Toggle Spiro:S
+CharView.Menu.Point.Tools.Activate Spiro:S
 +CharView.Menu.Point.Tools.G2:1
 +CharView.Menu.Point.Tools.G4:2
 +CharView.Menu.Point.Tools.Corner:3
--- a/share/default
+++ b/share/default
@@ -73,7 +73,7 @@ CharView.Menu.Edit.Select.Width:No Shortcut
 CharView.Menu.Edit.Select.VWidth:No Shortcut
 CharView.Menu.Edit.Select.Select Points Affected by HM:No Shortcut
 CharView.Menu.Edit.Unlink Reference:Ctrl+U
-CharView.Menu.Edit.Remove Undoes:No Shortcut
+CharView.Menu.Edit.Remove Undoes...:No Shortcut
 
 
 CharView.Menu.Point.Tools.Ruler:R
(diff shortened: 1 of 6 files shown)


ctrlcctrlv:
Fantastic work!&mdash;I **approve** of this pull, and especially invite the change made to ` fontforgeexe/startui.c ` as necessary.

ctrlcctrlv:
@skef I've just noticed something. Is there a reason you've decided to give the Spiro points `Ctrl+1`, `Ctrl+2` etc rather than `1`, `2` etc., other than that's how it was in the broken original?

skef:
I don't believe those are Control-modified -- it's just the bare characters. And no, it's just the consequence of the assigned hotkeys working now.

share/default
```diff
@@ -105,8 +105,8 @@ CharView.Menu.Point.HVCurve:Ctrl+5
 +CharView.Menu.Point.G2 Curve:Ctrl+1
 +CharView.Menu.Point.G4 Curve:Ctrl+2
 +CharView.Menu.Point.Corner:Ctrl+3
-+CharView.Menu.Point.Prev Constraint:Ctrl+4
-+CharView.Menu.Point.Next Constraint:Ctrl+5
```
ctrlcctrlv: It is not the bare numbers @skef&hellip;

skef: Those are different controls directly on the Point menu. They convert selected points between those types. The tools submenu switches between palette tools (a more common operation, so a sensibly simpler key combo).

skef:
(De)activate Spiro was still wrong, so I made it a bare "S". I imagine that might piss off some people, but round-trips through Sprio that don't modify are "safe", right?

ctrlcctrlv:
I use zero, but yes they are safe.

skef:
Put this on hold for now.

ctrlcctrlv:
Why? What's wrong with it?

skef:
Nothing; it's just probably going to be superseded.

jtanx:
Generally looks ok to me. Was it ever possible to edit hotkeys interactively? Otherwise I'm not sure either why there's a need to save the hotkeys file

skef:
This is probably best superseded by #3637, but in case that stalls out let's leave it open for a bit.

davelab6:
> Was it ever possible to edit hotkeys interactively?

I don't think so.

skef:
> This is probably best superseded by #3637, but in case that stalls out let's leave it open for a bit.

Prescience!

PR #3608, opened 2019-03-30
Fix Coverity reports

serval2412:
1083078
1083627
1082959
1082960
1082963
1082964
1082965
1082966

### Important

### Motivation and Context

### Types of changes

### Description

### Final checklist

### DIFF
--- a/fontforge/lookups.c
+++ b/fontforge/lookups.c
@@ -5041,6 +5041,10 @@ return( xasprintf( _("%s is not a class name for the forward classes." ), parsed
 	}
       break;
       default:
+	  for ( i=0; i<cnt; i++ ) {
+	      free( parsed[i].entity );
+	      free( parsed[i].replacements );
+	  }
 return( copy( _("Bad FPST format")) );
     }
     if ( fpst->format!=pst_reversecoverage ) {
--- a/fontforge/lookups.c
+++ b/fontforge/lookups.c
@@ -5041,6 +5041,9 @@ return( xasprintf( _("%s is not a class name for the forward classes." ), parsed
 	}
       break;
       default:
+        for ( i=0; i<cnt; ++i )
+	    free( parsed[i].entity );
+        free(parsed);
 return( copy( _("Bad FPST format")) );
     }
     if ( fpst->format!=pst_reversecoverage ) {
--- a/fontforge/lookups.c
+++ b/fontforge/lookups.c
@@ -5041,9 +5041,6 @@ return( xasprintf( _("%s is not a class name for the forward classes." ), parsed
 	}
       break;
       default:
-        for ( i=0; i<cnt; ++i )
-	    free( parsed[i].entity );
-        free(parsed);
 return( copy( _("Bad FPST format")) );
     }
     if ( fpst->format!=pst_reversecoverage ) {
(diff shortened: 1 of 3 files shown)


fontforge/lookups.c
```diff
@@ -5041,6 +5041,9 @@ return( xasprintf( _("%s is not a class name for the forward classes." ), parsed
 	}
       break;
       default:
+        for ( i=0; i<cnt; ++i )
+	    free( parsed[i].entity );
```
skef: One would have to read the algorithm above very closely to know whether an `entity` could wind up NON-NULL at this point, so freeing them here makes sense, given that whenever one is "used" at the end the pointer is set to NULL. 

Similarly, one would have to read the algorithm above very closely to know whether a `replacements` could wind up non-NULL at this point, so freeing them makes just as much sense (to me), also given that whenever one is "used" the pointer is set to NULL. 

So either you've got a better understanding of the algorithm (which is quite possible) or you could also free `replacements` here).

serval2412: Thank you for your feedback, I reverted this part.

skef: My comment was probably too rhetorical. It should be fine to just change it to:
```
for ( i=0; i<cnt; i++ ) {
    free( parsed[i].entity );
    free( parsed[i].replacements );
}
```
I was just nattering on about why you might see one of those lines without the other. But even if there are no `replacements` to free because of how the loop above works, this will be fine -- they'll all be NULL).

ctrlcctrlv:
I'm sorry, I have just another question about Coverity friends, I hope it will be more educated than my concern that it was something like Coveralls, when it has no relation to that at all. Egg met face.

So, in looking at the Coverity website, it seems like it also automatically finds segmentation faults, not just leaks.

Is there a reason that you mostly focus on leaks and not crashes reported by Coverity?

fontforge/lookups.c
serval2412: No pb, as I said I'm just trying to fix low hanging fruits, there are already some Coverity reports about Resource leak I put aside because it needs a real understanding of the code.

serval2412:
@ctrlcctrlv : could you give an example of cid about a segfault so I can give it a look?
Most of Ressource leak reports seem easy to fix, just add some free or fclose at the right location. So for someone like me who doesn't know anything about Fontforge and isn't expert in C, it's quite practical.

ctrlcctrlv:
@serval2412 As I understand it, the faults in categories "Out-of-bounds read" such as 1083341, "Out-of- bounds access" such as 1083325, and all the issues in the "Null pointer dereference" supercategory can cause segmentation faults.

I'm basing my comments also on this chart:
npd

When I look at  the files that these faults were automatically found in, it makes me wonder if fixing them could actually have a big benefit on users by avoiding real world possible crashes.

skef:
I won't speak for @serval2412 but fixing a typical SEGV requires more domain-specific knowledge than fixing a memory leak. With leaks the goal is to match up allocations and deallocations. A SEGV is often due to something returning NULL, bad pointer arithmetic, or bad bounds checking. Knowing that there is a problem doesn't necessarily tell you much about how to fix it.

Issue #3615, opened 2019-04-01
COMBINING CEDILLA (U+0327) not considered by "Build Accented Glyph"


ctrlcctrlv:
Arch Linux, https://github.com/fontforge/fontforge/commit/3e641ec84d141d291aa75bd84897aee04bfe5428

FontForge prioritizes `comma` (,) over what I affectionately term `cedillacomb`, that is, COMBINING CEDILLA (U+0327), ◌̧ .

In fact, `cedillacomb` seemingly isn't considered at all.

Not sure how to fix this, but will look into it.

Edit: Seems like there are some data files used by `SFIsSomethingBuildable` and `SCBuildComposit` I have to edit.

ctrlcctrlv:
OK I found the problem. It's this bizarre function, SFGetAlternate. Including comments like:

https://github.com/fontforge/fontforge/blob/3e641ec84d141d291aa75bd84897aee04bfe5428/fontforge/fvcomposite.c#L1198

Is this really valid? To use George Williams' interpretation of the appearance of glyphs in a font over what the standard defines? I think most of this function should be gutted, but welcome comments from other esteemed contributors as always.

JoesCat:
Might have been true for unicode 2 or 3. Charts get refined with time. Old errors out, new ones in.
I had to override an error in unicode 11.0 - Unicode/makeutype.c:798 - capital ß, and this is in 2018

It may be a case that some stuff got corrected and refined with time, so some code may indeed be removed and allow the unicode chart values to be the ones coming through instead.
You may want to gut this in small chunks instead of one big sweeping removal. The smaller chunks are easier to see, validate.

ctrlcctrlv:
With the benefit of a good night's sleep, I think what I'll attempt is to write a PR which will add an option, "UnicodeSupremacy", to optionally bypass SFGetAlternate at first and perhaps other functions in the future.

PR #3617, opened 2019-04-02
Add "UnicodeStrict" option

ctrlcctrlv:
In many places in the code, George has put his own opinions on the
appearance of Unicode characters above the Unicode standard itself. He
may have done so rightly in some cases, but the problem is that he was
wrong in others (at least under the current version of the Unicode
Standard).

This PR adds a new option, "UnicodeSupremacy", which users to use the
standard above George's opinions when building characters via "Build
Accented Glyph"—there are likely more places it could be added, but this
is a start.

Close fontforge/fontforge#3615

Video showing what this does: [USO.zip](https://github.com/fontforge/fontforge/files/3032714/USO.zip)

### DIFF
--- a/fontforgeexe/prefs.c
+++ b/fontforgeexe/prefs.c
@@ -72,6 +72,7 @@ extern Encoding *default_encoding;
 extern int autohint_before_generate;
 extern int use_freetype_to_rasterize_fv;
 extern int use_freetype_with_aa_fill_cv;
+extern int UnicodeStrict;
 extern int OpenCharsInNewWindow;
 extern int ItalicConstrained;
 extern int accent_offset;
@@ -313,6 +314,7 @@ static struct prefs_list {
 	{ N_("AutoSaveFrequency"), pr_int, &AutoSaveFrequency, NULL, NULL, '\0', NULL, 0, N_( "The number of seconds between autosaves. If you set this to 0 there will be no autosaves.") },
 	{ N_("RevisionsToRetain"), pr_int, &prefRevisionsToRetain, NULL, NULL, '\0', NULL, 0, N_( "When Saving, keep this number of previous versions of the file. file.sfd-01 will be the last saved file, file.sfd-02 will be the file saved before that, and so on. If you set this to 0 then no revisions will be retained.") },
 	{ N_("UndoRedoLimitToSave"), pr_int, &UndoRedoLimitToSave, NULL, NULL, '\0', NULL, 0, N_( "The number of undo and redo operations which will be saved in sfd files.\nIf you set this to 0 undo/redo information is not saved to sfd files.\nIf set to -1 then all available undo/redo information is saved without limit.") },
+	{ N_("UnicodeStrict"), pr_bool, &UnicodeStrict, NULL, NULL, '\0', NULL, 0, N_( "Whether or not the Unicode Standard should be strictly adhered to even in cases its contents may be dubious, controversial, or even wrong.\nThis currently only changes certain build procedures when using “Make Accented Glyph”, but can likely be extended to other parts of FontForge.") },
 	PREFS_LIST_EMPTY
 },
   new_list[] = {
--- a/fontforge/fvcomposite.c
+++ b/fontforge/fvcomposite.c
@@ -44,6 +44,7 @@
 
 int accent_offset = 6;
 int GraveAcuteCenterBottom = 1;
+int UnicodeStrict = false;
 int PreferSpacingAccents = true;
 int CharCenterHighest = 1;
 
@@ -1155,6 +1156,10 @@ return( adobes_pua_alts[base-0xf600]);
     if ( base==-1 || base>=65536 || unicode_alternates[base>>8]==NULL ||
 	    (upt = unicode_alternates[base>>8][base&0xff])==NULL )
 return( SFAlternateFromLigature(sf,base,sc));
+
+    if (UnicodeStrict) {
+        return upt;
+    }
 
 	    /* The definitions of some of the greek letters may make some */
 	    /*  linguistic sense, but I can't use it to place the accents */


fontforgeexe/prefs.c
```diff
@@ -313,6 +314,7 @@ static struct prefs_list {
 	{ N_("AutoSaveFrequency"), pr_int, &AutoSaveFrequency, NULL, NULL, '\0', NULL, 0, N_( "The number of seconds between autosaves. If you set this to 0 there will be no autosaves.") },
 	{ N_("RevisionsToRetain"), pr_int, &prefRevisionsToRetain, NULL, NULL, '\0', NULL, 0, N_( "When Saving, keep this number of previous versions of the file. file.sfd-01 will be the last saved file, file.sfd-02 will be the file saved before that, and so on. If you set this to 0 then no revisions will be retained.") },
 	{ N_("UndoRedoLimitToSave"), pr_int, &UndoRedoLimitToSave, NULL, NULL, '\0', NULL, 0, N_( "The number of undo and redo operations which will be saved in sfd files.\nIf you set this to 0 undo/redo information is not saved to sfd files.\nIf set to -1 then all available undo/redo information is saved without limit.") },
+	{ N_("UnicodeSupremacy"), pr_bool, &UnicodeSupremacy, NULL, NULL, '\0', NULL, 0, N_( "Whether or not the Unicode Standard should have supremacy even in cases its contents may be dubious, controversial, or even wrong.\nThis currently only changes certain build procedures when using “Make Accented Glyph”.") },
 	PREFS_LIST_EMPTY
```
JoesCat: "Supremacy" is probably the wrong impression here, plus the word does conjure up negative impressions of racism. What you probably want to say here is have it default to Unicode standards, and enable backwards compatibility with George's edits for users that want to maintain backwards compatibility for these glyphs. You also want to indicate what you are modifying (like in this case it's combiners I think).
Maybe OldCombiners, "Enable backwards compatible combiner codes for users maintaining older fonts"
...probably not the best choice of words, but suggests we're defaulting to unicode values, and giving users a switch to enable older values if they need it.
I would prefer use pr_int instead of pr_bool for consistency, plus ints are able to convey error codes while bools cannot.
...just my 2 cents review

dscorbett:
That U+0327 COMBINING CEDILLA should sometimes look like a comma is not merely one developer’s opinion. See [the sections “Variation in Diacritical Marks” and “Latvian Cedilla” of chapter 7 of <i>The Unicode Standard</i>](https://www.unicode.org/versions/Unicode12.0.0/ch07.pdf#page=5) and [the Latin Extended-A code chart](https://www.unicode.org/charts/PDF/U0100.pdf). An option to trust Unicode’s decompositions uncritically, against Unicode’s own advice, does not seem useful. Why not just fix the specific cases where George was wrong?

JoesCat:
That is excellent advice plus reference too. Thanks for correcting me.

ctrlcctrlv:
@dscorbett According to the same document you cited:

>  U+010F and U+0165 are *often* rendered

> In  typewritten  or  handwritten  documents,  or  in  didactic  and  pedagogical  material, **glyphs with háčeks [i.e. carons] are preferred**

It doesn't say that they are **always** rendered, and my option is certainly not "against Unicode's own advice". When making a handwriting or typewriter ("monospace") font, it rather says that the compositions with commas/apostraphes are simply _wrong_!

I admit that you may have to change this option depending on the type of font you're working on.

skef:
@dscorbett 's comment, it seems fine to have some preferred work-arounds. On the other hand I'm dubious on the idea of there being a) a single right answer that b) differs from the Unicode spec. It raises the "who the hell are you?" question.  So I think the flag as @ctrlcctrlv has coded it is reasonable. 

A word just as colorful, with significantly less baggage, and probably a little bit more suited to the case is "sovereignty"

ctrlcctrlv:
I must disagree. There is nothing negative or racist about the term supremacy, and it is often used in legal language. For example, you can say that _parliament has supremacy over the king_ or _the will of the dictator is supreme over his electorate_ or even _in our country's government the supremacy of democracy shall not be infringed_. Supremacy is the right word to use, "sovereignty" makes no sense. Unicode isn't more or less sovereign if the option is enabled, both it and FontForge can be sovereign at the same time. No, rather the sovereignty of Unicode is _supreme_ over that of FontForge if it's enabled.

ctrlcctrlv:
Because there is no single right answer is why I've coded it as I have. If you don't want Unicode's decompositions to be, dare I say it lest I be accused of being a racist, *SUPREME*, then enable it. If not, don't.

image

skef:
I'll defer on the "suited to the case" question. The other issue isn't whether the term is racist but with the associations some people have with the word. There's no reason to raise those associations when they can be avoided. (Or rather the reason "those people are wrong and this word is the most appropriate because X, Y, Z" has its *own* associations, and is a hill whose defense is better tackled by individuals than by community projects.)

ctrlcctrlv:
OK, I find no reason to argue the matter, so I'll think of a better option name.

BTW, I have found more proof...

2019-04-03-080135_3172x1724_scrot

I didn't make this font, I found it on [some website](http://fonty.condak.cz/detail.php?font=typechev&lang=2) in Czech.

ctrlcctrlv:
How about `MakeUnicodeSupreme`? It's the same root as _supremacy_ but you really can't think _supreme_ is also racist, right? I mean it's used all the time in advertising and (I believe, I don't drive) is a type of gasoline.

skef:
What about just "UnicodeStrict"?

ctrlcctrlv:
Oh, I like that more than `Supremacy`. I'll update it.

ctrlcctrlv:
Alright, it's changed! Sorry, my first force-push was done incorrectly.

ctrlcctrlv:
I just wanted to say to the reviewers, @skef and @JoesCat, that I apologize for my initial choice of word. I understand that many different people use FontForge, with all levels of English ability, and all levels of exposure to racism and discrimination, and that even though there's nothing inherent to the word _supremacy_ that makes it racist, some people might not like to see it in any context due to their experiences. I should not have chosen it.

skef:
I wasn't offended, and I don't personally think you did anything wrong or that anything bad has happened here. I don't even think someone who uses this or that term *specifically to invite that kind of argument* (which I'm not at all saying you had in mind) would *necessarily* be wrong, although I'm wary of people looking for arguments. 

This is more about prudence than anything else: If the issue is there, and it's easy to work around, you might as well.

JoesCat:
@ctrlcctrlv - don't take it personally, some phrases are regional, and although there is nothing wrong with the word supremacy, looking at the recent problems we've had world-wide, it's simply one of them words that maybe/probably could be substituted. For example we say car trunk, while the British refer to the boot, and we've seen words that were positive in meaning take on entirely different meaning, for example, there's a difference between a hacker and a cracker, but now we frown on anyone calling themselves hackers.
Looking at the function, I initially thought it was more of, let's depricate the existing method and move on to unicode, but as @dscorbett points-out, there is reason for it being the way it is/was, so now it makes more sense instead of depricating, to make it either/or, and @skef suggestion is a really good one that conveys the meaning == either/or

ctrlcctrlv:
Yes, I acknowledge that for some (most?) fonts it has a reason to be the way it is. I think that the "correct" OpenType response is to put a substitution in for `D caroncomb` to `D cedillacomb`...but am not sure. The `ccmp` table ideally has something to do with this, right?

But for now, I wrote this in my description of the option:

> Whether or not the Unicode Standard should be strictly adhered to **even in cases its contents may be dubious, controversial, or even wrong.**

I think that's enough to get this over the line and into master. I'm not saying using Unicode is preferable or correct in all cases. But it's definitely needed for some glyphs in some cases, and may even be enabled/disabled multiple times while developing the same font.

JoesCat:
> I think that's enough to get this over the line and into master.

Almost there.... :-)
Understandably, comments in code is one thing, but we should try keep it technically neutral when we get to the GUI. There are going to be a lot of users that look at this and have no idea about PR#3617.
Take a couple days to look at something else, then come back and look at this, you may say, yeah, perhaps the same.  For myself, I'm holding back from, Let's Get This Merged" due to this part...even in cases its contents may be dubious, controversial, or even wrong.
My thought is needs more toning down, be more dry technical lingo....
Take a couple days, and look back, happens to all of us... :-)

ctrlcctrlv:
How about this tooltip:

"When building accented glyphs, FontForge has two options: (a) it can either build the glyph solely based on its Unicode decompositon, or (b) it can build the glyph based on how its representation looks in most fonts. For example, "dcaron", LATIN LETTER SMALL D WITH CARON, is commonly, but not always, represented as a d with an apostrophe rather than a d with a caron. If this option is enabled, FontForge will always go with (a), otherwise, it will go with (b) whenever possible."

dscorbett:
>_There is no right answer_

That’s true for the Latin characters affected by this function. Can you confirm it is true for the characters in other scripts?

ctrlcctrlv:
I don't see why I should have to confirm that there is a font style in which the Unicode decomposition is preferable to the overridden form for every single character affected. 

In an attempt to try to answer the question, I would say that for, e.g. Greek, there are indeed many fonts, mostly of the display or handwritten variety, including ones I've authored, where there's no difference between the tonos and the acute accent, and that using the acute accent in the building of Greek characters is the prudent thing to do.

However I would say to @dscorbett finally that I should have no obligation to prove to him or anyone that for every single character this change affects there is a reason someone would want to build it with the Unicode decomposition rather than George's decomposition. That bar is ridiculously high.

JoesCat:
Hi @dscorbett - do you think it is a worthwhile option to add?
Can you recommend something that says what needs to be said?
The shorter the better, is informative, and doesn't convey opinion one way or another.
No rush, take your time, and hopefully something everyone can agree with - thanks.

dscorbett:
This option would be useful for some characters. It might not be useful for all affected characters, which is why I asked about them. I don’t think a single option to prefer Unicode’s decompositions is good UI design, because (for example) whether U+0123 LATIN SMALL LETTER G WITH CEDILLA appears to have a cedilla is not related to whether the tonos looks like an oxeia, or to whatever effect this option has on Arabic and Hangul. Such an option requires the user to know details of NFD that might not be fair to assume.

The following is brief and neutral.
>Use Unicode canonical decompositions when building accented
>and composite glyphs, rather than the representations used in
>most fonts.

Unfortunately, it is not true. There are other places in FontForge that override Unicode when building composite glyphs. For example, even with this option enabled, FontForge still tries to build U+00EC LATIN SMALL LETTER I WITH GRAVE using U+0131 LATIN SMALL LETTER DOTLESS I. Without fully knowing in which user-visible ways this option prefers Unicode and in which it doesn’t, though, I can’t propose a more accurate tooltip.

ctrlcctrlv:
Thanks for the comment. I hadn't even considered the `dotlessi` thing. The long discussions we've had on this topic make me think there might be a better way to do it, which would satisfy all. It's harder to implement, but I need a challenge.

I'll remove UnicodeStrict and make it so you can set the composition used for building the glyph in the "Glyph Info" dialog on a glyph by glyph basis, with it defaulting to how things are done now.

Does this satisfy all parties?

JoesCat:
I'm ok. It was a learning experience for everyone. Thanks for your feedback @dscorbett

ctrlcctrlv:
Yeah, and I'm going to request that this remain _open_, because while I am working on it and have gotten some done:

```diff
diff --git a/fontforgeexe/charinfo.c b/fontforgeexe/charinfo.c
index 498d64a7c..89d136043 100644
--- a/fontforgeexe/charinfo.c
+++ b/fontforgeexe/charinfo.c
@@ -3918,10 +3918,19 @@ static void CIFillup(CharInfo *ci) {
        ubuf[0] = '\0';
        GGadgetSetTitle(GWidgetGetControl(ci->gw,CID_Components),ubuf);
     } else {
-       unichar_t *temp = malloc(11*u_strlen(bits)*sizeof(unichar_t));
+       unichar_t *temp = malloc(18*u_strlen(bits)*sizeof(unichar_t));
        unichar_t *upt=temp;
        while ( *bits!='\0' ) {
-           sprintf(buffer, "U+%04x ", *bits );
+           sprintf(buffer, "U+%04x (", *bits );
+           uc_strcpy(upt,buffer);
+           upt += u_strlen(upt);
+           if (iscombining(*bits)) {
+               upt += 1;
+               upt[-1] = 0x25CC; // DOTTED CIRCLE ◌
+           }
+           upt += 1;
+           upt[-1] = *bits;
+           sprintf(buffer, ") ");
            uc_strcpy(upt,buffer);
            upt += u_strlen(upt);
            ++bits;
```

This is a big challenge (for me) and some parts are proving much harder (for me) than I anticipated. If this is merged and I do get a better solution in, I'll just revert that in my PR.

What I would say is, I still think this should be merged, _as to not make the perfect the enemy of the good (enough)._

ctrlcctrlv:
This took me entirely too long but I've learned a lot so far. I am so happy to have gotten this far, I considered giving up numerous times. 

[w00t2.zip](https://github.com/fontforge/fontforge/files/3053898/w00t2.zip) (video)

image

ctrlcctrlv:
SUPERSEDED by #3645, REDUNDANT.

Issue #3553, opened 2019-03-17
Native scripting language crashes FontForge


dscorbett:
Using the native scripting language crashes FontForge. To reproduce, open the “File” menu, select “Execute Script”, select “FF”, and click “Call...”. I am using FontForge 20190317 on macOS Mojave 10.14.3.

skef:
This isn't reproducing for me. Any chance of a stacktrace?

dscorbett:
[Here is the crash report.](https://github.com/fontforge/fontforge/files/2976162/fontforge_2019-03-17-223500.txt)

skef:
Would you be willing to share the file you were attempting to open? The code that crashed is trying to determine its MIME type, and while that *may* not matter, it could be an interaction between the contents of your script, `g_content_type_guess()` (not our code) and how we call that function.

dscorbett:
It was a new unsaved font without any changes, i.e. what you get by just starting FontForge. The script itself was empty.

skef:
So to be clear, you open a new font, pick File -> Execute Script, press the FF radio button (if necessary), and then what? The trace implies a file was being opened, perhaps with the "Call..." button. If the file was empty did it have a name?

ctrlcctrlv:
Is the script empty as in zero-bytes, or empty in another sense (only whitespace, only BOM, etc)? You might want to upload an archive of the script if you don't know how to figure out which.

dscorbett:
I open a new font, pick “File”, pick “Execute Script”, select the “FF” radio button, and click the “Call...” button. FontForge immediately crashes. The script is empty as in I didn’t type anything into the text box, but it doesn’t matter: if I do type something into the text box, it crashes with the same stack trace.

What file are you referring to that might have a name? There is no file. There is only a text box.

skef:
I'm not really familiar with this area of the code, so I'm trying to nail things down. In fact, I should have deduced this answer as `GFileChooserScanDir()` is in the stacktrace.

I'm afraid I'm going to need some help to have any hope of debugging this. Would it be possible to:
1. Run FontForge
2. Get up to right before the error happens (with the script dialog window open)
3. Open a terminal
4. In that terminal type `sudo dtruss -n fontforge` and authenticate yourself for `sudo`. (This will work if you're logged into an account with Administrator privileges.) 
5. Press the button to make FontForge crash
6. Pack up and attach the output of `dtruss`
?

JoesCat:
[...]

PR #3622, opened 2019-04-03
g_giomime.c: Don't call into g_content_type_guess with an empty buffer

jtanx:
To work around #3553 and #3609


JoesCat:
Thanks for taking care of this, looked like easy ones that could have been handled earlier.

Issue #3613, opened 2019-04-01
Crash when opening auto-save file


ctrlcctrlv:
Arch Linux, https://github.com/fontforge/fontforge/commit/3e641ec84d141d291aa75bd84897aee04bfe5428

```
Recovering from /home/fred/.config/fontforge/autosave/auto007231-1.asfd... Recovering from /home/fred/.config/fontforge/autosave/auto00195a-11.asfd...  Done
17:42:6.844 WARNING: _GGDKDraw_CreateWindow (ggdkdraw.c:601) - Could not call GGDKDrawSetTransientFor on a restricted window (0x555555aaf9b0 Loading...), restrict count 0, transient window count 0
```

```
Program received signal SIGSEGV, Segmentation fault.
CVPaletteCheck (cv=0x15554a338010) at cvpalettes.c:3761
3761        if ( cv->b.sc->parent->multilayer && cvlayers2==NULL ) {
```

```
#0  0x000015555520d72e in CVPaletteCheck (cv=0x15554a338010) at cvpalettes.c:3761
#1  0x000015555520da34 in _CVPaletteActivate (cv=0x15554a338010, force=0, docking_changed=0) at cvpalettes.c:3809
#2  0x000015555520e093 in CVPaletteActivate (cv=0x15554a338010) at cvpalettes.c:3905
#3  0x000015555519d39e in cv_e_h (gw=0x555555d2dce0, event=0x7fffffffafd0) at charview.c:6326
#4  0x0000155555014966 in _GWidget_Container_eh (gw=0x555555d2dce0, event=0x7fffffffafd0) at gcontainer.c:403
#5  0x0000155555015f1e in _GWidget_TopLevel_eh (gw=0x555555d2dce0, event=0x7fffffffafd0) at gcontainer.c:745
#6  0x0000155555090cd2 in _GGDKDraw_CallEHChecked (gw=0x555555d2dce0, event=0x7fffffffafd0, eh=0x155555015590 <_GWidget_TopLevel_eh>)
    at ggdkdraw.c:336
#7  0x00001555550934f0 in _GGDKDraw_DispatchEvent (event=0x555555da0000, data=0x55555565fcb0) at ggdkdraw.c:1211
#8  0x000015555384bc86 in  () at /usr/lib/libgdk-3.so.0
#9  0x000015555387dab5 in  () at /usr/lib/libgdk-3.so.0
[...]

Issue #3569, opened 2019-03-23
Spiro crash when merging points


ctrlcctrlv:
Arch Linux, 3aabfd2
```
Program received signal SIGABRT, Aborted.
0x00001555526f1d7f in raise () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00001555526f1d7f in raise () at /usr/lib/libc.so.6
#1  0x00001555526dc672 in abort () at /usr/lib/libc.so.6
#2  0x0000155552734878 in __libc_message () at /usr/lib/libc.so.6
#3  0x000015555273b18a in  () at /usr/lib/libc.so.6
#4  0x000015555273ca26 in _int_free () at /usr/lib/libc.so.6
#5  0x0000155554b13161 in SplineFree (spline=0x555555a520b0) at splineutil.c:159
#6  0x0000155554b13464 in SplinePointsFree (spl=0x555555c542e0) at splineutil.c:226
#7  0x0000155554b134d9 in SplineSetBeziersClear (spl=0x555555c542e0) at splineutil.c:238
#8  0x0000155554a816ec in SSRegenerateFromSpiros (spl=0x555555c542e0) at spiro.c:286
#9  0x00001555551d04ce in CVMouseMoveSpiroPoint (cv=0x555555c731a0, p=0x7fffffffba50)
    at cvaddpoints.c:702
#10 0x00001555551d0559 in CVMouseMovePoint (cv=0x555555c731a0, p=0x7fffffffba50) at cvaddpoints.c:716
#11 0x00001555551992ae in CVMouseMove (cv=0x555555c731a0, event=0x7fffffffbcf0) at charview.c:5446
#12 0x000015555519a5e1 in v_e_h (gw=0x555555cb8920, event=0x7fffffffbcf0) at charview.c:5763
#13 0x0000155554fffde6 in _GWidget_Container_eh (gw=0x555555cb8920, event=0x7fffffffbcf0)
    at gcontainer.c:403
#14 0x000015555507fd89 in dispatchEvent (gdisp=0x55555562f0c0, event=0x7fffffffbea0) at gxdraw.c:3293
#15 0x0000155555080332 in GXDrawEventLoop (gd=0x55555562f0c0) at gxdraw.c:3401
#16 0x0000155555004906 in GDrawEventLoop (gdisp=0x55555562f0c0) at gdraw.c:787
#17 0x00001555553a2282 in fontforge_main (argc=2, argv=0x7fffffffe1c8) at startui.c:1396
[...]

PR #3623, opened 2019-04-03
Coverity: several reports

serval2412:
Coverity indicates this (eg: cid#1083113)
1530  toolsfont = GDrawInstanciateFont(NULL,&rq);
8. noescape: Resource toolsfont is not freed or pointed-to in GResourceFindFont. [show details]
CID 1083113 (#1 of 1): Resource leak (RESOURCE_LEAK)9. overwrite_var: Overwriting toolsfont in toolsfont = GResourceFindFont("ToolsPalette.Font", toolsfont) leaks the storage that toolsfont points to.
1531  toolsfont = GResourceFindFont("ToolsPalette.Font",toolsfont);

Let's investigate:
GDrawInstanciateFont returns an object "toolfont" (which is the result of an allocation but can be NULL).

Then GResourceFindFont is called and one of its argument is "toolsfonts";
Here's its definitione:
FontInstance *GResourceFindFont(char *resourcename,FontInstance *deffont) {
    char *val = GResourceFindString(resourcename);
    if ( val==NULL )
return( deffont );
return( GResource_font_cvt(val,deffont));
}

so if "ressourcename" is found, val is not null and we call "GResource_font_cvt"
this function  (in ggadgets.c) has this declaration:
void *GResource_font_cvt(char *val, void *def)

if we focus on "def" var, we've got in the definition of this function this:
    if ( def!=NULL )
	GDrawDecomposeFont((FontInstance *)def, &rq);
    if ( fi==NULL )
return( def );
return( (void *) fi );

So if "fi" is not NULL, we don't return "def" (which corresponds to "toolsfont") but "fi"
and "def" is leaked since the return of the function will be assigned to "toolfonts"

### Important

### Motivation and Context

### Types of changes

### Description

### Final checklist

### DIFF
--- a/gdraw/ggadgets.c
+++ b/gdraw/ggadgets.c
@@ -255,6 +255,7 @@ void *GResource_font_cvt(char *val, void *def) {
 
     if ( fi==NULL )
 return( def );
+    free(def);
 return( (void *) fi );
 }
 


serval2412:
@skef: I took a look to Travis failing build but didn't find anything related to the patch (perhaps missed it)
Any thoughts here?

ctrlcctrlv:
@serval2412 I looked at the log and wouldn't worry about it. Our Travis CI builds are notorious for failing due to network issues...AppVeyor also routinely fails because it can't find a signing key sometimes.

skef:
The analysis of where the memory is isn't wrong, but I'm not comfortable with the change at the present time. `GResourceFindFont()` is written poorly and `GResource_font_cvt()` is called in a way I don't fully understand. I'm worried this will make a confusing function even more confusing.

skef:
@jtanx could you briefly weigh in on this? `GResourceFindFont` and `GResource_font_cvt` are called with an argument that might or might not be the return value. If it is, well -- it is. If it isn't, there are now two things instead of one from a `malloc` perspective. 

Callers mostly call like `foo = GResourceFindFont(bar, foo)`, and therefore one could (as this push request does) handle the difference internally by freeing the parameter `foo` when its different. But with functions that work like this it's better to use a double pointer with the argument rather than returning a value. I could ask @serval2412 to look into making that change, but `GResource_font_cvt` is called in some weird ways in `ggadget` I'm not comfortable with.

jtanx:
I agree that this might be better as passing in a double pointer, but the way it's been used makes that very hard to change.

I'll state this now, this fix isn't correct. There are cases where the passed in font instance is `_ggadget_default_font`, in which case it should never be freed, as this is a default value (and one of this initialise-once lifetime variables). In which case, trying to free this leads right back to the issue saw in #3624.

As for why the function signature is hard to change, the reason that parameter is a void* and not a FontInstance* is because it's the function prototype for a conversion function (e.g. https://github.com/fontforge/fontforge/blob/master/gdraw/gresedit.c#L2576 and https://github.com/fontforge/fontforge/blob/master/gdraw/ggadgets.c#L303), which means that you'd have to change how that value is passed in everywhere it's used like that - harder because of the additional layers of indirection, but perhaps not too bad in this case, since I only see it in ggadgets.c and gresourceedit.c.

Without some serious yak-shaving, this one's going to be hard to fix, and I'm not sure it's worth the effort.

skef:
@serval2412 the sort of reasons @jtanx raised about the use of the function is what made me nervous about accepting this. Basically this is a poorly defined interface that should be rearchitected, but before then it's just better to let it leak. It happens sometimes.

serval2412:
Refactoring, removing dead parts, useless variables is useful, it helps for having a code easier to maintain.
Reading http://pfaedit.org/ff-history-redux.html, I understand those who prefer refactoring first before adding new features.
So about yak-shaving, do you have in mind some easy tasks to begin with?

jtanx:
That's for you to figure out; one possible solution is to make it so it always returns a copy, and never a shared instance, so that it's always safe to free the result. Another is to figure out if it's possible to detect when it's safe and when it's not to free the return value. You're changing the ownership semantics of the returned value in doing so.

Anyway, can I ask that you test/exercise the code somehow for all changes that you make? I tested #3633 to make sure it did what I thought it did. It's great that you want to make these changes, but I can't be spending my time trying to track down regressions (#3613 caused by #3593), when it could have been easily caught by actually running what would trigger that condition. Otherwise, I'll be vetoing all future changes that haven't actually been tested.

serval2412:
jtanx: to try to reproduce the crash, I put again SplineFontFree(sf) + include line.
Then I runned "make clean" and followed INSTALL_md with spiro, unifont, etc.
I succeeded in launching Fontforge.
make check gave:
=> 71 tests successful, 20 tests skipped (I don't know why)

I tried fontforge -new too, no crash.

Could you provide a minimal step by step process to reproduce this? (or perhaps provide auto007231-1.asfd ?)
Sorry for bringing some regressions, if I can run some tests to detect them early, don't hesitate to tell me.
As I indicated in a comment, I don't know anything in internals of Fontforge and on font creation/edition. My goal is just to try to contribute to FOSS.

ctrlcctrlv:
What @jtanx is getting at is that he _shouldn't have to_ provide the autosave file. He's saying you should make your own autosave file which will hit the condition in the code for the express purpose of _testing_ PRs before they're submitted.

I understand your eagerness to contribute, but I think @jtanx is really not asking too much from you. Read the code, test files/command line switches/dialogs/etc that use the part of the code you touched.

serval2412:
@ctrlcctrlv: it seems a too steep learning curve for me, I give up.
Sorry for all the annoying problems my patches brought.

ctrlcctrlv:
@serval2412 Sorry, hope to see you in a few years. If you see my history with this project, in 2015 I was literally so bad at C I had to offer to pay people to help me patch FontForge (https://github.com/fontforge/fontforge/issues/2390#issuecomment-118758198), and now I've gotten several commits in and have tracked down quite a few Spiro / UI bugs, currently #3569 is my nemesis.

What I'm saying is, things will get better if you keep studying and applying what you know.

jtanx:
@serval2412 I apologise if I was a bit glib in my previous reply.

I just wanted to make it clear, that I am supportive of these kind of changes (it's not great that we're leaking memory), but there needs to be some care around fully thinking through what these changes will mean.

For #3593 (and this case) in particular - do you know what's worse than leaked memory? Using it after it has been freed. I guess you might not be inclined any more, but the way that you can repro what I saw from #3593 is to start editing a font, and FontForge will generate autosave files. Kill FontForge and start it back up - it should try to load them. The (sparse) tests that FontForge has won't pick this up.

As I alluded to earlier, it's not particularly sustainable for us to be second-guessing every single change, because it was made without really understanding what's happening.

So what makes our lives so much easier, is if you try to spend the time to really figure out what's going on, what the potential side-effects will be, and most ideally, be able to test/execute the code path that you're changing - this will significantly reduce the chance of regressions like we saw. I'm not saying this is easy, but it is something that you get better at with time, as @ctrlcctrlv pointed out.

Towards that, my suggestion is to get familiar with both gdb and valgrind - gdb in particular will help you to actually see what code is executing and how your changes will impact on that. valgrind is useful to pick up blatant memory access errors, although it can generate a lot of noise, particularly so when python scripting is enabled.

serval2412:
@jtanx: exactly, what I needed, a quick way to reproduce the crash.
I know how to use basics from gdb with "bt", "frame" and "p" commands to display vars.
I also know how to retrieve a Valgrind trace but know few about how to interpret it.
Sometimes we can't reproduce a crash with Valgrind because this one monitors memory.
However, it can still show problems in its long traces.
A useful thing of Valgrind is cachegrind for perf, badfully even with kcachegrind, I don't understand how to find bottlenecks.

Here's an example of bt I got when giving a try to 3569:
0x00007ffff7d1bc85 in CVInfoDrawText (cv=cv@entry=0x555555905cb0, pixmap=pixmap@entry=0x55555597a070) at charview.c:4049
4049		    strncpy(layername,cv->b.sc->parent->layers[idx].name,layernamesz);
(gdb) bt
#0  0x00007ffff7d1bc85 in CVInfoDrawText (cv=cv@entry=0x555555905cb0, pixmap=pixmap@entry=0x55555597a070) at charview.c:4049
#1  0x00007ffff7d1bf81 in CVInfoDraw (cv=0x555555905cb0, pixmap=0x55555597a070) at charview.c:4179
#2  0x00007ffff7d1d65f in CVUpdateInfo (event=0x7fffffffbbd0, event=0x7fffffffbbd0, cv=0x555555905cb0) at charview.c:3295
#3  0x00007ffff7d1d65f in CVNewScale (cv=cv@entry=0x555555905cb0) at charview.c:3313
#4  0x00007ffff7d1d9c0 in CVResize (cv=cv@entry=0x555555905cb0) at charview.c:6046
...
Here are the lines:
   4043     const int layernamesz = 100;
   4044     char layername[layernamesz+1];
   4045     strncpy(layername,_("Guide"),layernamesz);
   4046     if(cv->b.drawmode!=dm_grid) {
   4047         int idx = CVLayer((CharViewBase *) cv);
   4048         if(idx >= 0 && idx < cv->b.sc->parent->layer_cnt) {
   4049             strncpy(layername,cv->b.sc->parent->layers[idx].name,layernamesz);
   4050         }
   4051     }

So here I would change line 4049 with:
strncpy(layername,cv->b.sc->parent->layers[idx].name,layernamesz - strlen(layername));

I just thought I should build with enable-debug, it may help to avoid "value has been optimized out" in gdb.

Also I gave a try to "enable-gcc-warnings" and it spat a lot of warnings in console. I suppose a lot are harmless but how much may help to tackle a nasty bug? Perhaps a good start cleaning work?

jtanx:
What you've proposed is unlikely to fix it, disclaimer that I haven't scoped this out fully, but I'd suspect that this segfault is a secondary side-effect only, and that you simply can't free the splinefont as you've done, because it's still being used by the call into `FontViewCreate` prior.

For easier debugging, the `enable-debug` flag should work, although I normally just do a `CFLAGS="-g -O0" ./configure`

serval2412:
About the bt part I provided, it's just some crash I encountered. I didn't mean to question your revert about splinefont.
I didn't get the same bts as Frederic or you retrieved, here's another one:
__memmove_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:416
416	../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S: Aucun fichier ou dossier de ce type.
(gdb) bt
#0  0x00007ffff6d1bbf2 in __memmove_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:416
#1  0x00007ffff74bc6c5 in BdfCFromBitmap (bitmap=0x7fffffffb760, bitmap_left=-5156113, bitmap_top=-26311296, pixelsize=72, depth=4, sc=0x5555558547f0, metrics=0x0) at freetype.c:407
#2  0x00007ffff74bfa39 in SplineCharFreeTypeRasterizeNoHints (sc=0x5555558547f0, layer=1, ptsize=72, dpi=72, depth=4) at freetype.c:1101
#3  0x00007ffff767a19d in BDFPieceMeal (bdf=0x5555558bad40, index=0) at splinefill.c:1697
#4  0x00007ffff767a365 in BDFPieceMealCheck (bdf=0x5555558bad40, index=0) at splinefill.c:1721
#5  0x00007ffff7d2cf4a in FVRefreshChar (fv=0x5555558726d0, gid=0) at fontview.c:5946
#6  0x00007ffff7d2d13d in FVRegenChar (fv=0x5555558726d0, sc=0x5555558547f0) at fontview.c:5977
#7  0x00007ffff7c33b6f in SC_UpdateAll (sc=0x5555558
[valgrind.log.gz](https://github.com/fontforge/fontforge/files/3050498/valgrind.log.gz)
547f0) at charview.c:3168
I suppose negative numbers aren't appreciated by memcpy but I don't know where they may come from.
Here's a Valgrind trace when trying to repro to 3569. (retrieved with valgrind --leak-check=yes fontforge /tmp/jul/test.sfd &> /tmp/valgrind.log)

Issue #3621, opened 2019-04-03
Freeze! Execute script dialog, paste, try to double click to select line, freeze in gtextfield_mouse


ctrlcctrlv:
Arch Linux, https://github.com/fontforge/fontforge/commit/26a0b528801d2e1281a009b8a36338124eb6f7c1

Reproduction steps:
1. File &rarr; Execute Script&hellip;
1. Write multiple lines, for example `test\ntest\ntest` (Where `\n` is a new line.)
2. Paste something long that goes over one line between the first and second line. For example, try pasting `test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test/test`.
3. As in #3620, keep clicking on the new line you pasted. Click click click, and drag a bit while clicking sometimes.
4. You'll get a freeze!

On killing the program, you'll get this:
```
Program received signal SIGINT, Interrupt.
0x0000155555073754 in gtextfield_mouse (g=0x555555dc9530, event=0x7fffffffae30)
    at gtextfield.c:1761
1761                for ( j=0; gt->lines[i+1]!=-1 && gt->sel_base>=gt->lines[i+1]; ++j );
(gdb) bt
#0  0x0000155555073754 in gtextfield_mouse
    (g=0x555555dc9530, event=0x7fffffffae30) at gtextfield.c:1761
#1  0x00001555550130ab in _GWidget_Container_eh
    (gw=0x555555e07ae0, event=0x7fffffffae30) at gcontainer.c:302
#2  0x0000155555014f1e in _GWidget_TopLevel_eh
    (gw=0x555555e07ae0, event=0x7fffffffae30) at gcontainer.c:745
#3  0x000015555508fcd2 in _GGDKDraw_CallEHChecked
    (gw=0x555555e07ae0, event=0x7fffffffae30, eh=
    0x155555014590 <_GWidget_TopLevel_eh>) at ggdkdraw.c:336
#4  0x00001555550924f0 in _GGDKDraw_DispatchEvent
    (event=0x555555646e60, data=0x555555662320) at ggdkdraw.c:1211
#5  0x000015555384ac86 in  () at /usr/lib/libgdk-3.so.0
#6  0x000015555387cab5 in  () at /usr/lib/libgdk-3.so.0
#7  0x0000155552a3f7bf in g_main_context_dispatch ()
    at /usr/lib/libglib-2.0.so.0
[...]

Issue #3620, opened 2019-04-03
Printed to stdout repeatedly: "About to crash"


ctrlcctrlv:
Arch Linux, https://github.com/fontforge/fontforge/commit/26a0b528801d2e1281a009b8a36338124eb6f7c1

```
The program being debugged has been started already.
Start it from the beginning? (y or n) y
Starting program: /usr/local/bin/fontforge Chomsky.sfd
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
Copyright (c) 2000-2018 by George Williams. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Based on sources from 06:12 UTC  1-Apr-2019-ML-D.
 Based on source from git with hash: 3e641ec84d141d291aa75bd84897aee04bfe5428
About to crash
About to crash
About to crash
About to crash
About to crash
About to crash
About to crash
About to crash
About to crash
[snip]
```

Reproduction steps:
1. File &rarr; Execute Script&hellip;
2. Write "test"
3. Just keep clicking on the word test, try to drag a bit while clicking. You'll get this output. 

If you want a video of this, just me know and I'll provide one.

skef:
@ctrlcctrlv Is this still an issue?

ctrlcctrlv:
Nope, I can't repro it anymore, #3652 solved it

PR #3632, opened 2019-04-05
Fix some crashes/weird behavior in gtextfield

ctrlcctrlv:
(Please carefully review this before merging.)

This closes fontforge/fontforge#3621 and closes fontforge/fontforge#3620,
which are both weird crashes in the "Execute Script" dialog.

It's extremely pointless to print "About to crash" before a crash, so I
totally removed that. Furthermore, despite the message, I'd never
experienced a crash after that was printed. The message goes back to
2001, in commit 78bdbeb3ef349c35e3b8f64b0689a4574391d686...I think it
can be easily seen to be outdated.

As far as the crash goes, I *believe* this totally fixes it while
retaining all previous functionality.

### DIFF
--- a/gdraw/gtextfield.c
+++ b/gdraw/gtextfield.c
@@ -1746,22 +1746,30 @@ return( true );
 	if ( gt->pressed==NULL )
 	    gt->pressed = GDrawRequestTimer(gt->g.base,200,100,NULL);
 	if ( gt->sel_start > u_strlen( gt->text ))	/* Ok to have selection at end, but beyond is an error */
-	    fprintf( stderr, "About to crash\n" );
-	_ggadget_redraw(g);
-return( true );
+        _ggadget_redraw(g);
+        return( true );
     } else if ( gt->pressed && (event->type == et_mousemove || event->type == et_mouseup )) {
 	int refresh = true;
 
 	if ( gt->drag_and_drop ) {
 	    refresh = GTextFieldDoDrop(gt,event,end-gt->text);
 	} else if ( gt->linesel ) {
-	    int j, e;
+        int j = 0;
+        int e = 0;
 	    gt->sel_start = gt->lines[i]; gt->sel_end = gt->lines[i+1];
-	    if ( gt->sel_end==-1 ) gt->sel_end = u_strlen(gt->text);
-	    for ( j=0; gt->lines[i+1]!=-1 && gt->sel_base>=gt->lines[i+1]; ++j );
-	    if ( gt->sel_start<gt->lines[i] ) gt->sel_start = gt->lines[i];
-	    e = gt->lines[j+1]==-1 ? u_strlen(gt->text): gt->lines[j+1];
-	    if ( e>gt->sel_end ) gt->sel_end = e;
+
+        while ( gt->lines[i+1]!=-1 && gt->sel_base>=gt->lines[i+1] && j<sizeof(gt->lines) ) {
+            if ( gt->sel_end==-1 ) gt->sel_end = u_strlen(gt->text);
+            j++;
+        }
+
+	    e = gt->lines[j+1]==-1 ?
+            u_strlen(gt->text) :
+            gt->lines[j+1];
+
+        if ( e>gt->sel_end ) {
+            gt->sel_end = e;
+        }
 	} else if ( gt->wordsel )
 	    GTextFieldSelectWords(gt,end-gt->text);
 	else if ( event->u.mouse.button!=2 ) {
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
Any thoughts @skef ?

gdraw/gtextfield.c
```diff
@@ -1746,22 +1746,30 @@ return( true );
...
+        int e = 0;
 	    gt->sel_start = gt->lines[i]; gt->sel_end = gt->lines[i+1];
-	    if ( gt->sel_end==-1 ) gt->sel_end = u_strlen(gt->text);
-	    for ( j=0; gt->lines[i+1]!=-1 && gt->sel_base>=gt->lines[i+1]; ++j );
-	    if ( gt->sel_start<gt->lines[i] ) gt->sel_start = gt->lines[i];
-	    e = gt->lines[j+1]==-1 ? u_strlen(gt->text): gt->lines[j+1];
-	    if ( e>gt->sel_end ) gt->sel_end = e;
+
+        while ( gt->lines[i+1]!=-1 && gt->sel_base>=gt->lines[i+1] && j<sizeof(gt->lines) ) {
+            if ( gt->sel_end==-1 ) gt->sel_end = u_strlen(gt->text);
```
skef: This loop isn't modifying `gt->text`, and the only line that modifies `gt->sel_end` is this one. Putting it into the loop a) doesn't accomplish anything and b) makes it O(n*m) instead of O(n+m).

gdraw/gtextfield.c
```diff
@@ -1746,22 +1746,30 @@ return( true );
...
 
 	if ( gt->drag_and_drop ) {
 	    refresh = GTextFieldDoDrop(gt,event,end-gt->text);
 	} else if ( gt->linesel ) {
-	    int j, e;
+        int j = 0;
+        int e = 0;
 	    gt->sel_start = gt->lines[i]; gt->sel_end = gt->lines[i+1];
-	    if ( gt->sel_end==-1 ) gt->sel_end = u_strlen(gt->text);
-	    for ( j=0; gt->lines[i+1]!=-1 && gt->sel_base>=gt->lines[i+1]; ++j );
```
skef: What the hell?

gdraw/gtextfield.c
```diff
@@ -1746,22 +1746,30 @@ return( true );
...
-	    if ( gt->sel_start<gt->lines[i] ) gt->sel_start = gt->lines[i];
-	    e = gt->lines[j+1]==-1 ? u_strlen(gt->text): gt->lines[j+1];
-	    if ( e>gt->sel_end ) gt->sel_end = e;
+
+        while ( gt->lines[i+1]!=-1 && gt->sel_base>=gt->lines[i+1] && j<sizeof(gt->lines) ) {
+            if ( gt->sel_end==-1 ) gt->sel_end = u_strlen(gt->text);
+            j++;
```
skef: This doesn't seem to address the burning question of line 1761. We're incrementing `j`in a loop with a guard that doesn't include `j` or anything modified by `j`. You've added a check but the whole structure reduces to `j=sizeof(gt->lines)`.

gdraw/gtextfield.c
```diff
@@ -1746,22 +1746,30 @@ return( true );
 	if ( gt->pressed==NULL )
 	    gt->pressed = GDrawRequestTimer(gt->g.base,200,100,NULL);
 	if ( gt->sel_start > u_strlen( gt->text ))	/* Ok to have selection at end, but beyond is an error */
-	    fprintf( stderr, "About to crash\n" );
-	_ggadget_redraw(g);
-return( true );
+        _ggadget_redraw(g);
```
skef: The comment has a point, even if it might be better as an informed logged error. If you have a reproducible case that enters this code block it would be better to fix it. It would also be better to know why it doesn't cause a crash.

gdraw/gtextfield.c
ctrlcctrlv: I have no idea how to reply. I removed this code to replace it with what I thought to be a more readable alternative&hellip;

gdraw/gtextfield.c
ctrlcctrlv: I didn't put anything into the loop that wasn't already there.

gdraw/gtextfield.c
ctrlcctrlv: Perhaps...I just thought a `while` was more readable than a `for`.

gdraw/gtextfield.c
ctrlcctrlv: I do have a reproducible case, it's #3620

gdraw/gtextfield.c
ctrlcctrlv: Given this comment and https://github.com/fontforge/fontforge/issues/3569#issuecomment-480476944 I'm dubious as to whether I should keep trying to contribute if my commits/stack traces, etc only cause bewilderment :slightly_frowning_face:

ctrlcctrlv:
@skef Should I close this?

skef:
I don't think the current code gets at the problems involved. If the other issues weren't filed to track this problem I would recommend leaving it open, but they are so that's probably sufficient and this can be closed.

ctrlcctrlv:
Okay :) This does fix the crash, but probably not in the best way.

Issue #3614, opened 2019-04-01
Crash when closing font


ctrlcctrlv:
Perhaps, but not positively, related to #3613 

```
Recovering from /home/fred/.config/fontforge/autosave/auto007231-1.asfd.zip... Recovering from /home/fred/.config/fontforge/autosave/auto007231-1.asfd...  Done
17:53:46.675 WARNING: _GGDKDraw_CreateWindow (ggdkdraw.c:601) - Could not call GGDKDrawSetTransientFor on a restricted window (0x5555558ee610 Loading...), restrict count 0, transient window count 0

Program received signal SIGSEGV, Segmentation fault.
0x000015555529dee5 in FontView_Free (fv=0x5555559513f0) at fontview.c:7583
7583                for ( prev = (FontView *) (fv->b.sf->fv); prev->b.nextsame!=&fv->b; prev=(FontView *) (prev->b.nextsame) );
(gdb) bt
#0  0x000015555529dee5 in FontView_Free (fv=0x5555559513f0) at fontview.c:7583
#1  0x000015555529c748 in FontViewRemove (fv=0x5555559513f0) at fontview.c:7170
#2  0x000015555529ca7a in fv_e_h (gw=0x555555818af0, event=0x7fffffffbd80) at fontview.c:7251
#3  0x0000155555014966 in _GWidget_Container_eh (gw=0x555555818af0, event=0x7fffffffbd80) at gcontainer.c:403
#4  0x0000155555015f1e in _GWidget_TopLevel_eh (gw=0x555555818af0, event=0x7fffffffbd80) at gcontainer.c:745
#5  0x0000155555090cd2 in _GGDKDraw_CallEHChecked (gw=0x555555818af0, event=0x7fffffffbd80, eh=0x155555015590 <_GWidget_TopLevel_eh>)
    at ggdkdraw.c:336
#6  0x000015555509621b in GGDKDrawPostEvent (e=0x7fffffffbd80) at ggdkdraw.c:2212
#7  0x00001555550906c7 in _GGDKDraw_OnWindowDestroyed (data=0x555555818af0) at ggdkdraw.c:218
#8  0x0000155552a3fe53 in  () at /usr/lib/libglib-2.0.so.0
#9  0x0000155552a40661 in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#10 0x0000155552a42739 in  () at /usr/lib/libglib-2.0.so.0
[...]

PR #3634, opened 2019-04-05
Fix autosave recovery

jtanx:
Regression from #3593

### DIFF
--- a/fontforge/autosave.c
+++ b/fontforge/autosave.c
@@ -28,7 +28,6 @@
 #include "autosave.h"
 
 #include "baseviews.h"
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
Did you test this against #3613 or is that another problem?

jtanx:
This was just something I observed locally, looking at that stack trace, it looks like it's the same cause

ctrlcctrlv:
@jtanx Hey&mdash;I confirm this closes both #3613 and #3614. Your patchset averts both crashes.

jtanx:
Thanks for confirming

Issue #3512, opened 2019-03-09
Mass cleanup, build system reforms, fixing those compiler warnings


jtanx:
After the current release is done, I'll be looking at doing some things about how FontForge is currently built. There's a lot of cruft in the way of actually making improvements, and detecting potential issues early enough.

- Compiling with `-Wall` is a disaster; there's so many compiler warnings that it's hard to spot real issues
- Compilation itself is slow, or at least slower than it needs to be.
   - On Windows, build times are atrocious; running `./bootstrap` takes a good 10-15 minutes on my system, running configure another 10 minutes or so, and building it another 45 minutes or so. 
     - Linking in particular is very slow, although I'd put that down to libtool/mingw

Those two points in particular impedes on debugging issues, because it takes so long just to iterate on builds; worse still if those changes affect the build system in any way.

So because of that, at the top of my list is to:

- Delete a bunch of old stuff out of the repo, namely:
  - breakpad and the collab feature; both are seriously unmaintained, and never really worked all that well; as far as I know, breakpad in particular was only ever built to work on mac, and I have doubts that it's doing anything useful now.
  - The gtkui folder, no one ever uses it (besides, it'll still be in git history, that's what it's for)
- The nodejs (related to collab) and win folders. The win folder contains nothing useful (except maybe the icons, which can be moved out).
- Remove gnulib. Using gnulib is a mistake, for the things we use it for, we can just use glib, or write our own - more importantly, this means we can move off autotools.
- Switch to a different build system, either cmake or meson. It actually wasn't too bad to get something in cmake off the ground (see https://github.com/fontforge/fontforge/tree/cmake), but that's mostly because I'm already familiar with cmake.
- Get rid of the plugin system, no one uses it, and if I recall correctly, the only plugin that exists is gb12345, which again, is also broken. This means that we can not depend on libltdl, which is fairly unix specific.

Full wishlist here: https://github.com/jtanx/fontforge/wiki/Wishlist

skef:
After my exposure to the code base the past couple months I agree with just about all of this. 

An alternate possibility with collab is to cut it off at the roots rather than remove it entirely. If it were ever to be revived almost all of the existing code -- especially the zeromq stuff -- would belong in python, but the code close to the UI changes might be worth leaving in.

ctrlcctrlv:
Yes, please remove all the node.js stuff, I don't think it has a single user. It was a cute idea but never should have been in the core release, it should be an extension, and making it one isn't our responsibility.

[...]

PR #3619, opened 2019-04-03
Add `WriteModificationTime` option

ctrlcctrlv:
By default, FontForge writes a `ModificationTime` into every SFD file.
This can disturb merging when using a version control system to manage
your font versions. Merge conflicts are inevitable with this line and
very annoying.

This PR adds an option to remove the line. I tested it, and FontForge
opens SFD files without the line without complaint or line in the
Warnings dialog.

If this PR is merged, the default will be to write the line, as is the
current action.

### Types of changes
- [x] New feature (non-breaking change which adds functionality)

### Final checklist
- [x] My code follows the code style of this project found [here](https://github.com/fontforge/fontforge/blob/master/CONTRIBUTING.md#coding-style).
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the [**CONTRIBUTING**](https://github.com/fontforge/fontforge/blob/master/CONTRIBUTING.md) guidelines.

PR #3635, opened 2019-04-06
Remove plugin support

jtanx:
This removes plugin support. Related issue; #3512.

After reviewing gb12345, I've also removed it. It's literally just an encoding, which anyone so inclined to use it may do so via the 'Load Encoding' setting. An equivalent to what gb12345.c was doing may be found here: https://gist.github.com/jtanx/e154f9f47256ba52fedd8bb3039d1883

That gb12345 was also quite broken, the forward encoding doesn't even match the reverse encoding (I don't know the exact terminology, but going from uni to encoded, it uses the 'unpacked' form, but going the other way, it uses the 'packed' form (adding 0x8080)).

In fact, given that FontForge uses iconv, it's kind of redundant that it tries to provide/bundle its own iconv implementation with all the conversion tables; those probably contribute to 1/4 to 1/2 of FontForge's total size.


ctrlcctrlv:
I don't want to sound like I'm rejecting or even criticizing this, I'm not. Most probably the current infrastructure needs to go.

But long term _some_ form of plugins/scripts are definitely needed. There are so many useful FontForge scripts that could be shared in a similar way as [WeeChat](https://github.com/weechat/weechat), an IRC client, [shares its scripts](https://weechat.org/scripts/).

I've brought this up numerous times, most recently at #3619. I think @skef supports this idea but am not sure.

jtanx:
I agree, but not via the way this plugin system was designed (for reasons outlined on the mailing list).

Extensibility should come through the python/scripting interface.

PR #3637, opened 2019-04-07
Update to "new menus" project 

skef:
# Important Note

This is a push request **into the `menus-2019` branch**, not into `master`. I am submitting and self-approving this request for record-keeping purposes. 

This jumbo commit is my update of the `new-menus` branch last updated in 2015. It's preceded by a squash of that other branch. 

# Contents 

There are too many to mention -- see the commit for details. Highlights include:

1. Combined "Extensions" menu (was "Tools"): If and when `NATIVE_CALLBACKS` is supported, both kinds of scripts will be listed under the single menu, with a line between if there are some of both types.
2. BitmapView now has the new menu pattern as well, and hotkey support has been added.
3. Hotkeys have been remapped to match the pattern. (Similar to #3605, but this supercedes it. If this is accepted that should just be closed.)

# Changes from Earlier Documentation

I made a lot of judgment calls in completing this, so it doesn't completely match the [documentation](https://drive.google.com/drive/folders/0B-5mpXa8HrFVdkc3ZmhiZ3R4bzA). A systemic account of the differences is also probably best accomplished by comparing the docs to code, but I'll list some things here:

1. There are some name changes. For example I put/left "Feature File" under "Merge" because it fit better. 
2. There are also ordering and "grouping" (i.e. line placement) changes where I thought it made sense.
3. There are quite a few additions -- things in the existing menus not yet given a place in the new ones. These were the primary motivation for the name, ordering, and grouping changes. 
4. When an entire menu or submenu would always be blank in a given window, I omitted it and made the next level up a disabled entry.
5. Most of the menus are the same across all windows except for small details (like the label on "Find", or what's in the `Selection` menu). This was the evident design intent. However, I made the area below View -> Former Glyph specific to the window type. (The documentation didn't really indicate what to do about it, the existing implementation hadn't addressed it yet, and this was the way that made sense.)

I wrote some book-keeping helper scripts for myself, and while there are almost certainly typos and other mistakes still buried in here, I've confirmed that all functions present before in menus are present now. I've also gone through the Alt-accelerators (the underlines) and done my best to make them unique per-menu and (mostly) unified across the different windows. There are more hotkeys than supported in the current `master` because of the duplication of #3605 work.


jtanx:
Glad to see someone picking this up again

davelab6:
I love it :

PR #3643, opened 2019-04-10
Don't install the mac timer in gdk mode

jtanx:
It shouldn't be needed; gdk/quartz drives that event loop itself

(this greatly improves idle cpu usage)


frank-trampe:
The fact that it runs at all without the event timer seems to me to validate the change, no?

jtanx:
Yeah, more so that it still responds to dock open/close/activate events (which it does)

frank-trampe:
Let's merge.

PR #3645, opened 2019-04-11
Add user decompositions

ctrlcctrlv:
This closes fontforge/fontforge#3615.

This supersedes fontforge/fontforge#3617.

This allows users to set their own decomposition for each buildable
glyph.

A new line is added for each char in the SFD, `Decomposition`. It has
the same format as the `Comment` line.

Through the Python interface, users can set the decompositon using the
`glyph.user_decomp` attribute. This has been added to `python.html`.

`SFGetAlternate` now respects user decompositions if a `SplineChar` is
passed, as happens when characters are built.

Besides all the changes to "Glyph Info &rarr; Components", I unindented
"Components", so it is no longer under "Ligatures" in the menu, since
with this patch that no longer makes sense.

The "Components" section now looks like this:

glyph info

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -7276,6 +7276,11 @@ static int PyFF_Glyph_set_user_decomp(PyFF_Glyph *self,PyObject *value, void *UN
         return -1;
     }
 
+    if (strlen(newv) > 5) {
+        PyErr_WarnEx(PyExc_ValueError, "It doesn't make sense for the decomposition to be this long; truncated to five characters", 1);
+        newv[5] = '\0';
+    }
+
     free(self->sc->user_decomp);
     self->sc->user_decomp = utf82u_copy(newv);
 
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -7249,6 +7249,7 @@ static PyObject *PyFF_Glyph_get_user_decomp(PyFF_Glyph *self, void *UNUSED(closu
     else {
         char* out = u2utf8_copy(self->sc->user_decomp);
         return( PyUnicode_DecodeUTF8(out, strlen(out), NULL) );
+        free(out);
     }
 }
 
@@ -7929,12 +7930,20 @@ static PyGetSetDef PyFF_Glyph_getset[] = {
 /*  Glyph Methods  */
 /* ************************************************************************** */
 
-static PyObject *PyFFGlyph_Build(PyObject *self, PyObject *UNUSED(args)) {
+static PyObject *PyFFGlyph_Build(PyObject *self, PyObject *args) {
     SplineChar *sc = ((PyFF_Glyph *) self)->sc;
     int layer = ((PyFF_Glyph *) self)->layer;
+    int accent_hint = false;
+    PyObject *accent_hint_pyo;
+
+    if ( PyArg_ParseTuple(args, "|O", &accent_hint_pyo) ) {
+        if (accent_hint_pyo == Py_True) {
+            accent_hint = true;
+        }
+    }
 
     if ( SFIsSomethingBuildable(sc->parent,sc,layer,false) )
-	SCBuildComposit(sc->parent,sc,layer,NULL,true);
+	SCBuildComposit(sc->parent,sc,layer,NULL,true,accent_hint);
 
 Py_RETURN( self );
 }
@@ -9333,7 +9342,7 @@ static PyMethodDef PyFF_Glyph_methods[] = {
     { "autoInstr", PyFFGlyph_autoInstr, METH_NOARGS, "Guess at truetype instructions"},
     { "autoTrace", PyFFGlyph_autoTrace, METH_NOARGS, "Autotrace any background images"},
     { "boundingBox", (PyCFunction) PyFFGlyph_BoundingBox, METH_NOARGS, "Finds the minimum bounding box for the glyph (xmin,ymin,xmax,ymax)" },
-    { "build", PyFFGlyph_Build, METH_NOARGS, "If the current glyph is an accented character\nand all components are in the font\nthen build it out of references" },
+    { "build", PyFFGlyph_Build, METH_VARARGS, "If the current glyph is an accented character\nand all components are in the font\nthen build it out of references" },
     { "canonicalContours", (PyCFunction) PyFFGlyph_canonicalContours, METH_NOARGS, "Orders the contours in the current glyph by the x coordinate of their leftmost point. (This can reduce the size of the postscript charstring needed to describe the glyph(s)."},
     { "canonicalStart", (PyCFunction) PyFFGlyph_canonicalStart, METH_NOARGS, "Sets the start point of all the contours of the current glyph to be the leftmost point on the contour."},
     { "changeWeight", (PyCFunction) PyFFGlyph_changeWeight, METH_VARARGS, "Change the weight (thickness) of the stems of the glyph"},
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -7271,21 +7271,20 @@ static int PyFF_Glyph_set_user_decomp(PyFF_Glyph *self,PyObject *value, void *UN
         /* Need to force utf8 encoding rather than accepting the "default" */
         /*  which would happen if we treated unicode as a string */
         temp = PyUnicode_AsUTF8String(value);
-        newv = copy( PyBytes_AsString(temp));
+        udbuf = utf82u_copy(PyBytes_AsString(temp));
         Py_DECREF(temp);
     } else {
-        newv = copy( PyBytes_AsString(value));
+        udbuf = utf82u_copy(PyBytes_AsString(value));
     }
     
-    if ( newv==NULL ) return -1;
+    if ( udbuf==NULL ) return -1;
 
-    if (newv[0] == '\0') {
+    if (udbuf[0] == '\0') {
         if (self->sc->user_decomp != NULL) { free(self->sc->user_decomp); }
         self->sc->user_decomp = NULL;
         return 0;
     }
 
-    udbuf = utf82u_copy(newv);
 
     if (u_strlen(udbuf) > 5) {
          PyErr_WarnEx(PyExc_ValueError, "It doesn't make sense for the decomposition to be this long; truncated to five characters", 1);
@@ -7950,7 +7949,7 @@ static PyObject *PyFFGlyph_Build(PyObject *self, PyObject *args) {
     SplineChar *sc = ((PyFF_Glyph *) self)->sc;
     int layer = ((PyFF_Glyph *) self)->layer;
     int accent_hint = false;
-    PyObject *accent_hint_pyo;
+    PyObject *accent_hint_pyo = NULL;
 
     if ( PyArg_ParseTuple(args, "|O", &accent_hint_pyo) ) {
         if (accent_hint_pyo == Py_True) {
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -7243,46 +7243,57 @@ return( PyFF_Glyph_set_hints(self,true,value));
 }
 
 static PyObject *PyFF_Glyph_get_user_decomp(PyFF_Glyph *self, void *UNUSED(closure)) {
+    PyObject *ret;
+
     if ( self->sc->user_decomp==NULL ) {
         return( Py_BuildValue("s", "" ));
     }
     else {
         char* out = u2utf8_copy(self->sc->user_decomp);
-        return( PyUnicode_DecodeUTF8(out, strlen(out), NULL) );
+        ret = PyUnicode_DecodeUTF8(out, strlen(out), NULL);
         free(out);
+        return ret;
     }
 }
 
 static int PyFF_Glyph_set_user_decomp(PyFF_Glyph *self,PyObject *value, void *UNUSED(closure)) {
     char *newv;
     PyObject *temp;
+    unichar_t *udbuf;
 
     if (value == Py_None) {
+        if (self->sc->user_decomp != NULL) { free(self->sc->user_decomp); }
         self->sc->user_decomp = NULL;
         return 0;
     }
 
     if ( PyUnicode_Check(value)) {
-	/* Need to force utf8 encoding rather than accepting the "default" */
-	/*  which would happen if we treated unicode as a string */
-	temp = PyUnicode_AsUTF8String(value);
-	newv = copy( PyBytes_AsString(temp));
-	Py_DECREF(temp);
+        /* Need to force utf8 encoding rather than accepting the "default" */
+        /*  which would happen if we treated unicode as a string */
+        temp = PyUnicode_AsUTF8String(value);
+        newv = copy( PyBytes_AsString(temp));
+        Py_DECREF(temp);
     } else {
-        return -1;
+        newv = copy( PyBytes_AsString(value));
+    }
+    
+    if ( newv==NULL ) return -1;
+
+    if (newv[0] == '\0') {
+        if (self->sc->user_decomp != NULL) { free(self->sc->user_decomp); }
+        self->sc->user_decomp = NULL;
+        return 0;
     }
 
-    if ( newv==NULL ) {
-        return -1;
+    udbuf = utf82u_copy(newv);
+
+    if (u_strlen(udbuf) > 5) {
+         PyErr_WarnEx(PyExc_ValueError, "It doesn't make sense for the decomposition to be this long; truncated to five characters", 1);
+         udbuf[5] = '\0';
     }
 
-    if (strlen(newv) > 5) {
-        PyErr_WarnEx(PyExc_ValueError, "It doesn't make sense for the decomposition to be this long; truncated to five characters", 1);
-        newv[5] = '\0';
-    }
-
-    free(self->sc->user_decomp);
-    self->sc->user_decomp = utf82u_copy(newv);
+    if (self->sc->user_decomp != NULL) { free(self->sc->user_decomp); }
+    self->sc->user_decomp = udbuf;
 
     return 0;
 }
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -7242,6 +7242,45 @@ static int PyFF_Glyph_set_vhints(PyFF_Glyph *self,PyObject *value, void *UNUSED(
 return( PyFF_Glyph_set_hints(self,true,value));
 }
 
+static PyObject *PyFF_Glyph_get_user_decomp(PyFF_Glyph *self, void *UNUSED(closure)) {
+    if ( self->sc->user_decomp==NULL ) {
+        return( Py_BuildValue("s", "" ));
+    }
+    else {
+        char* out = u2utf8_copy(self->sc->user_decomp);
+        return( PyUnicode_DecodeUTF8(out, strlen(out), NULL) );
+    }
+}
+
+static int PyFF_Glyph_set_user_decomp(PyFF_Glyph *self,PyObject *value, void *UNUSED(closure)) {
+    char *newv;
+    PyObject *temp;
+
+    if (value == Py_None) {
+        self->sc->user_decomp = NULL;
+        return 0;
+    }
+
+    if ( PyUnicode_Check(value)) {
+	/* Need to force utf8 encoding rather than accepting the "default" */
+	/*  which would happen if we treated unicode as a string */
+	temp = PyUnicode_AsUTF8String(value);
+	newv = copy( PyBytes_AsString(temp));
+	Py_DECREF(temp);
+    } else {
+        return -1;
+    }
+
+    if ( newv==NULL ) {
+        return -1;
+    }
+
+    free(self->sc->user_decomp);
+    self->sc->user_decomp = utf82u_copy(newv);
+
+    return 0;
+}
+
 static PyObject *PyFF_Glyph_get_comment(PyFF_Glyph *self, void *UNUSED(closure)) {
     if ( self->sc->comment==NULL )
 return( Py_BuildValue("s", "" ));
@@ -7793,6 +7832,9 @@ static PyGetSetDef PyFF_Glyph_getset[] = {
     {(char *)"comment",
      (getter)PyFF_Glyph_get_comment, (setter)PyFF_Glyph_set_comment,
      (char *)"Glyph comment", NULL},
+    {(char *)"user_decomp",
+     (getter)PyFF_Glyph_get_user_decomp, (setter)PyFF_Glyph_set_user_decomp,
+     (char *)"Glyph user decompositon", NULL},
     {(char *)"glyphclass",
      (getter)PyFF_Glyph_get_glyphclass, (setter)PyFF_Glyph_set_glyphclass,
      (char *)"glyph class", NULL},
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -7257,7 +7257,6 @@ static PyObject *PyFF_Glyph_get_user_decomp(PyFF_Glyph *self, void *UNUSED(closu
 }
 
 static int PyFF_Glyph_set_user_decomp(PyFF_Glyph *self,PyObject *value, void *UNUSED(closure)) {
-    char *newv;
     PyObject *temp;
     unichar_t *udbuf;
 
@@ -7283,12 +7282,6 @@ static int PyFF_Glyph_set_user_decomp(PyFF_Glyph *self,PyObject *value, void *UN
         if (self->sc->user_decomp != NULL) { free(self->sc->user_decomp); }
         self->sc->user_decomp = NULL;
         return 0;
-    }
-
-
-    if (u_strlen(udbuf) > 5) {
-         PyErr_WarnEx(PyExc_ValueError, "It doesn't make sense for the decomposition to be this long; truncated to five characters", 1);
-         udbuf[5] = '\0';
     }
 
     if (self->sc->user_decomp != NULL) { free(self->sc->user_decomp); }
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -1707,6 +1707,11 @@ static void SFDDumpChar(FILE *sfd,SplineChar *sc,EncMap *map,int *newgids,int to
 	SFDDumpUTF7Str(sfd,sc->comment);
 	putc('\n',sfd);
     }
+    if ( sc->user_decomp != NULL ) {
+	fprintf( sfd, "Decomposition: " );
+	SFDDumpUTF7Str(sfd,u2utf8_copy(sc->user_decomp));
+	putc('\n',sfd);
+    }
     if ( sc->color!=COLOR_DEFAULT )
 	fprintf( sfd, "Colour: %x\n", (int) sc->color );
     if ( sc->parent->multilayer ) {
@@ -5834,6 +5839,8 @@ exit(1);
 	    sc->color = temp;
 	} else if ( strmatch(tok,"Comment:")==0 ) {
 	    sc->comment = SFDReadUTF7Str(sfd);
+	} else if ( strmatch(tok,"Decomposition:")==0 ) {
+	    sc->user_decomp = utf82u_copy(SFDReadUTF7Str(sfd));
 	} else if ( strmatch(tok,"TileMargin:")==0 ) {
 	    getreal(sfd,&sc->tile_margin);
 	} else if ( strmatch(tok,"TileBounds:")==0 ) {
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -1709,7 +1709,9 @@ static void SFDDumpChar(FILE *sfd,SplineChar *sc,EncMap *map,int *newgids,int to
     }
     if ( sc->user_decomp != NULL ) {
 	fprintf( sfd, "Decomposition: " );
-	SFDDumpUTF7Str(sfd,u2utf8_copy(sc->user_decomp));
+	char* temp_ud = u2utf8_copy(sc->user_decomp);
+	SFDDumpUTF7Str(sfd, temp_ud);
+	free(temp_ud);
 	putc('\n',sfd);
     }
     if ( sc->color!=COLOR_DEFAULT )
@@ -5840,7 +5842,9 @@ exit(1);
 	} else if ( strmatch(tok,"Comment:")==0 ) {
 	    sc->comment = SFDReadUTF7Str(sfd);
 	} else if ( strmatch(tok,"Decomposition:")==0 ) {
-	    sc->user_decomp = utf82u_copy(SFDReadUTF7Str(sfd));
+	    char* decomp = SFDReadUTF7Str(sfd);
+	    sc->user_decomp = utf82u_copy(decomp);
+	    free(decomp);
 	} else if ( strmatch(tok,"TileMargin:")==0 ) {
 	    getreal(sfd,&sc->tile_margin);
 	} else if ( strmatch(tok,"TileBounds:")==0 ) {
--- a/fontforgeexe/charinfo.c
+++ b/fontforgeexe/charinfo.c
@@ -74,25 +74,29 @@ typedef struct charinfo {
 #define CID_Cancel	1005
 #define CID_ComponentMsg	1006
 #define CID_Components	1007
-#define CID_Comment	1008
-#define CID_Color	1009
-#define CID_GClass	1010
-#define CID_Tabs	1011
+#define CID_ComponentTextField  1008
+#define CID_ComponentChangeMsg  1009
+#define CID_ComponentDefaultCB  1010
+#define CID_ComponentInterpMsg  1011
+#define CID_Comment	1012
+#define CID_Color	1013
+#define CID_GClass	1014
+#define CID_Tabs	1015
 
-#define CID_TeX_Height	1012
-#define CID_TeX_Depth	1013
-#define CID_TeX_Italic	1014
-#define CID_HorAccent	1015
+#define CID_TeX_Height	1016
+#define CID_TeX_Depth	1017
+#define CID_TeX_Italic	1018
+#define CID_HorAccent	1019
 /* Room for one more here, if we need it */
-#define CID_TeX_HeightD	1017
-#define CID_TeX_DepthD	1018
-#define CID_TeX_ItalicD	1019
-#define CID_HorAccentD	1020
+#define CID_TeX_HeightD	1020
+#define CID_TeX_DepthD	1021
+#define CID_TeX_ItalicD	1022
+#define CID_HorAccentD	1023
 
-#define CID_ItalicDevTab	1022
-#define CID_AccentDevTab	1023
+#define CID_ItalicDevTab	1024
+#define CID_AccentDevTab	1025
 
-#define CID_IsExtended	1024
+#define CID_IsExtended	1026
 #define CID_DefLCCount	1040
 #define CID_LCCount	1041
 #define CID_LCCountLab	1042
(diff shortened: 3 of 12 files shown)


khaledhosny:
I used to do something like this from Python to build composite glyphs with arbitrary bases and marks, no idea if it still works:
```python
    glyph.addReference(base)                                                    
    glyph.useRefsMetrics(base)                                                  
    for mark in marks:                                                          
        glyph.appendAccent(mark)                                                
        glyph.build()
```

ctrlcctrlv:
@khaledhosny That may work, but putting it into the UI is much more user friendly...as is saving the information along with the font. As I wrote in #3617, there are many fonts where overriding the decompositions is necessary.

ctrlcctrlv:
@khaledhosny I hope you're not rejecting these changes, because I took four days to figure out how to add this feature, and it received tentative approval in #3617. :sob:

fontforgeexe/charinfo.c
```diff
@@ -1225,6 +1229,7 @@ static SplineChar *CI_SCDuplicate(SplineChar *sc) {
     newsc->unicodeenc = sc->unicodeenc;
     newsc->orig_pos = sc->orig_pos;
     newsc->comment = copy(sc->comment);
+    newsc->user_decomp = sc->user_decomp;
```
jtanx: No need to copy?

fontforgeexe/charinfo.c
```diff
@@ -1436,6 +1441,7 @@ return( false );
 	ci->cachedsc = chunkalloc(sizeof(SplineChar));
 	ci->cachedsc->orig_pos = ci->sc->orig_pos;
 	ci->cachedsc->parent = ci->sc->parent;
+    ci->cachedsc->user_decomp = ci->sc->user_decomp;
```
jtanx: ditto? (I'm not actually sure here, you need to check how this is used)

ctrlcctrlv:
> I haven't reviewed this fully, so there's probably still more, but I'll leave it here for now

Thanks for the review, I'll fix those things, or comment on the ones I had reasons for doing. :)

fontforgeexe/charinfo.c
```diff
@@ -3909,26 +3979,82 @@ static void CIFillup(CharInfo *ci) {
...
+    GGadgetSetEnabled(cotf, false);
+    GGadgetSetEnabled(cola, false);
+    GGadgetSetEnabled(codcb, false);
+    GGadgetSetChecked(codcb, true);
     } else {
-	unichar_t *temp = malloc(11*u_strlen(bits)*sizeof(unichar_t));
+	unichar_t *temp = malloc(18*u_strlen(bits)*sizeof(unichar_t));
 	unichar_t *upt=temp;
+    // i.e. five six hex digit long codepoints with spaces after them
+    char* codepoints_as_hex = malloc((6 * 5) + 5);
```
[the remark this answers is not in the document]
ctrlcctrlv: So, I know the maximum valid thing given by FontForge will look something like `01d481` (𝒁), so, five of those, with spaces after them, FontForge will never give more than that. In most cases it won't even get close.

fontforgeexe/charinfo.c
```diff
@@ -3653,6 +3659,69 @@ static int CI_DefLCChange(GGadget *g, GEvent *e) {
...
+
+    return out;
+}
+
+char* CI_CreateInterpretedAsLabel(unichar_t* inp) {
+    char* lblprefix = "Interpreted as: ";
+    char* lblerror = "Error: wrong format";
+    // 2 makes it large enough to hold the error string
+    int factor = inp == NULL ? 2 : u_strlen(inp);
+    char* lblbuf = malloc(strlen(lblprefix)+(factor*4));
```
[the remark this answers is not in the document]
ctrlcctrlv: How so? The maximum UTF-8 bytes a unichar_t can produce is four, and the factor is the length of the input. In the opposite case, the factor is just enough to hold the difference between `lblerror` and `lblformat`.

khaledhosny:
It was just an observation.

skef:
@ctrlcctrlv It looks like @jtanx is giving you good review comments and I'll take a look at this in more detail too, probably later today. However, with a high-level glance I think something is missing. I'm going to argue this in a little depth because I want the issues to be clear and to not just jerk you around with an arbitrary code review.  

There are various different places to put bits of configuration data. Given it's infrastructure FontForge offers two currently convenient places: the prefs file or in individual font files. One could also always add any number of other places in the form of additional files (which is, when you think about, pretty much the status of a feature file relative to FontForge).

One question raised by this change is "is the configuration data being put in the appropriate place?"  My own answer to this is: "probably and who cares anyway?". Lots of things in FontForge work like this already and it's fine and appropriate for this to also.

However, one argument against it is "Won't lots of users want to have the exact same mappings in all of their fonts?" This is a genuine concern worth addressing, and *given your choice of implementation* there's a conventional way of addressing it: You include a "merge this data from another font" feature. This is what is currently missing. A user could write a little python to do this work, but since your approach makes this a "top-level" feature, it's more proportional to make it part of the system. 

There's no exact, already existing model, but at the python level `importLookups` is close. I haven't looked in enough detail to see if you would need any arguments besides the other font, but you probably already have some idea. There should also presumably be a UI, which would be some variation on an "Import ..."  dialog. Maybe there's already an appropriate dialog in which this could be added as an option, or maybe there are a bunch of things that should have such a dialog but don't yet. (In the latter case there's an argument for you not being forced to do what others have avoided and this might be added to a "todo" list.)

ctrlcctrlv:
>You include a "merge this data from another font" feature.

I think there's already a way to merge fonts in FontForge. Do you mean that I should make it so this data carries over during a merge that way, or something else?

I see what you mean, that a user may want to use the same user decompositions for all their fonts, but although I might sound like an Apple CEO in saying so, they're using it wrong. As explained in #3617, for most glyphs _it really depends on the type of font you're making._ Especially &dcaron;, if you're making a handwriting or typewriting font, you are supposed to use the caron, ◌̌, as ď (might look wrong depending on font).

That's why I chose to attach it to the font and not the user's settings. In my own case, there are fonts where I'd want to use a certain set of decompositions, and other fonts where the defaults are correct and appropriate.

skef:
You should definitely make it so the data comes over when you merge an entire font. But you should also add a means of merging *just* this data from another font. When someone who doesn't "script" uses your UI to get their substitutions just right, they should have an easy way of moving that data to their other files without affecting anything else.

skef:
> As explained in #3617, for most glyphs it really depends on the type of font you're making.

True, but there's still the usual situation of Regular, Bold, Italic, BoldItalic and so forth. All of those are typically going to have the same compositions.

ctrlcctrlv:
I think I understand, so you think I should add "Export decompositions" and "Import decompositions" functions. Right? I can certainly work on that tomorrow while fixing the memory leaks I inadvertently added.

skef:
> so you think I should add "Export decompositions" and "Import decompositions" functions. Right?

Yes, exactly. That would make this implementation proportional to (most of) the other similar features in FontForge.

The UI may be tricky; you might ask for some feedback from The Powers That Be on the subject.

skef:
Oh, wait, let me correct my "Yes exactly": If you do what I'm suggesting you don't need an "Export", only an "Import", because the source would just be another font file. 

You could *also* decide to make a new file format to export to and import from, but for this case that seems like overkill to me. Just offering the option of grabbing the data from another `.sfd` seems entirely sufficient. A user who wants a separate file for the data could always just open a "blank" `.sfd`, add their data, and import from that.

khaledhosny:
What about a “composition” file, similar to encoding files? That would be easier to share and bulk edit, and in line with similar functionality.

skef:
> What about a “composition” file, similar to encoding files? That would be easier to share and bulk edit, and in line with similar functionality.

Does anyone actually do this? I haven't run across one in the wild, and I just tried doing some searches to turn one up and didn't find any.

frank-trampe:
How do other products deal with this (entry, storage, and manipulation of decomposition overrides), if at all? @davelab6?

skef:
Various contributors here might prefer to imagine that if @ctrlcctrlv had asked the dev user list how to implement this, he would have gotten an answer. This is doubtful -- questions there often go without response. 

He could have submitted an issue asking for something to solve this problem, too. That may have gone without response as well, but more likely would have garnered a "we're open source -- please consider submitting a pull request". There are dozens of these. 

Now he's submitted a pull request, with an implementation. And only now, in project-typical fashion, are people giving him implementation advice -- much of which is inevitably pointing in a very different direction than the one he took. 

This is how this project is managed. No one has the time or inclination to give substantive advice other than to suggest that people should implement solutions to their own problems, until it looks like there's a risk that something might actually happen, when they swoop down to explain How Things Should Really Be Done.

skef:
To be clear, I am not endorsing an anything-goes policy. I pushed back on #3596 for a few reasons, with the primary one being that it wasn't what the author really wanted anyway, and I was able to construct something very close without much effort from existing "materials" and he was able to determine what else was needed to solve his problem. 

This request is not like that. There are scripting hooks that can duplicate the actions triggered with the UI, but not those to influence those actions. The author has taken one of the two or three general approaches to the problem. To square-one him on this is to treat him as a remote robot or servant forced to implement someone else's design. This is rude and unnecessary at a project level even if not (necessarily) at the individual level. I am here to point at the rude and unnecessary phenomenon in the hope that people will take the larger picture into account and stop it.

frank-trampe:
@skef, the question of how to store and to manipulate these values did not come up in the issue thread, you were the one who brought the question up on the pull request, and it was a good thing to ask about. The issue thread was not neglected in any way, and it's inescapable that some things don't come to mind until there is already code in place. Some things just arise later. In this case, the question relates to follow-on enhancements unnecessary to merge the pull request, so it does not put anybody back at square one.

skef:
https://github.com/fontforge/fontforge/pull/3645#issuecomment-482096408 was my attempt to preemptively argue for the validity of his approach, along the way of pointing out one thing that it called for. Questioning all aspects of the approach is one form of square-oneing:

> How do other products deal with this (entry, storage, and manipulation of decomposition overrides), if at all?

If you didn't intend that to put all issues back on the table you could have worded it better. @ctrlcctrlv's [recent experience](https://github.com/fontforge/fontforge/pull/3619#issuecomment-479398890) trying to help with these practical problems has been frustrating enough as it is. And this arose from an [attempt to add a flag to treat the Unicode standard as standard](https://github.com/fontforge/fontforge/pull/3617) that didn't seem to be going anywhere either.

khaledhosny:
I knew I was going to regret commenting here again.

skef:
@khaledhosny you've explicitly made yourself not part of the project and on that basis I think you are fine being curmudgeonly from the sidelines. @ctrlcctrlv knows your deal. Not having a veto puts you in a different category. 

And anyway, I apologize for the way I went off on you earlier. I had a really short fuse from the documentation problems I was facing and that wasn't your fault or your responsibility.

frank-trampe:
@khaledhosny, I'm actually glad that we all get to discuss this while it has momentum. Thank you for participating.

frank-trampe:
@skef, I still don't think my comments would lead one to believe that I was advocating throwing this out and starting over. I'm sorry if they appeared that way to you.

skef:
> I'm sorry if they appeared that way to you.

A phrasing that is never uttered to communicate to the referent of "you".

frank-trampe:
@skef, I don't understand. I was responding to you, and "you" referred to you.

skef:
@frank-trampe seriously? How I perceive things is not and cannot be your responsibility. That phrasing is the canonical [non-apology apology](https://en.wikipedia.org/wiki/Non-apology_apology).

frank-trampe:
I know, but it seemed rude just to say "that's not what I meant." I'd normally say at this point that I'm sorry that my attempt to make my response sound nice offended you by its hollow niceties, but that would admittedly be ridiculous. :smile:

skef:
I accept that that was your intention and will drop this.

skef:
Looking at the code and poking around the UI, one thing this needs is a bit more thought about the scope of the functionality, and how radical you want to make it.

If I were a certain kind of font designer and came across this functionality I might be disappointed that I can't use this system to make, for example, precomposed petite-caps glyphs. Of course:

1. It's not your "job" to make that font designer happy -- you're addressing a more specific problem and there are other resources that person can draw on, such as the approach @khaledhosny noted. 
2. Making that font designer happy would actually be pretty tricky, if for no other reason than you'd have to allow the use of glyph names instead of unicode numbers in the implementation. That opens a big can of worms. 

So let's assume you want the system to have the scope that's currently implied: It's to specify alternative compositions for glyphs with unicode values that correspond to precomposed characters. Hence the dialog being disabled in other cases. That's a sensible scope, but I think there's some more work to do to maintain it:

1. There's currently no unicode value checking (that I can see) at the python level. What if someone assigns to a glyph with a non-precomposed unicode value? Or a unicode value of -1? 
2. There's also no string validity checking (that I can see) at either the UI or python level. What if a user enters the unicode for two "base" characters? Shouldn't it always be one base character followed by some number of mark characters? (I'm actually not sure about this.) 
3. The field should probably cleared if the user changes the unicode value of the glyph, or given a warning and an opportunity to cancel. Probably the former.
4. It might or might not also be worth validating the string and the unicode value when reading it in from an `.sfd`. If there's a function to do this at the python level it might be easy to apply there as well.

Given this scope, it's probably OK that you only support reading in Unicode strings in the python interface, because such a string is needed to specify a "mark"/combining character. 

On a different issue, you should verify an older version of the software can skip over `Decomposition: ` lines, and consider bumping the SFD version number. Doing both would be ideal, but if the program can skip over unrecognized lines and will complain and/or barf on encountering an SFD with a higher version than it recognizes, it could be a reasonable compromise to skip the number change. (I don't know the answer to any of these questions.)

ctrlcctrlv:
@skef Just to answer one of the many questions you posed me since it's the simplest for me to answer right now, I purposely made it so you could enter multiple base characters. For example, in my [Some Time Later font](https://github.com/ctrlcctrlv/some-time-later), &aring; is built out of `a` and `o`&mdash;of course, this is really only possible in very "wacky" fonts. I tested multiple bases and it works as I intended.

skef:
> I purposely made it so you could enter multiple base characters. For example, in my Some Time Later font, å is built out of a and o—of course, this is really only possible in very "wacky" fonts. I tested multiple bases and it works as I intended.

In that case the limitation in the python setter is incorrect, and you also need to read in non-unicode strings and up-convert them. ("ao" would just be a normal string, at least in Python 2.) 

It also at least raises questions about whether the functionality should be limited to precomposed unicode characters, as it seems to be now. "Potential wackiness" is, in effect, one kind of generality. Do you want to allow -1 encoded characters to be precomposed as well? If you can compose two base characters, should you be able to compose unencoded characters? (Again, the latter would likely be much more tricky to implement.)

ctrlcctrlv:
@skef I'm not sure what you mean, in my fonts if I'm going to have "weird" characters (KJV1611 has two stylistic alternates for the capital "B"), I just encode them in the PUA. That's what I'd expect people to do if they wanted to use them in this dialog. Is that wrong?

skef:
It goes against the advice most often given in other issues that I've run across. But no, it's not wrong, or not obviously wrong. More knowledgeable folks can weigh in on the pluses and minuses of PUA encodings and their "interactions". It could be an acceptable workaround for "wacky" cases.

ctrlcctrlv:
Right...the only way I can think of to prevent that would be to make it so that you must refer to characters by their names and not their encodings, then from Python, have the type be a tuple and/or list of glyph names. Of course, that's quite a big change, and I think especially from the Python side this is much more obvious.

skef:
On the PUA front: you don't currently appear to allow setting components for PUA characters. A more wacky implementation would support that.

ctrlcctrlv:
@skef Yes, I agree that I should enable setting components even for characters FontForge doesn't think have any. :)

skef:
Well, one could make an argument for supporting that in the PUA space but not outside of it. It's a matter of how much rope you want to hand to users. Is there really a good reason for automatic composition at unicode codepoints for base or mark glyphs, is the question.

ctrlcctrlv:
> It's a matter of how much rope you want to hand to users.

Enough to hang themselves with and then some! :laughing: :sunglasses:

ctrlcctrlv:
To be serious though, while prototyping this I thought, "this should really work for &oelig; and &aelig;", and your comments just cement that idea into me that I should open up user decompositions to all characters.

skef:
Note that more rope doesn't always mean fewer checks, it can mean different checks. For example: I'm reasonably confident a user should not be able to compose a character with itself. Or create cycles of any kind. Part of the role of the existing Unicode structure is to rule out cases like that.

skef:
FontForge already has a Unicode-based ligature composition system. Is that implemented using the same system you're modifying?

ctrlcctrlv:
Yes, as far as I can tell the distinction between composing an accented glyph or a ligature only exists in the UI, Python `build()` will do whatever feels appropriate based on return of `SFGetAlternate`, which `sc->user_decomp` now influences.

ctrlcctrlv:
And I don't know if you've built the branch yet and tried it out, but if you enter `ao` as the decomposition for &aring;, the UI will treat it like a ligature, unless `o` has an anchor point (I'm pretty sure&mdash;haven't tested recently)

skef:
In that case it would be nice if one could add compositions to -1 encoded glyphs, even if the characters in the composition have to have proper encodings.

ctrlcctrlv:
I can't get `ao` to build as an accented glyph even by adding anchor points, so I'll have to fix that as well. Thanks for the ideas!

ctrlcctrlv:
Oh&mdash;I just wanted to say&mdash;earlier versions of FontForge can handle the `Decomposition` line just fine. I'll probably even push a version of `Chomsky.sfd` soon that has decompositions, the current one is just depending on `UnicodeStrict` which I'm no longer using in favor of this.

ctrlcctrlv:
@skef I forgot to ask this, but why are we still targeting Python 2 compatibility for new features? I'm not totally against working around Python 2's weird Unicode system, but I suppose what I'm saying is, like a child asked by their parent to complete an unpleasant homework assignment, _do I have to?_

frank-trampe:
@ctrlcctrlv, although Python 2 is officially going over a cliff soon, a lot of people still use it, and some of our package builds still default to it, so we can't break it quite yet.

ctrlcctrlv:
@frank-trampe I won't break the build on Python 2, but I'm asking if I really have to make this new feature available/easy to use on Python 2, given its imminent removal.

frank-trampe:
I don't think that we'd get flamed on the mailing list for leaving out Python 2 support for a new feature, but FontForge is likely to support Python 2 for a while past the official cliff, so a lot of people might appreciate having it. I'd prefer to see this merged without Python 2 support rather than to have it languish awaiting more free time. We can always add later.

fontforgeexe/charinfo.c
```diff
@@ -74,25 +74,29 @@ typedef struct charinfo {
...
+#define CID_ComponentInterpMsg  1011
+#define CID_Comment	1012
+#define CID_Color	1013
+#define CID_GClass	1014
+#define CID_Tabs	1015
+
+#define CID_TeX_Height	1016
+#define CID_TeX_Depth	1017
+#define CID_TeX_Italic	1018
+#define CID_HorAccent	1019
```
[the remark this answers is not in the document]
ctrlcctrlv: Sorry about that , me too to be honest. The problem is, OK, I could add them at the bottom. But then the new Composition labels, checkbox etc isn't under its peers, that looks weird. I could add them under the older Composition UI stuff but just with higher numbers, but then future programmers won't have an obvious way to know where to start with theirs, and things become a jumbled mess. So I chose to increment. I know it sucks, but it's the least worst option.

ctrlcctrlv:
Just an end of day report.

@jtanx&hellip;I cleared up all the memory leaks you pointed out, I think.
@skef&hellip;I added a basic Python truncation and got `ao` to build as an accented glyph provided the base and the mark had anchor points. See comment on https://github.com/fontforge/fontforge/pull/3645/commits/07bedbe911fbbf8a50132f5685f95d66b791c485 for more about that.

I believe this could be merged right now and not be a detriment. However, I'm willing to keep discussing the importing of decompositions from other fonts. Still figuring out the best way to do that, I have _no idea_ where I should put that in the UI.

ctrlcctrlv:
Oh! And @frank-trampe, I made sure this builds with Python 2, and scripts that don't use this feature work, and this feature even works under some circumstances but not all the cases I tested in Python 3.

fontforge/python.c
```diff
@@ -7242,6 +7242,45 @@ static int PyFF_Glyph_set_vhints(PyFF_Glyph *self,PyObject *value, void *UNUSED(
 return( PyFF_Glyph_set_hints(self,true,value));
 }
 
+static PyObject *PyFF_Glyph_get_user_decomp(PyFF_Glyph *self, void *UNUSED(closure)) {
+    if ( self->sc->user_decomp==NULL ) {
+        return( Py_BuildValue("s", "" ));
+    }
+    else {
+        char* out = u2utf8_copy(self->sc->user_decomp);
+        return( PyUnicode_DecodeUTF8(out, strlen(out), NULL) );
```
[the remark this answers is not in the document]
jtanx: This is not fixed, because you return before you free, store the return value temporarily, free then return

fontforge/python.c
```diff
@@ -7887,12 +7935,20 @@ static PyGetSetDef PyFF_Glyph_getset[] = {
...
 /* ************************************************************************** */
 
-static PyObject *PyFFGlyph_Build(PyObject *self, PyObject *UNUSED(args)) {
+static PyObject *PyFFGlyph_Build(PyObject *self, PyObject *args) {
     SplineChar *sc = ((PyFF_Glyph *) self)->sc;
     int layer = ((PyFF_Glyph *) self)->layer;
+    int accent_hint = false;
+    PyObject *accent_hint_pyo;
+
+    if ( PyArg_ParseTuple(args, "|O", &accent_hint_pyo) ) {
```
jtanx: Why not just use `"|p"` and pass `accent_hint` directly into the parser?

fontforgeexe/charinfo.c
```diff
@@ -3653,6 +3659,73 @@ static int CI_DefLCChange(GGadget *g, GEvent *e) {
 return( true );
 }
 
+// Turn the user's entered decomposition into the format returned by SFGetAlternate
+unichar_t* CI_ParseUserDecomposition(char* inp) {
+    unsigned long len = strlen(inp);
+    char* end;
+    unichar_t* out = malloc(len*sizeof(unichar_t)+1);
```
jtanx: `(len+1)*sizeof(unichar_t)`

fontforgeexe/charinfo.c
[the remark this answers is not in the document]
jtanx: If that's the case, then aren't you missing one for the null terminator?

Anyway, why I say this is dodgy is not so much because you *can* inspect what the max will be, it's that it's a fragile approach if the assumptions change (e.g. what if someone came along and decided to change that error message), and you can (and probably did, as it is) get it wrong, vs. just computing the length and allocating after the fact.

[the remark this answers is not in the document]
jtanx: Another problem with this is that it's not gettext'ed, so it's not translatable, use the `_()` macro (although I forget how you use it with format variables - you'll need to look that one up)

fontforge/python.c
```diff
@@ -7242,6 +7242,51 @@ static int PyFF_Glyph_set_vhints(PyFF_Glyph *self,PyObject *value, void *UNUSED(
...
+	Py_DECREF(temp);
+    } else {
+        return -1;
+    }
+
+    if ( newv==NULL ) {
+        return -1;
+    }
+
+    if (strlen(newv) > 5) {
```
jtanx: 5 bytes or 5 characters?

fontforge/python.c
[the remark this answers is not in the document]
ctrlcctrlv: derp. sorry

fontforge/python.c
ctrlcctrlv: @jtanx I can't use `p` because it's "New in version 3.3". @frank-trampe said to _at least_ make sure this builds under 2.x, and that old scripts keep running under 2.x.

fontforge/python.c
ctrlcctrlv: @jtanx I thought since I'm only allowing Unicode, it would be 5 characters. Turns out that assumption was incorrect, good catch :-)

ctrlcctrlv: Fixed in https://github.com/fontforge/fontforge/pull/3645/commits/a3135465fd4aa05d984d7f06870d5b13601fb71b (I hope)

fontforgeexe/charinfo.c
ctrlcctrlv: Fixed in https://github.com/fontforge/fontforge/pull/3645/commits/a3135465fd4aa05d984d7f06870d5b13601fb71b (I hope)

ctrlcctrlv:
https://github.com/fontforge/fontforge/pull/3645/commits/a3135465fd4aa05d984d7f06870d5b13601fb71b _should_ fix everything requested of me so far. Thanks for all the comments @jtanx, helps a lot.

frank-trampe:
@ctrlcctrlv, no need to wait on the import/export stuff. Let's merge this as soon as @jtanx clears it so that it makes the next release.

jtanx:
As mentioned prior, I'm not targeting new features for this release

fontforge/python.c
```diff
@@ -7242,6 +7242,62 @@ static int PyFF_Glyph_set_vhints(PyFF_Glyph *self,PyObject *value, void *UNUSED(
...
+        if (self->sc->user_decomp != NULL) { free(self->sc->user_decomp); }
+        self->sc->user_decomp = NULL;
+        return 0;
+    }
+
+    if ( PyUnicode_Check(value)) {
+        /* Need to force utf8 encoding rather than accepting the "default" */
+        /*  which would happen if we treated unicode as a string */
+        temp = PyUnicode_AsUTF8String(value);
+        newv = copy( PyBytes_AsString(temp));
```
jtanx: Get rid of `newv`, just assign to `udbuf` directly here (by calling `utf82u_copy`), otherwise you're leaking `newv` down the line as you never free it

fontforge/python.c
```diff
@@ -7887,12 +7946,20 @@ static PyGetSetDef PyFF_Glyph_getset[] = {
 /*  Glyph Methods  */
 /* ************************************************************************** */
 
-static PyObject *PyFFGlyph_Build(PyObject *self, PyObject *UNUSED(args)) {
+static PyObject *PyFFGlyph_Build(PyObject *self, PyObject *args) {
     SplineChar *sc = ((PyFF_Glyph *) self)->sc;
     int layer = ((PyFF_Glyph *) self)->layer;
+    int accent_hint = false;
+    PyObject *accent_hint_pyo;
```
jtanx: Default assign this to NULL, I'm fairly sure that `PyArg_ParseTuple` doesn't touch it if that optional argument isn't present

skef:
It's sensible to be conservative in a short-span fixit release like this one. I'm glad to see both a mac work-around and a solid candidate for the GDK transient difficulties.

fontforgeexe/charinfo.c
```diff
@@ -3653,6 +3659,74 @@ static int CI_DefLCChange(GGadget *g, GEvent *e) {
...
+
+    out[j] = '\0';
+
+    return out;
+}
+
+char* CI_CreateInterpretedAsLabel(unichar_t* inp) {
+    char* lblprefix = _("Interpreted as: ");
+    char* lblerror = _("Error: wrong format");
+    // 2 makes it large enough to hold the error string
```
jtanx: delete this comment

fontforgeexe/charinfo.c
```diff
@@ -3653,6 +3659,74 @@ static int CI_DefLCChange(GGadget *g, GEvent *e) {
...
+        }
+    }
+
+    if (inp != NULL && inp[0] != 0 && valid) {
+        char* inp_l = u2utf8_copy(inp);
+        lblbuf = malloc(strlen(lblprefix)+strlen(inp_l)+1);
+        sprintf(lblbuf, "%s%s", lblprefix, inp_l);
+        free(inp_l);
+    } else {
+        lblbuf = malloc(strlen(lblerror)+1);
```
jtanx: Can simplify this to just be `lblbuf = copy(lblerror);`

ctrlcctrlv:
Sure, make your planned release first, then merge this, then while that's going on, I'll think more about the import/export, and open a new PR/issue for that seeking comment.

I've also finally subscribed to the SourceForge mailing lists in case anyone posts a comment/question for me there.

fontforgeexe/charinfo.c
```diff
@@ -3909,28 +3984,81 @@ static void CIFillup(CharInfo *ci) {
...
 	    ++bits;
 	}
 	upt[-1] = '\0';
 	GGadgetSetTitle(GWidgetGetControl(ci->gw,CID_Components),temp);
 	free(temp);
+
     }
 
+    // i.e. five six hex digit long codepoints with spaces after them
+    char* codepoints_as_hex = malloc((6 * 5) + 5);
```
jtanx: Again, supposing it really maxes out, isn't this still missing space for the null terminator?

fontforgeexe/charinfo.c
```diff
@@ -3653,6 +3659,74 @@ static int CI_DefLCChange(GGadget *g, GEvent *e) {
...
 }
 
+// Turn the user's entered decomposition into the format returned by SFGetAlternate
+unichar_t* CI_ParseUserDecomposition(char* inp) {
+    unsigned long len = strlen(inp);
+    char* end;
+    unichar_t* out = malloc((len+1)*sizeof(unichar_t));
+
+    int j = 0;
+    for (unsigned long i = strtoul(inp, &end, 16); inp != end; i = strtoul(inp, &end, 16)) {
```
jtanx: So one thing I don't quite understand - there's references to this being limited to at most 5 characters, but this seems to allow for an arbitrary length?

ctrlcctrlv: I didn't think this was a big problem, but it is inconsistent, so I limited it.

fontforge/python.c
```diff
@@ -7271,21 +7271,20 @@ static int PyFF_Glyph_set_user_decomp(PyFF_Glyph *self,PyObject *value, void *UN
         /* Need to force utf8 encoding rather than accepting the "default" */
         /*  which would happen if we treated unicode as a string */
         temp = PyUnicode_AsUTF8String(value);
-        newv = copy( PyBytes_AsString(temp));
+        udbuf = utf82u_copy(PyBytes_AsString(temp));
```
jtanx: Now that `newv` is unused, remove the declaration at the top too

fontforgeexe/charinfo.c
jtanx: One area where it might have been a problem is `codepoints_as_hex`, since there's an assumption of a max of 5

fontforgeexe/charinfo.c
```diff
@@ -3659,28 +3659,43 @@ static int CI_DefLCChange(GGadget *g, GEvent *e) {
 return( true );
 }
 
+struct ParsedDecomp_s {
+    unichar_t* decomp;
+    bool truncated;
+};
+
+typedef struct ParsedDecomp_s ParsedDecomp;
+
```
jtanx: Personal preference, but I don't think defining a struct for this is necessary. I'd do something like

```c
int CI_ParseUserDecomposition(char *inp, unichar_t **decomp) {
    *decomp = malloc(blah);
    ....
    return truncated;
}
```

fontforge/sfd.c
```diff
@@ -5834,6 +5841,10 @@ exit(1);
 	    sc->color = temp;
 	} else if ( strmatch(tok,"Comment:")==0 ) {
 	    sc->comment = SFDReadUTF7Str(sfd);
+	} else if ( strmatch(tok,"Decomposition:")==0 ) {
+	    char* decomp = SFDReadUTF7Str(sfd);
+	    sc->user_decomp = utf82u_copy(decomp);
```
jtanx: Limit to 5 chars too?

The other option is to just make it safe to have an arbitrary length

ctrlcctrlv:
@jtanx When I was originally working on this, I didn't know

ctrlcctrlv:
Oops

ctrlcctrlv:
@jtanx When I was originally working on this I didn't know https://github.com/fontforge/fontforge/pull/3645#issuecomment-482424131 ... So that's why I added the five character limit, I thought it was a sane limit for *accented* characters. However, as I've been adding it to more and more of the UI with this in mind it's making less and less sense especially for ligatures. I have a font with a four character long ligature and can easily imagine one with six being useful.

So when I get home today I'll remove all the limits for sanity's sake. :)

fontforgeexe/charinfo.c
ctrlcctrlv: Probably just Rust influence on me :) I prefer this way, but will change it if you insist.

ctrlcctrlv:
@jtanx Removed the limits. What I just sent to the fontforge-devel ML (subject: Gdraw - setting popup message?) is not a blocker for this FYI so you can still merge it if you want to and if I fixed everything to your satisfaction :)

fontforgeexe/charinfo.c
```diff
@@ -3909,27 +3994,92 @@ static void CIFillup(CharInfo *ci) {
...
+    // i.e. five six hex digit long codepoints with spaces after them, + '\0'
+    int d_length;
+    if (sc->user_decomp != NULL) {
+        d_length = u_strlen(sc->user_decomp);
+    } else {
+        d_length = u_strlen(bits2);
+    }
+
+    const int MAX_UNICHAR_T_BYTES = 4;
+    char* codepoints_as_hex = malloc(((2 * MAX_UNICHAR_T_BYTES) * d_length) + d_length + 1);
```
jtanx: Free this

fontforgeexe/charinfo.c
```diff
@@ -3800,13 +3872,26 @@ static int CI_PickColor(GGadget *g, GEvent *e) {
 return( true );
 }
 
+int BytesNeeded(unichar_t in) {
```
jtanx: Might as well mark this as static

fontforgeexe/charinfo.c
```diff
@@ -3909,27 +3994,92 @@ static void CIFillup(CharInfo *ci) {
...
 	    upt += u_strlen(upt);
+
 	    ++bits;
 	}
 	upt[-1] = '\0';
 	GGadgetSetTitle(GWidgetGetControl(ci->gw,CID_Components),temp);
 	free(temp);
     }
+    if (bits2 == NULL) {
+        bits2 = (unichar_t*) L"\0";
```
jtanx: Don't do this, `wchar_t` has an implementation defined width (i.e. on windows it's 2 bytes, not 4 bytes). Given the use case below, something like this should work:

```c
if (d_ptr == NULL) d_ptr = bits2;

int d_length = d_ptr ? u_strlen(d_ptr) : 0;
```

Especially since you check for `d_ptr` being null in the while loop

fontforgeexe/charinfo.c
```diff
@@ -3909,27 +3994,90 @@ static void CIFillup(CharInfo *ci) {
...
+
 	    ++bits;
 	}
 	upt[-1] = '\0';
 	GGadgetSetTitle(GWidgetGetControl(ci->gw,CID_Components),temp);
 	free(temp);
     }
+    if (d_ptr == NULL) d_ptr = bits2;
+
+    int d_length = d_ptr ? u_strlen(d_ptr) : 0;
```
jtanx: Ah sorry, should have been clearer. With this, you can delete lines 4037-4040, since you've already computed the length here.

ctrlcctrlv:
Should I make a null commit to cause AppVeyor to try again? The failure has nothing to do with my code, it's just a network error that happens sometimes to every PR...

jtanx:
Nah that's fine. Just one more I think, update SplineCharFreeContents in splineutil.c to free the decompositon if it's present

davelab6:
Just my opinion, but I think lifting a finger with regards to anything for
python 2 support is a bit bizarre. It's gotta die. Help it go :)

PR #3648, opened 2019-04-12
GDK: Yet another transient fix attempt

### LINKED ISSUE #3611: GDK modality freeze sequence
This is with a recent build on Arch Linux with GDK:

1. Run `fontforge tests/fonts/ItalicHand.sfd` (I don't think the font matters but it's the one I was using.
2. Open Metrics -> Kern Pair Close-up, enter "a" and "b" as glyphs. 
3. Pick "New Lookup Subtable" from the lower menu, make it a "kern" and press OK, accepting the default lookup name.

At this point the "Could not call GGDKDrawSetTransientFor" warning prints, but more importantly all windows are "grabbed", leaving the program effectively frozen.

jtanx:
This is an attempt to fix #3611, while preserving other behaviour otherwise.

### DIFF
--- a/gdraw/ggdkdraw.c
+++ b/gdraw/ggdkdraw.c
@@ -192,14 +192,6 @@ static gboolean _GGDKDraw_OnWindowDestroyed(gpointer data) {
 
     if (!gw->is_pixmap) {
         if (gw != gw->display->groot) {
-            if (gw->is_toplevel) {
-                for (int i = ((int)gw->display->transient_stack->len) - 1; i >= 0; --i) {
-                    GGDKWindow tw = (GGDKWindow)gw->display->transient_stack->pdata[i];
-                    if (tw->transient_owner == gw) {
-                        GGDKDrawSetTransientFor((GWindow)tw, NULL);
-                    }
-                }
-            }
             if (!gdk_window_is_destroyed(gw->w)) {
                 gdk_window_destroy(gw->w);
                 // Wait for it to die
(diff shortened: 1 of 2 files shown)


ctrlcctrlv:
Do you want help testing this, or are you about to merge it into master?

jtanx:
Yeah for sure, if you want to test it, that'd be great, thanks

ctrlcctrlv:
Cool, will do :-)

gdraw/ggdkdraw.c
```diff
@@ -593,23 +608,30 @@ static GWindow _GGDKDraw_CreateWindow(GGDKDisplay *gdisp, GGDKWindow gw, GRect *
...
-            } else if (gdisp->top_window_count > 0) {
-                Log(LOGWARN, "Could not call GGDKDrawSetTransientFor on a restricted window (%p %s), restrict count %d, transient window count %d",
-                    nw, nw->window_title, gdisp->restrict_count, (int)gdisp->transient_stack->len);
-            }
+        } else if (nw->restrict_input_to_me && gdisp->toplevel_windows->length > 0) {
+            GGDKDrawSetTransientFor((GWindow)nw, (GWindow) - 1);
         }
         nw->isverytransient = (wattrs->mask & wam_verytransient) ? 1 : 0;
-        nw->is_toplevel = true;
+
```
skef: This use of `g_list_append` might be worth a comment something like "Create a list link element pointing to the window"

gdraw/ggdkdraw.c
```diff
@@ -1573,7 +1612,13 @@ static void GGDKDrawSetTransientFor(GWindow transient, GWindow owner) {
...
 
     if (owner == (GWindow) - 1) {
-        ow = gdisp->last_nontransient_window;
+        for (GList_Glib *pw = gdisp->toplevel_windows->head; pw != NULL; pw = pw->next) {
+            GGDKWindow tw = (GGDKWindow)pw->data;
+            if (tw != gw && tw->is_visible) {
+                ow = tw;
+                break;
+            }
+        }
```
skef: So to confirm, the idea hear is to find *some* window for this one to be transient to *now*, regardless of whether there's an actual arrangement and even if that window may be destroyed before this one, because GDK needs it for `gdk_window_set_transient_for`? And we won't wind up with a "dangling transient" because of the loop in `_GGDKDraw_OnWindowDestroyed` (which calls back into here)?

jtanx: The intent is to select the most recently focused window, and it shouldn't matter what gets destroyed first, because of the transient reparenting code. 

In terms of there being an 'actual arrangement' - there's no way at this point to know what the arrangement should be (and is one of my gripes with GDraw - it should have made everything explicit about what it's transient to), so selecting the most recently focused window is a best-effort attempt at that.

This is basically what the `last_nontransient_window` was trying to do, but it also works in the case where the `last_nontransient_window` is being destroyed, which causes `last_nontransient_window` to be unset (and before it gets set again) - this is the bulk of the real fix.

To be honest though, I wouldn't expect the transient reparenting code to ever run, but in the case it does happen, we can't destroy the transient window, because it's more than likely the caller will call GDrawDestroyWindow on it (and you can't destroy more than twice).

skef: Ah, OK, so the theory is that the most recently focused window is the best candidate for a transient parent for a new window, and the fact that it's also the target of the reparenting code (because it calls the same function) is more or less irrelevant, because that code is expected to run rarely or never.

jtanx: Yep, that's my current theory

skef:
My comment quibbles aside, this looks good, and although I can't run through all the cases in my head, reasonably likely to fix the known issues.

jtanx:
Sorry, did a bit of renaming, plus I also moved the removal step into `GGDKDrawDestroyWindow` instead of waiting until the final cleanup step ([here](https://github.com/fontforge/fontforge/compare/e4819d44e9bc34b9fe02f027ac56fbc2845f5370..7d0b8e502e577088fd6c99b59568fd943ed83027)) - why I had it there before was because I thought I could replace `top_window_count` with the length of that toplevel window list, but because the semantics are different, I couldn't do that.

Since that's the case, there's no reason to keep it in the list for longer than necessary (and may prevent some of that potential for churn noted)

skef:
Easier to review these smaller changes without the squash

skef:
Don't see any new issues

jtanx:
> Easier to review these smaller changes without the squash

Sorry, yeah was a bit lazy

PR #3651, opened 2019-04-12
Fix for #3649 Error when generating fonts with MathKerning

### LINKED ISSUE #3649: Error when generating fonts witn MathKerning
- [x] Screenshot

Ver: Build 2019-3-17; Windows 10.
To reproduce:
- Create a new file, draw anything in glyph `A` and `B`;
- Create a `Math Kerning` for `A`, and link `B` as a `vert. Variants` of`A`;
- Save file;
- Generate OTF.

And the `Math Kerning` and `Vert. Variants` info is missing in the OTF.
This seems to be the problem of the newest version, the old version does not have this issue.
The `sfd` file is attached here: [bug.zip](https://github.com/fontforge/fontforge/files/3074072/bug.zip)

Screen shots:
Create and Save
image
image
image
image
Error (open the generated OTF):
image
image

skef:
Set a vertical variant and it doesn't export, even though it seemed to in the past. Why? 

Turns out there's a big block of constants in `struct MATH` off of `SplineFont` that starts out NULL. If in a given font you open the "MATH Info" dialog at least once it is initialized to reasonable values, which you might or might not change. From that point on FontForge will save that table in an SFD and export it when the relevant flags are set. 

However, it's possible to add character-level math values, such as vertical variants, without ever opening that dialog. And the current check for whether any data are exported is whether the `MATH` pointer is non-NULL. So if you ever opened that dialog, even years ago without changing any values, your vertical variants will export. But you never opened it they won't. 

It might be valid to skip the block of constants when outputting by setting the table offset to 0, but a) some implementations might expect it there if the table exists at all, and b) it's not like FontForge being super-careful about the table anyway. So this change just allocates a fresh table if there isn't one already, using the already-existing function for that purpose. And since there's no opportunity to change it, it's freed rather than kept to avoid writing out default values in the `.sfd` unnecessarily.   

Closes #3649 

### Types of changes

### Final checklist

### DIFF
--- a/fontforge/tottfgpos.c
+++ b/fontforge/tottfgpos.c
@@ -36,6 +36,7 @@
 #include "splinesaveafm.h"
 #include "splineutil.h"
 #include "tottf.h"
+#include "mathconstants.h"
 #include <utype.h>
 #include <ustring.h>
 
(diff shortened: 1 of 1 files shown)


frank-trampe:
The change looks safe to me, in terms of not leaking or crashing and also in terms of the exported file. @YuanshengZhao, are you able to test?

skef:
Low risk, reviewed, and the submitter has confirmed.

PR #3657, opened 2019-04-14
Always enable glif hashing and switch from uthash to glib

jtanx:
As I was working on trying out cmake/meson, it became apparent that it doesn't actually build without uthash present (i.e. without FF_UTHASH_GLIF_NAMES defined, things break).

So the intent of this PR is twofold:
* Always enable glif name hashing, since it's basically a hard requirement (and likely has been for years). At the same time, make the implementation of the glif name hash opaque, so that it's easy to switch out the implementation
* Switch to using glib over uthash. glib is already a hard dependency in FontForge, and using it means that we don't have to continue bundling uthash with the source distribution. It's also just one less dependency to manage.

Tested manually and also with the following:

```c
#include "glif_name_hash.h"
#include <stdio.h>
#include <string.h>

int main() {
    for (int k = 0; k < 100; k++) {
        struct glif_name_index *idx = glif_name_index_new();
        
        for (int i = 0; i < 100; i++) {
            char buf[100];
            sprintf(buf, "ENTRY %d", i);
            glif_name_track_new(idx, i, buf);
        }
        
        for (int i = 0; i < 100; i++) {
            char buf[100];
            sprintf(buf, "ENTRY %d", i);

            struct glif_name *entry =  glif_name_search_glif_name(idx, buf);
            if (entry == NULL) {
                printf("ERROR: Not found: %d (%s)\n", i, buf);
                return 1;
            }
            
            if (entry->gid != i) {
                printf("ERROR: gid mismatch: %ld vs %d (%s vs %s)\n", entry->gid, i, entry->glif_name, buf);
                return 2;
            }
            
            if (strcmp(entry->glif_name, buf)) {
                printf("ERROR: name mismatch: %s vs %s (%d)\n", entry->glif_name, buf, i);
                return 3;
            }
        }

        for (int i = 100; i < 200; i++) {
            char buf[100];
            sprintf(buf, "ENTRY %d", i);
            
            struct glif_name *entry =  glif_name_search_glif_name(idx, buf);
            if (entry != NULL) {
                printf("ERROR: Unexpected: %d (%s) - got: (%ld, %s)\n", i, buf, entry->gid, entry->glif_name);
                return 1;
            }
        }
        
        glif_name_index_destroy(idx);
    }

    return 0;
}
```

Compiled with `gcc -g -std=c99 -Ifontforge -I/usr/local/include  -I/usr/local/include/fontforge test.c  -L/usr/local/lib -lfontforge -otest -Wall`

### DIFF
--- a/fontforge/ufo.c
+++ b/fontforge/ufo.c
@@ -68,9 +68,7 @@
 #undef extended			/* used in xlink.h */
 #include <libxml/tree.h>
 
-#ifdef FF_UTHASH_GLIF_NAMES
 #include "glif_name_hash.h"
-#endif
 
 /* The UFO (Unified Font Object) format ( http://unifiedfontobject.org/ ) */
 /* is a directory containing a bunch of (mac style) property lists and another*/
@@ -219,11 +217,7 @@ char * ufo_name_mangle(const char * input, const char * prefix, const char * suf
 }
 
 char * ufo_name_number(
-#ifdef FF_UTHASH_GLIF_NAMES
 struct glif_name_index * glif_name_hash,
-#else
-void * glif_name_hash,
-#endif
 int index, const char * input, const char * prefix, const char * suffix, int flags) {
         // This does not append the prefix or the suffix.
         // The specification deals with name collisions by appending a 15-digit decimal number to the name.
@@ -239,7 +233,7 @@ int index, const char * input, const char * prefix, const char * suffix, int fla
         }
         char * name_base = same_case(input); // This is in case we need a number added.
         long int name_number = 0;
-#ifdef FF_UTHASH_GLIF_NAMES
+
         if (glif_name_hash != NULL) {
           if (strlen(input) > (255 - 15 - strlen(prefix) - strlen(suffix))) {
             // If the numbered name base is too long, we crop it.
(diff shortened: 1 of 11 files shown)


fontforge/ufo.c
```diff
@@ -3409,14 +3301,7 @@ static void MakeKerningClasses(SplineFont *sf, struct ff_glyphclasses *group_bas
   // This silently ignores already extant groups for now but avoids duplicates unless group_base has internal duplication.
   int left_count = 0, right_count = 0, above_count = 0, below_count = 0;
   int left_start = 0, right_start = 0, above_start = 0, below_start = 0;
-#ifdef FF_UTHASH_GLIF_NAMES
-    struct glif_name_index _class_name_hash;
-    struct glif_name_index * class_name_hash = &_class_name_hash; // Open the hash table.
-    memset(class_name_hash, 0, sizeof(struct glif_name_index));
-    HashKerningClassNames(sf, class_name_hash);
```
jtanx: This was a little bit of a mystery; it does this hashing, but then never uses the hash later on. I'm pretty sure `HashKerningClassNames` has no side effects on `sf` too.

frank-trampe: The plan was that this function would check old names and not add duplicates, but this seems not to have happened, probably because it was unnecessary in the current use case. So my preceding comment about ignoring extant groups is probably wrong.

jtanx:
(just restarting the ci builds)

PR #3658, opened 2019-04-16
Conform to Adobe feature file spec section 7.b

skef:
A single substitution is arguably a degenerate case of a multiple, ligature, or "alternate" substitution. In #3350 I extended the already-existing treatment of a single as a multiple to include ligature and alternate substitutions, so that lookups of those types could also (conveniently) include single substitution entries. 

The existing code did this in both lookup and feature blocks, but in the later case did not verify that all types of a lookup matched. As it seemed strange to convert singles to multiples without that check, I added it. This turned out [to break at least one feature file](https://github.com/fontforge/fontforge/pull/3350#discussion_r275148092). 

Looking into this, my original thoughts were both right and wrong. It was right that converting the type of singles without checking if all the types matched was strange. It was wrong that any of that should be happening. The specification [says](https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#7b-ordering-of-lookups-and-subtables):

> A lookup in the OpenType font will be created from each named lookup block (§4.e) or each run of rules with the same feature, script, language, lookupflag and lookup type attribute.

Note the term "run of". 

Ignoring the checks I added, the existing implementation doesn't do this. And it's confusing, so bear with me. The relevant code (in `featurefile.c:fea_ApplyFeatureList`) is:

```
          case ft_pst:
          case ft_pstclass:
          case ft_ap:
          case ft_fpst:
            if ( f->type==ft_ap && f->u2.ap->type==at_mark ) {
                struct feat_item *n, *a;
                /* skip over the marks */
                for ( n=f; n!=NULL && n->type == ft_ap && n->u2.ap->type==at_mark; n=n->next );
                    /* omitted irrelevant stuff */
                }
                ltype = fea_LookupTypeFromItem(a);
            } else
                ltype = fea_LookupTypeFromItem(f);
            start = f;
            f = fea_SetLookupLink(start,ltype);
            for ( f=start; f!=NULL && f->ticked; f=f->next );
            otl = fea_ApplyLookupList(tok,start,lookup_flags);
            fea_AttachFeatureToLookup(otl,feature_tag,sl);
    continue;
```
The existing `fea_SetLookupLink`, when used this way, goes through the entries in the list starting with the first argument looking for those of the type of the second argument, and adding them to the `lookup_next` list of the first argument and marking them as `ticked`. It stops when it reaches an element type that indicates the list is finished, and returns that element. 

What is confusing about this is that the outer loop of the switch statement iterates through `f`, and this is setting `f` to the value at the end, while only copying some entries to be processed by `fea_ApplyLookupList()`. But that `f` value is irrelevant because of the loop immediately after, which begins not at the current value of `f` but at `start`. So the overall implementation iterates from the start of the list to the first un`ticked` element, and the next iteration of the loop will start there, collecting together entries of that type, queueing up the third type, and so forth. 

The problem with this is that it groups all substitutions of a given type together, potentially changing their order. There are cases where this won't matter and cases when it will, the latter most clearly with mixed contextual lookups. And regardless regrouping is contrary to the spec, as quoted above. 

This push request doesn't touch this confusing code (perhaps it should, eliminating that assignment to `f`), but alters `fea_ApplyLookupList` to stop on encountering the first (relevant) element with a type different from `ltype`. Because it is still ticking the elements it "consumes", the higher-level function can use it the same way. (It would be more efficient to advance the search with the return value, but the code suggests that some elements (`mark` anchor points) may be needed in multiple lookups, so it's not safe to just advance the list beyond them.) It also, accordingly, removes the single-substitution adjustment and the feature-level type match check. 

I'm a little uneasy about this change because the spec is arguably too permissive here, and it's very likely different implementations will do different things. For example, some may treat a run of 10 multiple substitutions with one single substitution in the middle as a single lookup, while others might break that into three lists. It might therefore be better to force a user to implement their intention rather than just dividing up the list, either with sub-lookups or using the `subtable` keyword. So maybe there should be a flag or preference to influence the behavior. 

This doesn't have that, and simply tries to match the spec as stated. As a result, opening the UFO of the font referenced in the link above now yields the same `frac` subtables as converting it with `ufo2ft` and opening the resulting OpenType font (as far as I can tell). Further verification is warranted. 

### Types of changes

### Final checklist

### DIFF
--- a/fontforge/featurefile.c
+++ b/fontforge/featurefile.c
@@ -5525,56 +5525,6 @@ return;
     }
     fea_end_statement(tok);
 
-    /* Any of a multiple, alternate, or ligature substitution may have */
-    /* a single destination. In either case it will look just like a single */
-    /* substitution. So if there are both multiple/alternate/ligature and */
-    /* single subs in a lookup translate all the singles into the */
-    /* corresponding type */
-    has_single = has_multiple = has_ligature = has_alternate = false;
-    for ( item=tok->sofar ; item!=NULL && item->type!=ft_feat_start; item=item->next ) {
-	enum otlookup_type cur = fea_LookupTypeFromItem(item);
-	if ( cur==gsub_multiple )
-	    has_multiple = true;
-	else if ( cur==gsub_alternate )
-	    has_alternate = true;
-	else if ( cur==gsub_ligature )
-	    has_ligature = true;
-	else if ( cur==gsub_single )
-	    has_single = true;
-    }
-    if ( has_single ) {
-	enum possub_type psttype = pst_pair;
-	if ( has_multiple && !( has_alternate || has_ligature ) )
-	    psttype = pst_multiple;
-	else if ( has_alternate && !( has_multiple || has_ligature ) )
-	    psttype = pst_alternate;
-	else if ( has_ligature && !( has_multiple || has_alternate ) )
-	    psttype = pst_ligature;
-	if ( psttype!=pst_pair ) {
-	    for ( item=tok->sofar ; item!=NULL && item->type!=ft_feat_start; item=item->next ) {
-		enum otlookup_type cur = fea_LookupTypeFromItem(item);
-		if ( cur==gsub_single )
-		    item->u2.pst->type = psttype;
-	    }
-	}
-    }
-
-    /* Make sure all entries in this lookup of the same lookup type */
-    lookuptype = ot_undef;
-    for ( item=tok->sofar ; item!=NULL && item->type!=ft_feat_start; item=item->next ) {
-	enum otlookup_type cur = fea_LookupTypeFromItem(item);
-	if ( cur==ot_undef )	/* Some entries in the list (lookupflags) have no type */
-	    /* Tum, ty, tum tum */;
-	else if ( lookuptype==ot_undef )
-	    lookuptype = cur;
-	else if ( lookuptype!=cur ) {
-	    LogError(_("All entries in a lookup must have the same type on line %d of %s"),
-		    tok->line[tok->inc_depth], tok->filename[tok->inc_depth] );
-	    ++tok->err_count;
-	    break;
-	}
-    }
-
     item = chunkalloc(sizeof(struct feat_item));
     item->type = ft_feat_end;
     item->u1.tag = feat_tag;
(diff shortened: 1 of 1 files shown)


skef:
@jtanx you more or less requested work in this area. Any thoughts?

jtanx:
Thanks for this. To be honest, I'm not super familiar with this area, so it's a bit hard for me to say either way. I'll take a look though, and see what I can find.

fontforge/featurefile.c
```diff
@@ -6208,30 +6158,39 @@ return( type==ft_lookup_end || type == ft_feat_end ||
...
-    continue;
-	}
-	found_type = fea_LookupTypeFromItem(nested);
-	if ( type==ot_undef || found_type == ot_undef || found_type == type ) {
-	    if ( nested->type!=ft_ap || nested->u2.ap->type!=at_mark )
-		nested->ticked = true;		/* Marks might get used in more than one lookup */
-	    if ( prev!=NULL )
-		prev->lookup_next = nested;
-	    prev = nested;
+    while ( nested!=NULL && !fea_FeatItemEndsLookup(nested->type) ) {
```
frank-trampe: No logical change here (moving the break condition into the while condition), right?

fontforge/featurefile.c
```diff
@@ -6208,30 +6158,39 @@ return( type==ft_lookup_end || type == ft_feat_end ||
...
+	if ( !nested->ticked ) {
+	    found_type = fea_LookupTypeFromItem(nested);
+	    if ( type==ot_undef || found_type == ot_undef || found_type == type ) {
+		if ( nested->type!=ft_ap || nested->u2.ap->type!=at_mark )
+		    nested->ticked = true; /* Marks might get used in more than one lookup */
+		if ( prev!=NULL )
+		    prev->lookup_next = nested;
+		prev = nested;
+	    } else if ( type!=ot_undef && found_type!=type )
```
frank-trampe: Is this (breaking on type mismatch) the only logical change in this function?

frank-trampe:
I'm satisfied that the code changes are safe. Determining the best way to approach edge case feature files is less clear. I think that it is important to distinguish edge cases that occur for some good reason and edge cases that occur by accident and could be written better. But I don't know feature files well enough to define those cases any further.

fontforge/featurefile.c
skef: Correct. I found the other syntax to be less easily understood, so as long as I was "in here" I figured I would change it to this.

fontforge/featurefile.c
skef: Yes, although it's worth noting that that makes the return type rather different when not passing `ot_undef`. Before it would go through the whole list of lookups and return the first `feat_item` after. Now it returns intermediate values. As it happens that doesn't matter because the return value isn't used when the function is called this way.

skef:
Is there an existing convention for issuing a warning in a "[the user] may want to think twice about this" situation?

fontforge/featurefile.c
jtanx: > As it happens that doesn't matter because the return value isn't used when the function is called this way.

In `fea_ApplyFeatureList` as you mention, for the case in question it doesn't matter, but I do see it would have an effect on the `ft_lookup_start` case. I'm still trying to understand this, but from what I get so far, we're now breaking up the lookup/definition when the type changes, vs. carrying on and applying that single substitution adjustment?

jtanx:
> Is there an existing convention for issuing a warning in a "[the user] may want to think twice about this" situation?

Perhaps `ff_post_warning` or `LogError` might be handy? It's been a while though since I've looked at exactly what these do, I think you want the former.

skef:
If it were just me deciding, I would add a warning against any mixing of substitution types at the feature level, that encouraged either in-place lookups or use of the `subtable` keyword as an alternative. (We'd need to verify that the latter actually works.) Even though truly ambiguous cases are probably not common, the whole thing smells to me like an invitation for unexpected behavior.

skef:
Let's do this: If someone merges or officially approves this, I'll code up a version with a warning as a separate pull request and that part of the discussion can proceed there. I think this code conforms to the spec and there seems to be no disagreement on that point, so either way it should go in.

fontforge/featurefile.c
jtanx: Oh right, I see, it's called with `ot_undef` in that case

jtanx:
Yeah, I'd be happy with that, as best I can tell the changes look fine.

fontforge/featurefile.c
skef: This is all heavy spaghetti. I believe the other cases correspond to lookup-level directives rather than feature-level directives. So in those the types are supposed already be verified as uniform (that check is still there), and you'd get the same result here even if I changed the logic for `ot_undef`. But I didn't anyway.

skef:
I played around for a while trying to make a good warning. Unfortunately it would need a suggestion of how to work around the problem, and those means raise their own ugly issues. So I think it may be best to just integrate this and be done with it.

skef:
These changes have been individually approved of twice, without quite being Approved, for understandable reasons: the code is complex and although the spec is mostly clear, there are potential ambiguities. I tried to come up with a good warning, but the obvious workarounds have their own issues*. It seems like this stuff is on the edge of what the current parsing infrastructure can handle, and doing better might require a lot of research and rewriting. 

So I'm going to let this comment sit for a day and then merge if there are no objections. 

* What are those issues? The two candidates for workarounds are the `subtable` directive (to explicitly mark the boundary between subtables) and inline `lookup` tables. The problem with the former is that the spec says that parsers are free to ignore them in many cases, which makes them a poor means of disambiguation. The problem with the latter is that sub-lookups work fine as long as every substitution is in one, but the current code *sometimes* reports an error when all but one run of substitutions is "wrapped", and I don't know enough to judge whether those errors are spec-accurate (although I suspect they aren't).

PR #3660, opened 2019-04-16
Remove the gtkui and win folders, update the .po translation files from the source

jtanx:
It's probably best to refer to the individual commits for what's happening. It's a bit unfortunate that the gettext commands create such a large diff, but it's all autogenerated, so I don't expect any issues with it. Related issues; #3512


.travis.yml
```diff
@@ -92,16 +92,12 @@ install:
...
-        wget 'https://github.com/zeromq/czmq/releases/download/v3.0.2/czmq-3.0.2.tar.gz' -O - |tar -zxf -
         wget 'https://github.com/fontforge/libspiro/releases/download/0.5.20150702/libspiro-0.5.20150702.tar.gz' -O - | tar -zxf -
         wget 'https://github.com/fontforge/libuninameslist/releases/download/20190305/libuninameslist-dist-20190305.tar.gz' -O - | tar -zxf -
         git clone --branch v1.0.2 https://github.com/google/woff2 --depth 1
         wget --tries 1 "http://download.savannah.gnu.org/releases/freetype/freetype-$FTVER.tar.gz" || \
           wget "https://sourceforge.net/projects/freetype/files/freetype2/$SFFTVER/freetype-$FTVER.tar.gz"
 
-        pushd zeromq-4.1.5 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
-        pushd czmq-3.0.2 && CFLAGS="$CFLAGS -Wno-format-truncation" ./configure --prefix=$DEPSPREFIX && make && make install && popd
         pushd libspiro-0.5.20150702 && ./configure --prefix=$DEPSPREFIX && make && make install && popd
```
skef: Not great to be sneaking in cleanup from one mass-removal into another, but acceptable I suppose.

po/en_GB.po
```diff
@@ -119,7 +124,8 @@ msgstr "Next CP Colour"
 
 #: charview.c:140
 msgid "The color used to draw the \"next\" control point of an on-curve point"
-msgstr "The colour used to draw the \"next\" control point of an on-curve point"
+msgstr ""
+"The colour used to draw the \"next\" control point of an on-curve point"
```
skef: What's going on here? How should a reviewer approach these changes?

jtanx: Yeah, it's not great, I'll play around with the config options to see if this can be improved, but the diff is always going to be huge, at least because of all the location/line number comments.

At a guess, I think it's running into some line wrapping limit and bumping it to the next line. I could be wrong though.

In terms of verifying the change, setting the env variable e.g. 'LANGUAGE=fr' and seeing that it still translates is one approach.

As for why I want to do this - clear out outdated translation strings, and to add new ones (untranslated) that have been added over time.

skef: Oh, I don't mean to question the intent. But short of trusting the tools I'm not sure how to review.

jtanx: It's still difficult to review, because they haven't really been kept up to date with the source changes, but try having a look at https://gist.github.com/jtanx/488c5c0ad57a2e73d239c16601cdb60c for a more manageable diff; this was done by following the steps in the second answer to https://stackoverflow.com/questions/2006351/gettext-po-files-under-version-control

skef: This was helpful. It's not like I read through every line, but I looked at enough to see that the overall content seems sound and about as expected, and to recognize some of the strings from the hotkey fix.

PR #3661, opened 2019-04-16
Remove graphicore and IPython checks

skef:
This gets rid of `pycontrib/graphicore` and the related checks for IPython. This is in preparation for/part of removing `pycontrib/gdraw`. 

I'll update the Wiki once I have a request number. 

### Types of changes

### Final checklist


skef:
Wiki updated

ctrlcctrlv:
I want to maybe complain a bit about this. I might not understand it fully so let me be conscientious in my complaint so as not to offend.

As far as IPython goes, I have no objection to it's removal.

But robust GUI access from Python seems a necessity if we want to make Python plugins users can install á la WeeChat.

So I must question your removal of graphicore. Maybe not all is necessary, but certainly some is. Plugins need to be able to make dialogs, right? And access gdraw literals...

ctrlcctrlv:
I retract my comment, I was confusing graphicore with gdraw.py / _gdraw.py.

Sorry for my ignorance, I hope it did not trouble you

skef:
First, "graphicore" has always been a misnomer. That directory just contains `ipython_view.py` and some supporting files to open the window, not the `gdraw` interfaces. 

But second, the gdraw interfaces are really gxdraw interfaces. `pycontrib/gdraw` is, in effect, a trick that tries to kludge GTK event processing into the X version of gdraw. As I remember you're one of the advocates for switching to GGDKDraw as the default interface, and the two approaches are entirely incompatible. 

It would be a good thing for FontForge to support some kind of graphics layer for extensions, and GTK is  the most viable candidate. I have a version of that more or less working [over here](https://github.com/jtanx/fontforge/pull/2). However, some aspects need more thought.

ctrlcctrlv:
@skef Sure, let's make gdraw.py use GDK! :)

I feel like to have a successful scripting ecosystem that's easy for users and installable just by clicking e.g. "Install Plugins" in FontForge, we're going to need some way to create UI elements from within scripts.

skef:
@ctrlcctrlv After this is merged I'm going to submit a gdraw removal too, so you can paste your concerns into that one ...

skef:
@ctrlcctrlv if you have access to GTK why would you use GDraw? Remember that what "GDraw" refers to is GW's custom graphics layer. If FontForge could service a GTK event loop there's no reason an extension would have to or want to do anything with GDraw.

ctrlcctrlv:
Removing gdraw.py without a plan to replace it seems like a step backwards IMO...

Are you aware of Inkscape's extensions? You can create dialogs and get much information from the user. There's hardly a script we'd want to put in the Tools menu where that wouldn't be useful.

ctrlcctrlv:
> if you have access to GTK why would you use GDraw?

You wouldn't, but under GDK do Python scripts _really_ have access to GTK?

ctrlcctrlv:
I'm sorry we're chatting about this in an unrelated issue btw, I definitely deserve the Dunce of the Day award for thinking graphicore was the, ahem, _graphics core_, :desktop_computer: ...

skef:
@ctrlcctrlv You're more or less asking for `pycontrib/gdraw` to be left as a ritualistic representation of some other actually viable interface.

jtanx:
I 100% support removing gdraw.py, as mentioned, it's a giant kludge that doesn't work under ggdkdraw anyway.

I have some plans to reimplement that gdk stuff under gtk anyway, and potentially some of skef's work on that gtk bridge might come in handy too. *Once* that is in place, we can consider how we want to allow interop from python.

skef:
@jtanx I'm staging this removal separately from `pycontrib/gdraw` mostly for record-keeping purposes.

ctrlcctrlv:
`pycontrib/gdraw` is aspirational. I fear just removing it instead of replacing it will send the wrong message, that we don't want to be able to create GUI elements in Python.

What harm is caused by leaving it alone until its replacement is ready?

skef:
> What harm is caused by leaving it alone until its replacement is ready?

Among other things it attempt to link in a shared library that probably won't be built after @jtanx's build infrastructure changes.

PR #3662, opened 2019-04-16
Remove pycontrib/gdraw and GTK-specific configure checks

### LINKED ISSUE #3656: Remove X-based GTK interface
In the spirit of #3512 and #3655, another thing that needs to go is the old python GTK interface for gxdraw. This officially Does Not Make Sense Now, even if it could be revived. With a GDK-based GDraw (or some successor) there's no sense in mixing python GTK with the X version of the software -- anyone trying to do something like that would not choose the X stack. 

Main targets:

1. The relevant portions of `fontforgeexe/pythonui.c`
2. The files in `pycontrib/gdraw`
3. `pycontrib/graphicore` (might as well remove this too -- virtually all parts of it are moribund one way or another)

Note that `pycontrib/gdraw/_gdraw.py` tries to import `libgdraw.so.5`, making it an obstacle to library reduction projects. The whole thing is the very archetype of a "fragile" interface. The chances of anyone currently making independent use of it are close to nil, because it's based on the old `pygtk` module rather than the current PyGObject interface.

skef:
This removes the `pycontrib/gxdraw` interfaces as well as the related hooks in `fontforgeexe/pythonui.c`. 

It's X- and gtk2-only and based on the old `pygtk` interfaces. It loads the `gdraw.so` shared object using an interface built with a tool in 2013. FontForge would greatly benefit from a system or convention for adding graphical elements from python, but this isn't it. 

Closes #3656

### Types of changes

### Final checklist


skef:
[There are reservations](https://github.com/fontforge/fontforge/pull/3661#issuecomment-483621084) about this change, but @frank-trampe is [confident that the functionality hasn't been used in years](https://github.com/fontforge/fontforge/issues/3656#issuecomment-482907480). This kind of approach to compatibility is intrinsically fragile -- requiring constant monitoring and updating to keep alive. That didn't wind up happening, and the clearest path to something comparable now goes through direct support of GTK without any gdraw interaction on the part of a user. This functionality has little to no relevance to a future solution.

Issue #3612, opened 2019-04-01
Require `--without-libuninameslist` to compile when `libuninameslist` is not present


ctrlcctrlv:
(This issue started out as a different report that could be resolved by building with libuninameslist installed. I'm updating the description with a comment from below.)

Suppose we were to change the compilation process to that it fails if:

1. libuninameslist is not present, and
2. `--without-libuninameslist` is not specified.

That amounts to a strong hint to installers that libuninameslist should be installed, that you really want to build with it rather than without it. And we can say as much in the instructions. However, installers who don't care about it can still just add `--without-libuninameslist` and have the same effect you get now when compiling without it.

It's a nudge, basically.

skef:
I take it part of this criticism is that the tooltip isn't enough help?

ctrlcctrlv:
Unicode's name for the glyph doesn't appear in the tooltip for me

skef:
Oh, wait, I think I misunderstand. This is really about the namelist more than anything. Even in the best Adobe list a lot are just referred to by codepoint. Are you thinking of names like "Combining Acute Accent"?

skef:
Or (to pick a more poetic example) U+3014 "Left Tortoise Shell Bracket"?

ctrlcctrlv:
I am thinking of names like that yes. Since I refer to these glyphs in feature files I'm also giving them descriptive names where they don't have them, but to get that far I still have to look it up the first time.

skef:
It does seem like more useful information than the Unicode block name.

ctrlcctrlv:
I think the block name should be in the tooltip or in "Glyph Info&hellip" if mentioned at all. I can't think of a time where knowing the block name helped me know what to draw, and certainly it's never been more helpful than Unicode's name for the glyph. :wink:

skef:
I was wondering about just swapping the glyph name in and the block name out. Having both could be awkward length-wise, but the block would be more appropriate as a label in the "scroll stream", which isn't currently possible because the glyph-group FontView interface (that I recall reading about somewhere) never made it "intro production". 

Hmm. From `fontforgeexe/fontview.c:SCPreparePopup` it looks like the code may already be designed to do this when https://github.com/fontforge/libuninameslist is installed.

skef:
I just tested and it works. Let me try to whip up a `PKGBUILD` for you.

skef:
[...]

PR #3665, opened 2019-04-19
Nudge users to install libuninameslist

ctrlcctrlv:
This will require `--without-libuninameslist` to be provided to compile
when `libuninameslist` is not present.

I'm not the best with autotools, but I think I did this the best way
possible given my starting position.

This closes fontforge/fontforge#3612.


skef:
There's no hurry on this because it mostly relates to the next actual "release", whenever that winds up being. 

I haven't looked at the code yet, but I favor the change as long as that timing is far enough "out". (If there were another quick maintenance release it would be counter-productive to change things on the distros.)

ctrlcctrlv:
Bump

JoesCat:
FreeBSD made a 10_year jump in update:
https://www.freshports.org/textproc/libuninameslist

skef:
I don't think there was a question, but I just verified the nudge behavior is preserved with the cmake build, which is good.

PR #3666, opened 2019-04-19
Update list of supported MIME types

alexhenrie:
This adds font/woff and font/woff2 and renames application/x-font-otf
and application/x-font-ttf to font/otf and font/ttf.

### Important

### Motivation and Context

### Types of changes

### Description

### Final checklist

### DIFF
--- a/desktop/org.fontforge.FontForge.desktop
+++ b/desktop/org.fontforge.FontForge.desktop
@@ -20,5 +20,5 @@ StartupWMClass=fontforge
 Terminal=false
 Type=Application
 Categories=Graphics;
-MimeType=application/vnd.font-fontforge-sfd;application/x-font-ttf;application/x-font-otf;application/x-font-type1;application/x-font-bdf;application/x-font-pcf;application/x-font-tex;
+MimeType=application/vnd.font-fontforge-sfd;application/x-font-ttf;application/x-font-otf;application/x-font-type1;application/x-font-bdf;application/x-font-pcf;application/x-font-tex;application/font-woff;font/otf;font/ttf;font/woff;font/woff2;
 Keywords=font;fonts;editor;TTF;OTF;typeface;


skef:
Are you perhaps updating this list based on MIME types currently recommended for labeling new content? MIME-labeled content can stick around for long periods (in email archives, for example), so the list that FontForge accepts should include the old and the new strings for a given type.

ctrlcctrlv:
@skef The XDG MIME specification uses a distribution provided MIME database. So really this patch's correctness is very hard to gauge. You're right that to support more distros we should keep the old ones though.

https://wiki.archlinux.org/index.php/XDG_MIME_Applications#New_MIME_types

skef:
If there's something in the database itself to map from old x- to new types internally, then it's definitely less important that FontForge keep the old ones. But not necessarily of no importance. Unless keeping both is harmful somehow it seems best to keep all of them.

alexhenrie:
Right, the shared-mime-info package handles the translation from deprecated names like "application/x-font-ttf" to the standard name "font/ttf". The standard names [font/ttf and font/otf](https://tools.ietf.org/html/rfc8081) were created in Feburary 2017, and both [TTF](https://gitlab.freedesktop.org/xdg/shared-mime-info/commit/c50e192b11b46092859f0d9ca261e09a054aebe4) and [OTF](https://gitlab.freedesktop.org/xdg/shared-mime-info/commit/c96c8e7634ba2aac05ecc6684607f591d27086f7) were updated in shared-mime-info immediately.

The big thing for me is getting font/woff and font/woff2 on the list. If you prefer, I can change the patch to simply append these two types and leave the rest alone.

skef:
Can we just make it a superset of what you've added and what was there before (e.g. keep `application/x-font-ttf` while adding `font/ttf`)? I realize this "desktop" distribution mechanism is fairly new, but that would keep systems with outdated MIME DBs working as before. (Someone can always come in in another five years and clean out the cruft.)

ctrlcctrlv:
Agreed with @skef...distros from 2017 aren't THAT old, are they?

alexhenrie:
@skef, I just force-pushed a new commit to my branch that does exactly what you asked.

skef:
"We" have to think about `gutils/g_giomime.c` too, but I don't see how that's @alexhenrie's problem.

alexhenrie:
I just force-pushed again because I realized that you're going to want application/font-woff on the list too for the benefit of old distros.

skef:
Strange but sure enough: no "x-" on that one.

PR #3679, opened 2019-04-26
 Save quadratic spline in quadratic layer 

### LINKED ISSUE #3667: [Spiro] “Internal Error: Invalid 2nd order spline in SplineRefigure2” when opening an SFD containing Spiros on a “Q” Quadratic layer
Arch Linux, 71cde275d083dbea87b1e8d0a6baaf7144189ea3

sr2
weird S

Reproduction steps:

1. Create a new font
2. Change the layers to “Q”&mdash;Quadratic
3. Draw any spiro spline that contains at least one curve
4. Save your font
5. Re-open, observe this bug.

Alternatively, you may just open my [S.zip](https://github.com/fontforge/fontforge/files/3099771/S.zip) generated by following those steps

It seems that you can just move one of the spline's points and put it back and the weird shape goes away (for that glyph only), but this glitchiness makes me very weary to trust FontForge with a big project involving quadratic layers and spiros.

### LINKED ISSUE #3668: ⟦Spiro⟧ Bad behavior—cannot cleanly copy Spiro from “Q” Quadratic layer to “C” Cubic layer!
Arch Linux, 71cde275d083dbea87b1e8d0a6baaf7144189ea3

##  Reproduction steps:

1. New font
2. Switch to background, making sure it's “C” Cubic
2. Draw a Spiro that has at least one curve (image 1)
3. Make foreground layer “Q” Quadratic
4. Select all on background, cut, paste to foreground
5. Receive weird Spiro shown in image 2

## Video

[Qbug.zip](https://github.com/fontforge/fontforge/files/3099782/Qbug.zip)

## Screenshots

image1
image2

### LINKED ISSUE #3670: ⟦Spiro⟧ Glyphs with spiros are saved as cubic splines even in quadratic layers
Arch Linux, 71cde27

This may be related to #3667 but I'm not sure. 

I made some files to prove this. Please download and peruse [SavingBug.zip](https://github.com/fontforge/fontforge/files/3099881/SavingBug.zip), else I fear this report will make no sense.

I drew two Spiro circles in two different files, using the &ldquo;Get Info&rdquo; dialog to make them exactly the same. All that's different is whether or not I drew them on a &ldquo;Q&rdquo; Quadratic layer or a &ldquo;C&rdquo; Cubic layer&mdash;`Quadratic.sfd` &amp; `Cubic.sfd`.

I then did the same, I drew two circles made of B&eacute;zier splines, only differing in their respective layer types&mdash;`QuadraticB.sfd` &amp; `CubicB.sfd`.

I then made two diffs&mdash;`SpiroDiff.diff` & `BézierDiff.diff`.

I really think the results speak for themselves. Drawing Spiros on quadratic layers is fundamentally broken, it saves in the cubic format. I think this is the root cause of #3667.

### SpiroDiff.diff
This is essentially a noop. _That's bad!_

```diff
diff --git a/Quadratic.sfd b/Cubic.sfd
index 8f70a41..d79a061 100644
--- a/Quadratic.sfd
+++ b/Cubic.sfd
@@ -14,13 +14,13 @@ Descent: 200
 InvalidEm: 0
 LayerCount: 2
 Layer: 0 0 "Back" 1
-Layer: 1 1 "Fore" 1
+Layer: 1 0 "Fore" 0
 XUID: [1021 575 -1819899406 628701]
 OS2Version: 0
 OS2_WeightWidthSlopeOnly: 0
 OS2_UseTypoMetrics: 1
 CreationTime: 1555747562
-ModificationTime: 1555747779
+ModificationTime: 1555747871
 OS2TypoAscent: 0
 OS2TypoAOffset: 1
 OS2TypoDescent: 0
@@ -50,6 +50,7 @@ Encoding: 79 79 0
 Width: 1000
 VWidth: 0
 InSpiro: 1
+Flags: HO
 LayerCount: 2
 Fore
 SplineSet
```

### BézierDiff.diff

They're actually saved in different formats depending on  their type. _That's good!_

```diff
diff --git a/QuadraticB.sfd b/CubicB.sfd
index e0ee6b8..6050776 100644
--- a/QuadraticB.sfd
+++ b/CubicB.sfd
@@ -14,13 +14,13 @@ Descent: 200
 InvalidEm: 0
 LayerCount: 2
 Layer: 0 0 "Back" 1
-Layer: 1 1 "Fore" 0
+Layer: 1 0 "Fore" 0
 XUID: [1021 575 -1819899406 9896654]
 OS2Version: 0
 OS2_WeightWidthSlopeOnly: 0
 OS2_UseTypoMetrics: 1
 CreationTime: 1555747893
-ModificationTime: 1555748292
+ModificationTime: 1555748270
 OS2TypoAscent: 0
 OS2TypoAOffset: 1
 OS2TypoDescent: 0
@@ -49,18 +49,15 @@ StartChar: Q
 Encoding: 81 81 0
 Width: 1000
 VWidth: 0
+Flags: HO
 LayerCount: 2
 Fore
 SplineSet
-400 700 m 128,-1,1
- 288 700 288 700 194 606 c 128,-1,2
- 100 512 100 512 100 400 c 128,-1,3
- 100 288 100 288 194 194 c 128,-1,4
- 288 100 288 100 400 100 c 128,-1,5
- 512 100 512 100 606 194 c 128,-1,6
- 700 288 700 288 700 400 c 128,-1,7
- 700 512 700 512 606 606 c 128,-1,0
- 512 700 512 700 400 700 c 128,-1,1
+400 700 m 0
+ 250 700 100 550 100 400 c 4
+ 100 250 250 100 400 100 c 0
+ 550 100 700 250 700 400 c 0
+ 700 550 550 700 400 700 c 0
 EndSplineSet
 EndChar
 EndChars
```

cc: @skef

ctrlcctrlv:
This closes #3670.
This closes #3667.
\#3668 remains outstanding for now.

This patch makes sure that quadratic splines are saved in quadratic
layers, and that the splines associated with a Spiro are quadratic if
the layer is quadratic.

This patch works when the file is saved, so old files will remain broken
until saved over. Doing it this way side steps the performance issues
@skef brought up in #3670, but it also doesn't fi​x #3668—I'll have to do
that another time.

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -669,12 +669,15 @@ static void SFDDumpHintMask(FILE *sfd,HintMask *hintmask) {
     }
 }
 
-static void SFDDumpSplineSet(FILE *sfd, SplineSet *spl, int order2) {
+static void SFDDumpSplineSet(FILE *sfd, SplineSet *spl, int want_order2) {
     SplinePoint *first, *sp;
+    int order2 = spl->first->next==NULL || spl->first->next->order2;
+    int reduce = (want_order2 && !order2);
+    if (order2 && !want_order2) IError("Asked for cubic when had quadratic");
     SplineSet *nspl;
 
     for ( ; spl!=NULL; spl=spl->next ) {
-	if (order2) {
+	if (reduce) {
 	    nspl = SSttfApprox(spl);
 	} else {
 	    nspl = spl;
@@ -769,7 +772,7 @@ static void SFDDumpSplineSet(FILE *sfd, SplineSet *spl, int order2) {
 	if ( spl->start_offset ) {
 	    fprintf( sfd, "  PathStart: %d\n", spl->start_offset );
 	}
-    if (order2) free(nspl);
+    if (reduce) SplinePointListFree(nspl);
     }
     fprintf( sfd, "EndSplineSet\n" );
 }
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -668,9 +668,8 @@ static void SFDDumpHintMask(FILE *sfd,HintMask *hintmask) {
     }
 }
 
-static void SFDDumpSplineSet(FILE *sfd,SplineSet *spl) {
+static void SFDDumpSplineSet(FILE *sfd, SplineSet *spl, int order2) {
     SplinePoint *first, *sp;
-    int order2 = spl->first->next==NULL || spl->first->next->order2;
 
     for ( ; spl!=NULL; spl=spl->next ) {
 	first = NULL;
@@ -937,7 +936,7 @@ void SFDDumpUndo(FILE *sfd,SplineChar *sc,Undoes *u, const char* keyPrefix, int 
             }
 	    if( u->u.state.splines ) {
                 fprintf(sfd, "SplineSet\n" );
-                SFDDumpSplineSet( sfd, u->u.state.splines );
+                SFDDumpSplineSet( sfd, u->u.state.splines, u->was_order2 );
             }
             break;
 
@@ -1637,7 +1636,7 @@ static void SFDDumpChar(FILE *sfd,SplineChar *sc,EncMap *map,int *newgids,int to
 	    SFDDumpImage(sfd,img);
 	if ( sc->layers[i].splines!=NULL ) {
 	    fprintf(sfd, "SplineSet\n" );
-	    SFDDumpSplineSet(sfd,sc->layers[i].splines);
+	    SFDDumpSplineSet(sfd,sc->layers[i].splines,sc->layers[i].order2);
 	}
 	SFDDumpRefs(sfd,sc->layers[i].refs,newgids);
 	SFDDumpGuidelines(sfd, sc->layers[i].guidelines);
@@ -2782,7 +2781,7 @@ static int SFD_Dump( FILE *sfd, SplineFont *sf, EncMap *map, EncMap *normal,
 	if ( sf->grid.order2 )
 	    fprintf(sfd, "GridOrder2: %d\n", sf->grid.order2 );
 	fprintf(sfd, "Grid\n" );
-	SFDDumpSplineSet(sfd,sf->grid.splines);
+	SFDDumpSplineSet(sfd,sf->grid.splines,false);
     }
     if ( sf->texdata.type!=tex_unset ) {
 	fprintf(sfd, "TeXData: %d %d", (int) sf->texdata.type, (int) ((sf->design_size<<19)+2)/5 );
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -679,6 +679,7 @@ static void SFDDumpSplineSet(FILE *sfd, SplineSet *spl, int want_order2) {
     for ( ; spl!=NULL; spl=spl->next ) {
 	if (reduce) {
 	    nspl = SSttfApprox(spl);
+	    order2 = true;
 	} else {
 	    nspl = spl;
 	}
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -769,8 +769,8 @@ static void SFDDumpSplineSet(FILE *sfd, SplineSet *spl, int order2) {
 	if ( spl->start_offset ) {
 	    fprintf( sfd, "  PathStart: %d\n", spl->start_offset );
 	}
-    }
     if (order2) free(nspl);
+    }
     fprintf( sfd, "EndSplineSet\n" );
 }
 
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -2789,7 +2789,7 @@ static int SFD_Dump( FILE *sfd, SplineFont *sf, EncMap *map, EncMap *normal,
 	if ( sf->grid.order2 )
 	    fprintf(sfd, "GridOrder2: %d\n", sf->grid.order2 );
 	fprintf(sfd, "Grid\n" );
-	SFDDumpSplineSet(sfd,sf->grid.splines,false);
+	SFDDumpSplineSet(sfd,sf->grid.splines,sf->grid.order2);
     }
     if ( sf->texdata.type!=tex_unset ) {
 	fprintf(sfd, "TeXData: %d %d", (int) sf->texdata.type, (int) ((sf->design_size<<19)+2)/5 );
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -41,6 +41,7 @@
 #include "psread.h"
 #include "splinefont.h"
 #include "splinefill.h"
+#include "splineorder2.h"
 #include "splinesaveafm.h"
 #include "splineutil.h"
 #include "splineutil2.h"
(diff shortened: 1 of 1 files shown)


fontforge/sfd.c
```diff
@@ -764,6 +770,7 @@ static void SFDDumpSplineSet(FILE *sfd,SplineSet *spl) {
 	    fprintf( sfd, "  PathStart: %d\n", spl->start_offset );
 	}
     }
+    if (order2) free(nspl);
```
frank-trampe: Are you sure that this is in the right block?

ctrlcctrlv: @frank-trampe Sorry, fixed in a3aff8a6e1390a3fdee8b318a28d6b8f47a2211d

frank-trampe:
This looks fine to me, but we ought to get a final look from @skef, since he's spent a lot more time thinking about the root issue than I have.

fontforge/sfd.c
```diff
@@ -2782,7 +2789,7 @@ static int SFD_Dump( FILE *sfd, SplineFont *sf, EncMap *map, EncMap *normal,
 	if ( sf->grid.order2 )
 	    fprintf(sfd, "GridOrder2: %d\n", sf->grid.order2 );
 	fprintf(sfd, "Grid\n" );
-	SFDDumpSplineSet(sfd,sf->grid.splines);
```
frank-trampe: Wait. One other question. Why is this always false?

ctrlcctrlv: The guide layer is always cubic, and usually only contains straight lines.

ctrlcctrlv: Although now that you mention it I do see `sf->grid.order2`, but there's no way to change the guide layer's type in the UI so it must always be cubic.

ctrlcctrlv: Changed in https://github.com/fontforge/fontforge/pull/3679/commits/4f5de27ed2a8f0ea54919aa82f0f1b39470b4329

skef:
I might just not understand, but at first inspection this doesn't look right to me. The master version of `SFDDumpSplineSet` calculates `order2` as `spl->first->next==NULL || spl->first->next->order2;` -- that is, from the spline passed in. That suggests that either a quadratic or cubic spline could be "dumped" (which, of course, makes sense). The new code passes in `order2` and does this:

```
	if (order2) {
	    nspl = SSttfApprox(spl);
	} else {
	    nspl = spl;
	}
```
But what if the spline passed in is already quadratic? And, further, what if it is quadratic and the `order2` passed in is false?

I would have expected leaving the `order2` calculation, adding a parameter `want_order2` (defined and used as the third parameter is in your code), and then converting when the two differ (presumably by calling `SSPSApprox()` or `SSttfApprox()`).

ctrlcctrlv:
@skef The change here is making it so the layer is authoritative as to which type of spline should be saved, each spline should not be able to decide on its own whether or not it's cubic or quadratic; if it's in a quad layer it should be saved as a quad. #3670 is the motivation for this, and it breaks no tests.

The only time you get a cubic spline on a quad layer is when you have a Spiro, and that's what this handles. The opposite doesn't happen because a quad Spiro is impossible, and it's also impossible to have mismatched B&eacute;zier splines.

skef:
In that case there is a substantive problem and a theoretical/interface concern. 

The theoretical/interface concern is that by adding that extra parameter you make it "look like" it will influence the output, but then the function doesn't implement that. It may never be called with a cubic spline and `order2` set to true, but someone looking at it in the future may well assume it can be called that way. So if that possibility is left unimplemented you should add a error or strong warning for that case. 

The substantive concern is that `SSttfApprox()` allocates a bunch of memory and does a bunch of work, and *may* be harmless to call on an already quadratic spline but that isn't obvious from a glance at the code. In non-Spiro cases there's no good reason to pass the already quadratic splines into `SSttfApprox()` and a number of reasons not to. So even if the function doesn't need to convert quadratic to cubic, it should avoid converting quadratic to quadratic.

ctrlcctrlv:
What would you have me do? What can I say? This works, that's what I know, and it's been a necessity for my latest font project, working name _Typewriter 2020_.

https://twitter.com/HW_BEAT_THAT/status/1124652979468947456
https://twitter.com/HW_BEAT_THAT/status/1123913145750315013
https://twitter.com/HW_BEAT_THAT/status/1123882795141607424

I'm making a generative typewriter font. There's a smooth master then a script to generate as many rough ones from it as you want, with contextual alternates to avoid the problem I describe in [this Quora answer](https://www.quora.com/What-are-the-disadvantages-of-typography/answer/Fredrick-Brennan).

This patch, along with a bunch of others I've contributed, have made this possible.

ctrlcctrlv:
@skef I'm sorry, I don't want to be combative, but what I'm trying to say is that this fixes the bug, which is really a rather serious longstanding bug. You don't like the way I fixed it but it seems like you've got a good idea of how you would fix it yourself now that you've seen my patch, so feel free to modify my patch so it fits your theoretical scenario...

fontforge/sfd.c
skef: Unresolving: nspl is a linked list that needs to be freed with `SplinePointListFree()` (SplinePointList and SplineSet have the same typedef I believe.)

skef:
What I'd have you do is:

1. Change the name of the new parameter to `want_order2`
2. Add the calculation of `order2` from the deleted code back in.
3. Add a boolean `int reduce = want_order2 && !order2`
4. Use `reduce` in your new code the way you use `order2` now (call `SSttfApprox()` and (see other comment) `SplinePointListFree()`
5. Print a warning if `order2` is false and `want_order2` is true.

skef:
This isn't just a theoretical scenario. I don't know what `SSttfApprox()` will do when called on an already-quadratic spline. It could potentially either corrupt or distort it. In the latter case saving a file with quadratic splines repeatedly could end up distorting the shapes.

ctrlcctrlv:
With all due respect, since you have such a clear vision of how it should look, I think you should just push those changes to the patch yourself. It would save a lot of back and forth in case I misunderstand one of your directions and you have to make another comment, ping pong, etc.

But if it's still sitting here unmerged in a few days time I'll get on it.

skef:
@ctrlcctrlv that isn't how the push/review process typically works. I've occasionally done that on this project with "extra" changes that required significant, separable work (like cleaning up flag definitions), but this isn't such a case.

ctrlcctrlv:
I'm just trying to be efficient with my time as well as yours, but if you're really against it I guess I can do it right now and see if I understand your directions

skef:
There are just three goals beyond the current code:

1. Only convert the spline if your new parameter is true **and** it's not already quadratic.
2. If your new parameter is false and the spline is quadratic, print a strong warning and then proceed.
3. Use the correct freeing-function.

ctrlcctrlv:
lol

For the last nine days seems like I've been calling `SSttfApprox` on already quadratic splines yet it hasn't corrupted my font

ctrlcctrlv:
I think instead of a warning I'll just
```c
int reduce = (want_order2 && !order2) && !(want_order2 && order2);
```

ctrlcctrlv:
no wait i can just delete the warning since reduce won't be true in that case

skef:
You don't need a warning for the case when you are converting, the warning is for the case you're not handling (because we don't think it currently arises): `order2` is false but the existing spline is quadratic.

ctrlcctrlv:
o ok

ctrlcctrlv:
@skef 000022a40 should do what you want.

fontforge/sfd.c
```diff
@@ -668,13 +669,21 @@ static void SFDDumpHintMask(FILE *sfd,HintMask *hintmask) {
...
+static void SFDDumpSplineSet(FILE *sfd, SplineSet *spl, int want_order2) {
     SplinePoint *first, *sp;
     int order2 = spl->first->next==NULL || spl->first->next->order2;
+    int reduce = (want_order2 && !order2);
+    if (order2 && !want_order2) IError("Asked for cubic when had quadratic");
+    SplineSet *nspl;
 
     for ( ; spl!=NULL; spl=spl->next ) {
+	if (reduce) {
+	    nspl = SSttfApprox(spl);
```
skef: OK, just one more thing: set `order2` to true here so that the methods that use it below get the right value. After that @frank-trampe can take another glance at the new changes, but it basically looks good to me.

ctrlcctrlv: good idea

ctrlcctrlv: Resolved in https://github.com/fontforge/fontforge/pull/3679/commits/8e2198411b453a5426faec0cd9d60f4a129a78fb

skef:
@frank-trampe: I more or less approve this but since I asked for some changes give it a look when you get a chance.

ctrlcctrlv:
?

PR #3709, opened 2019-05-25
Fix packaging metadata.

frank-trampe:
This fixes problems building Debian packages due to incompatibilities with the news names and locations of desktop and icon files. It also fixes a compilation problem on older versions of gcc caused by use of an incompatible option.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -547,8 +547,10 @@ AM_CONDITIONAL([FONTTOOLS],[test x"${fontforge_extras}" = xyes])
 #--------------------------------------------------------------------------
 # Create and Set FontForge Compiler Flags
 
+FONTFORGE_CFLAG_CHECK([WARNING_CFLAGS], [-Werror=implicit-function-declaration])
+FONTFORGE_CFLAG_CHECK([WARNING_CFLAGS], [-Werror=int-conversion])
+
 my_cflags="${my_cflags} ${WARNING_CFLAGS}"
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -548,7 +548,7 @@ AM_CONDITIONAL([FONTTOOLS],[test x"${fontforge_extras}" = xyes])
 # Create and Set FontForge Compiler Flags
 
 my_cflags="${my_cflags} ${WARNING_CFLAGS}"
-my_cflags="${my_cflags} -Werror=implicit-function-declaration -Werror=int-conversion"
```
jtanx: What version of gcc is this?

frank-trampe: Whatever ships on Ubuntu 14.04.

frank-trampe: I got the error on Launchpad. I don't have an active Trusty machine anymore.

jtanx: I'd prefer not disabling this, since it is useful, and has caught issues in the past. I'll see if I can modify this to detect if the flag is supported and only add it if it is.

frank-trampe: Okay.

jtanx: @frank-trampe take a look

frank-trampe: I like that.

jtanx: Apologies, I didn't realise that AC_MSG_ERROR prevents configure from continuing - I've updated it to not error out in that case (and I've also spun up a trusty VM to test against)

PR #3711, opened 2019-05-26
tottf.c: AssignTTFBitGlyph: Don't clobber assignments made by AssignNotdefNull

### LINKED ISSUE #3692: Wrong character positions in OTB font
When I convert a BDF to an OTB font, the glyphs are wrongly positioned in the resulting OTB file. All glyphs seem to be shifted by 2 characters, e.g. glyph "A" is placed on character "C".

Fontforge version: 2019-04-15
Operating system: Void Linux

The attached zip file includes a sample BDF font and the OTB font generated from it:
[fonts.zip](https://github.com/fontforge/fontforge/files/3144517/fonts.zip)

jtanx:
Fixes #3692

For reference, there's a similar check in `AssignTTFGlyph` (the non-bitmap path equivalent).

When writing out the cmap, FF always adds .null, .notdef and .nonmarkingreturn, even if they're not explicitly defined. However, in this case, two out of the three *are* defined. `AssignNotdefNull` correctly sets `ttf_glyph`  and `bygid` based on this, but this gets clobbered right after, causing the gid assigned to be off by 2.

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -1443,7 +1443,12 @@ static int AssignTTFBitGlyph(struct glyphinfo *gi,SplineFont *sf,EncMap *map,int
 	for ( j=0; bsizes[j]!=0 && ((bsizes[j]&0xffff)!=bdf->pixelsize || (bsizes[j]>>16)!=BDFDepth(bdf)); ++j );
 	if ( bsizes[j]==0 )
     continue;
-	for ( i=0; i<bdf->glyphcnt; ++i ) if ( !IsntBDFChar(bdf->glyphs[i]) )
+	/* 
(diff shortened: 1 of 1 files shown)


frank-trampe:
Given the opacity of the code and the fact that you probably just made a substantial time investment to understand what was going on, a few comments might be in order, no?

joshuakraemer:
This fixes a segfault for me, see https://github.com/fontforge/fontforge/issues/3692#issuecomment-491608574
Thank you very much, @jtanx!

jtanx:
@frank-trampe ok to merge? I've added in some comments

frank-trampe:
It's enough to send somebody in the right direction.

PR #3712, opened 2019-05-26
Make comb. chars always superior to spacing chars

### LINKED ISSUE #3708: Combining dieresis not seen as superior to dieresis, even if “Glyph Info→Components” used
A combining dieresis should always be seen as superior to a regular dieresis for the purpose of glyph building.

Even if you don't agree, certainly it should be seen as superior when custom components are in use!

Yet it isn't, see [DieresisIssue.zip](https://github.com/fontforge/fontforge/files/3219364/DieresisIssue.zip) for a minimal example.

ctrlcctrlv:
Combining characters are superior in some cases to spacing characters
but not others. For example, combining macron (U+304) is superior to
macron (U+AF), yet combining dieresis (U+308) is inferior to dieresis
(U+A8).

This situation is untenable. It makes it impossible, while designing a
font, to know which character will be used when FontForge builds an
accented glyph without digging through the source code. Furthermore, the
logical case, to use the combining character first, is rarely done.

Apparently this is due to some restriction of Type1 fonts according to
George in 2001. I fixed his reported restriction in this same commit by
just dissolving the reference if the character referred to is between
0x300 and 0x345.

Some, such as @frank-trampe, might say that dissolving the reference
“breaks” something. But nothing is broken. No function of a Type 1 font
changes if a reference is used or not. Type 1 fonts never were a first
class format for storing font data—and certainly haven't become one—
that's what SFD is for. The only discernible difference therefore is a
small increase in filesize.

I confirmed:
* generated PFA (PostScript type 1) fonts still work; FontForge shows
the dissolved references as expected when a problem character is
referred to.
* if you want to use `dieresis` you can tell FontForge to in “Glyph Info
→ Components” and it will work.
* that this closes #3708, the issue is solved.

I encourage my fellow members to support this change. As an author of
five and counting open source fonts made only using FontForge and
FontTools, I can unequivocally say that sacrificing the file size of a
subset of Type 1 fonts is worth the benefit this change provides.

### DIFF
--- a/fontforge/dumppfa.c
+++ b/fontforge/dumppfa.c
@@ -977,8 +977,8 @@ void SC_PSDump(void (*dumpchar)(int ch,void *data), void *data,
     int i, last, first;
     SplineSet *temp;
 
-    if ( sc==NULL ) return;
-
+    if ( sc==NULL )
+return;
     last = first = layer;
     if ( layer==ly_all )
 	first = last = ly_fore;
@@ -986,21 +986,7 @@ void SC_PSDump(void (*dumpchar)(int ch,void *data), void *data,
 	first = ly_fore;
 	last = sc->layer_cnt-1;
     }
-
     for ( i=first; i<=last; ++i ) {
-
-    // Before we do anything, convert references to glyphs between 0x300 and
-    // 0x345 to splines. This is because Type1 fonts don't support references
-    // to them—or at least didn't in 2001—this format is ancient and retained
-    // primarily for backwards compatibility.
-    if ( sc->layers[i].refs!=NULL ) {
-        for ( ref = sc->layers[i].refs; ref!=NULL; ref=ref->next ) {
-            if (ref->unicode_enc >= 0x300 || ref->unicode_enc <= 0x345) {
-                SCRefToSplines(sc, ref, i);
-            }
-        }
-    }
-
 	if ( sc->layers[i].splines!=NULL ) {
 	    temp = sc->layers[i].splines;
 	    if ( sc->layers[i].order2 ) temp = SplineSetsPSApprox(temp);
--- a/fontforge/dumppfa.c
+++ b/fontforge/dumppfa.c
@@ -977,8 +977,8 @@ void SC_PSDump(void (*dumpchar)(int ch,void *data), void *data,
     int i, last, first;
     SplineSet *temp;
 
-    if ( sc==NULL )
-return;
+    if ( sc==NULL ) return;
+
     last = first = layer;
     if ( layer==ly_all )
 	first = last = ly_fore;
@@ -986,7 +986,21 @@ return;
 	first = ly_fore;
 	last = sc->layer_cnt-1;
     }
+
     for ( i=first; i<=last; ++i ) {
+
+    // Before we do anything, convert references to glyphs between 0x300 and
+    // 0x345 to splines. This is because Type1 fonts don't support references
+    // to them—or at least didn't in 2001—this format is ancient and retained
+    // primarily for backwards compatibility.
+    if ( sc->layers[i].refs!=NULL ) {
+        for ( ref = sc->layers[i].refs; ref!=NULL; ref=ref->next ) {
+            if (ref->unicode_enc >= 0x300 || ref->unicode_enc <= 0x345) {
+                SCRefToSplines(sc, ref, i);
+            }
+        }
+    }
+
 	if ( sc->layers[i].splines!=NULL ) {
 	    temp = sc->layers[i].splines;
 	    if ( sc->layers[i].order2 ) temp = SplineSetsPSApprox(temp);
--- a/fontforge/fvcomposite.c
+++ b/fontforge/fvcomposite.c
@@ -50,20 +50,17 @@ int CharCenterHighest = 1;
 #define BottomAccent	0x300
 #define TopAccent	0x345
 
-/* for accents between 0x300 and 345 these are some synonyms */
-/* type1 wants accented chars built with accents in the 0x2c? range */
-/*  except for grave and acute which live in iso8859-1 range */
 /*  this table is ordered on a best try basis */
 static const unichar_t accents[][4] = {
-    { 0x2cb, 0x300, 0x60 },	/* grave */
-    { 0x2ca, 0x301, 0xb4 },	/* acute */
-    { 0x2c6, 0x302, 0x5e },	/* circumflex */
-    { 0x2dc, 0x303, 0x7e },	/* tilde */
-    { 0x2c9, 0x304, 0xaf },	/* macron */
+    { 0x300, 0x2cb, 0x60 },	/* grave */
+    { 0x301, 0x2ca, 0xb4 },	/* acute */
+    { 0x302, 0x2c6, 0x5e },	/* circumflex */
+    { 0x303, 0x2dc, 0x7e },	/* tilde */
+    { 0x304, 0x2c9, 0xaf },	/* macron */
     { 0x305, 0xaf },		/* overline, (macron is suggested as a syn, but it's not quite right) */
-    { 0x2d8, 0x306 },		/* breve */
-    { 0x2d9, 0x307, '.' },	/* dot above */
-    { 0xa8,  0x308 },		/* diaeresis */
+    { 0x306, 0x2d8 },		/* breve */
+    { 0x307, 0x2d9, '.' },	/* dot above */
+    { 0x308, 0xa8 },		/* diaeresis */
     { 0x2c0 },			/* hook above */
     { 0x2da, 0xb0 },		/* ring above */
     { 0x2dd },			/* real acute */
@@ -74,9 +71,9 @@ static const unichar_t accents[][4] = {
     { 0 },			/* cand... */		/* 310 */
     { 0 },			/* inverted breve */
     { 0x2bb },			/* turned comma */
-    { 0x2bc, 0x313, ',' },	/* comma above */
+    { 0x313, 0x2bc, ',' },	/* comma above */
     { 0x2bd },			/* reversed comma */
-    { 0x2bc, 0x315, ',' },	/* comma above right */
+    { 0x315, 0x2bc, ',' },	/* comma above right */
     { 0x316, 0x60, 0x2cb },	/* grave below */
     { 0x317, 0xb4, 0x2ca },	/* acute below */
     { 0 },			/* left tack */
@@ -86,8 +83,8 @@ static const unichar_t accents[][4] = {
     { 0 },			/* half ring */
     { 0x2d4 },			/* up tack */
     { 0x2d5 },			/* down tack */
-    { 0x2d6, 0x31f, '+' },	/* plus below */
-    { 0x2d7, 0x320, '-' },	/* minus below */	/* 320 */
+    { 0x31f, 0x2d6, '+' },	/* plus below */
+    { 0x320, 0x2d7, '-' },	/* minus below */	/* 320 */
     { 0x2b2 },			/* hook */
     { 0 },			/* back hook */
     { 0x323, 0x2d9, '.' },	/* dot below */
--- a/fontforge/fvcomposite.c
+++ b/fontforge/fvcomposite.c
@@ -2250,16 +2250,6 @@ return;
 	_BCCenterAccent( bdf,sc->orig_pos,rsc->orig_pos,ch,basech,italicoff,pos,sf->ascent+sf->descent );
 }
 
-static void SCCenterAccent(SplineChar *sc,SplineChar *basersc, SplineFont *sf,
-	int layer, int ch, BDFFont *bdf,int disp_only,
-	real ia, int basech, char *dot ) {
-    int invert = false;			/* invert accent, false==0, true!=0 */
-    SplineChar *rsc = GetGoodAccentGlyph(sf,ch,basech,&invert,ia,dot,sc);
-
-    /* find a location to put an accent on this character */
-    _SCCenterAccent(sc,basersc,sf,layer,ch,bdf,disp_only,rsc,ia,basech,invert,FF_UNICODE_NOPOSDATAGIVEN);
-}
-
 static void _BCPutRefAfter( BDFFont *bdf,int gid,int rgid,int normal,int under ) {
     BDFChar *bc, *rbc;
     int ispacing;
@@ -2803,6 +2793,8 @@ void SCBuildComposit(SplineFont *sf, SplineChar *sc, int layer, BDFFont *bdf, in
     const unichar_t *pt, *apt; unichar_t ch;
     real ia;
     char *dot;
+    SplineChar *rsc;
+    int invert;
     /* This does not handle arabic ligatures at all. It would need to reverse */
     /*  the string and deal with <final>, <medial>, etc. info we don't have */
 
@@ -2874,8 +2866,15 @@ return;
 	else if ( sc->width == base->sc->width )
 	    base->use_my_metrics = true;
 	while ( iscombining(*pt) || SCUserDecompAccent(sc, *pt, accent_hint) || (ch=='L' && *pt==0xb7) ||	/* b7, centered dot is used as a combining accent for Ldot but as a lig for ldot */
-		*pt==0x384 || *pt==0x385 || (*pt>=0x1fbd && *pt<=0x1fff) )	/* Special greek accents */
-	    SCCenterAccent(sc,base!=NULL?base->sc:NULL,sf,layer,*pt++,bdf,disp_only,ia,ch,dot);
+		*pt==0x384 || *pt==0x385 || (*pt>=0x1fbd && *pt<=0x1fff) ) {	/* Special greek accents */
+	    invert = false;
+	    if ( sc->user_decomp!=NULL )
+		rsc = SFGetChar(sf,*pt,NULL);
+	    else
+		rsc = GetGoodAccentGlyph(sf,*pt,ch,&invert,ia,dot,sc);
+	    _SCCenterAccent(sc,base!=NULL?base->sc:NULL,sf,layer,CanonicalCombiner(*pt++),bdf,disp_only,rsc,ia,
+	                    ch,invert,FF_UNICODE_NOPOSDATAGIVEN);
+	}
 	while ( *pt ) {
 	    if ( base!=NULL ) base->use_my_metrics = false;
 	    SCPutRefAfter(sc,sf,layer,*pt++,bdf,disp_only,dot);
--- a/fontforge/fvcomposite.c
+++ b/fontforge/fvcomposite.c
@@ -61,67 +61,67 @@ static const unichar_t accents[][4] = {
     { 0x306, 0x2d8 },		/* breve */
     { 0x307, 0x2d9, '.' },	/* dot above */
     { 0x308, 0xa8 },		/* diaeresis */
-    { 0x2c0 },			/* hook above */
-    { 0x2da, 0xb0 },		/* ring above */
-    { 0x2dd },			/* real acute */
-    { 0x2c7 },			/* caron */
-    { 0x2c8, 0x384, 0x30d, '\''  },	/* vertical line, tonos */
+    { 0x309, 0x2c0 },			/* hook above */
+    { 0x30a, 0x2da, 0xb0 },		/* ring above */
+    { 0x30b, 0x2dd },			/* real acute */
+    { 0x30c, 0x2c7 },			/* caron */
+    { 0x30d, 0x2c8, 0x384, '\'' },	/* vertical line, tonos */
     { 0x30e, '"' },		/* real vertical line */
-    { 0 },			/* real grave */
-    { 0 },			/* cand... */		/* 310 */
-    { 0 },			/* inverted breve */
-    { 0x2bb },			/* turned comma */
+    { 0x30f },			/* real grave */
+    { 0x310 },			/* cand... */		/* 310 */
+    { 0x311 },			/* inverted breve */
+    { 0x312 },			/* turned comma */
     { 0x313, 0x2bc, ',' },	/* comma above */
-    { 0x2bd },			/* reversed comma */
+    { 0x314, 0x2bd },			/* reversed comma */
     { 0x315, 0x2bc, ',' },	/* comma above right */
     { 0x316, 0x60, 0x2cb },	/* grave below */
     { 0x317, 0xb4, 0x2ca },	/* acute below */
-    { 0 },			/* left tack */
-    { 0 },			/* right tack */
-    { 0 },			/* left angle */
-    { 0 },			/* horn, sometimes comma but only if nothing better */
-    { 0 },			/* half ring */
-    { 0x2d4 },			/* up tack */
-    { 0x2d5 },			/* down tack */
+    { 0x318 },			/* left tack */
+    { 0x319 },			/* right tack */
+    { 0x31a },			/* left angle */
+    { 0x31b },			/* horn, sometimes comma but only if nothing better */
+    { 0x31c },			/* half ring */
+    { 0x31d, 0x2d4 },			/* up tack */
+    { 0x31e, 0x2d5 },			/* down tack */
     { 0x31f, 0x2d6, '+' },	/* plus below */
     { 0x320, 0x2d7, '-' },	/* minus below */	/* 320 */
-    { 0x2b2 },			/* hook */
-    { 0 },			/* back hook */
+    { 0x321, 0x2b2 },			/* hook */
+    { 0x322 },			/* back hook */
     { 0x323, 0x2d9, '.' },	/* dot below */
     { 0x324, 0xa8 },		/* diaeresis below */
     { 0x325, 0x2da, 0xb0 },	/* ring below */
     { 0x326, 0x2bc, ',' },	/* comma below */
-    { 0xb8 },			/* cedilla */
-    { 0x2db },			/* ogonek */		/* 0x328 */
+    { 0x327, 0xb8 },			/* cedilla */
+    { 0x328, 0x2db },			/* ogonek */		/* 0x328 */
     { 0x329, 0x2c8, 0x384, '\''  },	/* vertical line below */
-    { 0 },			/* bridge below */
-    { 0 },			/* real arch below */
+    { 0x32a },			/* bridge below */
+    { 0x32b },			/* real arch below */
     { 0x32c, 0x2c7 },		/* caron below */
     { 0x32d, 0x2c6, 0x52 },	/* circumflex below */
     { 0x32e, 0x2d8 },		/* breve below */
-    { 0 },			/* inverted breve below */
+    { 0x32f },			/* inverted breve below */
     { 0x330, 0x2dc, 0x7e },	/* tilde below */	/* 0x330 */
     { 0x331, 0xaf, 0x2c9 },	/* macron below */
     { 0x332, '_' },		/* low line */
-    { 0 },			/* real low line */
+    { 0x333 },			/* real low line */
     { 0x334, 0x2dc, 0x7e },	/* tilde overstrike */
     { 0x335, '-' },		/* line overstrike */
     { 0x336, '_' },		/* long line overstrike */
     { 0x337, '/' },		/* short solidus overstrike */
     { 0x338, '/' },		/* long solidus overstrike */	/* 0x338 */
-    { 0 },
-    { 0 },
-    { 0 },
-    { 0 },
-    { 0 },
-    { 0 },
-    { 0 },
+    { 0x339 },
+    { 0x33a },
+    { 0x33b },
+    { 0x33c },
+    { 0x33d },
+    { 0x33e },
+    { 0x33f },
     { 0x340, 0x60, 0x2cb },	/* tone mark, left of circumflex */ /* 0x340 */
     { 0x341, 0xb4, 0x2ca },	/* tone mark, right of circumflex */
     { 0x342, 0x2dc, 0x7e },	/* perispomeni (tilde) */
     { 0x343, 0x2bc, ',' },	/* koronis */
-    { 0 },			/* dialytika tonos (two accents) */
-    { 0x37a },			/* ypogegrammeni */
+    { 0x344 },			/* dialytika tonos (two accents) */
+    { 0x345, 0x37a },			/* ypogegrammeni */
     { 0xffff }
 };
 


skef:
The changes to `SC_PSDump()` in this request suggest that the combined process of a) constructing precombined characters followed by b) exporting a Type 1 font has always been broken. Is that true? 

I think I agree, contra to my #3708 notes, that just unlinking problematic references is a better solution than adding unmapped glyphs. So the spirit of this fix is a good one. But as coded this is a pretty blunt instrument. What if `x0308` (for example) is mapped and output in the font? If references don't work at all, why not add code to always remove them?

ctrlcctrlv:
References work at least in some cases, I did a few tests with the generated PFA from my font which showed that to be the case. To be honest, I think George may have just had a problem with some 90's-early 2000's era software given when he committed the message and declared that a PostScript font which refers to characters in that range won't work.

However, as I wrote:
> Type 1 fonts never were a first class format for storing font data—and certainly haven't become one—
that's what SFD is for. The only discernible difference therefore is a small increase in filesize.

Therefore we could literally dissolve every reference and not be doing anything wrong. FontForge's OTF (CFF mode) generation code seems to do just that and no one has complained.

fontforge/dumppfa.c
```diff
@@ -977,16 +977,30 @@ void SC_PSDump(void (*dumpchar)(int ch,void *data), void *data,
...
     for ( i=first; i<=last; ++i ) {
+
+    // Before we do anything, convert references to glyphs between 0x300 and
+    // 0x345 to splines. This is because Type1 fonts don't support references
+    // to them—or at least didn't in 2001—this format is ancient and retained
+    // primarily for backwards compatibility.
+    if ( sc->layers[i].refs!=NULL ) {
+        for ( ref = sc->layers[i].refs; ref!=NULL; ref=ref->next ) {
+            if (ref->unicode_enc >= 0x300 || ref->unicode_enc <= 0x345) {
+                SCRefToSplines(sc, ref, i);
```
skef: So: Starting a separate, totally independent code review conversation here ...

If this works at all it's because `SCRefToSplines()` modifies the `sc` argument, because there's no other way for it to have an effect.  Does this function somehow always get called with a copy of the font? Because otherwise exporting and then saving is going to unlink references, which is at least counter-intuitive (and probably just "wrong").

skef: (Added: Happily, there's already a code "model" of how to accomplish what's needed below in the `refs_to_splines` stanza. It's not exactly intuitive, but it's there.)

skef:
I would say the larger message here is that something along the lines of this push is an appropriate fix. That leaves the smaller issue being whether something more general (e.g. less "hard-coded") could be arrived at and how hard that would be. Perhaps everyone can wind up content with a fix. 

Just jawboning here: `SC_PSDump()` isn't called in that many places. Suppose that `refs_to_splines` were to become an array pointer, where NULL means "convert no refs", 0x1 (or whatever other sentinel pointer value) means "convert all refs", and an actual pointer to a null-terminated array means "convert these refs". That would give the necessary level of control in the parameters to do whatever was desired with refs. It would require rewriting the `refs_to_splines` if/else to work per-reference, but both parts loop over the reference list so it's probably not that hard to turn the code inside-out. 

Then if, when outputting a Type 1 font, you could make a list of every glyph that 1) is used as a reference and 2) won't actually be written into the font, that list could be passed in the NULL-terminated array. 

Assuming that were possible, would it make everyone happy? @frank-trampe?

frank-trampe:
It still seems to me that it would be simpler just to have conditional selection of the accent scheme ("regular" or "legacy") to use, possibly based on export format such that we would not need to expose it in the user interface or the script bindings. That provides exactly what @ctrlcctrlv needs and preserves existing behavior too.

skef:
@frank-trampe How do you know the export format at the time the precombined glyph is constructed?

frank-trampe:
Oh, indeed. I guess it would need an option. Let me see where we would need to expose it.

skef:
But who would this option be "for"? The only user I can think of is someone building a second Type 1 font from scratch. *No one* is going to think about this issue in advance, so the person building their first Type 1 font will wind up being told "you should have selected X -- clear your chars and hope that doesn't screw up any kerning groups you've set up." And it still doesn't do anything to help the person trying to generate a Type 1 font from a source in a different format.

ctrlcctrlv:
Once again, adding an option is totally pointless and I oppose it on principle. No one would use it. On the off chance someone is *relying* on whether or not a type 1 font contains references, which affects its function in no way, let them use an old version of FontForge. We don't need to add an option every time we change some minute function of the software. You are too conservative.

ctrlcctrlv:
Why can't you guys see that it doesn't matter whether or not the generated font contains references or splines?? Even users of this format will never notice or care! George cared twenty years ago due to tiny hard drives and RAM. Optimizing references is pointless, I'm a broken record but my points just keep being ignored.

skef:
> Why can't you guys see that it doesn't matter whether or not the generated font contains references or splines?? Even users of this format will never notice or care! George cared twenty years ago due to tiny hard drives and RAM. Optimizing references is pointless, I'm a broken record but my points just keep being ignored.

If this is accurate then there is a one-line fix for the Type 1 issue, which is to call `SC_PSDump()` with `refs_to_splines` set to true.

frank-trampe:
@ctrlcctrlv, I understand your frustration here. You're not getting good output here and there's an easy way to get good output for your case.

But there are a lot of other people using FontForge with other needs, and breaking one of their workflows isn't just a disaster for them but costly for us since we need to spend time hunting down the problem (more difficult with binary formats).

I can't know all the use cases, but, as an example, some Linux distributions use FontForge to generate fonts. They need to be able to build from source but to get the same binary result each time. To accommodate that, we added options to override the generation timestamp. We also made a set of other changes, if I recall correctly, to get them to give up their forks or deep patch stacks. This need for bit-for-bit reproducibility may also exist in validation tests in other workflows.

There are several ways that we can accommodate you and them without one of you forking the project, and I would like to pursue those before we settle for less.

ctrlcctrlv:
I package for Arch Linux. Sometimes the output of a program varies version to version even when the input stays the same. _This is totally normal and not a problem at all._ Sometimes the old version was broken in a subtle way or the new version writes some new metadata. Expecting FontForge to produce fonts with exactly the same hash version to version is completely ridiculous, we cannot change anything without adding tons of pointless settings with that approach. Quit obstructing this completely benign patch.

ctrlcctrlv:
The idea that a hash of an output font changing version to version is a disaster stretches the bounds of the word hyperbole to the breaking point. WTF.

frank-trampe:
A large part of the magnitude of a problem is in how easy it is for you to fix it. A different hash result might not be a disaster for you in this case since you understand what's going on. For somebody else, though, who doesn't work on FontForge, even a good engineer, it can quickly become one and run through a lot of time.

I really don't think there's reason to be terribly angry here. You have a compatible fork that works for you, and you can probably merge from master to get other patches until we merge a more accommodating fix there, in which case you will get your desired results from the mainline.

ctrlcctrlv:
Anyone sensible would just change the hash when they upgrade. It wouldn't be a problem for anyone ever in history. This discussion is tiresome.

ctrlcctrlv:
So are you saying that going forwards any patch which changes FontForge cannot change the hash of any output font in any situation? Is that seriously your position? There's ample reason to be angry at your intransigence.

frank-trampe:
That's not my position. See my UFO changes from the last release. But removing an entire class of data, currently stored correctly, from a binary format comes with the risks I've stated. The discussion is indeed tiresome, but some of that may just be the time of night. Let's get some sleep and regroup tomorrow.

ctrlcctrlv:
I live in UTC+8. You've found a way to tire me out by 11AM. It's not an "entire class of data"&mdash;what does that even mean? How many times do I have to say that a Type 1 font with references functions identically to one without for it to sink in? I'm not even unlinking every reference, just ones to the Unicode combining characters which this ancient font format isn't even supposed to be using.

fontforge/dumppfa.c
ctrlcctrlv: @skef SCRefToSplines modifies `sc`, the code after this then uses the modified `sc` to write out the font.

ctrlcctrlv:
Oh, and by the way @frank-trampe, Type 1 fonts are not a binary only or even binary usually format, it's extremely easy to just save as &ldquo;Type 1 (Ascii)&rdquo; and compare the output of two FontForge versions. :man_facepalming: 

But no one would ever need to do that since it doesn't matter **_TO THE FONT'S FUNCTION_** if the font contains references or not, as I've stated over and over again to no effect. :rage: 

But yet I'm the problem, I'm told I should just maintain a fork of FontForge so you don't have to merge something which you've arbitrarily decided not to like.

fontforge/dumppfa.c
skef: Right, but `sc` is (unless a copy is being made) a data structure maintained by FontForge as the in-memory representation of the glyph. So modifying `sc` is modifying the internal font representation.

ctrlcctrlv: Oh, OK, I see what you mean. I can confirm that it is called with a copy of the font; when I save a PFA it doesn't cause the reference to be dissolved in my open SFD, no.

ctrlcctrlv: If you want though @skef, to make this completely bulletproof since it's called in multiple places, I can change the code so everything works on a copy of `sc` and not the original. Let me know. :-)

jtanx:
Folks, this discussion is getting a bit heated, let's take a step back please, and keep this discussion civil.

I haven't been following this at all, and this is definitely not my area of expertise, so bear with me.

For this change, I agree *in principle* with it, on the following assumptions:
* Functionally, the font behaves the same (sounds like it does, since it seems like we've just flattened the problematic references). Even better if you can extend the test suite to check for this.
* No one really actively uses type 1 fonts any more (is this true? I guess so, but I haven't done an extensive search). But even if this wasn't the case, as long as the first point is true, I don't really see much of an argument otherwise - of course, if there is a better way, that somehow easily preserves the references, then that's fine too. 

As for the reproducability argument - I also agree with @ctrlcctrlv, reproducibility *between* versions of software is not something that needs to be enforced. Sure, decreasing the amount of churn is nice, but sometimes it's unavoidable.

imo, that's *completely* separate to reproducibility with the *same* version, run with the same inputs - that's just nondeterminism in the output. I think that's closer to what reproducible builds is trying to target, and not changes in output between versions of software.

ctrlcctrlv:
According to my research, the only place Type 1 fonts remain in active use is in (very) old PDF files. Since version 1.6 of the PDF standard, OpenType fonts are supported as fonts embedded in PDF files. Even in cases where you'd want to make a Type 1 font for embedding in a PDF, this patch has no effect on that, glyphs would still look the same.

Some software might downconvert an OpenType font to a Type 1 font in order to print to an (old) PostScript printer. Some software might downconvert an OpenType font to put it into an EPS file (I think). However, again, this patch has no effect on that either, users won't notice any references that have disappeared.

The format is ancient. Adobe, who made the format, [stopped developing new Type 1 fonts in 1999](https://blog.typekit.com/2005/10/06/phasing_out_typ/). No one is making new Type 1 fonts today unless they're a hobbyist just doing it "because they can" to play with some old hardware. Even if they are, it's not a first class format so it doesn't matter if references get dissolved.

I apologize for getting angry but that hash argument was made in bad faith. That's never been the standard for patch acceptance and we all know that's never been the standard. I'm willing to assume malice wasn't intended if @frank-trampe doesn't come back making the same argument tomorrow.

Once again, I encourage all members to support the _principle_ of this change, if not the code itself, which might have bugs, which I will fix if they are expounded upon clearly. Once again, I reiterate that there's absolutely nothing wrong with this change in principle.

frank-trampe:
@ctrlcctrlv, I'm aware of the pfa format. FontForge was once called pfaedit, in fact! It's still about as readable as PostScript, though.

I can't speak for others, but, I have 740 pfbs and 3 pfas in various places under /usr on my main computer (running Ubuntu 18.04, not IRIX or anything like that), many from TeX packages, but also various installed typefaces. I also have some in packages that I use under wine. Somebody maintains each of those, and, more often than not, he probably uses FontForge.

As FontForge's original use case and the occasional query we get on the Facebook page demonstrate, people still routinely debug and edit "compiled" typefaces. I'm not advocating a perfect-round-trip crusade as I've pursued on UFO for the last few years, but I think that it remains important to avoid eroding whatever round-trip capability we have. If a typeface has references, it is by the choice of some designer, and it's not our place to drop those gratuitously when the user is trying to make a small adjustment.

fontforge/dumppfa.c
skef: Oh: `if (ref->unicode_enc >= 0x300 || ref->unicode_enc <= 0x345) {` should be `if (ref->unicode_enc >= 0x300 && ref->unicode_enc <= 0x345) {`, right?

skef: If you were to:

1. Change the top guard to be `if ( sc->layers[i].refs!=NULL && !refs_to_splines ) {`
2. Confirm that every call path for the function that doesn't set `refs_to_splines` to true calls with a copy of `sc` (or an `sc` that will always be discarded for some other reason).
3. Add a comment about the need to call with a copy.

Then I think your approach would be fine. There aren't that many callers, but this function is used (for example) to export Encapsulated PostScript glyphs (with `refs_to_splines` set to true), and I don't know if it uses a copy of `sc` in that case. 

However, as I indicated somewhere else, it might be most "elegant" to invert the `refs_to_splines` clause in the later part of the function. Since the code is "right there" I guess my recommendation would be to try fiddling with that for a half-hour or so, and see if it works straightforwardly.

To be clearer, I'm recommending eliminating your loop and changing what now looks like:
```
            if ( refs_to_splines ) {
                if ( !pdfopers || !sc->parent->multilayer ) {
                    /* In PostScript, patterns are transformed by the page's */
                    /*  transformation matrix. In PDF they are not. */
                    /* Of course if we have no patterns we can still use this code */
                    for ( ref = sc->layers[i].refs; ref!=NULL; ref=ref->next ) {
                        A;
                    }
                } else {
                    /* If we get here we are outputting pdf, are type3, refs_2_splines */
/*  We need different gradients and patterns for different transform */
/*  matrices of references */
                    int j;
                    for ( ref = sc->layers[i].refs; ref!=NULL; ref=ref->next ) {
                        B;
                    }
                }
            } else {
                dumpstr(dumpchar,data,"    pop -1\n" );
                for ( ref = sc->layers[i].refs; ref!=NULL; ref=ref->next ) {
                    C;
                }
            }

```
to something like:
```
    int ref_to_dump = 0;

            for ( ref = sc->layers[i].refs; ref!=NULL; ref=ref->next ) {
                int j;
                if ( refs_to_splines || (ref->unicode_enc >= 0x300 && ref->unicode_enc <= 0x345) ) {
                    if ( !pdfopers || !sc->parent->multilayer ) {
                        /* In PostScript, patterns are transformed by the page's */
                        /*  transformation matrix. In PDF they are not. */
                        /* Of course if we have no patterns we can still use this code */
                        A;
                    } else {
                        /* We are outputting pdf, are type3, converting a ref to splines */
                        /* We need different gradients and patterns for different transform */
                        /* matrices of references */
                        B;
                    }
                } else
                    ref_to_dump = 1;
            }
            if ( ref_to_dump ) {
                dumpstr(dumpchar,data,"    pop -1\n" );
                for ( ref = sc->layers[i].refs; ref!=NULL; ref=ref->next ) {
                    if ( ref->unicode_enc >= 0x300 && ref->unicode_enc <= 0x345 )
                        continue;
                    C;
                }
            }
```
Now **this will reorder**. But given that all the splines are output first anyway, I think it will reorder in the same way your modification does (in that it converts certain references to splines).

skef:
I guess I should clarify: In the code review thread I'm working with @ctrlcctrlv on potential issues *with this particular approach*. That's separate from the discussion on the merits of that approach. So maybe I/we should suspend the review until there's more clarity at the top level. 

At that top level, I continue to think that an option that influences what glyphs accent references are pulled from (or the priority ordering of those glyphs) strikes me as close to useless, because basically no one will know that option exists at the time they need to. 

Since discussion seems to have moved from #3708 to here, let me ask: Do we **know** that there is still a problem if we just reorder the array, and if so **what** that problem is? Some font has content encoded as (for example) 0x308. That content is then used by reference in a precombined character. The user exports that font to a pfa. What, specifically, goes wrong? Is it that the 0x308 glyph isn't output, so the reference dangles? Is it that the 0x308 glyph is output, but in a way that prevents it from being used as a reference? *What is the problem this code needs to solve?*

frank-trampe:
@skef, that's a good question. I had assumed that the code change would result in a change to pfb output, but it might not now that you mention it. Perhaps we could generate a pfb from the sample file that @ctrlcctrlv provided in #3708 on each of the two possible accent mappings and see how it works? @ctrlcctrlv, does that, based on what you've seen so far, seem likely to give meaningful results?

frank-trampe:
@ctrlcctrlv, would you describe the part of your workflow relevant to this issue so that I can explore options for exposing an accent map switching option?

@skef, I realize that the accent map mode option is of limited utility if somebody does not know to look for it. What might work is to store it in the splinefont (stored in sfd also), defaulting to the new accent map but set to the old mode when importing from a Type 1 file. The other upside of storing in the splinefont is that we would not need to add arguments to any functions.

frank-trampe:
Oops. I submitted before I was done. But anyway, setting the flag on imported pfbs takes care of the round-trip case, and we could probably reach the people who have automated workflows with a bulletin on the mailing list as we did for the UFO 3 changes. Another option would be to have the no-option default be the old way but to populate the new behavior in SFD, which would be even less likely to disrupt old workflows.

skef:
> What might work is to store it in the splinefont (stored in sfd also), defaulting to the new accent map but set to the old mode when importing from a Type 1 file.

But isn't the conversion path from type 1 to, say, OpenType at least as likely a scenario if not more likely? Who is opening PostScript fonts, adding new precombined characters, and saving the result as a PostScript font? Changing the priority for "precombining" and leaving all the other code as it is will not change FontForge's *conversion* behavior (by definition). 

I still don't understand what GW's actual concern was in 2001-6. If we don't know why the priority needed to be that way when PostScript fonts were common it's hard to judge how it needs to be now.

skef:
@ctrlcctrlv: On yet another topic, in #3708 you imply, correctly in my view, that custom components (your helpful addition) should have priority over these selection heuristics. But this pull request doesn't enforce that priority. Have you thought about that, and how you might implement it?

frank-trampe:
I think that the expectation of correct mapping of all data is a lot lower for conversion than for round-tripping. The least reversible damage is that done to source files, and those get round-tripped repeatedly. As you mention, though, there is a very good chance that just changing the order of the accent map would not affect round-tripping of pfbs. In that case, I would be fine having the accent map option obscured and emitting a warning when exporting to Type 1 with an invalid accent. I'm sure that we could get fair warning to the most important downstream users to check their workflows.

khaledhosny:
How are references handled in the Type 1 code, using `seac` operator? If it is the case, then I’d suggest doing away with accent operator altogether as it has many limitations and was generally a bad idea even for Type 1 era. Subroutinization is a more general technique (even superior to references as it can be used for any part of the path) and FontForge should already support it.

Here is how `seac` operator is documented in Type 1 spec:
> **seac** ⊢ asb adx ady bchar achar **seac** (12 6) ⊢
>
> for *standard encoding accented character*, makes an accented character from two other characters in its font program. The *asb* argument is the *x* component of the left sidebearing of the accent; this value must be the same as the sidebearing value given in the **hsbw** or **sbw** command in the accent’s own charstring. The origin of the accent is placed at (*adx*, *ady*) relative to the origin of the base character. The *bchar* argument is the character code of the base character, and the *achar* argument is the character code of the accent character. Both *bchar* and *achar* are codes that these characters are assigned in the Adobe StandardEncoding vector, given in an Appendix in the *PostScript Language Reference Manual*. Furthermore, the characters represented by *achar* and *bchar* must be in the same positions in the font’s encoding vector as the positions they occupy in the Adobe StandardEncoding vector. If the name of both components of an accented character do not appear in the Adobe StandardEncoding vector, the accented character cannot be built using the **seac** command.
>
>The **FontBBox** entry in the font dictionary must be large enough to accommodate both parts of the accented character. The **sbw** or **hsbw** command that begins the accented character must be the same as the corresponding command in the base character. Finally, **seac** is the last command in the charstring for the accented character because the accent and base characters’ charstrings each already end with their own **endchar** commands.
>
> The use of this command saves space in a Type 1 font program, but its use is restricted to those characters whose parts are defined in the Adobe StandardEncoding vector. In situations where use of the **seac** command is not possible, use of **Subrs** subroutines is a more general means for creating accented characters.

skef:
@khaledhosny The postscript-specific code in question (so far) isn't conditioned on whether something is an accent. All it "knows" is that some glyph being output includes a reference to another glyph, which then needs to be included (in effect) either by value or reference. When by value, it just writes the strokes and fills more or less like it would for a spline. When by reference it winds up writing a command string like:
```
1 index /CharProcs get /[glyphname] get exec
```

skef:
OK, more info: `SC_PSDump()` is only (non-recursively) called with `refs_to_splines` set to false in one case, in `dumpproc` on line 1177 of `dumppfa.c`. So on my reading *all* of this reference stuff is strictly a Type 3 issue, and has nothing to do with Type 1 at all (at least in FontForge).

skef:
Does anyone have any evidence, at this point, that reordering the accents table and making no other change would cause a problem? @ctrlcctrlv indicates he included his loop in light of GW's comment, not because of any problem he was having.

ctrlcctrlv:
> Do we know that there is still a problem if we just reorder the array, and if so what that problem is?

No. I only changed `dumppfa.c` based on George's 2001 advice, I didn't experience anything that led me to feel it was necessary firsthand.

> On yet another topic, in #3708 you imply, correctly in my view, that custom components (your helpful addition) should have priority over these selection heuristics. But this pull request doesn't enforce that priority. Have you thought about that, and how you might implement it?

This patch will fix that as a side effect, no change to my User Components (Glyph Info&hellip;&rarr;Components) feature is needed if this is pulled. The problem with just using my added User Components feature, which has been merged, wrongly assumes that FontForge will always prefer the Unicode decomposition when building an accented charter. That is to say, when I wrote that code I assumed combining characters were always superior to spacing characters for the purposes of building a glyph, so it makes no sense to specify you want to build with the combining character because my code assumes that that will always happen if the combining character exists in the font in the first place. 

That is to say, I wrongly thought that the Unicode decomposition would be the first thing FontForge tried when building a glyph, so it's impossible to specify you want to use a &ldquo;non-default&rdquo; decomposition which is equal to the Unicode decomposition.

ctrlcctrlv:
>@ctrlcctrlv, would you describe the part of your workflow relevant to this issue

I have a font in development (TT 2020) which has glyphs for both a spacing dieresis and a combining dieresis. It's a typewriter font that will enforce (pseudo) randomness using contextual alternates, so that e.g. in the word `Beekeeping`, the repeating `e`'s will not look exactly the same. I am writing GIMP scripts to accomplish this, I've sketched out one already [here](https://gist.github.com/ctrlcctrlv/b73de5420cc712893dc7b01cdd9e0c2d). This project is highly ambitious and I've been patching FontForge and writing scripts with it in mind for months. I made my font Chomsky with the main purpose of getting User Decompositions accepted and making a build system for TT 2020.

My build system doesn't work right if combdieresis is not superior to dieresis. To put it mildly, it fucks up all my precombined characters. Chomsky didn't have a spacing dieresis so I didn't catch this problem with it, but TT 2020 does. My User Decompositions feature wasn't expecting spacing characters to ever be seen as superior so it's no help either.

This is a massive impediment for me. I don't want to have to maintain a fork forever due to this obvious bug. Just accept the principle of the patch Frank... :disappointed:

ctrlcctrlv:
I'm going to be making at least four automatically distressed styles plus the base font, which is based on, but not identical to, Smith Corona's so-called &ldquo;Pica 10 Pitch&rdquo;. I've also been influenced by Remington's Greek faces and Olympia's Elite Italic. Bitstream's "Pica 10 Pitch" is another influence. I purposefully avoided Courier because it's overdone and there are many other typewriter faces, once prominent, which now languish in obscurity and deserve their time in the sun.

TT

All of the glyphs are built from Spiros in the back layer. `i`, `j`, `period` have circles in a special &ldquo;No Stroke&rdquo; layer which are copied. I also made Greek, Cyrillic, Hebrew and Armenian. That's my project. It is complicated and ambitious and this will probably need other tweaks to FontForge before I'm done.

skef:
> That is to say, I wrongly thought that the Unicode decomposition would be the first thing FontForge tried when building a glyph, so it's impossible to specify you want to use a “non-default” decomposition which is equal to the Unicode decomposition.

I guess I'm saying something stronger, which is that a decomposition chosen by hand should override any other type of selection. So, for example, *after* your change to the `accents` array, if I wanted FontForge to pull from the spacing characters rather than the combining characters, I would assume I could just specify the spacing character manually. But that still won't work.

skef:
@ctrlcctrlv I suggest re-pushing without the changes to `SC_PSDump()`. I'm not positive but I suspect there's momentum in the direction of just reordering the `accents` array. I can always file a separate issue on the priority of manual decompositions.

frank-trampe:
I agree with that on the stated merits but also on the basis that it makes it a lot easier for somebody to work around a problem like the underlying one here.

skef:
OK, I've filed #3713 to get that issue out of the way of this push.

fontforge/fvcomposite.c
```diff
@@ -50,20 +50,17 @@ int CharCenterHighest = 1;
 #define BottomAccent	0x300
 #define TopAccent	0x345
 
-/* for accents between 0x300 and 345 these are some synonyms */
-/* type1 wants accented chars built with accents in the 0x2c? range */
-/*  except for grave and acute which live in iso8859-1 range */
 /*  this table is ordered on a best try basis */
```
skef: Now, as reviewers, I hope the folks on this thread can help out with the other issue posed by this push: what this table should look like. 

My question is: Shouldn't every entry either start with 0 or with the corresponding 0x3?? value? (Or maybe just the latter, for clarity's sake?) @ctrlcctrlv has been conservative in just reordering the entries and not adding anything, but aren't "hook above" and "reversed comma" even more problematic?

frank-trampe:
As for this more limited effort, I think that we still need a way to enable the old accent map in case the problem that George described does actually exist. At this point, it's easier just to assume the worst than to try to nail down precisely what's going on with the accent map. On the flip side, it is of sufficiently narrow scope that I think that it's okay for the option to be non-obvious. We can send a notice out on the announcement list describing the change, the reason for it, and the option to preserve the old behavior, and I would expect it to get to most of the right people. If it turns out that the new default does somehow cause widespread strife, we can just change the default back to the old way but preserve the option to use the correct priorities. Either way, having the option also lets us be a lot more aggressive about reworking the map. It is also worth looking at the other maps to see whether they have logical issues.

So something like this.
* A parallel accent map or maybe parallels for all maps in fvcomposite.c.
* A new value sf->accent_mode or something more general if we work on all maps, stored in SFD and defaulting to logically correct prioritization.
* A switch in all functions in fvcomposite.c that use the accent array (with an extra argument added to CanonicalCombiner since it does not take sf or sc as an argument).
* A checkbox in the Font Info dialog to enable legacy accenting or the more general thing.
* Appropriate getters and setters in the scripting interfaces.

frank-trampe:
And we would not need to unlink the references that way, of course. But does that sound like a satisfactory plan?

skef:
@frank-trampe I have an alternative proposal:

1. Fix #3713 
2. I write a python script that recreates the effect of the old `accents` array using manual overrides. (That is, it checks each entry in order for content, and if it finds content sets that manual override.)

Then anyone with legacy issues can use the script instead. We could even stick it in `contrib`.

fontforge/fvcomposite.c
frank-trampe: @skef, do all of the accents in the table have combining marks?

skef: I believe it does by definition. It's indexed like `apt = accents[uni-BottomAccent]`, so what the table represents is the set of combining characters between 0x300 and 0x345.  There aren't any "empties" in that range: https://vazor.com/unicode/b0300.html

frank-trampe:
@skef, I've stared at the ceiling for a few minutes to think it through, and that seems defensible to me. In effect, the current prioritization of the non-combining accents is a hack, not inherently correct behavior, and we're simply replacing it with another more flexible if slightly less accessible hack.

fontforge/fvcomposite.c
frank-trampe: I would concur with your suggestion then. George may know something that we don't here, but it seems to me that we ought to be consistent across the whole range. @ctrlcctrlv?

skef:
@frank-trampe that's more or less my thinking. I doubt anyone will wind up using the script but I can see the argument for having it available. And if we can put the solution in an outside script that seems far preferable (to me) to adding a bunch of C code and some UI for something we're not *sure* is an issue. (It raises difficult questions of how to test that code, for example.)

fontforge/dumppfa.c
skef: I believe we're heading towards eliminating this code, so I'm resolving this part of the discussion.

skef:
Come to think of it, it would be both easier and closer to the existing implementation to implement the work-around with a "hook". Suppose one could register a python function that gets "second crack" at specifying what glyphs to pull references from. (With manual overrides getting "first crack" and FontForge's internal heuristics coming third.) It gets called with the unicode point of the "target" and a reference to the font, and it either returns a set of glyphs or "None" to punt to FontForge's choice. That would be a handy hook to have for a variety of reasons.

frank-trampe:
That would be really neat. I would be surprised if it turned out to be less work, though.

skef:
@frank-trampe Similar amount of work but much more utility to the project for the effort.

frank-trampe:
Quite true.

ctrlcctrlv:
As long as combining characters come first in the `accents` array, as I wrote, #3713 will be fixed as a consequence of merging this. If you require a thorough explanation of the code to see why I can provide one, but since that takes time I'll wait for you to ask for one first.

skef:
> As long as combining characters come first in the accents array, as I wrote, #3713 will be fixed as a consequence of merging this.

I read this and figured "that's easy enough to test!" So I pulled your branch, cleared `Edieresis`, and first built using the defaults, which had the expected result:



Then I cleared the glyph again, used the GlyphInfo -> Components dialog to set a custom decomposition of 45 a8, and rebuilt. And you're right -- it does use the expected references, albeit in a strange way:

a8

Given that the glyph built using 45 and a8 before the accents array was reordered looked close to ideal and how the 45 308 glyph looks after, I think it would be worth going into the detailed explanation, because at first glance it seems like the order in which the heuristics are being applied is important and maybe needs to change.

skef:
Anyway, that conversation can happen in the #3713 issue. There was a bunch of energetic conversation about the changes in this push, and I think there is now momentum for

1. Removing the PS export changes
2. Making the combining character itself be the first entry in each `accents[]` line. 
 
@ctrlcctrlv do you have any objections to this approach? And if not are you interested in making those changes or would you prefer someone else do that "draft"? 

Depending on how @frank-trampe feels it may be appropriate to make a temporary branch to do this work in, so that the master doesn't change until everything envisioned is ready. Or maybe the pushes could be left approved but unintegrated.

frank-trampe:
In order to push to master, we would also need the scripted replacement for the presently hard-coded hack that @skef has mentioned, and we would need to send out a technical bulletin slightly in advance of the merge. I'm happy to write the bulletin, but it will probably require help from both of you. I propose that we copy this branch into master and continue development and testing without squashing until we're ready to merge.

fontforge/dumppfa.c
ctrlcctrlv: Per our discussion I've just decided to revert `dumppfa.c` to its state in master.

ctrlcctrlv:
User decompositions are implemented by changing `SFGetAlternate`. I added a new property to the `SplineChar` of type `unichar_t` called `user_decomp`. When `SFGetAlternate` is passed a `SplineChar`, `sc`, it checks if that property exists, and returns it if so.

This approach wrongly assumes that FontForge will always prefer the Unicode decomposition. Why? Because `SFGetAlternate` always returns the Unicode decomposition if there's no user decomposition! So telling `SFGetAlternate` to return a user decomposition equal to what it would already return is... :drum: :drum: **pointless!**

I wrongly assumed `GetGoodAccentGlyph` was sane and would always prefer Unicode decompositions&hellip;I was wrong and this bug made it past review.

That's the technical explanation. Again, if you just merge this, you'll also fix #3713.

OK, so why with my branch does the user decomp E&#xa8; (U+45 U+A8) render weirdly? A quick fix is just to add anchors, but I doubt that will satisfy you; after all, it does build right in master. The problem is of course `_SCCenterAccent`. The lack of position data in `ff_unicode_utype2` in `utype.c` is what causes what you see&mdash;`fvcomposite.c:2215`. 

In master, `SCBuildComposit` gives `_SCCenterAccent` U+A8 to center, but U+308 as the data source. You can see on `fvcomposite.c:2847` where it gets U+308 as the data source by calling `SFGetAlternate` and getting the Unicode decomposition. With user decompositions it would get U+A8 there. `SCCenterAccent` then calls `GetGoodAccentGlyph`, setting it up to position U+A8, but use the data from U+308.

Now look, my answer here is simple. Use anchors! Hopefully you can now see why fixing this would be hard. It's not guaranteed with user decompositions that the user would want to pass the position hint, in this case `FF_UNICODE_ABOVE`&hellip;maybe they want to compose a glyph that Unicode says has an accent above with an accent below. Why? I don't know, but the possibility is there. We could just change `ff_unicode_utype2` to make U+A8 have hint `FF_UNICODE_ABOVE`, but again, that's not good, U+A8 is just a spacing dieresis, I'm not convinced it's a spacing dieresis **above**. _Sîn-pet-sṳ_ is a word in Hakka Chinese, note the combining dieresis below.

Perhaps I could let the user choose the hint to pass, defaulting to the Unicode hint? But again, this is suboptimal, what do we do when there are more than two codepoints? Seems like now we need a whole domain specific language! And will users understand it? Is it a waste of developer time? (My answer is yes.)

frank-trampe:
Agh. We'll sort this out somehow. I'll need to explore the code a bit to understand fully what you're saying.

skef:
@ctrlcctrlv This sounds like a case of the existing code being organized questionably, and in a way that makes it difficult to extend as you have. The mixing of data sources is pretty good evidence of that kind of problem. So this certainly isn't your fault, and it's only your problem to the extent that developers trying to do new things always get stuck with a certain amount of this maintenance and updating work. 

On the other hand, your prediction that some will see "just use anchors" as a sufficient answer is right.

@frank-trampe has said he'll take a look at the code, and I may have some time to do that too. We should at least be able to judge how much work it would be to separate the construction stages properly.

ctrlcctrlv:
I just spent an hour writing a long proposal, which kept getting longer and longer as I kept finding more and more edge cases, therefore I throw in the towel, a domain specific language is the most flexible way to fix this and the only way that solves all the edge cases. Luckily I don't need another hour of writing to describe it:

# &#x261e; Proposal

Right now our "language" is just numbers, one after the other, in base 16, which explain one Unicode decomposition, for example, the NFD normalization form of Ự is `55 31b 323`, while FontForge's decomposition for that letter is `1af 323`. Prefixed zeroes are ignored.

Let's formalize and legitimize the current behavior&mdash;so we're talking about *overriding* FontForge's Unicode-based data here rather than *defining* it&mdash;the default, when a character has no combining class, is to just place the character's on top of each other semi-haphazardly.

Composition classes can be differentiated from codepoints by prepending them with `~`. When I say _composition class_, I mean the composition classes defined beginning at `Unicode/utype.h:116`, at least for now, and not Unicode composition classes, which seem to be something else entirely. I can't figure out how to turn a Unicode composition class into a FontForge composition class&mdash;they are presumably somewhat related since @JoesCat updated them to Unicode 9.0 in 2016 in #2963 (which is the last time anyone has done so), perhaps he can help explain the relationship between the two, and why it doesn't seem like it's an easy to task to convert one to the other.

I believe that this is enough to provide a script which will replicate the current behavior on later FontForge versions. 

For example, the perennial E&#xa8; can be written:

```
45 ~100 a8
```

Or, as it is seen by default by `_SCCenterAccent`, even though flags below 0x100 have no meaning (I'm assuming @JoesCat knows why):

```
45 ~130 a8
```

Put another way:

* A digit is `[a-z0-F]`, case-insensitive.
* A codepoint consists of between one and eight digits.
* A (FontForge composition class) marker is `~`.
* A marker followed by a number is a (FontForge composition) class.
* A number _x_ followed by a class followed by a number _y_ applies the class when combining y to x.

skef:
I'm surprised there are so many ways of combining characters that FontForge automates (or easily could) that a composition class is needed for this. Suppose it were to only support "accent-style" above, below, to-the-right-of, and to-the-left-of as positioning options. And suppose those were indicated by prefixing the character to be attached with `^`, `_`, `>` or `<` respectively, so your example would be encoded as `45 ^a8`. Wouldn't that solve 95% of the issues, with the remaining issues still being solvable manually?

ctrlcctrlv:
Yes. Probably 99% if multiple flags could be applied and with a flag meaning "attached". But if I have to do this anyway I may as well expose everything, right?

skef:
It partly depends on what "expose everything" means. And I think the answer in general is "no": if extending an interface to cover all cases makes it overly complicated or involved, it's usually better to keep things simple. Remember that the use case here is auto-construction absent shared anchor points. Most of those scenarios will call or some hand-tuning anyway.

JoesCat:
> they are presumably somewhat related since @JoesCat updated them to Unicode 9.0 in 2016 in #2963 (which is the last time anyone has done so), perhaps he can help explain the relationship between the two, and why it doesn't seem like it's an easy to task to convert one to the other.

While doing the update (manually), I saw pretty early in the process that it made more sense to automate the build of combiners.h vs doing it manually, however, what was more important at the time was to build fontforge with Unicode 9.0 and introducing edits like this was going to create a lot of churn before a release, so it didn't happen yet. The most logical place I can think of, is to have makeutype.c build combiners.h instead of importing the existing manually built file.

Combiners 3.0 was created at a time when FontForge was default less than 65536 chars, and now users are asking for the full Unicode set. Combiners as it currently is, is a lot of empty space and could probably be built in a similar way as ligatures which span the entire code set plus also has overrides.

makeutype.c isn't ready for 65535+ lists (which is why ligatures was composed before hitting the code that needs attention at 65535+).

> Or, as it is seen by default by _SCCenterAccent, even though flags below 0x100 have no meaning (I'm assuming @JoesCat knows why):

Perhaps unicode has an answer :-)
http://www.unicode.org/reports/tr44/#Canonical_Combining_Class_Values

ctrlcctrlv:
So is last byte just the Unicode composition class Joe?

JoesCat:
Yes.
in inc/utype.h, the value mask is: #define FF_UNICODE_COMBININGCLASS	0xff
...but also consider all this other stuff too because it's manually created and related:
#define FF_UNICODE_ABOVE		0x100
#define FF_UNICODE_BELOW		0x200
#define FF_UNICODE_OVERSTRIKE		0x400
#define FF_UNICODE_LEFT			0x800
#define FF_UNICODE_RIGHT		0x1000
#define FF_UNICODE_JOINS2		0x2000
#define FF_UNICODE_CENTERLEFT		0x4000
#define FF_UNICODE_CENTERRIGHT		0x8000
#define FF_UNICODE_CENTEREDOUTSIDE	0x10000
#define FF_UNICODE_OUTSIDE		0x20000
#define FF_UNICODE_LEFTEDGE		0x80000
#define FF_UNICODE_RIGHTEDGE		0x40000
#define FF_UNICODE_TOUCHING		0x100000

data all located in Unicode/utype.c in table ___utype2[]
FYI ...we had to isolate these behind functions since someone posted several buffer overruns using the earlier defines.

ctrlcctrlv:
OK, so. Either use `~` to refer to an exact FontForge combining class, or:

Constant name | Code | Symbol
--------------- | ---- | -------
FF_UNICODE_ABOVE | 0x100 |↑
FF_UNICODE_BELOW | 0x200 |↓
FF_UNICODE_OVERSTRIKE | 0x400 |⁄
FF_UNICODE_LEFT | 0x800 |←
FF_UNICODE_RIGHT | 0x1000 |→
FF_UNICODE_JOINS2 | 0x2000 |⋈
FF_UNICODE_CENTERLEFT | 0x4000 |⮠
FF_UNICODE_CENTERRIGHT | 0x8000 |⮡
FF_UNICODE_CENTEREDOUTSIDE | 0x10000 |⮔
FF_UNICODE_OUTSIDE | 0x20000 |⥁
FF_UNICODE_LEFTEDGE | 0x80000 |⇤
FF_UNICODE_RIGHTEDGE | 0x40000 |⇥
FF_UNICODE_TOUCHING | 0x100000 |▕

Symbols can be composable. Some symbols might not make much sense, but the field is a `unichar_t`, and almost no one outside of compatibility scripts is going to use them, so why not?

JoesCat:
The flags table above only has 13 definitions and cannot define some info, like 234, possibly 10...199 too. The exact code holds a lot more information, but I don't recall having seen any sort of code that made use of it (was a while ago, and maybe I missed it).
makeutype.c, you'll want to run a printf and/or a check to ensure you are grabbing the right value, George's code does scanf for values but I found it grabbing wrong values here and there and took note of a couple places where it happened. The code needs to be re-written to separate the values at ";" otherwise scanf for values can/will grab wrong codes in a couple of places.

ctrlcctrlv:
@JoesCat I understand that, that's what `~` is for. :-)

@skef @frank-trampe Can you guys accept or comment on this proposal so I can start working on it? Then we can make a Python script to make things work as before with user decomps, and push this. Let me remind you that user decompositions are buggy without this...

skef:
@ctrlcctrlv I'm not sure I can provide effective advice without backing up a level. We've talked about a bunch of things in the context of this push, and I've lost the thread of how this particular proposal relates to this particular push. Or maybe it doesn't. 

Whether it does or not, I don't see that supporting it is necessary in the strong sense. It could be a nice thing to have, but I'm not sure that FontForge *needs* it. So to the extent I have feedback, part of it is raising the question of whether this is where resources should be invested. Of course they're mostly your resources, and you're free to choose where to invest them. 

Regardless, as far as I know this push is waiting on the questions raised in the code review comments on `fontforge/fvcomposite.c`.

fontforge/fvcomposite.c
ctrlcctrlv: @skef You're right, I was overly conservative, and I've remedied that in my latest commit to this branch.

ctrlcctrlv:
@skef Sorry about that, here you go. You're right that talk about user decompositions has crept into this issue, because we're talking about how to bring the old automatic compositions into new versions of FontForge so that @frank-trampe approves this.

skef:
OK, good! I verbally approve the changes in this push at this point. 

As far as I can tell the issues of direct contention have been worked through. However, that doesn't necessarily mean this should be merged yet. Ideally the master branch should be kept as ready to push a new version (potentially for a bug fix) as feasible.

There are two identified problems. One is the positioning bug I refer to above. However, that's not an issue the changes in this push request *creates*, and neither does this make it worse. So it's not a "blocking" problem.

The other problem is backward compatibility for people who want to build Type 1 fonts under the old circumstances. That one *is* blocking. I've discussed adding a hook to handle that case, as it would be nice to have the hook anyway. So my suggestion is to:

1. Consider this push request complete.
2. Leave it unmerged for now, waiting on the complementary hook work.
3. I'll start taking a look at that, and can be considered "on the hook for it" (although I'm not going to make any promises schedule-wise). When it's done, and I've written a hook implementation that emulates the old `accents` array, this can be merged. 
4. Someone, maybe @frank-trampe, maybe @ctrlcctrlv, should look into the problems causing the bug illustrated in the pictures above. 
5. @ctrlcctrlv can pursue adding "combining class specification" separately as he sees fit. 

Is this more-or-less OK with all parties?

frank-trampe:
This was going to be a difficult week for me, but it has turned out to be vastly more so than expected. I will not be able to get up to speed on this with a clear and well-rested mind until Tuesday. We're definitely on the right track, and I look forward to doing so.

skef:
@frank-trampe Since there's no immanent release anyway, as long as @ctrlcctrlv is OK with the all-but-approved status for this sub-part of the problem I don't see why there's any time pressure.

ctrlcctrlv:
I'm okay with it as long as my changes to User Decompositions are not done.  Once those are done and merged this should be merged, hooks or no hooks...

skef:
@ctrlcctrlv Remember that very few users are grabbing the code continuously out of the master branch. When one isn't personally blocked in virtue of compiling one's own code, and there are no objections to one's proposals going into the next release, the impact of delayed merges is quite modest.

skef:
I went quite a ways toward doing a substantial rewrite of this module, but in the end I decided it would be too much work for me to do and test. 

@ctrlcctrlv I've pushed a small additional commit to your branch that handles the custom case of an accented glyph a bit better when a non-combiner is selected. As fixes go it's modest. It also removes the `invert` logic in the custom case -- which may sometimes be disappointing but isn't obviously wrong. (Presumably if these are to be supported there should be some way of specifying themrather than having FontForge guess around.)

ctrlcctrlv:
@skef Cool&mdash;still working on a way to tell FontForge how you want it to try combining rather than it just using the Unicode combining class.
2019-06-17-165713_1282x848_scrot

skef:
The back-end to support that kind of feature was what I was looking at. Handling the general case of customization seems like a lot of work/big mess.

ctrlcctrlv:
It definitely is. I'm just hacking it in as I go. I'm thinking about whether or not I can easily make `SFGetAlternate` use (or at least accept) glyphs (so the local encoding) rather than the Unicode encoding. Then I just have to make a linked list or something for the positions and add it to SplineChar alongside `user_decomp`

skef:
If you want to chat sometime about arch ideas I'm up for that.

ctrlcctrlv:
Since `SFGetAlternate` isn't used in _that_ many places, as I see it all we need to do is change its return value to a struct that contains a unichar_t, an int (sc->orig_pos), and an enum declaring which one to use. Then it can work with user decompositions better and unmapped characters can be used. But is the time this is going to take worth it?

I could just make the Glyph Info window refuse to allow you to save names of unmapped characters&hellip;e.g., _Error: X is unmapped (Unicode value is -1) and cannot be a component. Press &laquo;Map&raquo; to put X into the first available Private Use Area slot, or &laquo;Cancel&raquo; to cancel adding X as a component._

Thoughts ??

skef:
What do you mean "an enum declaring which one to use"?

ctrlcctrlv:
-1, use the first one isn't NULL
0, use unichar_t
1, use orig_pos

skef:
Ah, hmm. I'll have to think about that a bit. By "use unichar_t" do you mean "base the position off the unicode value"? 

If `SFGetAlternate` will return (or otherwise modify) a structure instead of a simple array, that structure can include a boolean tracking whether the decomposition is custom, which is a good start. That allows the call to be moved up before `DoSpaces` in `SCBuildComposit` so it can override those four initial functions. 

I'll assume that there isn't enough interest in allowing customized Spaces, Rules, Rotation Dotlessness and Hangul. And I guess RightToLeft ligatures. (Supporting LeftToRight ligatures might be nice.) There's also a question about `invert`. 

This scheme could probably be made to work, but I'm dubious about the idea of exposing this structure outside of `fvcomposite.c`. In the places `SFGetAlternate` is called outside that file it's to pose one or another "question", such as "Is this 'duplicatable'"? I would recommend moving that code to small functions implemented in `fvcomposite.c` and then called directly. 

(I don't know why `scstyles.c` is calling `SFGetAlternate` as opposed to `SFIsSomethingBuildable`. It's annoying.) 

If `SFGetAlternate` is going to be the locus of customization I'll probably wind up adding a flag or two to support the python hook, which I also want to give the option of building the glyph directly. But you don't have to worry about that.

ctrlcctrlv:
`SFGetAlternate` already is the locus of customization; in master that's the function that actually uses `SplineChar.user_decomp`&hellip;I think that's a sane place to put additions to it.

But it returns `unichar_t`. Meaning it can never refer to unmapped glyphs. Meaning `SplineChar.user_decomp` must match its return type. 

So really I only see two possibilities:
1. Make `SFGetAlternate` return some complex struct so it can refer to glyphs by either codepoint or orig_pos (font's internal encoding)
1. Make `SFGetAlternate` always refer to glyphs by orig_pos

If we do (2) however we need to somehow account for the fact that `SFGetAlternate` is assumed to occasionally refer to glyphs that don't exist, which `GetGoodAccentGlyph` is supposed to find matches for...:dizzy_face: 

So we *have* to do (1), right? How else will `SCCenterAccent` get unmapped characters?

Seeing how much has to change is why I question whether or not this is worth it. I don't put unmapped characters in my fonts, I either use the BMP's PUA for simple fonts or the huge PUA-A and PUA-B for TT2020 which has tons of auto-generated alternates.

skef:
Here's another way of looking at the question, which I want to stress I'm not saying is a better way *given the scope what you want to do*. 

You've decided, rightly I think, that some structure representing a series of "transformations" is called for. But in terms of the overall design making `SFGetAlternate` return that kind of structure is somewhat arbitrary and limiting. The "what to do" decisions in this layer are already spread out all over the place and putting more "what to do" decisions in `SFGetAlternate` spreads them out more. It would be cleaner to:

1. Create a structure that encodes *all* the supported transformations, including "Spaces" and "Rules" and "Rotations" and such.
2. When any subsystem needs to ask any question about character composition, build that whole structure and then write a function to answer that question based on the structure. 
3. Reorganize the rest of the code to build the glyph based on the specification. 

Probably the most awkward thing about leaving `SFGetAlternate` as the locus of decision-making is that on the main path it's called twice, once in `SFIsCompositBuildable` and then again directly in `SCBuildComposit`. If I had a hook then the hook would also need to be called twice. That's not "wrong", but it is evidence that putting more decisions into `SFGetAlternate` makes the whole layer more lopsided in terms of design.

Of course it would be a big pain to design a UI to support every kind of transformation FontForge supports (e.g. a UI feature to make a dotless version of a glyph). However, the UI question is separate from the architecture question, or at least it can be considered separately. Just because the specification has some feature doesn't mean FontForge has to expose it. A better architecture would make it easier to expose additional transformations in the future. 

So basically: Putting more logic in `SFGetAlternate` seems like it will work for your purposes and be less work, but has the result of making the layer more convoluted and difficult to modify for whoever may follow up. But there may never be someone who follows up, so that could be fine.

ctrlcctrlv:
I'm not confident enough to make a structure for everything.

Since only being able to add mapped characters was good enough the first time around&hellip;I'm just going to assume it remains so&hellip;

I'll add a linked list (or something) of `uint32` placement flags to `SplineChar` and tell `_SCCenterAccent` to look for it before placing.

That's what I can do.

ctrlcctrlv:
Just a quick update, sorry for using this as my blog.

I fixed a bug in gmatrixedit.c that will be part of the upcoming branch. It fixes other UI issues, not just mine.
I made it so even multiple combining classes worked, and I can convert back and forth between a `char*` describing a combining class to its `uint32` equivalent.

Screenshot_20190620_224120

I'm very happy with the UI and will move on to the backend stuff tomorrow!

ctrlcctrlv:
Someone just complained about this bug on the mailing list. https://sourceforge.net/p/fontforge/mailman/message/36712219/

I pointed them to this thread. https://sourceforge.net/p/fontforge/mailman/message/36712234/

* I'm not the only one with this problem.
* Less or non-technical users are being affected. User in question can't figure out how to apply my patch and I don't have time to help them step by step.
* I'm not the only one who'd like to see this merged. He writes &ldquo;It's essential to me, to finish my job on a font...&rdquo;

ctrlcctrlv:
This should be merged now IMO. My work on getting accent placement hints into user decompositions (for glyphs without anchors) is going to take a while, there are many things at play here, see https://github.com/fontforge/fontforge/pull/3755#issuecomment-507847817. Users need this now.

skef:
> * He writes “It's essential to me, to finish my job on a font...”

The `accents` map is only used at the time of construction. The easy work-around is to temporarily un-unicode-map the spacing accents, build, and remap them.

ctrlcctrlv:
That's a PITA, but I suppose you should write that on the ML. I'm certainly not going to suggest that :stuck_out_tongue:

frank-trampe:
We can't merge exactly as-is since we can neither confirm nor deny the problem George mentioned in his comment. We could make it work with the kludge that I proposed previously (having an application-wide preference to select between the "legacy" and "correct" modes). Tomorrow is a disaster for me, but I could write that Wednesday. That would essentially close the bug, with the remaining value in the general transformation patch all on the feature side.

ctrlcctrlv:
I think such an option is pointless, but if you're willing to write it, I'll grudgingly accept it if it leads to this being merged.

frank-trampe:
Writing something that is probably pointless hurts a lot more than accepting it. :smile:

frank-trampe:
There is an existing preference called "PreferSpacingAccents" with the following description. "Use spacing accents (Unicode: 02C0-02FF) rather than\ncombining accents (Unicode: 0300-036F) when\nbuilding accented glyphs." Could we just add better handling for that in `GetGoodAccentGlyph`?

ctrlcctrlv:
Yes. Good catch, I didn't know about that option.

frank-trampe:
This has been quite the adventure. I found in the course of due diligence that the original change would have broken `SCAppendAccent` and `CanonicalCombiner`, which depend upon the spacing accent being in the first position in order to convert to the generic accent number or the combining accent. The second of those functions seems to allow us to use the list both ways without retaining both copies of it. It scans the first cell of each row for a matching spacing accent and uses the row index to infer the combining accent (adding 0x300). I'm working on an experimental patch based upon these discoveries, but I do not expect to finish tonight.

frank-trampe:
I've thrown together some code in #3774. It is likely to require a little more work, and I plan to come back to it tomorrow. @ctrlcctrlv, do you have any tests handy?

frank-trampe:
Correction. The change before would not actually break `CanonicalCombiner`. It does scan the whole row.

ctrlcctrlv:
I tested this just by trying to build accented characters in fonts which include both spacing and non-spacing characters, e.g. `DejaVuSans.ttf`.

frank-trampe:
Hmm. It seems to work as-is. Could you test it a bit?

ctrlcctrlv:
Voluntarily closing in favor of #3774.

Issue #3566, opened 2019-03-22
Latest release in PPA does not install for Ubuntu Bionic


devosb:
Thank you for making a release in Mar 2019. That is much appreciated.

Unfortunately, this new release will not install from the FontForge PPA since the fontforge in Ubuntu (starting with Bionic) has a version number with an epoch (`1:`) which caused versions without an epoch to be seen as older.

In the attached screen shot, you can see the new release of FontForge version `20190317-0ubuntu1~bionic` says there is a newer release, which links to FontForge version `1:20170731~dfsg-1`.

The date in the version shows which is newer, but the epoch causes the date to be overridden.
![fontforge-ppa](https://user-images.githubusercontent.com/12450072/54857531-75322680-4cc5-11e9-8dd7-8bb168c45e0c.png)

ctrlcctrlv:
This is a packaging issue, it is not within our purview. Report the problem to the PPA maintainer, the Ubuntu packagers or the Ubuntu developers, depending on the process they advocate.

khaledhosny:
Does anyone knows who is the PPA maintainer? The account is simply named fontforge and no further details are provided.

khaledhosny:
FWIW, I found a different PPA that has a working Boinic package (and built with Python 3, as a bonus). https://launchpad.net/~silnrsi/+archive/ubuntu/smith-py3/

jtanx:
@frank-trampe maybe?

frank-trampe:
Sorry for the delay. There is an issue on Bionic and later that inexplicably breaks rather simple tests (no floating point) only on i386. See the test results near the end of [this log](https://launchpadlibrarian.net/425396904/buildlog_ubuntu-bionic-i386.fontforge_20190413-27-g1acfefa-0ubuntu1~bionic_BUILDING.txt.gz). I am not in a position to set up a test environment for over two weeks due to travel. Until this is fixed, our PPA support for those later versions will be unofficial, but you can try the current amd64 packages.

jtanx:
Got a vm running, it's segfaulting.

```
(gdb) bt
...
warning: Error reading shared library list entry at 0xffff8e80
warning: Error reading shared library list entry at 0xfffff730
Warning: Mac string is a subset of the Windows string in the 'name' table
 for the Sample Text string in the English (US) language.
The glyph named dotlessj is mapped to U+F6BE.
But its name indicates it should be mapped to U+0237.
Warning: Mac string is a subset of the Windows string in the 'name' table
[...]

PR #3715, opened 2019-06-01
scripting changes on passing arrays by reference

jtanx:
It's probably easiest to review this commit by commit.

I was contemplating a few approaches to fixing the issue mentioned in #3566, but they were just too involved to get right. The `arraynew` method I added isn't strictly required; it was a leftover from another approach, but I left it in because it reduced the amount of code that was floating around to initialise arrays.

I've opted for just adding a flag to the `Val` struct to replace the `dontfree` array. It serves a similar purpose, but means that dodgy pointer arithmetic isn't required.

The `v_arrfree` type could also probably be removed (so there's just one array type, `v_arr`) by adding a similar flag for it, but it's just too much effort to match the same semantics, so I'm leaving that alone for now.

As a side note, this has zero memory overhead on 64-bit operating systems, because the flag now occupies what was previously padding in the struct. I could have made it zero overhead for 32-bit too by making the type and flag 2 byte values, but ¯\\_(ツ)_/¯

### DIFF
--- a/fontforge/scripting.c
+++ b/fontforge/scripting.c
@@ -9661,8 +9661,8 @@ docall_expectint:   ScriptError(&sub,"Expected integer argument");
 		}
 		if ( sub.script==NULL )
 		    *pt = '\0';
+		sub.script = fopen(sub.filename,"rb");
 	    }
-	    sub.script = fopen(sub.filename,"rb");
 	    if ( sub.script==NULL ) {
 		ScriptErrorString(c, "No built-in function or script file", name);
 	    } else {
--- a/fontforge/scripting.c
+++ b/fontforge/scripting.c
@@ -10776,11 +10776,16 @@ _Noreturn void ProcessNativeScript(int argc, char *argv[], FILE *script) {
 			ff_backuptok(&c);
 			ff_statement(&c);
 		}
-		if ((c.script != stdin) && (c.script != script) && (c.script !=NULL)) fclose(c.script);
+		// calldatafree will close c.script; don't do that in these cases
+		if (c.script == stdin || c.script == script) c.script = NULL;
     }
-	// Free previously copied arguments.
-    for ( i=0; i<c.a.argc; ++i )
-		free(c.a.vals[i].u.sval);
+
+    calldatafree(&c);
+    // calldatafree doesn't free the first arg
+    if (c.a.vals[0].type == v_str)
+	free(c.a.vals[0].u.sval);
+    else if (c.a.vals[0].type == v_arr || c.a.vals[0].type == v_arrfree)
+	arrayfree(c.a.vals[0].u.aval);
     free(c.a.vals);
     free(c.dontfree);
     exit(0);
--- a/fontforge/scripting.c
+++ b/fontforge/scripting.c
@@ -253,6 +253,7 @@ static void DicaNewEntry(struct dictionary *dica,char *name,Val *val) {
     }
     dica->entries[dica->cnt].name = copy(name);
     dica->entries[dica->cnt].val.type = v_void;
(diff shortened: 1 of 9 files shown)


fontforge/scripting.c
```diff
@@ -6448,9 +6439,7 @@ static void bMMAxisBounds(Context *c) {
     axis = c->a.vals[1].u.ival;
 
     c->return_val.type = v_arrfree;
-    c->return_val.u.aval = malloc(sizeof(Array));
-    c->return_val.u.aval->argc = mm->axis_count;
```
jtanx: Fairly sure this was a copy pasta error

fontforge/splinefont.h
```diff
@@ -127,19 +127,22 @@ typedef struct ibounds {
 
 
 enum val_type { v_int, v_real, v_str, v_unicode, v_lval, v_arr, v_arrfree,
-		v_int32pt, v_int16pt, v_int8pt, v_void };
+		v_void };
+
+enum val_flags {
+    vf_none = 0,
+    vf_dontfree = (1 << 1)
```
frank-trampe: Is this interacting with another set of flags, or are we just reserving the low bit?

fontforge/splinefont.h
```diff
@@ -127,19 +127,22 @@ typedef struct ibounds {
...
+    enum val_flags flags;
     union {
 	int ival;
 	real fval;
 	char *sval;
 	struct val *lval;
 	struct array *aval;
-	uint32 *u32ptval;
-	uint16 *u16ptval;
-	uint8  *u8ptval;
```
frank-trampe: All unused?

fontforge/scripting.c
```diff
@@ -265,15 +266,17 @@ static void calldatafree(Context *c) {
     /* child may have freed some args itself by shifting,
        but argc will reflect the proper values none the less */
     for ( i=1; i<c->a.argc; ++i ) {
-        if ( c->a.vals[i].type == v_str ) {
-            free( c->a.vals[i].u.sval );
-            c->a.vals[i].u.sval = NULL;
-        }
-	if ( c->a.vals[i].type == v_arrfree || (c->a.vals[i].type == v_arr && c->dontfree[i]!=c->a.vals[i].u.aval )) {
+	if ( c->a.vals[i].flags & vf_dontfree )
+	    c->a.vals[i].flags ^= vf_dontfree;
```
frank-trampe: Why is it necessary to unset the flag?

frank-trampe: Ah, I see. It's the sort of thing that would normally fit a reference count pattern, but I guess that that would represent an unnecessarily large change from the old logic.

frank-trampe:
Using a constructor sure makes the code look a lot nicer.

fontforge/splinefont.h
jtanx: There's no interaction with other flags, so this is the only flag in this set at the moment.

I was thinking about adding another flag to get rid of the `vf_arrayfree` type, but I left that out for now, since that required more thought than I'm bothered investing in right now.

fontforge/splinefont.h
jtanx: Yep

fontforge/scripting.c
jtanx: Yeah, I actually tried refcounting first, but it's a bit of a headache to get right, because none of the current usages assume that, which makes it difficult exactly where to increment/decrement the counts.

frank-trampe:
I'm fine with this. @skef?

skef:
I've read through this and can't sign off on all of it with confidence because I haven't done the relevant research. But I can sign off on 90% of it. To the extent I understand the issues it looks right to me.

Issue #3470, opened 2019-02-12
Cursive Attachment(curs) feature don't works precisely in the metric view


AlirezaAlipour:
I need cursive attachment for a cursive script that I'm designing. In order to achieving this feature for my font I have created a GPOS table with Cursive Position type and Cursive Attachment(curs) feature. Cursive Attachment(curs) feature after generation works correctly and each glyph attach to each other precisely, also in the metric view Cursive Attachment(curs) feature works, but don't work precisely.

ctrlcctrlv:
Us having screenshots and SFD files are essential to helping you.

AlirezaAlipour:
Thank you very much for your help

[Cursive_Feature_test.zip](https://github.com/fontforge/fontforge/files/2855605/Cursive_Feature_test.zip)

ctrlcctrlv:
The `curs` feature isn't selected. Hold down Shift and choose `curs` from the pane on the left in the Metrics view to add it to the enabled OT features.

AlirezaAlipour:
> The `curs` feature isn't selected. Hold down Shift and choose `curs` from the pane on the left in the Metrics view to add it to the enabled OT features.

Thank you very for your assistance

If you look at left sidebar, there is Required Feature(RQD) feature, I added "RQD" feature in addition to Cursive Attachment(curs) feature to my lookup table because "curs" feature is not selected by default and with using "RQD" feature, it make my lookup table be applied, I also can select it using the Contextual Chaining Position.

ctrlcctrlv:
Could it be that FontForge only shows the `curs` feature in Metrics for Urdu? I see Urdu mentioned in the manual. Or is it never showed? Do you know of a font that for sure has `curs` working in applications? Try opening it in FontForge.

AlirezaAlipour:
> Could it be that FontForge only shows the `curs` feature in Metrics for Urdu? I see Urdu mentioned in the manual. Or is it never showed? Do you know of a font that for sure has `curs` working in applications? Try opening it in FontForge.

It works, but not precisely.

ctrlcctrlv:
I know this doesn't solve the specific possible issue in FontForge in general, but from my understanding, you don't even need `curs` for Western-style cursive; `kern` and `calt` are sufficient. What makes `curs` special is that it's similar to `mark`, you can move the y positions of letters as well as the x.

ctrlcctrlv:
>It works, but not precisely.

[...]

PR #3721, opened 2019-06-05
Fix metrics view update bug

### LINKED ISSUE #3720: Metrics window is not repainted when I type a new kerning value
When I type a new kerning value in the Metrics window (see the circled text field in my screen shot), the glyphs don't move. The kerning data is changed but the problem is only in the preview that is not repainted. If I type a new side-bearing value, there is no problem with repainting.

- [x] Screenshot
- [x] The FontForge version and the operating system you're using: FontForge-2019-04-13. Windows 7 and Windows 10.

FontForge_Metrics_bug

ctrlcctrlv:
This closes #3720.

This copies the MVRemetric function, already used in MVTextChanged, to
MV_KernChanged. I tested it with my Chomsky font and it works as
expected, no more having to change the text to see the new kern.

### Type of change

- **Bug fix**

### DIFF
--- a/fontforgeexe/metricsview.c
+++ b/fontforgeexe/metricsview.c
@@ -1548,6 +1548,7 @@ return( true );
 	    GDrawBeep(NULL);
 	else {
 	    MV_ChangeKerning(mv,which,val, false);
+	    MVRemetric(mv);
 	}
     } else if ( e->u.control.subtype == et_textfocuschanged &&
 	    e->u.control.u.tf_focus.gained_focus ) {


skef:
Eeesh - I recognize this code from #3470. 

Would it be possible to test this with `MVRefreshMetric(mv)` in place of where you've added `MVRemetric(mv)`? 

If that happens to work just as well, I would put an `int kerning_changed = 0;` at the top, and set it to 1 where you have `MVRemetric(mv)` now, and then make the guard below be:

```
if( haveClassBasedKerningInView(mv) || kerning_changed )
```
That way `MVRefreshMetric()` won't be called twice.

ctrlcctrlv:
@skef Sorry, that doesn't work. `MVRemetric` works, `MVRefreshMetric` does not.

ctrlcctrlv:
I honestly feel like it doesn't matter that MVRemetric is potentially called twice in some cases, AFAICT there's no performance penalty and it solves an annoying bug so meh

PR #3727, opened 2019-06-05
Always write a layercontents.plist when outputting UFO

### LINKED ISSUE #3726: missing "layercontents.plist" in export as UFO format
l'm a fontforge user for about 4 years and now im designing a variable font from two masters. I'm designing both masters inside fontforge and export them as UFO format. but the program i using ,fontmake, requires "layercontents.plist" file to generate variable font. i opened an issue for fontmake and explain the problem, they said this file is nessesary to fontmake, even if font have only one layer.

skef:
According to http://unifiedfontobject.org/versions/ufo3/layercontents.plist/ the `layercontents.plist` file is mandatory. This change just removes the `if (all_layers)` branch and changes the `for` loop over layers from 

```for (layer_pos = 0; layer_pos < sf->layer_cnt; layer_pos++) {```

to 

```for (layer_pos = (all_layers ? 0 : ly_fore); layer_pos < (all_layers ? sf->layer_cnt : ly_fore+1); layer_pos++) {```

I've tested with a single-layer font (the usual case) and the resulting output looks about right. @frank-trampe feel free to edit the branch as you see fit. 

Closes #3726 

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/ufo.c
+++ b/fontforge/ufo.c
@@ -2057,73 +2057,68 @@ int WriteUFOFontFlex(const char *basedir, SplineFont *sf, enum fontformat ff, in
     }
     glif_name_index_destroy(glif_name_hash); // Close the hash table.
     
-    if (all_layers) {
-      struct glif_name_index * layer_name_hash = glif_name_index_new(); // Open the hash table.
-      struct glif_name_index * layer_path_hash = glif_name_index_new(); // Open the hash table.
+    struct glif_name_index * layer_name_hash = glif_name_index_new(); // Open the hash table.
+    struct glif_name_index * layer_path_hash = glif_name_index_new(); // Open the hash table.
 
-      switch_to_old_locale(&tmplocale, &oldlocale); // Switch to the cached locale.
-      xmlDocPtr plistdoc = PlistInit(); if (plistdoc == NULL) return false; // Make the document.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by frank-trampe** [comment]
This reveals one remnant problem from before we differentiated UFO 2 from UFO 3, which is that true UFO 2 (rather than the hybrid that was standard before) does not have said file. We probably ought to fix that too, but let me think about it.

**Comment by skef** [comment]
I suppose there's a risk of a manifest being used to guess the version, and FontForge having saved the other files conservatively as UFO2.

Shall I do this?
```
--- a/fontforge/ufo.c
+++ b/fontforge/ufo.c
@@ -2113,7 +2113,8 @@ int WriteUFOFontFlex(const char *basedir, SplineFont *sf, enum fontformat ff, in
       free(glyphdir); glyphdir = NULL;
     }
     char *fname = buildname(basedir, "layercontents.plist"); // Build the file name for the contents.
-    xmlSaveFormatFileEnc(fname, plistdoc, "UTF-8", 1); // Store the document.
+    if (version >= 3)
+      xmlSaveFormatFileEnc(fname, plistdoc, "UTF-8", 1); // Store the document.
```
Easy enough ...

frank-trampe:
Yes. Let's do that.

frank-trampe:
Otherwise, it looks fine.

skef:
> Yes. Let's do that.

Done

PR #3728, opened 2019-06-08
Don’t generate MATH after deleting math data

khaledhosny:
After removing all `MATH` related data from Python, doing e.g.:

    font.math.clear()
    for glyph in font.glyphs():
        glyph.topaccent = 0x7fff
        glyph.italicCorrection = 0x7fff
        glyph.horizontalComponentItalicCorrection = 0x7fff
        glyph.verticalComponentItalicCorrection = 0x7fff
        glyph.isExtendedShape = False
        glyph.horizontalComponents = None
        glyph.horizontalVariants = None
        glyph.verticalComponents = None
        glyph.verticalVariants = None
        glyph.mathKern.bottomLeft = None
        glyph.mathKern.bottomRight = None
        glyph.mathKern.topLeft = None
        glyph.mathKern.topRight = None

FontForge would still generate a `MATH` table with empty glyph constructions. This seems to be caused by the checks in `MathBits()` that only checks if the constructions are `NULL`, but the Python code does not set them to `NULL` (as they might have other set members). Fix this by doing a deeper check for emptiness.

### Type of change
- **Bug fix**

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -7669,7 +7669,7 @@ return( 0 );
 }
 
 static PyObject *PyFF_Glyph_get_horizontalComponents(PyFF_Glyph *self, void *UNUSED(closure)) {
-    if ( self->sc->horiz_variants==0 || self->sc->horiz_variants->part_cnt==0 )
+    if ( self->sc->horiz_variants==NULL || self->sc->horiz_variants->part_cnt==0 )
 Py_RETURN_NONE;
 
 return( BuildComponentTuple(self->sc->horiz_variants ));
@@ -7697,7 +7697,7 @@ return( 0 );
 }
 
 static PyObject *PyFF_Glyph_get_verticalComponents(PyFF_Glyph *self, void *UNUSED(closure)) {
-    if ( self->sc->vert_variants==0 || self->sc->vert_variants->part_cnt==0 )
+    if ( self->sc->vert_variants==NULL || self->sc->vert_variants->part_cnt==0 )
 Py_RETURN_NONE;
 
 return( BuildComponentTuple(self->sc->vert_variants ));
--- a/fontforge/tottfgpos.c
+++ b/fontforge/tottfgpos.c
@@ -3603,6 +3603,14 @@ enum math_bits { mb_constants=0x01, mb_italic=0x02, mb_topaccent=0x04,
 	mb_gi=(mb_italic|mb_topaccent|mb_extended|mb_mathkern),
 	mb_gv=(mb_vertvariant|mb_horizvariant) };
 
+static int HasVariants(struct glyphvariants *gv) {
+    if ( gv!=NULL && (gv->variants!=NULL || gv->parts!=NULL ||
+	    gv->italic_correction!=0) )
+return( true );
+
+return( false );
+}
+
 static int MathBits(struct alltabs *at, SplineFont *sf) {
     int i, gid, ret;
     SplineChar *sc;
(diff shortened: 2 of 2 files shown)


ctrlcctrlv:
Is Sorts Mill still being maintained?

frank-trampe:
I'll need to test (probably early next week), but this might introduce round-trip problems since it essentially deviates output from internal state. Are you sure it wouldn't be easier just to have a Python function clearMath or something like that that affects the internal state?

skef:
> but this might introduce round-trip problems since it essentially deviates output from internal state.

SFDDumpGlyphVariants looks like this:
```
static void SFDDumpGlyphVariants(FILE *sfd,struct glyphvariants *gv,const char *name) {
    int i;

    if ( gv==NULL )
return;

    if ( gv->variants!=NULL )
        fprintf( sfd, "GlyphVariants%s: %s\n", name, gv->variants );
    if ( gv->part_cnt!=0 ) {
        if ( gv->italic_correction!=0 ) {
            fprintf( sfd, "GlyphComposition%sIC: %d", name, gv->italic_correction );
            if ( gv->italic_adjusts!=NULL ) {
                putc(' ',sfd);
                SFDDumpDeviceTable(sfd,gv->italic_adjusts);
            }
            putc('\n',sfd);
        }
        fprintf( sfd, "GlyphComposition%s: %d ", name, gv->part_cnt );
        for ( i=0; i<gv->part_cnt; ++i ) {
            fprintf( sfd, " %s%%%d,%d,%d,%d", gv->parts[i].component,
                    gv->parts[i].is_extender,
                    gv->parts[i].startConnectorLength,
                    gv->parts[i].endConnectorLength,
                    gv->parts[i].fullAdvance);
        }
        putc('\n',sfd);
    }
}
```
So except for a potential worry about parts and part_cnt getting out of sync, seems like any round-trip problem with this update would also be a round-trip problem through saving and loading. (Technically speaking there may be an SFD problem, in that `italic_correction` is inside the `part_cnt` test. But this push shouldn't affect that either way.)

khaledhosny:
>  If this is to be the chosen test, however, it seems that the first two loops in `ttf_math_dump_glyphvariants()` (below in the same file) should also be modified.

Makes sense, done.

khaledhosny:
> Is Sorts Mill still being maintained?

AFAIK no.

skef:
Looks good to me. I verbally approve but am leaving the status off for @frank-trampe's further investigation.

khaledhosny:
> I'll need to test (probably early next week), but this might introduce round-trip problems since it essentially deviates output from internal state.

Does it matter? The fonts will be functionally identical. And when saving to SFD, FontForge will discard the empty variants and the generated font will be identical with or without this change.

> Are you sure it wouldn't be easier just to have a Python function clearMath or something like that that affects the internal state?

That would be handy, but it wouldn’t fix the current issue.

ctrlcctrlv:
If the fonts are functionally identical, why is this change even needed

I mean it looks fine to me I guess, but you have to admit that a NULL comparison is better than a function like `HasVariants`, it's much more understandable and this issue is subtle so future authors are going to need to remember to use it

And maybe that's okay, but it seems like it would be better to just let Python set math NULL as @frank-trampe suggested

ctrlcctrlv:
why didn't you just have Python set the variants NULL when `font.math.clear()` is called?

skef:
> why didn't you just have Python set the variants NULL when `font.math.clear()` is called?

It's not just python -- you'd have to add a check at every instance a value is cleared in python or the GUI and get rid of the structure if it's the last one that was set. Another common and perfectly valid approach, but a bit more work and easier to miss a case now or in the future.

khaledhosny:
OK, no need for this PR then.

PR #3729, opened 2019-06-09
Header/misc cleanup

jtanx:
This started out as trying to figure out what to do with the way headers are laid out, and ended in being a yak shaving exercise to clean up how headers are declared and imported.

It's not perfect, and it's not finished, but the diff is getting large, so I'll leave it here for now. I also haven't addressed what I set out for (specifically, what headers get installed doesn't match the in-source layout, plus the fontforge/fontforgeexe split for headers is pretty awkward). But again, leaving that for now.

### Type of change

- **Non-breaking change**

### Main changes
- Removed unused headers
- Removed strlist and its usage from python.c - replaced with just using GPtrArray (let's stop reimplementing data structures...)
- Various header normalisations
- Ensure that all headers have include guards
   - Normalise the name of all header include guards
   - System includes use angle brackets, local includes use double quotes
   - Ensure all sources include fontforge-config.h first
   - Remove some redundant imports
   - Sort includes; local before system in most cases

This was done with a combination of ad-hoc python scripting and manual edits.

Related issues: #3512


gutils/unicodelibinfo.c
```diff
@@ -26,15 +26,11 @@
  * ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#ifndef _GUTILS_UNICODELIBINFO_C_
-#define _GUTILS_UNICODELIBINFO_C_
```
frank-trampe: Any idea why this was here in the first place?

inc/dynamic.h
```diff
@@ -1,55 +0,0 @@
-/* Vague support for several different shared library formats */
-/* Everyone but the mac uses a varient of dlopen, while the mac uses NSAddImage */
-/* Under cygwin the shared libs have very strange names which bare little */
-/*  resemblance to what we expect. GNU creates a .la file that tells us what */
-/*  the name should be, so indirect through that */
```
frank-trampe: When/how did this fall out of use?

jtanx: A combination of getting rid of runtime dynamic linking (with dlopen) and getting rid of plugins

gutils/unicodelibinfo.c
jtanx: 3aaa48c1c230840045bbae158093edef53d762d3 is the context, it looks erroneously added to me.

frank-trampe:
I guess we can do something like this now that we don't have active long-term forks. And it's fine for you to pick up all this "blame" since you understand the code in question better than the original authors. Having rational headers would certainly be a big improvement.

fontforge/views.h
```diff
@@ -1330,16 +1330,6 @@ extern void _CVMenuNameContour(CharView *cv);
...
-extern CharView* CharViewFindActive();
-extern FontViewBase* FontViewFindActive();
-extern FontViewBase* FontViewFind( int (*testFunc)( FontViewBase*, void* ), void* udata );
-
-extern int FontViewFind_byXUID(      FontViewBase* fv, void* udata );
-extern int FontViewFind_byXUIDConnected( FontViewBase* fv, void* udata );
-extern int FontViewFind_byCollabPtr(  FontViewBase* fv, void* udata );
-extern int FontViewFind_bySplineFont( FontViewBase* fv, void* udata );
-extern int FontViewFind_byCollabBasePort( FontViewBase* fv, void* udata );
-
```
frank-trampe: Are you sure that these functions (other than the collab one) are worthless (unlikely to be used)?

frank-trampe:
I did not look at the header changes line-by-line, by the way, but that is the sort of thing that is validated by successful compilation, so I'm not terribly concerned.

fontforge/views.h
jtanx: Yes, they were all introduced with collab support, but this was missed when removing it.

frank-trampe:
I'm happy with this. Anybody else? @skef, @JoesCat?

fontforge/nowakowskittfinstr.c
```diff
@@ -26,20 +26,23 @@
...
-#include <math.h>
-#include <utype.h>
-
-#include "ttf.h"
 #include "splinefont.h"
-#include "stemdb.h"
 #include "splineutil.h"
 #include "splineutil2.h"
 #include "stemdb.h"
+#include "stemdb.h"
```
skef: Duplicate before but not clear it's needed.

fontforge/python.c
```diff
@@ -19347,34 +19342,37 @@ extern void PyFF_FreePythonPersistent(void *python_persistent) {
...
     struct dirent *ent;
-    struct string_list *filelist=NULL;
-    struct string_list *item;
+    GPtrArray *filelist;
 
     diro = opendir(dir);
     if ( diro==NULL )		/* It's ok not to have any python init scripts */
 return;
 
+    filelist = g_ptr_array_new_with_free_func(free);
```
skef: I would recommend avoiding squashing these substantial python changes at merge. Keeping the commits separate would presumably be fine.

fontforge/featurefile.c
```diff
@@ -26,37 +26,36 @@
...
+#include "ttf.h"
+#include "ustring.h"
+#include "utype.h"
+
+#include <locale.h>
 #include <math.h>
-#include <stdlib.h>
+#include <stddef.h>
+
 #ifdef __need_size_t
```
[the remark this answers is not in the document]
jtanx: Good spot. Given it still compiles on Mac, I'm tempted to just remove that block, since it's clearly finding `offsetof` ok.

fontforge/python.c
jtanx: Yeah, usually I squash anything I want to squash before merging, and merge PRs as-is. I'm leaving the removal commits separate to the larger refactor commit.

fontforge/nowakowskittfinstr.c
jtanx: Yeah, so I sorted the includes but didn't search for duplicates. I've reran, searching for duplicates, and removed a couple more, in addition to this instance

Issue #2648, opened 2016-03-20
Dragging certain elements rapidly fills Undoes


mrkline:
Dragging some elements in the outline window (most notably the left bearing, though I believe I've seen this behavior in a few other cases) rapidly fills up the undoes list - it seems as though an undo "step" is created each time the window renders. Something similar happens when moving points using the keyboard - an undo event is saved for every unit the points move.

**Desired behavior:** An undo event is saved only once the mouse button is released, or the keyboard key is let up.

jtanx:
I have a feeling this is tied to the collab feature...

frank-trampe:
I have the same suspicion, although there's too much other stuff going on for me to check.

janpe2:
> I have a feeling this is tied to the collab feature...

Pull request #3655 has now removed the collab feature. Did it fix this issue?

ctrlcctrlv:
Not for me, I also have this issue

jtanx:
What other actions have you noticed this on? I've got a fix incoming for changing the lbearing

ctrlcctrlv:
@jtanx I've played around with the [`lbearing-undoes`](https://github.com/fontforge/fontforge/tree/lbearing-undoes) branch and haven't found any other things that cause the undoes to be rapidly filled up; great job as usual :+1:

PR #3731, opened 2019-06-12
Create only one undo when adjusting the lbearing

jtanx:
Adjusting the lbearing via dragging it is a little bit special, because we're reusing just calling into FVTrans, which sets up the undo state itself. 

When the mouse is dragged in the charview, CVMouseMovePointer already [preserves an undo state](https://github.com/fontforge/fontforge/blob/master/fontforgeexe/cvpointer.c#L1489-L1490), so ideally there's no need for the state preservation that FVTrans adds.

But when FVTrans is called, we make it adjust all the layers, and it also moves all the hints. CVMouseMovePointer  only preserves the active layer, and doesn't preserve the hint state.

So what I do is just remove the undo state added by CVMouseMovePointer, and get FVTrans to set up the state itself - but only once. Not the prettiest...

### Type of change

- **Bug fix**

Related issues: #2648


ctrlcctrlv:
I've tested this and it works for me, solves #2648 :smiley_cat:

frank-trampe:
I think it's good. This is the sort of thing that can only really be validated by use, though.

PR #3732, opened 2019-06-12
Fix exporting/applying kerning with RTL flag

khaledhosny:
FontForge is incorrectly interpreting RightToLeft lookup flag to have some effect on pair positioning lookups (kerning), however the spec is clear that it only affects cursive positioning:

From [OpenType spec][1]
> This bit relates only to the correct processing of the cursive attachment lookup type (GPOS lookup type 3). When this bit is set, the last glyph in a given sequence to which the cursive attachment lookup is applied, will be positioned on the baseline.
> Note: Setting of this bit is not intended to be used by operating systems or applications to determine text direction.

The [feature file specification][2] does not have any text that would suggest this flag affects kerning and both makeotf and feaLib don’t interpret it in any special way.

With the current code FontForge is incorrectly exporting and applying pair positioning lookups that has this flag set (setting this flag on non-cursive positioning lookups is pointless but should be harmless as
well).

Proper right to left kerning has to be set up by the user manually since OpenType model requires different handling for RTL and LTR kerning, but FontForge already provides enough flexibility to set the right pair positioning values, and there is no need for the current incorrect magic.

[1]: https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#lookupTbl:
[2]: https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#4.d

### Type of change
- **Bug fix**
- **Breaking change**

### DIFF
--- a/fontforge/featurefile.c
+++ b/fontforge/featurefile.c
@@ -509,9 +509,6 @@ return( NULL );
 		/* !!! Bug. Lose device tables here */
 		if ( isv )
 		    space->u.pair.vr[0].v_adv_off = kp->off;
-		else if ( otl->lookup_flags&pst_r2l )
-		    space->u.pair.vr[1].h_adv_off = kp->off;
-		else
 		    space->u.pair.vr[0].h_adv_off = kp->off;
 return( space );
 	    }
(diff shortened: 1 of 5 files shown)


fontforge/featurefile.c
```diff
@@ -509,9 +509,6 @@ return( NULL );
 		/* !!! Bug. Lose device tables here */
 		if ( isv )
 		    space->u.pair.vr[0].v_adv_off = kp->off;
```
skef: There's probably an `else` missing here, and if not the indentation should be adjusted.

khaledhosny: Right, deleted one line too many.

khaledhosny:
> In your judgment does this call for a new issue to address other uses of `pst_r2l`, such as those in `lookups.c`?

I didn't check other places where this flag is checked, but most likely they are wrong as well. RTL kerning right now is unusable in FontForge, but this change only fixes handling of it in feature files. The GUI, the import from/export to OpenType GPOS table, and possibly other places are still broken (with or without RightToLeft flag being set, as FontForge will do very "helpful" things as checking whether a character is a RTL one and change the glyph adjustments silently).

PR #3736, opened 2019-06-17
Remove OpenType Feature, Script, and Lookup limits (issue #3734)

### LINKED ISSUE #3734: No GSUB information available for a font with GSUB
The GSUB tab of “Font Information” for [NotoSerifTibetan-Regular.ttf](https://github.com/googlefonts/noto-fonts/blob/2a41389346abdb6b83f5e0c4efe598bcb8abbb42/phaseIII_only/unhinted/ttf/NotoSerifTibetan/NotoSerifTibetan-Regular.ttf) is empty, but that font has a GSUB table. I am using FontForge 20190413 on macOS Mojave 10.14.5.

skef:
This quick fix removes the stated arbitrary limits from `parsettfatt.c`. This could increase the chance of memory problems and such when opening fonts, although the chance of hitting a well-enough-formed encoding of such a large size seems remote. 

I didn't touch the limits on Apple-specific tables because I know nothing about those. (The 10000 limit on `mort` and `morx` is a bit mysterious given when the error is printed.)

Closes #3734

### DIFF
--- a/fontforge/parsettfatt.c
+++ b/fontforge/parsettfatt.c
@@ -46,6 +46,11 @@
 #include <math.h>
 #include <stdlib.h>
 
+// Disable maximum for these
+#define FF_TTF_SCRIPT_MAX -1
+#define FF_TTF_FEATURE_MAX -1
+#define FF_TTF_LOOKUP_MAX -1
+
 static uint16 *getAppleClassTable(FILE *ttf, int classdef_offset, int cnt, int sub, int div, struct ttfinfo *info) {
     uint16 *class = calloc(cnt,sizeof(uint16));
     int first, i, n;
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
This might be a bit controversial:

In 2019, Apple Advanced Typography is completely irrelevant. If support for it disappeared in FontForge it would be no worse for wear. So us patching it is pointless.

skef:
Whether or not it's *completely* irrelevant, it's relative irrelevance probably excuses any lack of attention to those particular limits before someone files an issue.

frank-trampe:
This seems fine to me.

frank-trampe:
@ctrlcctrlv, I'm on the same page about the utility of AAT. Even when it had unique features, their platform specificity limited their use to a few demonstration cases. At the same time, if we were to remove support (and only if we were to remove support), it would turn out that the one person whose hobby is making those demos uses FontForge.

skef:
I just read over my description and realized it was vague. Presumably these limits were to protect against corrupted fonts while allowing the non-corrupted parts to be opened and saved in FontForge. My "argument" is that if the length field gets corrupted some other part of the encoding will be be wrong enough for FontForge to abandon the table anyway, especially at record counts over 1000.

ctrlcctrlv:
It seems strange to prioritize possibly corrupted fonts over a Noto font. If it passes the test suite, let's wait for someone to complain about a font that opened before but now doesn't IMHO.

PR #3738, opened 2019-06-20
Replace tzset() with equivalent glib support (issue #10)

### LINKED ISSUE #10: timezone use not portable

skef:
and fix print.c randoms[] warnings.

Closes #10

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -511,7 +511,7 @@ AC_CHECK_FUNCS([mkdir munmap])
 AC_CHECK_FUNCS([nl_langinfo putenv rmdir])
 AC_CHECK_FUNCS([select setenv socket])
 AC_CHECK_FUNCS([tzset uname])
-AC_CHECK_FUNC([tzset],[:],[AC_DEFINE([_NO_TZSET],1,[Define if not using tzset.])])
+#AC_CHECK_FUNC([tzset],[:],[AC_DEFINE([_NO_TZSET],1,[Define if not using tzset.])])
 
 #--------------------------------------------------------------------------
 # Put ifndef wrapper on fontforge-config.h so we don't call it repeatedly.
--- a/fontforge/cvexport.c
+++ b/fontforge/cvexport.c
@@ -165,7 +165,8 @@ int _ExportPDF(FILE *pdf,SplineChar *sc,int layer) {
 /* TODO: Note, maybe this routine can be combined with print.c dump_pdfprologue() */
     DBounds b;
     time_t now;
-    struct tm *tm;
+    GDateTime *gdt;
+    GTimeSpan zoffset;
     int ret;
     char oldloc[24];
     int _objlocs[8], xrefloc, streamstart, streamlength, resid = 0, nextobj;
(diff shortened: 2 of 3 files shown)


configure.ac
```diff
@@ -511,7 +511,7 @@ AC_CHECK_FUNCS([mkdir munmap])
 AC_CHECK_FUNCS([nl_langinfo putenv rmdir])
 AC_CHECK_FUNCS([select setenv socket])
 AC_CHECK_FUNCS([tzset uname])
-AC_CHECK_FUNC([tzset],[:],[AC_DEFINE([_NO_TZSET],1,[Define if not using tzset.])])
+#AC_CHECK_FUNC([tzset],[:],[AC_DEFINE([_NO_TZSET],1,[Define if not using tzset.])])
```
ctrlcctrlv: Why comment this out instead of deleting it?

skef: I guess fear? I figured I would wait for someone to tell me to.

ctrlcctrlv: I've edited our autotools stuff before and let me tell you, stuff like this makes it harder to work with, as I don't know if that's something possibly useful that I could use as an example, or just cruft from 2008. Please just remove it, let the git log be the record, not comments; we're enlightened.

skef: Done

fontforge/print.c
```diff
@@ -2772,7 +2772,7 @@ unichar_t *PrtBuildDef( SplineFont *sf, void *tf,
 	rcnt = 0;
 	for ( s=0; s<therecnt; ++s ) if ( !ScriptInList(scriptsthere[s],scriptsdone,scnt)) {
 	    if ( ret ) {
-		if ( randoms[rcnt]!='\0' ) {
+		if ( randoms[rcnt]!=NULL ) {
```
jtanx: that warning is finally fixed!

fontforge/cvexport.c
```diff
@@ -228,27 +229,24 @@ int _ExportPDF(FILE *pdf,SplineChar *sc,int layer) {
     fprintf( pdf, "    /Creator (FontForge)\n" );
     now = GetTime();
     if (!getenv("SOURCE_DATE_EPOCH")) {
-	tm = localtime(&now);
+	gdt = g_date_time_new_from_unix_local((gint64)now);
```
jtanx: This is missing a `g_date_time_unref` somewhere

skef: Eeesh -- And I think of myself as a "programmer". Fixed.

PR #3739, opened 2019-06-20
Fix compilation with `--enable-write-pfm'.

vvp2014:
Correct the order of effective arguments in the call to WritePfmFile()
in WritePfm() according to the prototype that is declared in
fontforge/savefont.h

### DIFF
--- a/gdraw/gdraw.c
+++ b/gdraw/gdraw.c
@@ -36,6 +36,9 @@
 #  include <sys/select.h>
 #endif
 
+#if __FreeBSD__ || __NetBSD__ || __OpenBSD__ || __DragonFly__
+#  include <sys/select.h>
+#endif
 
 /* Functions for font metrics:
     rectangle of text (left side bearing of first char, right of last char)
--- a/fontforge/scripting.c
+++ b/fontforge/scripting.c
@@ -2397,7 +2397,7 @@ static void bWritePfm(Context *c) {
 
     t = script2utf8_copy(c->a.vals[1].u.sval);
     locfilename = utf82def_copy(t);
-    if ( !WritePfmFile(c->a.vals[1].u.sval,sf,0,c->curfv->map) )
+    if ( !WritePfmFile(c->a.vals[1].u.sval,sf,c->curfv->map,0) )
 	ScriptError(c,"Save failed");
     free(locfilename); free(t);
 }


gdraw/gdraw.c
```diff
@@ -36,6 +36,9 @@
 #  include <sys/select.h>
 #endif
 
+#if __FreeBSD__ || __NetBSD__ || __OpenBSD__ || __DragonFly__
+#  include <sys/select.h>
```
frank-trampe: It seems that it would make sense to combine this with the `__Mac` test given that that part of the Macintosh software is also BSD and that the resulting action is the same.

vvp2014: I put BSDs apart on purpose.  While MacOS does mimic BSDs here and there, there are more differences than similarities between them.  For instance, display rendering APIs are completely different, etc.  Such grouping prevents complex splits of preprocessor definitions in the future.

vvp2014: BTW, I have no idea why Github merged these two successive commits into the same pull request, this was not my intent.

frank-trampe: @vvp2014, if you push to the same branch, it ends up in the pull request associated with that branch (as long as the pull request is closed).

frank-trampe: Do you expect further divergent actions within these conditionals? Unless there is a likely imminent change in that direction, this provisioning seems excessive, particularly since it looks so much like an oversight. If we merge it like this, we'll probably get another pull request in the next year merging the two statements, and I'll be the one on the hook to defend this design decision.

vvp2014: Well, in my experience grouping platforms according to similarity is justified.  I am not going to argue about that.  At any rate this fix is the only one that is required for building a complete FF on the current OpenBSD.  Please proceed as you please.

vvp2014: Thanks for the branching tip!

frank-trampe: @jtanx, this bothers me but not very much, and I'm inclined to merge in the form proposed by @vvp2014. What do you think?

jtanx: No, just combine the checks into the same ifdef above, *if* at some point they start to differ then we'll deal with that then.

frank-trampe: @vvp2014, we cannot modify your branch. Would you consolidate the changes so that we can get this merged?

vvp2014: All right then, I am going to close this and submit two separate requests, as I intended.

PR #3741, opened 2019-06-20
Update CocoaDialog

MaddTheSane:
The CocoaDialog currently in the source tree is only 32-bit. This update fixes it.

Alternatively, we could set it to build from https://github.com/cocoadialog/cocoadialog, but this would require more changes to the make system.


jtanx:
Not too keen on adding a ~2mb binary blob into git history. 

Do you actually use it directly? Because I was actually planning on getting rid of it - it's [not even bundled](https://github.com/fontforge/fontforge/blob/cb6a3730128f648c2da4f8a88780e2100ba6ed07/travis-scripts/ffosxbuild.sh#L43-L45) in the current bundles that are made, since none of the uses in the repo actually apply to the bundle that gets made today.

MaddTheSane:
I don't blame you not wanting to add the big binary. More recent versions seem to have been hit by feature creep.

I have found an older version that doesn't use any external libraries. If it is not even needed, removal should be okay.
IIRC, it's only called once: urging the user to move the app bundle to the Applications directory.

jtanx:
Superseded by #3767, CocoaDialog has been removed.

Issue #3694, opened 2019-05-09
⟦Expand Stroke⟧ Stroking certain path results in weird degenerate spline


ctrlcctrlv:
# ☞ See below for test case.

![2019-05-09-160009_2560x1401_scrot](https://user-images.githubusercontent.com/838783/57437002-6e5e7380-7273-11e9-9c14-15db74f41e00.png)
![2019-05-09-160619_2204x1895_scrot](https://user-images.githubusercontent.com/838783/57437395-47ed0800-7274-11e9-8854-1d027bf0c897.png)

## stdout
```
Internal Error (overlap): The curve is too short.
Internal Error (overlap): The curve is too short.
Internal Error (overlap): monotonic is both needed and unneeded (451.785,167.938)->(361.457,140.992). y=140.992 (prev=323.091)
Internal Error (overlap): Winding number did not return to 0 when y=140.992
Internal Error (overlap): monotonic is both needed and unneeded (451.785,167.938)->(361.457,140.992). y=140.992 (prev=323.091)
Internal Error (overlap): Winding number did not return to 0 when y=140.992
Internal Error (overlap): monotonic is both needed and unneeded (451.785,167.938)->(361.457,140.992). y=140.992 (prev=323.091)
Internal Error (overlap): Winding number did not return to 0 when y=140.992
```

## Background
I've found another Expand Stroke bug. As an avid fan of Spiro and Expand Stroke, I of course constantly suffer from its bugginess while working on my fonts, and recently found a workaround to its most annoying bug in https://github.com/fontforge/fontforge/issues/3264#issuecomment-489408743, but this most recent bug seems to me something totally new and undiscovered.

# Glyph `qof` (ק)

Here is the spline that causes it:

<img src="https://user-images.githubusercontent.com/838783/57437328-1f650e00-7274-11e9-9d65-fef1f5f329b4.png" width="200px">

When simplified (_Element&rarr;Simplify&rarr;Simplify_), it no longer happens:

<img src="https://user-images.githubusercontent.com/838783/57437768-0ad54580-7275-11e9-8229-92c6ee5f7608.png" width="200px">

I'm not claiming that this ק is exceedingly beautiful, but certainly it's not so ugly that it should cause FontForge to behave this way, right? :wink:

Well, the spiral seems to be the main problem. I found other weird things:

# Glyph `poof`

If I remove a lot of the points and just leave the little Spiro, when I try to expand it it disappears (glyph `poof`):

[...]

Issue #3264, opened 2018-03-10
FontForge-2017-07-31-Windows-r2 “Expand Stroke”: Internal Error: Some fragments did not join in uni071B


ilyaza:
Instructions: load the attached file, do <kbd>Ctrl</kbd>-<kbd>A</kbd>, <kbd>Ctrl</kbd>-<kbd>Shift</kbd>-<kbd>E</kbd>.  Set `64.3×64.3,   Line Cap=Round,  Remove internal contour`
and start the operation.

All the glyphs in the attached `.sfd` file show the error above.  (This is, actually, an update to obsolete #1424.)

### When reporting a bug/issue:
- [ ] Screenshot
- [x] The FontForge version and the operating system you're using
- [x] The behavior you expect to see, and the actual behavior
- [x] Steps to reproduce the behavior 
- [ ] \(optional) Possible solution/fix/workaround

[filt-unifont-mono-x0.sfd.txt](https://github.com/fontforge/fontforge/files/1799159/filt-unifont-mono-x0.sfd.txt)

ilyaza:
**UPDATE:**:

This is not only a warning.  All these glyphs (and only them) are indeed found by “Find Errors” (<kbd>Ctrl</kbd>-<kbd>E</kbd>) with “Paths/Open paths” checked in.

ctrlcctrlv:
There's nothing _per se_ **wrong** with your test case @ilyaza , but it's not ideal due to how weird the glyphs are and how such (sorry if this offends you) ugly, weird glyphs are unlikely to occur in a real font.

A better test case for intrepid developers hoping to figure out this longstanding issue can be found in glyph `m` inside #3684. It's a much simpler shape yet has the same problem.

ctrlcctrlv:
I found a much simpler test case, doesn't even require Spiro, consists of only corners, and only has four points.

Drum roll please! 

# :drum: :drum: :drum: :drum: 
# [y.sfd](https://github.com/fontforge/fontforge/files/3145351/y.zip)!!

![y](https://user-images.githubusercontent.com/838783/57191769-bfacf100-6f5b-11e9-8054-dc5a3cebe5b1.png)
![y2](https://user-images.githubusercontent.com/838783/57191776-cd627680-6f5b-11e9-81eb-1d0255373734.png)

cc: @skef

skef:
[...]

PR #3742, opened 2019-06-20
Clean stray control points in stroking logic.

frank-trampe:
Per #3694, the primary overlap removal routines generate wild control points sometimes, and these break overlap removal, which the stroker calls after generating the preliminary strokes. This adds a cleaner to look for those wild control points, based on offset ratios, and to remove them. Only the stroking code calls it, and the valid points generated by the stroking code seem unlikely ever to stray outside the hard-coded offset ratio. So this is likely to be a non-breaking change.

There was another small problem in which the overlap remover would flag a monotonic as being malformed if its containing spline had zero displacement. This replaces that check with one on the displacement of the monotonic.

### DIFF
--- a/fontforge/splineoverlap.c
+++ b/fontforge/splineoverlap.c
@@ -755,6 +755,8 @@ ValidateMListTs_IF_VERBOSE(input1->monos)
   return;
 }
 
+#define evalCubicSpline(spline, t) ((((spline).a * (t) + (spline).b) * (t) + (spline).c) * (t) + (spline).d)
+
 static void AddSpline(Intersection *il,Monotonic *m,extended t) {
     MList *ml;
 
@@ -843,8 +845,13 @@ return;
         else if (Within16RoundingErrors(m->tstart, m->tend))
 	    SOError( "Attempt to subset monotonic rejoin inappropriately: m->tstart and m->tend are very close (%f = %f, t = %f)\n",
 		    m->tstart, m->tend, t );
-	else if (Within4RoundingErrors(m->s->from->me.x,m->s->to->me.x) && Within4RoundingErrors(m->s->from->me.y,m->s->to->me.y))
+	else if (m->s->from->nonextcp && m->s->to->noprevcp &&
+		Within4RoundingErrors(m->s->from->me.x,m->s->to->me.x) &&
+		Within4RoundingErrors(m->s->from->me.y,m->s->to->me.y))
 	    SOError( "The curve is too short.\n");
+	else if (Within4RoundingErrors(evalCubicSpline(m->s->splines[0], m->tstart), evalCubicSpline(m->s->splines[0], m->tend)) &&
+		Within4RoundingErrors(evalCubicSpline(m->s->splines[1], m->tstart), evalCubicSpline(m->s->splines[1], m->tend)))
+	    SOError( "The subcurve is too short.\n");
         else {
 	    /* It is monotonic, so a subset of it must also be */
 	    Monotonic *m2 = chunkalloc(sizeof(Monotonic));
--- a/fontforge/splineoverlap.c
+++ b/fontforge/splineoverlap.c
@@ -848,10 +848,10 @@ return;
 	else if (m->s->from->nonextcp && m->s->to->noprevcp &&
 		Within4RoundingErrors(m->s->from->me.x,m->s->to->me.x) &&
 		Within4RoundingErrors(m->s->from->me.y,m->s->to->me.y))
-	    SOError( "The curve is too short.\n");
+	    SOError( "The spline is straight and is too short to be meaningful.\n");
 	else if (Within4RoundingErrors(evalCubicSpline(m->s->splines[0], m->tstart), evalCubicSpline(m->s->splines[0], m->tend)) &&
 		Within4RoundingErrors(evalCubicSpline(m->s->splines[1], m->tstart), evalCubicSpline(m->s->splines[1], m->tend)))
-	    SOError( "The subcurve is too short.\n");
+	    SOError( "The monotonic curve is too short.\n");
         else {
 	    /* It is monotonic, so a subset of it must also be */
 	    Monotonic *m2 = chunkalloc(sizeof(Monotonic));
--- a/fontforge/splineutil.c
+++ b/fontforge/splineutil.c
@@ -7372,7 +7372,7 @@ int SplineRemoveWildControlPoints(Spline *s, bigreal distratio) {
 		bcdisplacement1 = DistanceBetweenPoints(&s->to->me, &s->to->prevcp);
 	if (!s->from->nonextcp && !s->to->noprevcp)
 		ccdisplacement = DistanceBetweenPoints(&s->from->nextcp, &s->to->prevcp);
-	bigreal basis = MIN(pdisplacement, ccdisplacement);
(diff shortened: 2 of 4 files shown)


frank-trampe:
@ctrlcctrlv, please test. @skef, please review.

fontforge/splineoverlap.c
```diff
@@ -843,8 +845,13 @@ return;
         else if (Within16RoundingErrors(m->tstart, m->tend))
 	    SOError( "Attempt to subset monotonic rejoin inappropriately: m->tstart and m->tend are very close (%f = %f, t = %f)\n",
 		    m->tstart, m->tend, t );
-	else if (Within4RoundingErrors(m->s->from->me.x,m->s->to->me.x) && Within4RoundingErrors(m->s->from->me.y,m->s->to->me.y))
+	else if (m->s->from->nonextcp && m->s->to->noprevcp &&
+		Within4RoundingErrors(m->s->from->me.x,m->s->to->me.x) &&
+		Within4RoundingErrors(m->s->from->me.y,m->s->to->me.y))
 	    SOError( "The curve is too short.\n");
```
skef: About a month ago I read enough about actual spline math that I have started to understand this stuff a little, including the macro above (relative to FontForge's representation, which I understand bits of). I take it here you're replacing a dodgy heuristic of just looking at the end point locations with an evaluation of the endpoints in the line case (because nothing interesting can happen between the very-close-together endpoints of a line) and the spline case (where far-away control points might stretch the spline enough away from the endpoints to make the t-interval "subcurve" interesting).

If that's right, might "The line is too short." be a more informative message in the former case?

fontforge/splineutil.h
```diff
@@ -57,6 +57,8 @@ extern int SplineExistsInSS(Spline *s, SplineSet *ss);
 extern int SplinePointIsACorner(SplinePoint *sp);
 extern int SplineSetIntersect(SplineSet *spl, Spline **_spline, Spline **_spline2);
 extern int SplineSetsRemoveAnnoyingExtrema(SplineSet *ss, bigreal err);
+extern int SplineRemoveWildControlPoints(Spline *s, bigreal distratio);
```
skef: Should I scold about this being speculatively non-`static`?

fontforge/splineutil.c
```diff
@@ -7353,6 +7353,52 @@ int SplineSetsRemoveAnnoyingExtrema(SplineSet *ss,bigreal err) {
...
+	bigreal bcdisplacement1 = 0.0;
+	bigreal ccdisplacement = 0.0;
+	if (!s->from->nonextcp)
+		bcdisplacement0 = DistanceBetweenPoints(&s->from->me, &s->from->nextcp);
+	if (!s->to->noprevcp)
+		bcdisplacement1 = DistanceBetweenPoints(&s->to->me, &s->to->prevcp);
+	if (!s->from->nonextcp && !s->to->noprevcp)
+		ccdisplacement = DistanceBetweenPoints(&s->from->nextcp, &s->to->prevcp);
+	bigreal basis = MIN(pdisplacement, ccdisplacement);
+	if (basis == 0 || MAX(bcdisplacement0, bcdisplacement1) / pdisplacement > basis) {
```
skef: Maybe this is something specific to the algorithm feeding into this check, but why should ccdisplacement==0 (via basis==0) be significant here? Consider a cubic curve with points in an equilateral triangle, with `from` and `to` *d* units apart `nextcp` and `prevcp` co-located *d* units part from each of `from` and `to`. Why convert to a line in that case?

fontforge/splineoverlap.c
frank-trampe: Agreed. Or maybe "This is a line segment, and it is too short.".

fontforge/splineutil.c
frank-trampe: Oh. This logic is wrong. `basis` was supposed to be the denominator and `distratio` the other side of the comparison. I made a bunch of changes that definitely merited a full review afterwards but got distracted by the great test results.

fontforge/splineutil.h
frank-trampe: Better that than scold me for breaking ancient convention.

fontforge/splineutil.c
skef: The `basis == 0` check makes more sense if its going to be the denominator (I was wondering about that). But the other comment still applies -- even if `ccdisplacement` is going to figure into the denominator I don't get the zero check for it.

frank-trampe: Ah, yes. `basis` ought to come from `MAX` rather than from `MIN`. These curves are most problematic when they are really sharp. The case I was trying to accommodate (by a rough heuristic that would let distant collinear control points slide) was a single-spline loop starting and ending at the same point with divergent control points.

skef:
Obviously the "wild" heuristic raises questions about the chosen cutoff and whether something else would be better, but I have no objection to the general approach. So pending an explanation of the `ccdisplacement` stuff the rest looks good at this point.

frank-trampe:
I think that I fixed the ratio basis problem.

frank-trampe:
Great. It is important to remember that this new routine is used only to clean stroking output, in which valid splines would ordinarily have non-"wild" values. I would not be surprised to see a counterexample, and @ctrlcctrlv might even have one handy from his Spiro adventures. I cannot think of a case that would yield one, though, and I expect any fix in a case like that to be easy.

frank-trampe:
I've made another provisional pull request with the squashed code. I'll await test results from @ctrlcctrlv and circle back tomorrow.

ctrlcctrlv:
My qof example is broken in the exact same way; if that's what this is meant to fix it doesn't seem to have worked for me. About box shows https://github.com/fontforge/fontforge/pull/3742/commits/2edecc13bc24e58d4203c590892aa465e6a256f2

wat

skef:
I think this push was suffering from some review response churn. Let me try something real quick ...

skef:
@ctrlcctrlv Try again with this new commit. 

The change to `bigreal basis = MAX(pdisplacement, ccdisplacement);` from `MIN` was an in-the-discussion decision that ultimately doesn't make sense. Normally in the case the code is trying to catch `ccdisplacement` will be large and therefore dividing by it will make the result small and the function won't do anything. (This is something I also should have thought of as the reviewer.) 

@frank-trampe explained:

> These curves are most problematic when they are really sharp. The case I was trying to accommodate (by a rough heuristic that would let distant collinear control points slide) was a single-spline loop starting and ending at the same point with divergent control points.

I don't have enough of a feel of these cases to offer immediate advice. Maybe he will after a fresh look.

ctrlcctrlv:
That (46b21e67f4926b0d5c1a442fdc3649a8796e6bc7) fixes all the cases I had, every case I have I've shown.

2019-06-21-173315_1204x454_scrot

#3264 is unaffected by the commit, but that's to be expected given the problem there is more in splinestroke.c than splineoverlap.c

frank-trampe:
Gee, I thought that I tested the code last night, but it is not working for me now.

Given the narrow use case of this function, the `ccdisplacement` test is not strictly necessary. For code clarity, it's probably best to drop it rather than to add logic to make it work. I have pushed a small change to the comments to reflect the change. I'll make a new pull request with the squashed code.

PR #3743, opened 2019-06-20
Fix compilation with `--enable-write-pfm'.

vvp2014:
Correct the order of effective arguments in the call to WritePfmFile()
in WritePfm() according to the prototype that is declared in
fontforge/savefont.h


frank-trampe:
This was closed via #3744.

PR #3744, opened 2019-06-20
Fix compilation of gdraw.c on modern BSD systems.

vvp2014:
All modern BSD derivatives, namely FreeBSD, NetBSD, OpenBSD, and DragonFly
require <sys/select.h> to be included.


frank-trampe:
This is all good per #3739.

PR #3745, opened 2019-06-21
Clean stray control points in stroking logic (final).

frank-trampe:
This closes #3742 and #3694 with squashed commits. Original text follows.

Per #3694, the primary overlap removal routines generate wild control points sometimes, and these break overlap removal, which the stroker calls after generating the preliminary strokes. This adds a cleaner to look for those wild control points, based on offset ratios, and to remove them. Only the stroking code calls it, and the valid points generated by the stroking code seem unlikely ever to stray outside the hard-coded offset ratio. So this is likely to be a non-breaking change.

There was another small problem in which the overlap remover would flag a monotonic as being malformed if its containing spline had zero displacement. This replaces that check with one on the displacement of the monotonic.

### DIFF
--- a/fontforge/splineoverlap.c
+++ b/fontforge/splineoverlap.c
@@ -755,6 +755,8 @@ ValidateMListTs_IF_VERBOSE(input1->monos)
   return;
 }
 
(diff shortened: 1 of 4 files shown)


frank-trampe:
There is a new squash in #3747.

PR #3746, opened 2019-06-21
ifdef-wrap call to nl_langinfo using HAVE_NL_LANGINFO define, instead of HAVE_LANGINFO_H

ViliusSutkus:
Android provides langinfo.h, however, nl_langinfo functions are only
available from Android-26+. Android-25 and earlier produce
HAVE_LANGINFO_H=1
HAVE_NL_LANGINFO=0

This oneliner fixes this particular build error on Android-25 and earlier.

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/noprefs.c
+++ b/fontforge/noprefs.c
@@ -648,7 +648,7 @@ static int DefaultEncoding(void) {
     const char *loc;
     int enc;
 
-#if HAVE_LANGINFO_H
+#if HAVE_NL_LANGINFO
     loc = nl_langinfo(CODESET);
     enc = encmatch(loc,false);
     if ( enc!=e_unknown )


ctrlcctrlv:
What about other operating systems? Do they define that too if they have nl_langinfo?

Common ones at least should be checked. So Mac, GNU Linux, FreeBSD, Windows (I think we distribute cygwin.dll with our Windows releases).

frank-trampe:
In theory, that ought to be the case. Per `config.h`, `HAVE_LANGINFO_H` indicates whether `langinfo.h` is present, and `HAVE_NL_LANGINFO` indicates whether the `nl_langinfo` function is present. The [manpage](http://man7.org/linux/man-pages/man3/nl_langinfo.3.html) says that `nl_langinfo` is declared in `langinfo.h`.

I have both defined on Linux (Ubuntu 18.04 on amd64). @jtanx, can you report on Windows and Macintosh? And, @vvp2014, would you help us out on BSD? Just bootstrap and configure and report those two values in `config.h`.

My concern is that things may not work the other way around. The point of the original check, which immediately precedes a call to `nl_langinfo`, is presumably that not all implementations of `langinfo.h` provide `nl_langinfo`. If the environment has `nl_getinfo` but `HAVE_NL_LANGINFO` is false, there is something wrong in the bootstrap/configure process.

ViliusSutkus:
HAVE_NL_LANGINFO comes from

```AC_CHECK_FUNCS([nl_langinfo putenv rmdir])```

Unless there is something wrong in the bootstrap/configure process, it should be there.

By the way, here is the excerpt from Android langinfo.h ,
```
#if __ANDROID_API__ >= 26
char* nl_langinfo(nl_item __item) __INTRODUCED_IN(26);
char* nl_langinfo_l(nl_item __item, locale_t __l) __INTRODUCED_IN(26);
#endif /* __ANDROID_API__ >= 26 */
```
Functions were added in Android 26, but the same header is used to compile for previous versions.

frank-trampe:
@ViliusSutkus, I don't understand. Are you saying that `langinfo.h` on Android does not declare `nl_langinfo` for the older versions but that it is somehow possible to compile and to link code that uses that function?

ViliusSutkus:
I might not have been very clear expressing it all. Let me explain it from the top.

Android-25 (and earlier) devices have no library containing nl_langinfo function.
Android-26 added liblanginfo.so (or whatever it's called) with a nl_langinfo function inside it.

When compiling for Android you choose a minimum version that you want to be able to run on (called minSdkVersion). Toolchain takes this number and exposes it as ```#define __ANDROID_API__ 25``` , which can be understood when parsing headers or other files.

Android toolchain provides a single set of headers for all Android versions. Functionality that is added after Android-1 is #ifdef guarded with that particular ANDROID_API version. This means that the header file is always there, but it could be empty, because of all the content being #ifdef'ed. This means that I always have langinfo.h and HAVE_LANGINFO_H is always set to 1, but the function definition is not always there.

HAVE_NL_LANGINFO detects whether I actually have that function for that particular Android version which I am targetting.
If I compile for Android-26+ I get HAVE_NL_LANGINFO=1
If I compile for Android-25 (or earlier) I get HAVE_NL_LANGINFO=0

fontforge/noprefs.c calls nl_langinfo(). That call is #ifdef guarded depending on whether I have langinfo.h
After the #ifdef there is some fallback code, that I assume does more or less the same, but in a different way.
I want that call to nl_langinfo() to be #ifdef guarded depending on whether I have that particular function available.

This proposed proposed patch would do just that. Let me know if there is any ambiguity.

vvp2014:
Apropos BSDs, here: https://man.openbsd.org/ you can find a very nice interface for searching man pages for all current BSDs, not only OpenBSD.  Here: https://man.openbsd.org/nl_langinfo is the man page that you seem to be interested in.

frank-trampe:
Oh. I was reading your patch backwards. That makes sense.

Issue #3740, opened 2019-06-20
Glyph names in 'post' are not used


dscorbett:
The glyphs of [NotoSerifTibetan-Regular.ttf](https://github.com/googlefonts/noto-fonts/blob/2a41389346abdb6b83f5e0c4efe598bcb8abbb42/phaseIII_only/unhinted/ttf/NotoSerifTibetan/NotoSerifTibetan-Regular.ttf) are displayed in the main grid with names generated by FontForge, like “glyph55” and “uni0F72_uni0F7E.abvs”, instead of the actual glyph names in the 'post' table, like “uni0F62.2.left” and “uni0F720F7E”. I am using FontForge 20190413 on macOS Mojave 10.14.5.

PR #3753, opened 2019-06-30
Fix reading glyph names from post table

khaledhosny:
Fixes https://github.com/fontforge/fontforge/issues/3740

FontForge seems to assume that glyphnames array is equal to the number of used name indices, but in NotoSerifTibetan-Regular.ttf this is not the case. The font has many unused name strings and the used ones start at an index higher than the number of font glyphs, so FontForge ends assigning the names to the glyph id 0 over and over again. Fix this by actually reading all names in the array tell the end of the table which seems to be what other tools do.

### Type of change
- **Bug fix**


fontforge/parsettf.c
```diff
@@ -5236,6 +5243,7 @@ static void readttfpostnames(FILE *ttf,struct ttfinfo *info) {
 		nm[j] = '\0';
 		if ( indexes[i]<info->glyph_cnt && info->chars[indexes[i]]!=NULL )
 		    info->chars[indexes[i]]->name = nm; /* Too many fonts have badly named glyphs to deduce encoding from name */
```
skef: While "we're here" (and since this change supports reading many more strings than there are glyphs), wouldn't an "`else free(nm);` be good here?

khaledhosny: Makes sense.

ctrlcctrlv:
This works well for me. I vote to merge.

frank-trampe:
Both changes look safe to me.

PR #3755, opened 2019-07-01
unicode 12.1 update

JoesCat:
Correction from unicode 12.1_delta to 12.1_final.


ctrlcctrlv:
Hey @JoesCat , this isn't related to this PR, but can I just talk to you? I have no other way to get in touch.

I know you manage the `Unicode/` stuff, and I need to add something to it for #3712 ...

https://gist.github.com/ctrlcctrlv/29c228bb5d05e93e3abf35d1615a76ed#file-combiningclassname-c

I also need to add of course a reverse function with signature `uint32 CombiningClassFromName(char* class)` but that's not done yet.

1. Can I add this to `Unicode/`? If so where?
2. Is it OK that a script generates my function?

JoesCat:
> this isn't related to this PR, but can I just talk to you? I have no other way to get in touch.

At the moment, few people do. I'm currently travelling out of town, with spotty wifi :-P
I only happen to be in Unicode/ now (I already ran through a couple other locations). I tried doing top-down programming, and just ran into too many branches that in my opinion needed attention and you get easily lost in the tangle of code. A lot of code could use a second look, therefore, I leave the stuff that's bothering everyone right now to those actively fixing splines and that sort of stuff, while I look at replacing the duct-tape patches lower down - it's one way that we all share in improving FontForge

> 1.Can I add this to Unicode/? If so where?

#3712 seems would make logical sense to build it within Unicode/ , therefore yes, I think it's probably the best place to do your improvements. This is probably the easiest place to import unicode.org files and then build the various files like utype.c, utype.h, etc and then move them to their intended homes.

> 2.Is it OK that a script generates my function?

The first version of is_Ligature.c was basically written by makeutype.c
However it started getting fairly complex pretty quickly, and @gioele pointed out an improvement by keeping code the same and passing values/definitions - turned-out to be a good decision.

In terms of scripts - you'll need to decide what language is used. The more different languages used the more dependencies required, and the fewer places FontForge can run. Or in this case, since you plan on "building functions" - be built.
If the scripts are only intended to be used for one specific point in time during development, like makeutype.c, dump.c, makebuildtables.c, and a few Unicode/charset/scripts - it will probably be okay, and seems probably the best home for this. If it's intended to be "run" during the build phase, like an end-user building a copy of FontForge, or a distro, then you run into a new set of problems probably better avoided altogether if you can. There was a time between 20120731 to 20140101 where we had some code that was hardcoded to run scripts and it caused various sorts of problems ranging from distros like debian, redhat, centos not being able to build FontForge since these distros were still on python2.6, while distros with more modern python being ok, to other sorts of problems.

If you're comfortable with C, I would suggest building Unicode/makeutype.c because that code is what currently builds utype.c and holds classes information. If you look at the main loop where is_Ligatures is built, this may probably be a good place to duct-tape more code so that you can go beyond the 65535 code problems in makeutype.c. If you're not comfortable with C, then I'd suggest put together the scripts you mention. It's easier for other developers to understand something seen in code, plus improve it further to the next level when there's something to look at. A good example is libuninameslist names2 was based on an initial coding in python, but now is built using the C build program.

jtanx:
@JoesCat when you did this PR, did you also check/refresh utype.c? I'm getting different results running makeutype using the data from ftp://ftp.unicode.org/Public/12.1.0/ucd/

JoesCat:
ls -l ~/unicode/UnicodeData.txt 
May  6 15:59 unicode/UnicodeData.txt
wget ftp://ftp.unicode.org/Public/12.1.0/ucd/UnicodeData.txt
diff ~/unicode/UnicodeData.txt UnicodeData.txt
no difference
https://www.unicode.org/Public/12.1.0/ucd/UnicodeData.txt
diff ~/unicode/UnicodeData.txt UnicodeData.txt
no difference
diff utype.c ~/fontforge/Unicode/utype.c
no difference
summary - built same result as current utype.c.

gcc - Ran on Mageia 7, 32bit
which compiler did you use?

jtanx:
gcc 5.4.0 on xenial and gcc 9.1.0 on windows.

See #3809, this is trivially verifiable that what you generated is dependent on undefined behaviour, because some of the ranges defined in combiners.h overflow the bounds of the lookup arrays, e.g. `poses300`.

PR #3769, opened 2019-07-08
Fix saving via «Save as…» w/Unicode file names

### LINKED ISSUE #3768: Unicode confuses FontForge when saving SFD…
# *L'erreur:* 

_Une boîte de dialogue d'avertissement apparaît._

```
You tried to save with the filename Ấ.sfd but it was saved as Ấ.sfd.sfd. 
Please choose File/Generate Fonts to save to other formats.
```

The SFD content doesn't matter. Ấ.sfd should be a valid file name; the rename to Ấ.sfd.sfd is not helpful and is a bug.

_&ldquo;Why would anyone want to name a file Ấ.sfd,&rdquo;_ you ask? Well, that was the most logical name for [helping someone on the mailing list.](https://sourceforge.net/p/fontforge/mailman/message/36711601/)

It seems like having non-ASCII anywhere in the file name causes the error&hellip;

# All of the following cause the unnecessary rename

* Ấ2.sfd
* 2Ấ2.sfd
* нет.sfd
* ¡No!.sfd

# References

This is possibly a regression, albeit a very old one. (2013) See #369 and  #1572 for the code that added this error message.

ctrlcctrlv:
This closes #3768.

Unicode in file names saved via «File&rarr;Save as…» names led to an extraneous `.sfd` always being added. This was because #1572 was using `char*` instead of `unichar_t` to deal with Unicode&hellip;which is a big no-no, at least in FontForge's codebase.

### DIFF
--- a/Unicode/ustring.c
+++ b/Unicode/ustring.c
@@ -974,6 +974,16 @@ char *chomp( char *line ) {
     return( line );
 }
 
+unichar_t *u_copytolower(const unichar_t *input)
+{
+    unichar_t* ret = u_copy(input);
+    unichar_t* p = ret;
+    for( ; *p; ++p ) {
+	*p = tolower(*p);
+    }
+    return ret;
+}
+
 char *copytolower(const char *input)
 {
     char* ret = copy(input);
@@ -1027,6 +1037,18 @@ int u_endswith(const unichar_t *haystack,const unichar_t *needle) {
     int nedlen = u_strlen( needle );
     if( haylen < nedlen )
 	return 0;
+    unichar_t* p = u_strstr( haystack + haylen - nedlen, needle );
+    return p == ( haystack + haylen - nedlen );
+}
+
+int u_endswithi(const unichar_t *haystackZ,const unichar_t *needleZ) {
+    unichar_t* haystack = u_copytolower(haystackZ);
+    unichar_t* needle   = u_copytolower(needleZ);
+    int haylen = u_strlen( haystack );
+    int nedlen = u_strlen( needle );
+
+    if( haylen < nedlen ) return 0;
+
     unichar_t* p = u_strstr( haystack + haylen - nedlen, needle );
     return p == ( haystack + haylen - nedlen );
 }
--- a/fontforgeexe/fontview.c
+++ b/fontforgeexe/fontview.c
@@ -593,7 +593,7 @@ static enum fchooserret _FVSaveAsFilterFunc(GGadget *g,struct gdirentry *ent, co
 
 int _FVMenuSaveAs(FontView *fv) {
     char *temp;
-    char *ret;
+    unichar_t *ret, *utemp;
     char *filename;
     int ok;
     int s2d = fv->b.cidmaster!=NULL ? fv->b.cidmaster->save_to_dir :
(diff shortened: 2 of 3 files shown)


fontforgeexe/fontview.c
```diff
@@ -656,16 +656,19 @@ int _FVMenuSaveAs(FontView *fv) {
...
 return( 0 );
-    filename = utf82def_copy(ret);
-    free(ret);
+    filename = u2utf8_copy(ret);
 
-    if(!(endswithi( filename, ".sfdir") || endswithi( filename, ".sfd")))
+    const unichar_t *u_sfdir = utf82u_copy(".sfdir");
+    const unichar_t *u_sfd = utf82u_copy(".sfd");
+
+    if(!(u_endswith( ret, u_sfdir ) || u_endswith( ret, u_sfd )))
```
ctrlcctrlv: Does anyone know why I have to do this? Why can't I just:

```c
if(!(u_endswith( ret, (unichar_t*) ".sfdir" ) || u_endswith( ret, (unichar_t*) ".sfd" )))
```

Unicode/ustring.c
```diff
@@ -1031,6 +1041,18 @@ int u_endswith(const unichar_t *haystack,const unichar_t *needle) {
     return p == ( haystack + haylen - nedlen );
 }
 
+int u_endswithi(const unichar_t *haystackZ,const unichar_t *needleZ) {
+    unichar_t* haystack = u_copytolower(haystackZ);
+    unichar_t* needle   = u_copytolower(needleZ);
```
skef: I didn't find where you call this. What is it for?

As written this will lean haystack and needle. And it seems like the remaining lines are exactly the same as `u_endswith()`. Why can't you do:

```
    int ret;
    unichar_t* haystack = u_copytolower(haystackZ);
    unichar_t* needle   = u_copytolower(needleZ);
    ret = u_endswith(haystack, needle);
    free(haystack);
    free(needle);
    return ret;
```
?

fontforgeexe/fontview.c
skef: C is a low-level language and you're casting a pointer here. A cast can modify the format of the variable itself but in the case of a pointer, not what is pointed to. If what is pointed to needs modification to match the new pointer type the cast should fail.

fontforgeexe/fontview.c
```diff
@@ -656,16 +656,19 @@ int _FVMenuSaveAs(FontView *fv) {
     }
 #endif
 
-    ret = GWidgetSaveAsFileWithGadget8(_("Save as..."),temp,0,NULL,
+    utemp = utf82u_copy(temp);
+    ret = GWidgetSaveAsFileWithGadget((unichar_t*)_("Save as..."),utemp,0,NULL,
```
skef: Not your fault but I hate that `ret` is being used as a variable name that isn't the function's return value.

jtanx:
Ahh darn, endswithi has been patched for years on windows builds with https://github.com/fontforge/fontforge/commit/540207a7d8978c422efb4a2e87c10d5d8799477e

I didn't realise this was a problem on linux too. I reckon if you just use that, this will be fixed.

Unicode/ustring.c
ctrlcctrlv: I do call it, I just forgot to put it in this branch, it ended up in my `git stash`, sorry. Fixed in https://github.com/fontforge/fontforge/pull/3769/commits/2ff0dc1ba00dbac63a1082a9fe6e8b834f9408e6

ctrlcctrlv:
@jtanx Not sure I follow, but the work I did does fix this&hellip;if you want to port the Windows branch into master and reject this PR that's your prerogative I guess

Unicode/ustring.c
skef: If it ain't in the branch it ain't \[something clever that rhymes with "branch"\]

jtanx:
@ctrlcctrlv  yes that's my preference, just use that commit instead of these changes. I much rather prefer to use existing functions (I used glib in that case) than reinvent the wheel, especially on these string manipulation functions.

fontforgeexe/fontview.c
ctrlcctrlv: @skef OK but why does `(unichar_t*) "example"` work in other areas of the codebase, e.g. metricsview.c:816?

ctrlcctrlv:
OK no problem. Open your own PR and close this one, glad I could help :)

fontforgeexe/fontview.c
jtanx: Do you have a link to that @ctrlcctrlv ? I saw this: https://github.com/fontforge/fontforge/blob/master/fontforgeexe/metricsview.c#L816

In that case most likely that's just some specialty in gdraw where inside gdraw it knows to treat it not as a unichar_t but as a char or something like that.

skef: If you mean `(unichar_t *) _("Name:")` it's because `_()` is actually a function that takes a `char *` as its parameter, used for internationalization. It (apparently) already returns a wide string.

ctrlcctrlv: Hmm...OK, thanks guys, just seems odd to me it's so hard to declare a constant (unichar_t*)

ctrlcctrlv:
Voluntarily closing in favor of #3770

PR #3770, opened 2019-07-08
Fix tolower usage on the `endswithi` functions.

jtanx:
You cannot expect `tolower` to work if it's operating on a multibyte (e.g.
UTF-8) string.

### Type of change

- **Bug fix**

Supersedes #3769. Sorry @ctrlcctrlv on the duplicated effort

### DIFF
--- a/Unicode/ustring.c
+++ b/Unicode/ustring.c
@@ -28,6 +28,7 @@
 #include <fontforge-config.h>
 
 #include "chardata.h"
+#include "ffglib.h"
 #include "ustring.h"
 #include "utype.h"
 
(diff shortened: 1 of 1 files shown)


jtanx:
So as far as I'm aware, this fixes all usages of `endswithi`, so there shouldn't be a need to create a wide version of this. i.e. this should fix #3768 without any further modifications.

One thing to be wary of, for file system paths, there is no actual guarantee that it will be utf-8 encoded (this is one of the reasons why windows builds are patched), so as it stands, using `utf82u_copy` and trying to do a `u_endswithi` is not strictly correct.

By comparison, `g_ascii_strdown` only cares about lowering ascii characters and ignores trying to do anything else for non-ascii characters. This is good enough to satisfy the requirements of how it is currently used.

skef:
Ah, OK, got it.

Issue #3771, opened 2019-07-09
Zoom in one tab should not apply to another tab


ctrlcctrlv:
Imagine you have a CharView open with two glyphs. In one glyph, you have an illuminated manuscript in the background from which you are sourcing many glyphs. In another one, you have the destination glyph. If you don't understand what I mean, see the video.

Basically, I'm saying, zoom and view area should be on a per-tab basis, not a per-CharView basis. Having it on a per-CharView basis makes it useless for most work of this kind.

# Video: [QRD.zip](https://github.com/fontforge/fontforge/files/3371377/QRD.zip)

ctrlcctrlv:
I've been looking at the code and this seems to be because a CharView has only one `scale`, `xoff` and `yoff`, while these elements should be on the tab...not sure if it should be on `cvtabs` or `tabs`.

ctrlcctrlv:
I didn't think I could fix this on my own, but now that I did I feel, as the kids say, &laquo;gangsta AF&raquo;. :sunglasses: 
[Fixed!!.zip](https://github.com/fontforge/fontforge/files/3372115/Fixed.zip)

PR #3772, opened 2019-07-09
Make zoom & view offset tab-based (CharView)

ctrlcctrlv:
This closes fontforge/fontforge#3771. See the video there, and compare it to
this commit, to see why this change is needed.

### Type of change

- **Bug fix**

### DIFF
--- a/fontforgeexe/bitmapview.c
+++ b/fontforgeexe/bitmapview.c
@@ -714,13 +714,6 @@ static void BVExpose(BitmapView *bv, GWindow pixmap, GEvent *event ) {
 
     CharView cvtemp;
     CharViewTab cvtabtemp;
-    GGadgetData ggdatatemp;
-    GTabInfo gtitemp[2];
-    memset(gtitemp,0,sizeof(gtitemp));
-    gtitemp[0].text = (unichar_t *) '\0';
-    gtitemp[0].text_is_1byte = true;
-    ggdatatemp.u.tabs = gtitemp;
-    GGadget *ggtemp = GTabSetCreate(pixmap, &ggdatatemp, NULL);
 
     GDrawPushClip(pixmap,&event->u.expose.rect,&old);
     GDrawSetLineWidth(pixmap,0);
@@ -780,7 +773,6 @@ static void BVExpose(BitmapView *bv, GWindow pixmap, GEvent *event ) {
 	cvtabtemp.xoff = bv->xoff/* *bv->scscale*/;
 	cvtabtemp.yoff = bv->yoff/* *bv->scscale*/;
 	cvtemp.cvtabs[0] = cvtabtemp;
-	cvtemp.tabs = ggtemp;
 
 	cvtemp.b.sc = bv->bc->sc;
 	cvtemp.b.drawmode = dm_fore;
--- a/fontforgeexe/bitmapview.c
+++ b/fontforgeexe/bitmapview.c
@@ -713,7 +713,6 @@ static void BVExpose(BitmapView *bv, GWindow pixmap, GEvent *event ) {
     extern Color widthcol;
 
     CharView cvtemp;
-    CharViewTab cvtabtemp;
 
     GDrawPushClip(pixmap,&event->u.expose.rect,&old);
     GDrawSetLineWidth(pixmap,0);
@@ -764,23 +763,21 @@ static void BVExpose(BitmapView *bv, GWindow pixmap, GEvent *event ) {
 		? (bv->bc->byte_data ? 0x008800 : 0x004400 )
 		: 0x00ff00;
 	memset(&cvtemp,'\0',sizeof(cvtemp));
-	memset(&cvtabtemp,'\0',sizeof(cvtabtemp));
 	cvtemp.v = bv->v;
 	cvtemp.width = bv->width;
 	cvtemp.height = bv->height;
 
-	cvtabtemp.scale = bv->scscale*bv->scale;
-	cvtabtemp.xoff = bv->xoff/* *bv->scscale*/;
-	cvtabtemp.yoff = bv->yoff/* *bv->scscale*/;
-	cvtemp.cvtabs[0] = cvtabtemp;
+	cvtemp.cvtabs[0].scale = bv->scscale*bv->scale;
+	cvtemp.cvtabs[0].xoff = bv->xoff/* *bv->scscale*/;
+	cvtemp.cvtabs[0].yoff = bv->yoff/* *bv->scscale*/;
 
 	cvtemp.b.sc = bv->bc->sc;
 	cvtemp.b.drawmode = dm_fore;
 
-	clip.width = event->u.expose.rect.width/cvtabtemp.scale;
-	clip.height = event->u.expose.rect.height/cvtabtemp.scale;
-	clip.x = (event->u.expose.rect.x-cvtabtemp.xoff)/cvtabtemp.scale;
-	clip.y = (cvtemp.height-event->u.expose.rect.y-event->u.expose.rect.height-cvtabtemp.yoff)/cvtabtemp.scale;
+	clip.width = event->u.expose.rect.width/cvtemp.cvtabs[0].scale;
+	clip.height = event->u.expose.rect.height/cvtemp.cvtabs[0].scale;
+	clip.x = (event->u.expose.rect.x-cvtemp.cvtabs[0].xoff)/cvtemp.cvtabs[0].scale;
+	clip.y = (cvtemp.height-event->u.expose.rect.y-event->u.expose.rect.height-cvtemp.cvtabs[0].yoff)/cvtemp.cvtabs[0].scale;
 	CVDrawSplineSet(&cvtemp,pixmap,cvtemp.b.sc->layers[ly_fore].splines,col,false,&clip);
 	for ( refs = cvtemp.b.sc->layers[ly_fore].refs; refs!=NULL; refs = refs->next )
 	    CVDrawSplineSet(&cvtemp,pixmap,refs->layers[0].splines,col,false,&clip);
--- a/fontforgeexe/bitmapview.c
+++ b/fontforgeexe/bitmapview.c
@@ -703,7 +703,6 @@ static void BVDrawGlyph(BitmapView *bv, BDFChar *bc, GWindow pixmap, GRect *pixe
 }
 
 static void BVExpose(BitmapView *bv, GWindow pixmap, GEvent *event ) {
-    CharView *cvtemp;
     GRect old;
     DRect clip;
     int i;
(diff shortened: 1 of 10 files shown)


fontforgeexe/bitmapview.c
```diff
@@ -761,23 +761,23 @@ static void BVExpose(BitmapView *bv, GWindow pixmap, GEvent *event ) {
...
-	cvtemp.b.drawmode = dm_fore;
-
-	clip.width = event->u.expose.rect.width/cvtemp.scale;
-	clip.height = event->u.expose.rect.height/cvtemp.scale;
-	clip.x = (event->u.expose.rect.x-cvtemp.xoff)/cvtemp.scale;
-	clip.y = (cvtemp.height-event->u.expose.rect.y-event->u.expose.rect.height-cvtemp.yoff)/cvtemp.scale;
-	CVDrawSplineSet(&cvtemp,pixmap,cvtemp.b.sc->layers[ly_fore].splines,col,false,&clip);
-	for ( refs = cvtemp.b.sc->layers[ly_fore].refs; refs!=NULL; refs = refs->next )
-	    CVDrawSplineSet(&cvtemp,pixmap,refs->layers[0].splines,col,false,&clip);
+	cvtemp = CharViewCreateExtended(bv->bc->sc, bv->fv, BVCurEnc(bv), 0);
```
jtanx: Why is this now needed? Because cvtabs isn't initialised? Also if we do this, we're now missing a free somewhere

fontforgeexe/charview.c
```diff
@@ -488,12 +488,13 @@ const int input_em_cnt = sizeof(input_em)/sizeof(input_em[0])-1;
 
 void CVDrawRubberRect(GWindow pixmap, CharView *cv) {
     GRect r;
+    int tabnumber = GTabSetGetSel(cv->tabs);
```
jtanx: This is fine, but one way to make this less verbose is to do something like

```c
CharViewTab* tab = &cv->cvtabs[GTabSetgetSel(cv->tabs)]; // I didn't check what the type of cvtabs  is
...
tab->xoff
...
```

fontforgeexe/bitmapview.c
ctrlcctrlv: Yes that's why. If I don't do this I get a segfault. If you don't set `show` to 0 you crash Xorg (at least on my system...), but doing it this way seems fine to me. I'll add the free.

fontforgeexe/charview.c
ctrlcctrlv: I would say, &ldquo;do I have to?&rdquo; Most of the substitutions were done automatically. It's the same to the computer, and people can see the git blame to see why it's that way.

fontforgeexe/bitmapview.c
jtanx: Hmm, sounds like a bit of a code smell tbh. Do you have a backtrace?

fontforgeexe/charview.c
jtanx: Yes, I'm fine with how it's currently done, just stating my preference.

fontforgeexe/bitmapview.c
ctrlcctrlv: Sorry, I learned the hard way that Xorg would crash so I don't.

fontforgeexe/bitmapview.c
```diff
@@ -779,6 +779,7 @@ static void BVExpose(BitmapView *bv, GWindow pixmap, GEvent *event ) {
 	for ( refs = cvtemp->b.sc->layers[ly_fore].refs; refs!=NULL; refs = refs->next )
 	    CVDrawSplineSet(cvtemp,pixmap,refs->layers[0].splines,col,false,&clip);
     }
+    CVUnlinkView(cvtemp);
     CharViewFree(cvtemp);
```
ctrlcctrlv: This still isn't perfect but at least it no longer crashes. However&hellip;it leads to a phantom `fontforge` process. I remember this happened before when we were talking about modals. Do you know why this might be happening @jtanx and how I can fix it?

jtanx: Sorry, it's getting late, I think I'll only be able to look at this tomorrow, unless someone else wants to take a look. Just noting I find this a bit odd, so I'd like to get a better understanding of why it was fine to not allocate a charview before/why it's needed now, and why these frees are causing issues.

ctrlcctrlv: No problem :-)

fontforgeexe/bitmapview.c
jtanx: I need to go to work soon, but I've done some further checking, and this solution isn't correct.  I didn't realise it earlier, but you're creating and destroying a new charview in the expose handler. This gets very expensive, you can see this if you try to interact/draw in that view.

So one thing was that tabbed feature was only ever implemented in the outline view and not the bitmap view, which is why this is breaking. Adding tabbed support to the bitmap view is probably going to be more work than just this, so you might be better off trying to figure out if there's a way to fix this without adding that support.

fontforgeexe/charview.c
jtanx: Thinking about this a bit more, and in consideration of that bitmap view issue, this actually may not be the best solution. Will need to look a bit further to think of what the alternatives are.

ctrlcctrlv:
Thanks a lot @skef!! I'll give that a shot, great idea! I didn't know what to do next, and you helped me get unstuck. I need this change a lot, this particular bug has annoyed me since I was working on my KJV1611 font (2017).

ctrlcctrlv:
In my humble opinion @jtanx the fix I just pushed to https://github.com/fontforge/fontforge/pull/3772/commits/836927c62a59f51985bff17431b396c9df0b0736 is the best so far.

* No crash when opening BitmapViews
* No performance hit when adding/removing pixels from BitmapViews
* Can open/close many BitmapView without any issue
* When exiting FontForge, the process quits as expected!

fontforgeexe/bitmapview.c
```diff
@@ -713,6 +712,16 @@ static void BVExpose(BitmapView *bv, GWindow pixmap, GEvent *event ) {
...
 
+    CharView cvtemp;
+    CharViewTab cvtabtemp;
+    GGadgetData ggdatatemp;
+    GTabInfo gtitemp[2];
+    memset(gtitemp,0,sizeof(gtitemp));
+    gtitemp[0].text = (unichar_t *) '\0';
+    gtitemp[0].text_is_1byte = true;
+    ggdatatemp.u.tabs = gtitemp;
+    GGadget *ggtemp = GTabSetCreate(pixmap, &ggdatatemp, NULL);
```
jtanx: Not a fan of this approach to be honest, it's a pretty big hack to create a tab set just for the purposes of getting a 0 out of `GTabSetGetSel`. You're also missing freeing this.

My suggestion, which I was pondering this morning, is more along the lines of:

* Create a function, e.g. `CVGetActiveTab` that returns you a pointer to the tab data for the active tab. For the bitmap view, it sees that `tabs` is NULL and returns a pointer to the first element, something like:


```c
CharViewTab* CVGetActiveTab(CharView *cv) {
    if (cv->tabs == NULL) {
        // This happens when used by the bitmap view, which doesn't support tabs
        return &cv->cvtabs[0];
    }
    int tab = GtabSetGetSel(cv->tabs);
    return &cv->cvtabs[tab];
}
```

ctrlcctrlv: Freeing it in BVExpose causes a segmentation fault because later code uses it :-)

jtanx: That makes me like this approach even less. What's holding onto the tabset after this 'temporary' charview goes out of scope?

ctrlcctrlv: I think CharViewFree will clean it up

ctrlcctrlv: But I do like your approach. I'm happy to change to using it. One moment please.

ctrlcctrlv:
My latest commit, https://github.com/ctrlcctrlv/fontforge/commit/32f8c1a0112ab1e873e98be5ca955ff08960dea7, following the sage advice of @jtanx, has all the benefits of https://github.com/fontforge/fontforge/pull/3772#issuecomment-509975297 but is much less hacky. :tada:

jtanx:
Hmm is GitHub being weird for you too @ctrlcctrlv? I can see the referenced commit in your comment but I don't see it in the actual PR

ctrlcctrlv:
I pushed it to the right branch so I'm not sure why it's not showing up here

jtanx:
Yeah, I think Github is broken... Anyway took a look at the commit, seems pretty good, the only point is don't dereference the whole tab in `CVDrawRubberRect`, dereference the individual members like you do elsewhere.

ctrlcctrlv:
Done :)

fontforgeexe/bitmapview.c
```diff
@@ -762,19 +764,23 @@ static void BVExpose(BitmapView *bv, GWindow pixmap, GEvent *event ) {
...
 	cvtemp.width = bv->width;
 	cvtemp.height = bv->height;
-	cvtemp.scale = bv->scscale*bv->scale;
-	cvtemp.xoff = bv->xoff/* *bv->scscale*/;
-	cvtemp.yoff = bv->yoff/* *bv->scscale*/;
+
+	cvtabtemp.scale = bv->scscale*bv->scale;
+	cvtabtemp.xoff = bv->xoff/* *bv->scscale*/;
+	cvtabtemp.yoff = bv->yoff/* *bv->scscale*/;
+	cvtemp.cvtabs[0] = cvtabtemp;
```
jtanx: Pretty sure you don't need this `cvtabtemp` any more, you can just assign directly to cvtemp.cvtabs[0]

fontforgeexe/cvdebug.c
```diff
@@ -2244,6 +2244,7 @@ return;
 }
 
 void CVDebugPointPopup(CharView *cv) {
+    int tabnumber = GTabSetGetSel(cv->tabs);
```
jtanx: Might as well keep it consistent and change these usages too

fontforgeexe/cvfreehand.c
```diff
@@ -77,12 +77,13 @@ static void TraceDataFree(TraceData *td) {
 }
 
 static void TraceDataFromEvent(CharView *cv, GEvent *event) {
+    int tabnumber = GTabSetGetSel(cv->tabs);
```
jtanx: And here too + the rest of the other files

ctrlcctrlv:
Sorry I keep missing things; I really am doing my best, I think this is the largest change I've tried to make so far, thanks for your patience and understanding. I'll fix that ASAP

ctrlcctrlv:
# :sweat_smile: 

All done this time I hope. Waiting for that :heavy_check_mark: :eyes:

skef:
Oh, I was assuming in my "quickie work-around" that the tabs structure wasn't fixed, but:
```
CharViewTab cvtabs[ charview_cvtabssz+1 ];
```
I would guess in that case that some of the problems in the earlier versions that prompted @ctrlcctrlv to add `CharViewCreateExtended` stemmed from that structure not being zeroed out.

skef:
No, that doesn't really make sense either. I'll shut up now. Glad this got worked through.

PR #3774, opened 2019-07-11
Allow preferring combining accents.

frank-trampe:
This is an attempt to address #3712 in a fashion sensitive to the "PreferSpacingAccents" preference.

### DIFF
--- a/fontforge/fvcomposite.c
+++ b/fontforge/fvcomposite.c
@@ -1799,6 +1799,9 @@ static SplineChar *GetGoodAccentGlyph(SplineFont *sf, int uni, int basech,
 	    else
 		ach = 0x302;
 	}
+	// Convert back to combining accents from spacing accents if requested.
+	if (!PreferSpacingAccents)
+		ach = CanonicalCombiner(ach);
     } else
 	ach = uni;
 
@@ -1841,6 +1844,9 @@ static SplineChar *GetGoodAccentGlyph(SplineFont *sf, int uni, int basech,
 		apt = accents[uni-BottomAccent]; end = apt+sizeof(accents[0])/sizeof(accents[0][0]);
 		while ( test==NULL && apt<end ) {
 		    int acc = *apt ? *apt : uni;
+		    // Convert back to combining accents from spacing accents if requested.
+		    if (!PreferSpacingAccents)
+			acc = CanonicalCombiner(acc);
 		    sprintf( buffer,"%.70s.%s", StdGlyphName(buffer,acc,ui_none,(NameList *) -1), suffixes[i]);
 		    if ( (test = SFGetChar(sf,-1,buffer))!=NULL )
 			rsc = test;
@@ -1859,6 +1865,9 @@ static SplineChar *GetGoodAccentGlyph(SplineFont *sf, int uni, int basech,
 	    apt = accents[uni-BottomAccent]; end = apt+sizeof(accents[0])/sizeof(accents[0][0]);
 	    while ( test==NULL && apt<end ) {
 		int acc = *apt ? *apt : uni;
+		// Convert back to combining accents from spacing accents if requested.
+		if (!PreferSpacingAccents)
+		    acc = CanonicalCombiner(acc);
 		sprintf( buffer,"%.70s.%s", StdGlyphName(buffer,acc,ui_none,(NameList *) -1), suffixes[i]);
 		if ( islower(buffer[0])) {
 		    buffer[0] = toupper(buffer[0]);
--- a/fontforge/fvcomposite.c
+++ b/fontforge/fvcomposite.c
@@ -1773,35 +1773,32 @@ static SplineChar *GetGoodAccentGlyph(SplineFont *sf, int uni, int basech,
 	    (dot!=NULL && haschar(sf,uni, NULL) &&
 		!SCDependsOnSC(GetChar(sf,uni,NULL),destination))) )
 	ach = uni;
-    else if ( uni>=BottomAccent && uni<=TopAccent ) {
+    else if ( uni>=BottomAccent && uni<=TopAccent && (PreferSpacingAccents || !haschar(sf,uni,NULL)) ) {
 	apt = accents[uni-BottomAccent]; end = apt+sizeof(accents[0])/sizeof(accents[0][0]);
 	while ( *apt && apt<end &&
 		(             GetChar(sf,*apt,dot) ==NULL || SCDependsOnSC(GetChar(sf,*apt,dot),destination)) &&
 		(dot==NULL || GetChar(sf,*apt,NULL)==NULL || SCDependsOnSC(GetChar(sf,*apt,NULL),destination)) )
 	    ++apt;
-	if ( *apt!='\0' && apt<end )
+	if ( *apt!='\0' && apt<end && PreferSpacingAccents)
 	    ach = *apt;
 	else if ( haschar(sf,uni,dot) && !SCDependsOnSC(GetChar(sf,uni,dot),destination))
 	    ach = uni;
 	else if ( uni==0x31b && haschar(sf,',',NULL))
 	    ach = ',';
-	else if (( uni == 0x32f || uni == 0x311 ) && haschar(sf,0x2d8,NULL) && ia==0 ) {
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
I would also change https://github.com/fontforge/fontforge/blob/master/share/prefs to make `PreferSpacingAccents` false by default, since that's not the behavior most expect or the behavior that makes sense in a modern context.

frank-trampe:
@ctrlcctrlv, that does seem like a good idea. I just sent a note to the user list to get feedback.

frank-trampe:
@skef, I would appreciate a review of the new logic.

frank-trampe:
I know that we're waiting on the preference change, but we can still test and review this. I'd like to get it buttoned up so that I can mentally move on to other things. The preference change is an easy thing to do later.

ctrlcctrlv:
Sorry, I will help review it, and will test it, the font which I was working on (TT2020) which had this problem is kind of on the back burner due to Expand Stroke crashes/poor stroking/etc.

skef:
From an initial look this seems off so I'm going to ask a question before further review.

The point of this function is to find the best accent glyph available for combining and the goal of this change is to affect what counts as "best". But it seems like in each case the "meat" of the search happens prior to the added `!PreferSpacingAccents()` check and then the canonical combiner is proposed instead. What if it's missing?  Shouldn't FontForge still choose the best available rather than moving forward with an empty entry? Or am I missing some aspect of the checks?

ctrlcctrlv:
I don't know the answer to @skef's question but I **confirm** this works for me.

ctrlcctrlv:
I'm going to merge this tomorrow if no one objects. @skef?

skef:
I have a review question that has not been addressed, so I guess I object.

If you (@ctrlcctrlv) want to help investigate it, what happens with this code when:

1. `PreferSpacingAccents` is false
2. You attempt to build an accented character in a font that has the spacing accent but is missing the combining accent

?

ctrlcctrlv:
Sure I'll help. Sorry, I'm just trying to tie up loose ends before the next release, and needed to think of something to say other than "bump".

ctrlcctrlv:
@skef I gave it a try, the Build menu is greyed out, it's not possible to build. SFD attached.

[Aacute.zip](https://github.com/fontforge/fontforge/files/3409975/Aacute.zip)

ctrlcctrlv:
By the way I agree that that's a big problem; this can't be merged as-is. So&hellip;I guess the option could be renamed to `UseSpacingAccents`, that describes the current behavior.

skef:
I can see an argument for such an option if the user didn't accidentally want to build a reference to a non-combining-character. But in most cases "prefer" would be what a user wants. `GetGoodAccentGlyph` is supposed to implement preferences not absolutes. 

Incidentally (i.e. for the record) the disabling is because `GetGoodAccentGlyph()` winds up calling `SFGetChar()` on a non-existent glyph and therefore returning NULL, and `SFIsCompositBuildable()` calls `GetGoodAccentGlyph()` to check if the char can be built. Those checks are all good -- better to disable the menu rather than build a bad glyph.

ctrlcctrlv:
So @skef you like the behavior of this and think it should be merged

For the record, I never want to build chars with spacing accents so this really doesn't matter to me...

frank-trampe:
So, to be clear, this is non-regressive, but it fails to return any accent when there is not an appropriate spacing accent?

skef:
When `PreferSpacingAccents` is *false*, and the corresponding combining accent is missing, it fails to return a corresponding spacing accent even when one is present. Since `PreferSpacingAccents` is apparently an already-existing configuration setting the change could be counted as a regression. 

The setting should influence the "search order", so that when it's false combining characters are preferred over spacing characters. The function should never return "NULL" when a character on the list is non-null (or, if we care to check, non-null with spline content or a reference).

frank-trampe:
It's easy to achieve that behavior. We can just bounds-check (`0x300 <= uni < 0x370`) in addition to the flag check before converting. I actually prefer the current behavior (since silently mixing accent modes seems undesirable), but I guess we do need to honor the preference name.

frank-trampe:
@skef, @ctrlcctrlv, does that sound right?

ctrlcctrlv:
I'm _pretty_ sure that's right. It will at least achieve the same result as #3712 when "PreferSpacingAccents" is _false_. My only worry would be that there are combining accents outside of 0x300-0x370. You might want to test `iscombining(uni)` instead.

frank-trampe:
There may be combining accents outside of that range, but they aren't in the table, so we would just get a null glyph in that case.

ctrlcctrlv:
What I'm saying is, we'd want `PreferSpacingAccents` to work outside that range right? If there are other spacing versions of combining accents not in the range? I'm specifically talking about the Armenian pativ, which is currently wrongly marked by Unicode as a spacing glyph, I'm writing a proposal which will see a combining version of it added,&dagger; at which point we'd want that used over the spacing pativ.

&dagger; Because Unicode can't delete glyphs, it's one of their promises to implementors.

skef:
>  does that sound right?

I'm not sure I understand the "bounds check" proposal. Given the current code it seems like the most "contained" hack would be to replace

```
    if (!PreferSpacingAccents)
        ach = CanonicalCombiner(ach);
```
with something like
```
    if (!PreferSpacingAccents)
        tmp = CanonicalCombiner(ach);
        if (IsARealChar(tmp))
             ach = tmp
```
Where "is a real char" is something like the current `sprintf()/SFGetChar` combo. But that would be a bunch of ugly code duplication. So it might be cleaner to have `ach_primary` and `ach_backup` (or `ach[2]` or whatever) and check for both, where the one checked first is set according to `CanonicalCombiner` (or left as -1) and the other is set the normal way. 

But maybe this is what you're already thinking ...

frank-trampe:
The problem is that the table is currently indexed by combining accents. If we need to support non-adjacent ranges, we would need to move to a more complicated data structure. Maybe an array of structs that have start encoding, length, and the accent table. We could call the replacement routine and the bounds-checker on that array.

ctrlcctrlv:
@frank-trampe I'd really like to see this in the next release. Why don't we just add a note to the option, "Note: This only affects the use of glyphs in the Combining Diacritical Marks block.", and call it a day for now? We can revisit this for more exotic cases when the time comes. The population who'd want to use the _wrong_ spacing mark over the correct combining mark in non-Latin writing systems is vanishingly small, they should not hold up this merger.

frank-trampe:
Oh, indeed. I wasn't proposing that we do that all now. I'll add the extra condition and a qualifier that fits tomorrow.

frank-trampe:
Here's a question. Would we ever want fvcomposite.c:1776 to evaluate true and the dependent code path to execute if `PreferSpacingAccents` is false and the existing encoding exists (`haschar(sf,uni,NULL)`)?

ctrlcctrlv:
No we wouldn't because that code path uses `accents`, which `PreferSpacingAccents` set to false disallows.

frank-trampe:
Alright. How about this then?

ctrlcctrlv:
Looks good. Once you change the preference default I think we can merge this.

skef:
Maybe I'm missing something about your question and this new version but wouldn't the relevant test be `(PreferSpacingAccents || !haschar(sf,CanonicalCombiner(uni),NULL))`? And then wouldn't the substitution still be needed somewhere below? I don't think `GetGoodAccentGlyph()` is guaranteed to be called with the canonical value.

frank-trampe:
The idea is to convert from combiner to spacer only if we prefer spacers or if the originally requested accent (which is a combiner if in the previously checked range) is unavailable. The logic works in my tests, presumably since upstream callers use the combiner Unicode value. If we find cases in which we need to be more aggressive (mapping back to combiners from a spacing input value), we can do so, but I don't want to be unnecessarily aggressive.

skef:
Ah, so in the cases in question the caller will supply the combining character unicode because that's what is in the decomposition. Makes sense.

frank-trampe:
@ctrlcctrlv, I merged this without the preference adjustment. Let's just do it in another pull request. I'll open it tomorrow if you don't get to it.

As you may recall, we asked for feedback on the preference change on the mailing list and got no response. As we are the only interested parties and the change seems positive and easily bypassed, we can safely push that through once tests pass.

PR #3781, opened 2019-07-11
Treat single-point splines as having desired order when dumping

### LINKED ISSUE #3775: Annoying warnings popping up: «Asked for cubic when had quadratic.»
I think this is a regression caused by a bug I fixed; at least the warning is something I introduced (on the advice of @skef)&mdash;#3679.

Just removing the warning works for me, but what causes it? I can't seem to figure it out, I think it's auto save related but would like help tracking it down.

To see the regression:

* Open a new font.
* Open any glyph.
* Enable Spiro.
* Put a G4 curve (circle shaped) point anywhere.

skef:
Closes #3775

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -674,9 +674,10 @@ static void SFDDumpHintMask(FILE *sfd,HintMask *hintmask) {
 
 static void SFDDumpSplineSet(FILE *sfd, SplineSet *spl, int want_order2) {
     SplinePoint *first, *sp;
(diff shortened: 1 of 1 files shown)


frank-trampe:
That looks right.

ctrlcctrlv:
Looks good to me (and works for me) but I think you should stick a comment because if I were reading this it'd be confusing to see `order2` being set to the value of `want_order2`, while we're here it's best to explain that single-point splines don't have a particular &laquo;order&raquo; and that's what the ternary operator is handling, single-point splines&hellip;

PR #3785, opened 2019-07-13
Version cleanup

jtanx:
As I was looking into CMake, I decided to clean up the number of defines that we have to set the version of FontForge. We seriously don't need as many as we do now. 

* Removed a bunch of redundant version strings
* Changed usages of FONTFORGE_VERSIONDATE (the date at which FontForge was built) to the version defined in `configure.ac`. The primary version that FontForge reports should **not** change based on when it was built. i.e. we tag 20190413, it should remain tagged as that independent of when it was built.

I've also fixed a couple of bugs/annoyances with the splash screen:
* I've reduced the text size from 12pt to 10pt, plus tried to make it a bit briefer. It was so large that it was practically taking up my full screen height: VirtualBox_xenial_13_07_2019_20_47_49, after for comparison 
image. Still pretty large, but better
* There was a bug that I'd see on startup where all the text would try to be rendered at the bottom of the image (where the splash screen was only the size of the splash image). So I've changed the expose handler to not render the text if rendering an area within the splash image.

### Type of change
- **Breaking change**
  - In as much as I've removed/consolidated previously defined version strings


jtanx:
@frank-trampe  can you please remove https://ci.appveyor.com/project/frank-trampe/fontforge, or at least not run it off the main repo? It's screwing up the build status.

fontforgeexe/startui.c
```diff
@@ -304,9 +294,6 @@ static void SplashLayout() {
     uc_strcat(pt,")");
```
frank-trampe: This is spurious now.

fontforge/dumppfa.c
```diff
@@ -1833,7 +1833,7 @@ static void dumpfontcomments(void (*dumpchar)(int ch,void *data), void *data,
     }
     if ( sf->comments!=NULL )
 	dumpascomments(dumpchar,data,sf->comments);
-    dumpf(dumpchar,data,"%% Generated by FontForge %d (http://fontforge.sf.net/)\n", FONTFORGE_VERSIONDATE_RAW);
+    dumpf(dumpchar,data,"%% Generated by FontForge %s (http://fontforge.sf.net/)\n", FONTFORGE_VERSION);
```
frank-trampe: This is a change in behavior, but I guess it's a good one.

fontforge/python.c
```diff
@@ -19221,20 +19218,20 @@ static void SetPythonModuleMetadata( PyObject *module ) {
...
     PyObject* pydate;
-    int dt = FONTFORGE_VERSIONDATE_RAW;
+    time_t dt = FONTFORGE_MODTIME_RAW;
+    struct tm* modtime = gmtime(&dt);
 
     /* Make __version__ string */
-    snprintf(ver, sizeof(ver), "%u.%u git:%s",
-	     FONTFORGE_LIBFF_VERSION_MAJOR,
-	     FONTFORGE_LIBFF_VERSION_MINOR,
+    snprintf(ver, sizeof(ver), "%s git:%s",
```
frank-trampe: I guess that this is okay since we encode the version number in the git tag as appropriate.

frank-trampe:
This all looks appropriate to me.

frank-trampe:
I have attempted to disable my Appveyor build. I would delete it, but I never created it in the first place, which makes me think that it gets created automatically somehow, so I would like to leave a disabled placeholder if possible.

ctrlcctrlv:
Assuming this builds it's all okay with me. I had some question as to how the current year thing I added to e.g. startui.c would work but verified that just by reading the source. I understand the desire to simplify the build in any way possible to make room for cmake.

travis-scripts/ffosxbuild.sh
```diff
@@ -96,10 +90,7 @@ sed -i -e "s/CFBundleGetInfoString.*/CFBundleGetInfoString = \"$CFBundleGetInfoS
 # Replace version strings in Info.plist
 sed -i -e "s/CFBundleShortVersionStringChangeMe/$CFBundleShortVersionString/g" $outdir/Contents/Info.plist
 sed -i -e "s/CFBundleGetInfoStringChangeMe/$CFBundleGetInfoString/g" $outdir/Contents/Info.plist
-sed -i -e "s/CFBundleVersionChangeMe/$FONTFORGE_VERSIONDATE_RAW/g" $outdir/Contents/Info.plist
-
-# Remove the temp file
-rm /tmp/fontforge-config-sorted
+sed -i -e "s/CFBundleVersionChangeMe/$FONTFORGE_VERSION/g" $outdir/Contents/Info.plist
```
skef: Since `VERSION_MAJOR` and `VERSION_MINOR` aren't going away it seems as if *VMa.VMi.V* would be [more compliant](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion) here. But I guess if we've been doing it wrong so far there's an argument for consistency.

jtanx: VERSION_MAJOR/MINOR are mostly on the way out too, I just didn't bother to remove it from here, because they're also used for the soversions.

As for why, those versions are/were inconsistently bumped, plus we've been using a date as the version for years now

frank-trampe: We are going to need to fix that someday. I am not happily anticipating our adventures with epochs.

jtanx: Yeah, that's fair. I'll leave fixing that out of this PR though.

PR #3786, opened 2019-07-14
Use getShareDir/getHelpDir everywhere

jtanx:
This consolidates the use of SHAREDIR/DOCDIR to gutils/fsys.c

### Type of change

- **Non-breaking change**


skef:
looks good to me also

frank-trampe:
Were we really doing the same thing four different ways?

jtanx:
Thanks all, yeah it was pretty iffy

Issue #3787, opened 2019-07-14
splash screen invisible


lemzwerg:
[da431519bab799cb76a1c199b9e615e81ad31c3a]
[openSUSE Leap 15.1]

Configuring fontforge with
```
./configure --enable-debug-raw-points \
            --enable-silent-rules \
            --enable-freetype-debugger=... \
            CFLAGS="-g -O0"
```
(this results in not using GTK), followed by calling `make && make install`, then starting FontForge without argument produces an empty splash screen.  Is this expected?  I have all necessary graphics libraries installed (only missing libspiro, whch shouldn't be relevant here).

jtanx:
Ugh this hurts my brain, so it seem like this is a bit of a bug only in the gxdraw/non-gdk backend.

If only an image is displayed, and no text rendered, it doesn't actually get flushed to the screen. When text does get written, the image finally gets shown.

The text doesn't get rendered on startup because of a change in #3785, due to another glitch in the text all being rendered at the bottom of the screen.

Thinking about it a bit more, that glitch seen in #3785 is probably because as/fh were 0 when the splash started (since those metrics are computed after), and because we removed the resizing during startup, it only ever gets rendered once.

It's getting late, so I'll look into this some other time, unless someone else fixes this first.

ctrlcctrlv:
Hi @lemzwerg - can you please try commit https://github.com/fontforge/fontforge/commit/a508d470f0093e9768199b3414a7cb41d457670b to see if the recent commit by @jtanx (https://github.com/fontforge/fontforge/commit/93605760bda7a318d925a35bbf194a4ea6468ed2) caused this issue?

jtanx:
Ok that's not quite correct, the changes in #3785 to not render the text don't cause the issue. It seems there's a requirement for the resize event to make it display properly (which used to happen when it did that resizing thing on the timer).

I also did actually manage to cobble a hack together.

jtanx:
Fixed by #3788

PR #3788, opened 2019-07-14
Fix splash screen on startup for gxdraw

jtanx:
A somewhat bad hack to fix the splash for gxdraw as seen in #3787 

### Type of change

- **Bug fix**

### DIFF
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -642,6 +642,16 @@ static int splash_e_h(GWindow gw, GEvent *event) {
 	}
 	GDrawPopClip(gw,&old);
       break;
+      case et_map:
+	// The splash screen used to gradually resize the longer it was displayed.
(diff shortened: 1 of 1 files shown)


lemzwerg:
This patch works, thanks.

ctrlcctrlv:
We definitely need a comment as this code looks like a mistake. Why can't this just be fixed in gxdraw? Is that not worth the effort as gxdraw is going to be deprecated?

jtanx:
If you want to fix gxdraw you're more than welcome to, but I won't be touching it. I have a hunch that it's this call, which is invoked on a resize: https://www.cairographics.org/manual/cairo-XLib-Surfaces.html#cairo-xlib-surface-set-size

That somehow fixes the cairo surface. Not sure why this isn't seen on other windows.

Anyway, I've added in some comments.

PR #3791, opened 2019-07-16
Further header cleanup

jtanx:
These address a couple more issues that I've found while looking into CMake. The inter-dependence of the headers is still pretty questionable, particularly between fontforge and fontforgeexe, but I think this is good enough for now.

Probably the biggest change is moving the modtime versioning information into a separate header. You notice it less with the current build system purely because *everything* takes so long, but separating out the modtimes means that each time the configuration script is run, most of it won't need to be recompiled, unlike now, where each time configure is invoked, it's practically a full rebuild.

### Type of change

- **Breaking change** in as much as moving some defines into a new header


gdraw/Makefile.am
```diff
@@ -40,7 +40,7 @@ libgdraw_la_SOURCES = choosericons.c ctlvalues.c drawboxborder.c	\
 	gwidgets.c gxdraw.c gxcdraw.c ghvbox.c gmatrixedit.c		\
 	gdrawable.c gspacer.c xkeysyms_unicode.c colorP.h gdrawP.h	\
 	gresourceP.h gxcdrawP.h fontP.h ggadgetP.h gpsdrawP.h		\
-	gwidgetP.h gxdrawP.h hotkeys.c hotkeys.h div_tables.c		\
+	gwidgetP.h gxdrawP.h hotkeys.c div_tables.c			\
```
skef: I take it that the incremental build dependency is now implemented via some chain ultimately ending in `pkginclude_HEADERS`?

frank-trampe:
I likewise don't see anything that could be dangerous, and I like the detachment of gdraw from libfontforge.

jtanx:
> I take it that the incremental build dependency is now implemented via some chain ultimately ending in pkginclude_HEADERS

Headers don't actually need to be part of this, gcc tracks header dependencies. The only reason to include the headers in the makefile is so autotools knows which headers to install, and also which headers to add to the distribution tarball


>  I like the detachment of gdraw from libfontforge

That sadly is still far from being achieved, I've only removed direct references, there's still indirect references via other included headers.

PR #3795, opened 2019-07-18
Splash screen title bar removal

happyrobotstudio:
Ensuring the splash screen is displayed for 7 seconds, yet making use of wam_nodecor to take away the title bar for aesthetic pleasure (click on image will close also, title bar is reduntant functionally and visually for splash screen).

### DIFF
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -1250,9 +1250,7 @@ int fontforge_main( int argc, char **argv ) {
     /*  the window around, which I can determine if I have a positioned */
     /*  decorated window created at the begining */
     /* Actually I don't care any more */
-    /* happyrobotstudio Jul_8_2019: updated to re-include wam_nodecor and wattrs.nodecoration = 1 */
-
-    wattrs.mask = wam_events|wam_cursor|wam_bordwidth|wam_backcol|wam_positioned|wam_nodecor; //|wam_utf8_wtitle|wam_isdlg;
+    wattrs.mask = wam_events|wam_cursor|wam_bordwidth|wam_backcol|wam_positioned|wam_nodecor;
     wattrs.event_masks = ~(1<<et_charup);
     wattrs.positioned = 1;
     wattrs.nodecoration = 1;
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -1255,7 +1255,6 @@ int fontforge_main( int argc, char **argv ) {
     wattrs.positioned = 1;
     wattrs.nodecoration = 1;
     wattrs.cursor = ct_pointer;
-    //wattrs.utf8_window_title = "FontForge";
     wattrs.border_width = 0;
     wattrs.background_color = 0xffffff;
 #ifdef FONTFORGE_CAN_USE_GDK
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -400,7 +400,7 @@ static void start_splash_screen(void){
     GDrawProcessPendingEvents(NULL);
     GDrawProcessPendingEvents(NULL);
 
-    splasht = GDrawRequestTimer(splashw,3200,1000,NULL);
+    splasht = GDrawRequestTimer(splashw,7000,1000,NULL);
 
     localsplash = false;
 }
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -1272,7 +1272,6 @@ int fontforge_main( int argc, char **argv ) {
     if ( unique && GDrawSelectionOwned(NULL,sn_user1)) {
 	/* Different event handler, not a dialog */
 	wattrs.is_dlg = false;
-   // GDrawCreateTopWindow(GDisplay *gdisp, GRect *pos, int (*eh)(GWindow,GEvent *), void *user_data, GWindowAttrs *wattrs)
 	splashw = GDrawCreateTopWindow(NULL,&pos,request_e_h,NULL,&wattrs);
 	PingOtherFontForge(argc,argv);
     } else {
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -393,9 +393,7 @@ exit( 0 );		/* But the event loop should never return */
 }
 
 static void start_splash_screen(void){
-
     GDrawSetVisible(splashw,true);
-
     GDrawSync(NULL);
     GDrawProcessPendingEvents(NULL);
     GDrawProcessPendingEvents(NULL);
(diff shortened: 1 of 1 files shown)


happyrobotstudio:
> (Obviously) this code looks fine too.

Hoorah =)

FonrForge === <3

fontforgeexe/startui.c
```diff
@@ -1248,12 +1250,15 @@ int fontforge_main( int argc, char **argv ) {
     /*  the window around, which I can determine if I have a positioned */
     /*  decorated window created at the begining */
     /* Actually I don't care any more */
-    wattrs.mask = wam_events|wam_cursor|wam_bordwidth|wam_backcol|wam_positioned|wam_utf8_wtitle|wam_isdlg;
+    /* happyrobotstudio Jul_8_2019: updated to re-include wam_nodecor and wattrs.nodecoration = 1 */
+
+    wattrs.mask = wam_events|wam_cursor|wam_bordwidth|wam_backcol|wam_positioned|wam_nodecor; //|wam_utf8_wtitle|wam_isdlg;
```
ctrlcctrlv: These comments aren't really needed&hellip`git blame` is good enough. Also, please refrain from making a new PR when changes are requested, please push your changes to the initial PR.

happyrobotstudio: Hiya,

Updated this item, 

Initially i was working on happyrobotstudio:fontforge:master

However thought to create a Branch in the usual fashion .. will update same branch in future

Cheers!

fontforgeexe/startui.c
```diff
@@ -393,7 +393,9 @@ exit( 0 );		/* But the event loop should never return */
 }
 
 static void start_splash_screen(void){
+
     GDrawSetVisible(splashw,true);
+
     GDrawSync(NULL);
```
ctrlcctrlv: Likewise, adding whitespace to unrelated areas of the code just makes `git blame` harder to read.

fontforgeexe/startui.c
```diff
@@ -1248,12 +1250,13 @@ int fontforge_main( int argc, char **argv ) {
...
     /* Actually I don't care any more */
-    wattrs.mask = wam_events|wam_cursor|wam_bordwidth|wam_backcol|wam_positioned|wam_utf8_wtitle|wam_isdlg;
+    wattrs.mask = wam_events|wam_cursor|wam_bordwidth|wam_backcol|wam_positioned|wam_nodecor;
     wattrs.event_masks = ~(1<<et_charup);
     wattrs.positioned = 1;
+    wattrs.nodecoration = 1;
     wattrs.cursor = ct_pointer;
-    wattrs.utf8_window_title = "FontForge";
-    wattrs.border_width = 2;
+    //wattrs.utf8_window_title = "FontForge";
```
ctrlcctrlv: This comment can be removed too.

Thank you for your contribution, by the way, we are always happy to have new contributors and hope for many more :-)

fontforgeexe/startui.c
happyrobotstudio: Oh gosh thought I got them all, yep, will figure out mah editor =PP

fontforgeexe/startui.c
happyrobotstudio: Thanks!

Yep, looking to work on areas as I learn where and what they are

fontforgeexe/startui.c
```diff
@@ -1248,12 +1248,12 @@ int fontforge_main( int argc, char **argv ) {
     /*  the window around, which I can determine if I have a positioned */
```
jtanx: This is kind of contradicting this comment now...

ctrlcctrlv: 1250: `    /* Actually I don't care any more */`

That's George, lol. Maybe delete the whole comment block?

jtanx:
I actually prefer if we kept the title bar. Now I can't move the splash screen at all, and that's a problem because it's basically  off the screen:

image

skef:
I guess I didn't review this carefully enough. I think it's a reasonable stylistic choice to avoid putting the title bar on the splash screen, but the about window should have one. I thought the change was limited to the former, but (unless something strange is happening) @jtanx's picture shows the latter. 

Could we iterate on this change to split the cases, and if we did that would @jtanx be OK with the result?

jtanx:
Ok, so the about dialog definitely should have a border, no questions about that.

If you can find a way to not have decor on the splash but on the about dialog then that's ok with me.

But I can already tell, you're going to run into problems, because those styles can only be specified on window creation and the splash window is overloaded, it's used for other tasks too apart from displaying the splash screen (e.g. autosaving and something about clipboard processing, not sure if that still works though.)

ctrlcctrlv:
Sorry, I also missed that bug, ugh. I've reverted this until it can be fixed&hellip;as @happyrobotstudio already deleted the branch most likely a new PR needs to be opened, I'm not sure that creating a new branch with the same name as a deleted branch works on Github.

happyrobotstudio:
Hmmm, potentially I can do some toying around with screen co-ords and ensure a nice placement of the screen?

I would really enjoy the splash without the title =PP

happyrobotstudio:
I will work on trying to get a reasonably placed window,

And also to not destroy the title bar for the about section

Ok, thanks everyone for feedback, will post changes for review soon

happyrobotstudio:
> Sorry, I also missed that bug, ugh. I've reverted this until it can be fixed…as @happyrobotstudio already deleted the branch most likely a new PR needs to be opened, I'm not sure that creating a new branch with the same name as a deleted branch works on Github.

Apologies for that, just getting sorted out with my git ... Will keep branches open in future

happyrobotstudio:
> I guess I didn't review this carefully enough. I think it's a reasonable stylistic choice to avoid putting the title bar on the splash screen, but the about window should have one. I thought the change was limited to the former, but (unless something strange is happening) @jtanx's picture shows the latter.
> 
> Could we iterate on this change to split the cases, and if we did that would @jtanx be OK with the result?

Indeed the about page should be as usual with a title bar, some further looking into the window creation to split out the cases may be needed

I'll keep at it

happyrobotstudio:
Ok, yes after carefully reading I see the difficulty, being that the splash window performs some other functions, and those may be affected by the window mask specified.

Hmmz.

From a user experience perspective, a solid loading screen makes me love a program =P I'll use this as my intro to X11 ... try and figure out something

I have another PR regarding the Splash Screen Timeout, simply adding an integer for the desired seconds to timeout the splash screen .. that seems ok as a simple change =P

I will cut my teeth with X11 on _NET_WM_WINDOW_TYPE_SPLASH etc etc and see what might work maybe sorta

jtanx:
> I will cut my teeth with X11 on _NET_WM_WINDOW_TYPE_SPLASH etc etc and see what might work maybe sorta

If it's anything x11 specific it won't be accepted. We're already moving away from that by going through gdk (and eventually gtk most likely)

Maybe splitting out a new window just for the startup splash might work, but that's some effort I won't undertake myself.

Issue #3055, opened 2017-04-15
FontForge shows very old splash screen


probonopd:
The latest version available in the PPA for Ubuntu trusty looks like this:

![screenshot from 2017-04-15 21-26-44](https://cloud.githubusercontent.com/assets/2480569/25066296/5c3b316e-2222-11e7-9e50-1bdb70660b0b.png)

Would look so much less 90s if it was using this look:

![fontforge](https://cloud.githubusercontent.com/assets/2480569/25069972/b857cbce-2290-11e7-8806-3a70da35ce1b.png)

JoesCat:
Make FontForge look good, ...or make FontForge work good... ;-)
If you are pointing-out a problem - then there are 800+ functional issues that have priority :-(
but if you are providing a patch - it could be reviewed by several maintainers :-)

probonopd:
Oh, and Gtk+ or Qt should be used... relating to look good _and_ work good... :-)

JoesCat:
Sometimes, what initially appears to be a ten minute fix, leads into something more involved ;-)
You may want to post some pngs for feedback.
Personally, I like the retro look, but then again, we don't know what you suggest, maybe better.
Take your time. The code itself is interesting as well.

ahyangyi:
Wait, I actually prefer the old splash screen...

By the way, what made you feel the splash screen as 90s?

probonopd:
@ahyangyi while this is (also) a matter of personal preference, the application and the website have no resemblance of each other...

jmckean83:
Regardless of whether you like or dislike the look of the current splash screen (personally, I think it's horrible), there are some real issues with the way it displays -- at least on Windows. The text is completely jumbled, and the graphic part contains outdated copyright information. Is it perhaps even technically a misrepresentation to include copyright attributions to George Williams at this point in its development? Of course functionality is the priority, but being mindful of superficial, aesthetic concerns now and then is perhaps not a bad idea--especially if we want to keep FontForge alive in part by appealing to new users. Despite the old adage, a lot of people DO judge books by their covers.
![2017-04-17](https://cloud.githubusercontent.com/assets/24866207/25104129/265c49a4-238d-11e7-91d6-d551b5868015.png)

frank-trampe:
It's not solely a Windows issue. I think that the leading character of the git hash gets cropped on other platforms. I have a lot of other stuff on my to-do list, but I would like to see at least that fixed.

khaledhosny:
[...]

PR #3799, opened 2019-07-18
Add settings options: Splash DIsplay Time in Seconds

happyrobotstudio:
Providing a new settings option 

"Splash Screen Seconds"

To allow more complete user preference regaring splash time

### DIFF
--- a/fontforge/noprefs.c
+++ b/fontforge/noprefs.c
@@ -113,6 +113,7 @@ extern int use_freetype_to_rasterize_fv;	/* in bitmapchar.c */
 /*  UI preference files without loss of data */
 static char *xdefs_filename;
 static int splash=1;
+static int splashSeconds=7000;
 static int cv_auto_goto=1;
 static int OpenCharsInNewWindow=1;
 static float arrowAmount=1;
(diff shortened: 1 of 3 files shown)


happyrobotstudio:
SlpashScreenSeconds

jtanx:
It's a splash screen, this doesn't need to be configurable...

ctrlcctrlv:
Agreed with @jtanx, this shouldn't be merged.

happyrobotstudio:
> It's a splash screen, this doesn't need to be configurable...

Fo sho =P

Just stepping through as a new user .. maybe ... (power) user .. =P

First things I notice are an excessively long loading screen that looks like its an actual window, and then the non-rememberance of the homedir 

Just my 2c =P

happyrobotstudio:
Ok, ill fork privately =P

JoesCat:
For everyone quick to say "no", I think it may be worth re-reading #3055, for example seconds==0 can be same as "off"

happyrobotstudio:
Ack, keep your 7 second mystery window .. ill fork my own way, ive got ideas regarding path based font creation that wont get past =P

Enjoy!

Issue #3751, opened 2019-06-30
⟦User decompositions⟧ `.` in glyph name leads to build failure


ctrlcctrlv:
Imagine a glyph `_ra_kudlit.i` (-1). This glyph is supposed to be built from `_ra` (U+170D) and `_kudlit.i` (U+1712). Its user decomposition is `170D 1712`.

It fails to build with glyph name `_ra_kudlit.i` . (Case Ⓐ)
It builds fine with glyph name `_ra_kudliti`. (Case Ⓑ)

Why? For some reason...SCMakeBaseReference (fvcomposite.c:1691), called by SCBuildComposit (fvcomposite.c:2877), behaves differently for glyph names containing periods. In Ⓑ, it looks up `uni170D` via `SFGetChar`, returning a pointer to `_ra`. But in Ⓐ, it looks up a non-existent character&hellip;`uni170D.i`.

I'm not sure why this feature is supposed to be useful. Perhaps for building accented small caps&hellip;although I'm not really sure why this would help with that I think I should disable it if a user decomposition exists, but to know if that's right I have to know what it's for.

What do you guys think? If no responses for a week I'll submit a PR to disable it for user decompositions.

skef:
In my view this is a sub-part of #3713.

ctrlcctrlv:
Agreed&hellip;I'll PR a fix for this, either on its own or as part of the new matrix-based User decompositions.

PR #3802, opened 2019-07-19
Close fontforge/fontforge#3751 (User decompositions period issue)

ctrlcctrlv:
This closes a minor known issue in user decompositions that would be good to merge before the next release.

### DIFF
--- a/fontforge/fvcomposite.c
+++ b/fontforge/fvcomposite.c
@@ -1694,7 +1694,7 @@ static int SCMakeBaseReference(SplineChar *sc,SplineFont *sf,int layer,int ch, B
     char *dot;
     char buffer[200], namebuf[200];
 
-    if ( (dot = strchr(sc->name,'.'))!=NULL ) {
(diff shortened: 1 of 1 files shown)


jtanx:
I've mentioned this before, but can you please make your commit messages more descriptive? It way easier to read off the git log without having to refer back to github what exactly the context of the commit was.

ctrlcctrlv:
Sorry @jtanx I did remember your request. I was just thinking that when I squash and merge I can write the more detailed commit message then. Squashing doesn't put my original commit onto master, but rather the merge commit and my initial commit become one commit where I will put the more detailed message. If putting the better message in at the time of squash-merging is not OK with you in future I'll put it both in my branch's commit and the squash-merge commit. If you have no issues with the change I can squash-merge right now

jtanx:
I'm fine with the change, if you're going to update the description then that's good.

My preference is for the commit to have the description in place already, that makes it clear in the pull request what each commit is. Plus even if you did do a squash merge, you most likely won't even have to write anything new, and it means there's no change to forget about it before merging.

PR #3804, opened 2019-07-20
Speed up the MacOS build, switch to python 3

jtanx:
### Type of change
- **New feature**
- **Breaking change**


.travis.yml
```diff
@@ -106,25 +107,26 @@ install:
...
-      export PYTHONPATH=$PYTHONPATH:$PREFIX/lib/python2.7/site-packages
+      export HOMEBREW_NO_AUTO_UPDATE=1
       # Disable fc-cache on fontconfig install. Because it's slow.
       sed -i.bak '/fc-cache/d' "$(brew --prefix)/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/fontconfig.rb"
-      brew unlink python || true
-      brew install ./travis-scripts/fontforge.rb --only-dependencies || true
-      export HOMEBREW_NO_AUTO_UPDATE=1
-      # Pin pango to this version, because of issues, see #3343
-      brew unlink pango && brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/ad91e4df4a843764901811a965a17f650f34e2c1/Formula/pango.rb && brew link pango || true
       # 10 billion years later...
```
skef: Is the removed install not what takes 10 billion years?

jtanx: Yeah, it was actually with reference to all of the above homebrew steps.

I'm ranting now, but package management on mac is so bad, the only reason homebrew is used is because there isn't much choice otherwise. Like the majority of the speedup here is just in uninstalling and reinstalling homebrew, which takes 40 seconds, vs. upgrading it, which takes 5 minutes (!!!)

skef: I figured that was the likely answer.

PR #3805, opened 2019-07-20
travis: Fix python selection and osx build on master

jtanx:
There's multiple versions of Python installed on the linux build host. 

### Type of change
- **Bug fix**


jtanx:
Yep, the top block is linux, the other block is mac (start of the if statement here: https://github.com/fontforge/fontforge/pull/3805/files#diff-354f30a63fb0907d4ad57269548329e3L85)

Issue #3783, opened 2019-07-12
I should be able to copy and paste objects without the CharView tab zoom resetting


ctrlcctrlv:
Not sure if this is a feature or a bug, guess it depends on how you're using FontForge. I definitely experience it as a bug.

Let's say we're drawing an `m`&hellip;

![2019-07-12-191708_3840x2160_scrot](https://user-images.githubusercontent.com/838783/61124461-894e9d80-a4d9-11e9-959a-f968898a5926.png)

We copy and paste the second stroke, and the zoom resets.

![2019-07-12-191809_3840x2160_scrot](https://user-images.githubusercontent.com/838783/61124512-aa16f300-a4d9-11e9-8c17-4176182de313.png)

D'oh&hellip;

khaledhosny:
That was the fix in sortsmill https://github.com/khaledhosny/sortsmill/commit/18fee9bd1f3538376f61d486f662840b1ef4d2a9, feel free to backport.

PR #3813, opened 2019-07-22
Don't fit CharView to content on paste

ctrlcctrlv:
This closes fontforge/fontforge#3783.

Users know what they copied and know where in the canvas it will appear. Fitting the canvas to all splines on every paste makes no sense and is actually harmful, see the initial issue for one concrete test case as to why.

This is less extreme than https://github.com/khaledhosny/sortsmill/commit/18fee9bd1f3538376f61d486f662840b1ef4d2a9; I decided to maintain the CharView fit on import; it makes sense there.

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -8666,7 +8666,14 @@ static void _CVPaste(CharView *cv) {
 static void CVPaste(GWindow gw, struct gmenuitem *UNUSED(mi), GEvent *UNUSED(e)) {
     CharView *cv = (CharView *) GDrawGetUserData(gw);
     _CVPaste(cv);
-    CVInkscapeAdjust(cv);
+    // Before July 2019, all pastes would reset the view such that all splines
+    // were visible. This is wrong in most cases, e.g. tracing many letters
+    // from one manuscript page, or copying and pasting parts of a glyph.
+    //
+    // A more complete fix would be to make it so only selected points were put
+    // into view, but it was deemed not worth it at the time, and this comment
+    // was left instead. See Github issue №3783 and PR №3813 for more details.
+    // CVInkscapeAdjust(cv);
 }
 
 static void _CVMerge(CharView *cv, int elide) {


ctrlcctrlv:
It's certainly not my fault no one commented, I gave ample time for them to do so :-)

It seems like&mdash;correct me if I'm wrong&mdash;I could just add an `int selected` to the declaration of `SplineSetQuickBounds`...and then to `SplineCharLayerQuickBounds`. That would check whether or not each `SplinePoint` considered is `selected` before setting `b->{min,max}{y,x}`.

Deal?

skef:
> It's certainly not my fault no one commented, I gave ample time for them to do so :-)

Issues here don't get nearly as much critical thought as push requests. But to say more might constitute behavior inappropriate in a professional setting. 

Seems like your proposal should work consistently given the prior call to `CVClearSel()` (and paste contents that don't include splines won't add any selected points). So, yes, that sounds like a good solution. 

It's a shame that `SplineCharLayerQuickBounds` doesn't appear to do anything about anchor points but I guess that's life.

frank-trampe:
I try to stay on top of the new issues, but I did miss that one somehow.

This is yet another case where either easy option is right in some cases and wrong in others but the choice is not important enough for a preference. Given the two easy choices, I personally prefer the post-change behavior, and, given the low stakes and lack of opposition, I think that it's worth trying it.

I would in this case, though, comment the old line out with an additional comment as breadcrumbs in case we revisit for a more complete fix in the future. There is also a chance that we're missing a more obvious reason for the behavior such that we would need to switch it back.

ctrlcctrlv:
@frank-trampe Done, breadcrumbs left :-)

khaledhosny:
I don’t know any other application that changes the zoom on paste, certainly not Inkscape. What usually happen is that the content gets pasted in the center of the current view.

skef:
@khaledhosny: Probably true, but in most 2d design scenarios the underlying coordinate system is arbitrary or irrelevant. Not so with fonts. If you copied the splines of some glyph to serve as the basis of another it would be a huge pain in the ass if it wasn't pasted at the same baseline-relative height. Left/right location is less important but preserving that is a good guess. Therefore the fact that none of the content may be visible when pasting makes it more sensible to zoom out until at least some of it is.

PR #3817, opened 2019-07-23
Make PreferSpacingAccents false by default

ctrlcctrlv:
The assumption that most people will want spacing accents to be used when building combining characters is wrong; this feature, which many interpret as a bug, should be false by default.

See also:
https://github.com/fontforge/fontforge/pull/3774#issuecomment-514072799
\#3712
\#3708

### DIFF
--- a/fontforge/fvcomposite.c
+++ b/fontforge/fvcomposite.c
@@ -48,7 +48,7 @@
 
 int accent_offset = 6;
 int GraveAcuteCenterBottom = 1;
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
>I merged this without the preference adjustment. Let's just do it in another pull request. [...] As we are the only interested parties and the change seems positive and easily bypassed, **we can safely push that through once tests pass.**

https://github.com/fontforge/fontforge/pull/3774#issuecomment-514072799

PR #3821, opened 2019-07-24
travis: osx: link gettext

jtanx:
If it's not linked in, then configure doesn't find libintl.h and falls back to not providing translations... This dependency should just be mandatory.

### Type of change

- **Bug fix**


jtanx:
Success for this is where it tries to detect `libintl.h` and it succeeds.

frank-trampe:
@jtanx, I'm away from my Macintosh for a bit, so you're the only one of us in any position to validate this. If it works, let's merge it.

jtanx:
This is testing the travis build, for which I need it to push to bintray. That only happens on master, so merging.

PR #3808, opened 2019-07-20
Add new blank hr.po file. Supplements info in issue #3797

JoesCat:
hr.po file prepared for Croatian translation.
Needs attention @milotype if ok to use.

PR #3797, opened 2019-07-18
Update fontstyles.html with Croatian

milotype:
### Type of change
- **New feature**: Croatian translation added

PR #3825, opened 2019-07-25
osx: build both python 2 and 3

jtanx:
Sigh, see #3808

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,13 @@ matrix:
       env:
         - OSX_ONLY=true
         - BINTRAY_DEPLOY=true
+        - PYTHON=python3
+    # Temporary measure, expect this to be gone by the August 1 release
+    - os: osx
+      env:
+        - OSX_ONLY=true
+        - BINTRAY_DEPLOY=true
+        - PYTHON=python
 
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


ctrlcctrlv:
Apple has no qualms over leaving its users in the dust to buy new laptops, why should we?

Why should we continue to support OS X versions even Homebrew no longer cares about?

Why should we have to keep using Python 2 because of that?

This is a problem that should remain unsolved.

jtanx:
Yeah, I'm on the fence on this one. See what I wrote in #3797 for more details. I'm going to think about this overnight, I may end up just not doing this after all.

frank-trampe:
This isn't about doing an unmerited favor for Apple. This is about helping its victims. People who design typefaces can't always afford to replace their computers as often as Apple likes. Further, Apple's aggressive withdrawal of support within its operating systems for older software means that even those who can afford financially to upgrade cannot actually do so.

It all seems absurd to us as users of non-abusive operating systems, but a lot of people, particularly type designers, play Tina Turner to Apple's Ike and can't see any world outside of that relationship. They are more deserving of pity than scorn.

jtanx:
So I've thought about this a little bit more overnight, and I'm ok with this until the next release, on the proviso that it's not part of the 'official' release, and only ever resides on bintray.

If we had someone actively developing on mac, then this might be a different story. But we don't, so this isn't sustainable. Given the lack of support for El Capitan in our build infrastructure, I'm not willing to spend any further resources doing things like this into the future.

frank-trampe:
Fine with me. You're already doing more to support an operating system you don't use than anybody can ask.

skef:
> People who design typefaces can't always afford to replace their computers as often as Apple likes. Further, Apple's aggressive withdrawal of support within its operating systems for older software means that even those who can afford financially to upgrade cannot actually do so.

I agree that Apple is overly aggressive about this but they're not *that* extreme. The hardware that cannot be (officially) upgraded past El Capitan was manufactured before 2010. That's quite a bit of time.

I would say the general convention is that once a system has reached stage where general purpose computing is at least questionable -- which would include the lack of security patches -- people are free to use it with software of its era but shouldn't expect to move further into the future with it. The issue isn't with FontForge running on this machine at all, its with the latest versions of FontForge doing so. And we're contemplating this change because Homebrew dropped support, so the Mac open source community is almost as aggressive.

PR #3826, opened 2019-07-25
Fix wordlist error w/glyph name matching ^uni & \.

### LINKED ISSUE #3820: Non-Latin alt glyph names in glyph window text unescaped with text typed after
When the preview text in Glyph window contains a non-Latin glyph (Cyrillic, Greek, or even some African Latin) with a period in the name (typically `uni0430.alt` etc.), and other text is typed after that glyph, it is converted into a sequence of the default glyph followed by the literal period glyph and the whole rest typed out, i.e. `/uni0430.alt` -> `/uni0430/period/a/l/t`. This is prevented by typing out slashes to escape the added letters so that they don't mess up the existing alt glyph, but this only works until the next time the text input gets focused.

Video recording of the issue: https://recordit.co/Nnji5v2NMd
FontForge version: 13-April-2019
OS: Windows 10

This is a severe bug as it interferes with testing alt glyphs (smallcaps, etc.) which in itself isn't easy or comfortable in FontForge. It might also happen with Latin glyphs, but I couldn't reproduce it at all times unlike with Cyrillic.

Steps to reproduce:

  1. Use the attached .sfd file [test.zip](https://github.com/fontforge/fontforge/files/3426655/test.zip), or
    1a. Create a base glyph e.g. `uni0430` and an alt glyph with period in the name, e.g. `uni0430.alt`
  2. Open the glyph with the period for edition by double-clicking
  3. Type at least two characters at the end of the input, or
    3a. Type one character, switch focus back and forth by clicking the drawing board and the input again
  4. Expect the text to be malformed

- [x] Screenshot

ctrlcctrlv:
This fixes a bug wherein the word list input in CharView was unable to understand input like `[/uni043b.alt]a`, fudging it into `[/uni043B]/periodalta`. This is because the parser was simply giving up after trying to find "uni043b.alta", instead of searching for "uni043b.alt", "uni043b.al" etc.

The meat and potatoes of this commit is the added for loop. The rest is just whitespace fixes to make this part of the code actually readable, the `if` statement under it was all messed up.

This closes #3820

### Type of change
- **Bug fix**

### DIFF
--- a/fontforgeexe/wordlistparser.c
+++ b/fontforgeexe/wordlistparser.c
@@ -289,16 +289,17 @@ u_WordlistEscapedInputStringToRealString_readGlyphName(
 	    {
 		unichar_t* endptr = 0;
 		long unicodepoint = u_strtoul( glyphname+3, &endptr, 16 );
-	            SplineChar* tmp = 0;
+		char c;
+		SplineChar* tmp = 0;
 		TRACE("uni prefix, codepoint: %ld\n", unicodepoint );
 		sc = SFGetChar( sf, unicodepoint, 0 );
 
-		for (int i = u_strlen(glyphname); i>=0; i--) {
-		    unichar_t* substr = calloc(i+1, sizeof(unichar_t));
-		    u_strncpy(substr, glyphname, i);
-		    tmp = SFGetChar( sf, -1, u_to_c(substr) );
-		    TRACE("looking for subst. char: %s\n", u_to_c(substr));
-		    free(substr);
+		for (int i = u_strlen(glyphname); i>0; i--) {
+		    c = glyphname[i+1];
+		    glyphname[i+1] = 0;
+		    tmp = SFGetChar( sf, -1, u_to_c(glyphname) );
+		    TRACE("looking for subst. char: %s\n", u_to_c(glyphname));
+		    glyphname[i+1] = c;
 		    if (tmp != NULL) {
 			TRACE("have subst. char: %s\n", tmp->name ); break;
 		    }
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
A potential problem with this is if the font contains a char named `uni043b.alta`, and the user enters `[/uni043b.alt]a`, that would be autocorrected to `[/uni043b.alta]`. This seems unavoidable given the way the parser works, and is very rare regardless...in some cases the user may even want that? It seems like we need a way to distinguish `[/uni043b.alt]a` and `[/uni043b.alta]` as user input, but given 239-247 of wordlistparser.c that is not possible as we aren't given the brackets...

fontforgeexe/wordlistparser.c
```diff
@@ -289,25 +289,35 @@ u_WordlistEscapedInputStringToRealString_readGlyphName(
...
+	            SplineChar* tmp = 0;
 		TRACE("uni prefix, codepoint: %ld\n", unicodepoint );
 		sc = SFGetChar( sf, unicodepoint, 0 );
-                if ((tmp = SFGetChar( sf, -1, u_to_c(glyphname) ))) {
-		    TRACE("have subst. char: %s\n", tmp->name );
-                    sc = tmp;
-                } else {
+
+		for (int i = u_strlen(glyphname); i>=0; i--) {
+		    unichar_t* substr = calloc(i+1, sizeof(unichar_t));
```
frank-trampe: It seems excessive to allocate this every time.

frank-trampe:
A comment in the code explaining what is happening might be worthwhile. It's not difficult to figure it out in this case, but tracing a problem takes a lot longer when the logic incorporates fifty such mystery blobs.

ctrlcctrlv:
FYI:

https://sourceforge.net/p/fontforge/mailman/message/36727023/

Sorry

skef:
What about this?

frank-trampe:
Looks good to me.

PR #3828, opened 2019-07-26
Menus 2019

mamoru9999:
### Type of change

- **Bug fix**
- **New feature**
- **Breaking change**
- **Non-breaking change**


skef:
While I have some admiration for the anti-rhetoric-as-rhetoric approach, any revival of this branch has to overcome the concerns of #3638. (And bringing the code up to date wouldn't hurt.)

frank-trampe:
@mamoru9999, we have not forgotten about the new menus. As @skef notes, there are issues blocking a merge of the code as-is. The biggest is that some of us do not consider the new menus an improvement as they stand right now, requiring some means to switch between them and the old ones until they mature. Given that the new menus are likely to require a year or so of experience-driven adjustment to reach potential, I think that the ideal approach would be to make the menu arrangement fully run-time-configurable (via an XML file or something). Unfortunately, this is non-trivial (per the discussion on the other issue), and more urgent issues keep drawing our attention, so it may be a bit before there is any movement on this front.

skef:
> some of us do not consider the new menus an improvement as they stand right now

@frank-trampe There are understandable issues with being familiar with something and not wanting it to change, "muscle memory", and so forth. You seem to be making a stronger claim here that this new layout is, in the eyes of some users including you, worse than the existing one setting aside those understandable issues. In other words "worse on the merits". I find that hard to understand, even to believe. Can you defend the existing layout over the proposed one? Can you produce these other users who will defend it on the merits?

frank-trampe:
I was certainly not saying that in the comment you quoted, and to be
honest, I have not used the new menu code in a long time. My point
elsewhere in my previous comment is that the current arrangement, for its
faults, has reached a stable point from years of progressively finer
adjustment. Any new arrangement is unlikely to have that level of polish,
and we need a way to get it there without pulling all users through the
process.

skef:
If that's the *only* reason, and there isn't a side requirement of keeping the existing menu layout as a permanent option, an optional beta program is sufficient to address it.

frank-trampe:
@skef, to be clear, that's not my only objection, but addressing it essentially takes care of the others.

I think that we have different notions of the length of time necessary to get the new menus to an acceptable state. Given the number of things that I found clunky the first time I tested, I was expecting this gradually to build momentum over a year or so. And, yes, we could just have a binary runtime option as a fallback, but evolution and polishing go a lot slower when every change requires engineering expertise.

I feel like my prior proposal to keep hard-coded menu items but to allow rearranging them via runtime configuration would achieve this without unnecessary complexity. If enough people prefer the new menu concept to make it worth my while, I can sketch out a code concept for that. It's a lot of work, but I think that it's our best option if we're to do this.

skef:
> If enough people prefer the new menu concept to make it worth my while, I can sketch out a code concept for that. It's a lot of work, but I think that it's our best option if we're to do this.

I don't understand what this means. And regardless I don't understand your argument about the schedule. 

If you are right that the new menu layout needs to be refined, and that that process will take a significant amount of time, then waiting to work on that until we've done more technical work gets things exactly wrong. If there's going to be a better default then we should make a branch and invite people into a beta of the layout now, so that when whatever development that is needed is finished the layout can be refined. 

I already did much more work on this project than will be required for further tweaks. It sounds to me 
("the first time I tested") that you haven't yet tried the new branch. I would be fine doing the work iterating as part of a beta. So that's not a reason for not going forward. 

I'm tired of hearing that there is some mass of work that needs to be done that either only you can do or that for some reason only you should do that is holding this up.

frank-trampe:
I'm certainly not claiming to be the only one capable of accomplishing the task in question. I'm just not in a position to ask other people to implement my proposals.

If you're willing to manage the beta process, perhaps just having a binary runtime option is sufficient.

ctrlcctrlv:
I'm closing this, as the person who opened it hasn't responded since having opened it, and it would be more logical to reopen other closed PRs than to merge this PR. 

Furthermore, @mamoru9999 opened a PR from `menus-2019`, which is a branch on fontforge/fontforge, which @mamoru9999 can't even push to, so they will _never_ be able to make the changes requested, much less other changes to get it past Frank (and me, if we're honest, as I'm not too keen on new menus either).

skef:
>  much less other changes to get it past Frank (and me, if we're honest, as I'm not too keen on new menus either

Another reminder that "as an existing user I don't like change" is not a good argument. The current menu layout is opaque and nonsensical. The layout worked out by @davelab6 is not perfect but much better. @davelab6 routinely works with new users of FontForge and is a good resource on this sort of thing.

ctrlcctrlv:
That may very well be so, yet I still think my reason for closing was valid.

frank-trampe:
We do have a path to accomplishing a universally agreeable end (optional use of new menus with the expectation that they would evolve over some period of time). The only remaining issue is staffing what doesn't seem to be a terribly high priority for anybody in this thread.

PR #3829, opened 2019-07-27
Make getShareDir work when invoked from the current directory

jtanx:
e.g. so executing ./fontforge works

When you execute`./fontforge`, `GResourceProgramDir` retains the `.` at the end of the path, so `getShareDir` was not going up one directory as it should have. 

There are so many things wrong with the fsys routines (i.e. don't have any paths longer than 1000 bytes), but I'll leave that for another time...

### Type of change

- **Bug fix**

### DIFF
--- a/gutils/fsys.c
+++ b/gutils/fsys.c
@@ -843,82 +843,58 @@ char *getShareDir(void) {
     static char *sharedir=NULL;
     static int set=false;
     char *pt;
-    int len;
 
     if ( set )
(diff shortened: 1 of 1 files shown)


jtanx:
Hmm, while this works, I thought that calling GFileGetAbsoluteName would normalise the path; it doesn't since it's already an absolute path. I'll go with just making GFileGetAbsoluteName  remove the trailing `/.` instead.

PR #3831, opened 2019-07-27
GFileGetAbsoluteName: Remove trailing /.

jtanx:
See #3829, this is to make executing FontForge from the current directory correctly detect the share directory.

e.g. this is to make GFileGetAbsoluteName return  `/path/to/bin` instead of `/path/to/bin/.` 

### Type of change
- **Bug fix**

### DIFF
--- a/gutils/fsys.c
+++ b/gutils/fsys.c
@@ -227,6 +227,9 @@ char *GFileGetAbsoluteName(const char *name, char *result, size_t rsiz) {
 		savestrcpy(spt,spt+1); /*  skipped past the :// of the machine name) */
 	    else if ( pt==spt+1 && spt[0]=='.' && *pt=='/' ) {	/* Noop */
 		savestrcpy(spt,spt+2);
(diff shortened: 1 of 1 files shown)


jtanx:
At present, I want to touch these fsys functions as little as possible. Given that this wouldn't change the behaviour if the path didn't end with `./`, e.g. if it was `/path/to/bin/` or `/path/to/bin` already, there should not be anything that relies on this leaving a trailing slash. 

Furthermore, this is breaking `getShareDir`, which assumes that there is no trailing slash. So I think it's fine...

There's actually a nice [glib](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-canonicalize-filename) function that will do most of what I want, but it's only in newer versions, so I don't want to deal with that old gem of trying to support older clients right now.

jtanx:
Actually, there is no trailing slash in this case, since it ends with `.`, which is  a no-op.

PR #3834, opened 2019-07-29
Croatian translations

milotype:
- GUI translation into Croatian
- Croatian font-styles update


ctrlcctrlv:
I tried these, they work. Not really much else to say - based on Google Translate the translator seems to have made a good faith effort to translate.

So I'll merge this :)

PR #3843, opened 2019-07-31
Use SplashScreen as pr_int to indicate splash delay in seconds

happyrobotstudio:
Allow the user to set the time in seconds that the Splash Screen (.. so called, that thing really needs to be de-coupled from other logic, its a combined everything ..) will last.

Let user know via hover message that setting to 0 is equivalent to turning Splash Screen off

### DIFF
--- a/fontforge/noprefs.c
+++ b/fontforge/noprefs.c
@@ -112,7 +112,7 @@ extern int use_freetype_to_rasterize_fv;	/* in bitmapchar.c */
 /* UI preferences which we don't use, but will preserve to so we can read/write */
 /*  UI preference files without loss of data */
 static char *xdefs_filename;
-static int splash=1;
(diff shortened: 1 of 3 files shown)


happyrobotstudio:
Duuude, nice.

Yeh the splash screen is so tightly coupled with other areas. Its a launcher and an about page.

It should be its own thing, and the other functionality kept in say, a hidden window ... ready to spring on unawares about page viewers.

For now, leaving it with a toolbar, though ugly, seems necessary .. having it with crisp edges sure would remove the confusion on start when the thing has a title bar ... its would be more 'splashscreen-esque'

For now your solution to simply give users more fine grained control over a bad splash situation seems good, good job bucko.

happyrobotstudio:
LOL, wrong repo.

PR #3852, opened 2019-08-02
 Warn users before discarding their unsaved scripts

### LINKED ISSUE #3846: FontForge should warn me if it has an unsaved script in memory
Yes, I know, I'm dumb, but I accidentally closed FontForge which had a script I was working on in &laquo;File&rarr;Execute Script&hellip;&raquo;. Needless to say, that script is gone now, in computer heaven. :cloud: :cry: 

FontForge should warn me if I've forgotten to export my script.

ctrlcctrlv:
This closes #3846.

I know, I could have written ten scripts or more in the time it took me to do this.

I hate losing my work. I had a pretty cool FontForge script, and I lost it on accident. I hate that. Yes I could do it again ten times in the time it took me to write this feature, but I'd rather never lose work than have to redo work once.

### DIFF
--- a/fontforgeexe/fontview.c
+++ b/fontforgeexe/fontview.c
@@ -544,7 +544,10 @@ static int AskScriptChanged() {
     buts[2] = _("_No");
     buts[3] = NULL;
     ret = gwwv_ask( _("Unsaved script"),(const char **) buts,0,2,_("You have an unsaved script in the «Execute Script» dialog. Do you intend to discard it?"));
-    if (ret == 1) warn_script_unsaved = false; SavePrefs(true);
+    if (ret == 1) {
+        warn_script_unsaved = false;
+        SavePrefs(true);
+    }
 return( ret );
 }
 
--- a/fontforgeexe/fontview.c
+++ b/fontforgeexe/fontview.c
@@ -534,6 +534,17 @@ static int AskChanged(SplineFont *sf) {
 return( ret );
 }
 
+static int AskScriptChanged() {
+    int ret;
+    char *buts[3];
+
+    buts[0] = _("_Yes");
+    buts[1] = _("_No");
+    buts[2] = NULL;
+    ret = gwwv_ask( _("Unsaved script"),(const char **) buts,0,1,_("You have an unsaved script in the «Execute Script» dialog. Do you intend to discard it?"));
+return( ret );
+}
+
 int _FVMenuGenerate(FontView *fv,int family) {
     FVFlattenAllBitmapSelections(fv);
 return( SFGenerateFont(fv->b.sf,fv->b.active_layer,family,fv->b.normal==NULL?fv->b.map:fv->b.normal) );
@@ -850,23 +861,26 @@ return( sf->changed );
 }
 
 static int _FVMenuClose(FontView *fv) {
-    int i;
+    int filesaved, scriptsaved;
     SplineFont *sf = fv->b.cidmaster?fv->b.cidmaster:fv->b.sf;
 
     if ( !SFCloseAllInstrs(fv->b.sf) )
 return( false );
 
+    if ( fv->script_unsaved ) scriptsaved = AskScriptChanged();
+
     if ( fv->b.nextsame!=NULL || fv->b.sf->fv!=&fv->b ) {
 	/* There's another view, can close this one with no problems */
     } else if ( SFAnyChanged(sf) ) {
-	i = AskChanged(fv->b.sf);
-	if ( i==2 )	/* Cancel */
-return( false );
-	if ( i==0 && !_FVMenuSave(fv))		/* Save */
-return(false);
-	else
-	    SFClearAutoSave(sf);		/* if they didn't save it, remove change record */
+	filesaved = AskChanged(fv->b.sf);
     }
+    if ( scriptsaved==1 ) /* No */
+return( false );
+	if ( filesaved==2 )	/* Cancel */
+return( false );
+	if ( filesaved==0 && !_FVMenuSave(fv))		/* Save */
+return(false);
+    SFClearAutoSave(sf);		/* if they didn't save it, remove change record */
     _FVCloseWindows(fv);
     if ( sf->filename!=NULL )
 	RecentFilesRemember(sf->filename);
--- a/fontforgeexe/fontview.c
+++ b/fontforgeexe/fontview.c
@@ -867,26 +867,26 @@ return( sf->changed );
 }
 
 static int _FVMenuClose(FontView *fv) {
-    int filesaved, scriptsaved;
+    int i;
     SplineFont *sf = fv->b.cidmaster?fv->b.cidmaster:fv->b.sf;
 
     if ( !SFCloseAllInstrs(fv->b.sf) )
 return( false );
 
-    if ( warn_script_unsaved && fv->script_unsaved ) scriptsaved = AskScriptChanged();
-
     if ( fv->b.nextsame!=NULL || fv->b.sf->fv!=&fv->b ) {
 	/* There's another view, can close this one with no problems */
+    } else if ( warn_script_unsaved && fv->script_unsaved && 
+                AskScriptChanged()==2 ) {
+        return false;
     } else if ( SFAnyChanged(sf) ) {
-	filesaved = AskChanged(fv->b.sf);
-    }
-    if ( scriptsaved==2 ) /* No */
+	i = AskChanged(fv->b.sf);
+	if ( i==2 )	/* Cancel */
 return( false );
-	if ( filesaved==2 )	/* Cancel */
-return( false );
-	if ( filesaved==0 && !_FVMenuSave(fv))		/* Save */
+	if ( i==0 && !_FVMenuSave(fv))		/* Save */
 return(false);
-    SFClearAutoSave(sf);		/* if they didn't save it, remove change record */
+	else
+	    SFClearAutoSave(sf);		/* if they didn't save it, remove change record */
+    }
     _FVCloseWindows(fv);
     if ( sf->filename!=NULL )
 	RecentFilesRemember(sf->filename);
--- a/fontforgeexe/fontview.c
+++ b/fontforgeexe/fontview.c
@@ -80,6 +80,7 @@ char *script_filenames[SCRIPT_MENU_MAX];
 extern int onlycopydisplayed, copymetadata, copyttfinstr, add_char_to_name_list;
 int home_char='A';
 int compact_font_on_open=0;
+bool warn_script_unsaved = true;
 int navigation_mask = 0;		/* Initialized in startui.c */
 
 static char *fv_fontnames = MONO_UI_FAMILIES;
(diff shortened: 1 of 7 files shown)


jtanx:
Some knee-jerk reactions, without looking too closely:

* Surely there's a way to do this without script hashing
* If we absolutely had to, don't add hashing to gutils, use [glib](https://developer.gnome.org/glib/stable/glib-Data-Checksums.htm)

ctrlcctrlv:
Other ways aren't as robust as this, and I definitely need hashing for my WIP SFDRefer branch. I'll look into glib in a half hour. I thought I made sure none of our dependencies could do this, but I may have missed something

jtanx:
Surely you can just set a boolean to flag that there is an unsaved script, I'm still not seeing what the purpose of hashing is.

ctrlcctrlv:
Because of the way gdraw works and because it's possible to save the script, then change it, then close the dialog. So saving it can't unset the flag and then closing it will set the flag.

I had to add a new `et_save` event just for this, as well.

We need to put our binary flag set (script changed) after `GDrawProcessOneEvent`, so that no matter how the window is closed, it will work. We can't put it in `et_close`, that's only for the close button, it won't work if OK/Cancel pressed&mdash;we can't make `SD_Cancel` post the event `et_close`, that causes an infinite loop.

jtanx:
It still sounds like the wrong approach to me. 

Let me clarify your workflow. You're saying that in the 'Execute script' window, you want to warn if there's an unsaved script (saving via a right click->Save in UTF8).

Supposing you had that `et_save` event (also not sure if this is appropriate, but let's go with this for now).

Why can't you just say extend `sd_data` with a dirty flag, which you set whenever the contents of the window changes, and unset whenever `et_save` is received? You can then propagate that as appropriate.

ctrlcctrlv:
Because after et_save is triggered the contents can change. I can et_save then write something else then close the window.

If I have no way of knowing whether or not the contents changed since the last save but before the last close, I either have to (a) always assume they have or (b) always assume they haven't, neither of which is always going to be correct.

jtanx:
> Because after et_save is triggered the contents can change. 

That's my point. An `et_save` occurs, you clear the flag. You receive an `et_charup` or whatever, changing the window contents. You set the flag again.

ctrlcctrlv:
No matter what the data type is I think it has to go on the FontView btw. `sd_data` isn't accessible by fontview.c, I think, and it needs it to do the warning when we're closing the window.

ctrlcctrlv:
Well, I didn't think of et_charup, I was putting the hash (or flag set or w/e) when the window was closed. I guess et_charup should  work :-)

jtanx:
Yeah, I suggested `sd_data` but you can propagate that as needed to the FontView. You'll probably need to look into `et_controlevent`, from a super quick skim.

ctrlcctrlv:
That event is confusingly named. Do you know its intended purpose? I thought it's maybe for like shift, alt, and ctl keys

jtanx:
Not sure, it's probably something to do with ggagdet interaction, since iirc technically it's the gtextfield that gets the key events.

ctrlcctrlv:
I'd still prefer hashes. What if they accidentally type something then backspace it?

ctrlcctrlv:
Maybe that's a rare event, but is it really that rare? To try something after saving, find it doesn't work, then delete it again and close the window?

ctrlcctrlv:
I'm happy to use libglib (makes it easier to use a stronger hash like SHA256) instead of the 1999 MD5 implementation I vendorized, but I'd like to keep hashes for this feature.

jtanx:
No, I just don't think hashing is required for this. Maybe for something else, but not this.

If you backspace, that still counts as the contents changing and you should be able to detect that. Some suggestions for prior work, see how the word list parser in the charview detects that it needs to update the charview, or even the same thing for the metrics window

ctrlcctrlv:
But the contents haven't changed. Consider the following flow of events:

* User opens window
* User types `print("H")` - flag gets set on first char `p` et_charup
* User executes
* User saves - et_save - flag unset
* User changes `print` to `prlint` - flag gets set on char `l` et_charup
* User executes
* User deletes the `l` - flag still set
* Contents unchanged, user closes window
* Erroneous error message about unsaved script

jtanx:
I don't think that matters, it's a heuristic. And that behaviour you just described is how all text editors work.

ctrlcctrlv:
I'm sure I could find one that uses hashes :-P

Besides, most text editors correctly understand "Undo" and keep a list of states, so it's more likely in a real text editor you'd undo, not delete the l

jtanx:
In this case using hashes is like throwing a hammer at the problem; it's just plain inefficient when you're already able to track if changes have been made.

> I'm sure I could find one that uses hashes :-P

Please do :)

ctrlcctrlv:
*checks Kate, vim, Geany...*
*nope...*

OK, you win, I'll remove the hashing.

ctrlcctrlv:
I think this is ready now :-)

fontforgeexe/scriptingdlg.c
```diff
@@ -189,6 +189,12 @@ return( true );
     
     if ( event->type==et_close ) {
 	SD_DoCancel( sd );
+    } else if ( event->type==et_charup ) {
+    char* contents = GGadgetGetTitle8(GWidgetGetControl(sd->gw,CID_Script));
```
jtanx: I can't take a closer look at the moment, but at a cursory glance this seems mighty inefficient. On every key press, you're reading the *full* contents of the script and performing at least one copy. You're then running over that again to find it's string length, *just* to see if it's unsaved or not.

Why can't you just always set `script_unsaved` to true? Also are you sure this is the right event to listen off?

jtanx: I skimmed `gtextfield.c` and you should be able to listen to `et_controlevent` with a subtype of `et_textchanged`. Then all you have to do is set `script_unsaved` to true in the case you receive that.

I reckon it makes more sense for that `et_save` event to also just be a subtype of the `et_controlevent` message.

fontforgeexe/fontview.c
```diff
@@ -534,6 +535,19 @@ static int AskChanged(SplineFont *sf) {
...
+static int AskScriptChanged() {
+    int ret;
+    char *buts[4];
+
+    buts[0] = _("_Yes");
+    buts[1] = _("Yes, and don't _remind me again");
+    buts[2] = _("_No");
+    buts[3] = NULL;
+    ret = gwwv_ask( _("Unsaved script"),(const char **) buts,0,2,_("You have an unsaved script in the «Execute Script» dialog. Do you intend to discard it?"));
+    if (ret == 1) warn_script_unsaved = false; SavePrefs(true);
```
jtanx: Don't put this all on one line and put this in brackets, you're otherwise always saving preferences

gdraw/gtextfield.c
```diff
@@ -916,6 +916,11 @@ static void GTextFieldSave(GTextField *gt,int utf8) {
     FILE *file;
     unichar_t *pt;
 
+    GEvent e;
+    e.type = et_save;
+    e.w = gt->g.base;
+    GDrawPostEvent(&e);
```
jtanx: Move this to the very end of the function. Also look at how `GTextFieldChanged` implements the send, and copy that (making et_save a subtype of et_controlevent)

fontforgeexe/scriptingdlg.c
ctrlcctrlv: I have to know if there's text in the field, otherwise if you write a script then delete it FontForge still thinks there's one there.

ctrlcctrlv:
Hey Jeremy, sorry, I've had a hard time focusing due to the recent mass shooting in Texas, I unfortunately have the crappy position of being the guy people email whenever 8chan is in the news, even though I haven't had anything to do with it for four years in April. I should be able to get to clear up your other comments in the next few hours, and if not then, then tomorrow.

fontforgeexe/scriptingdlg.c
jtanx: If you really want to avoid that, I can see a couple of options, extend the `et_textchanged` -> `tf_changed` event to indicate e.g. if the field is empty, or add a `GTextFieldIsEmtpy()` which does the same thing, e.g.

```c
bool GTextFieldIsEmpty(GTextField *gt) {
    return gt->text == NULL || *gt->text == '\0';
}
```

or

```c
		struct {
		    int from_pulldown;	/* -1 normally, else index into pulldown list */
		    int field_empty : 1;
		} tf_changed;

...

// In GTextFieldChanged
 e.u.control.u.tf_changed.field_empty = gt->text == NULL || *gt->text == '\0';
```

inc/gdraw.h
```diff
@@ -112,15 +112,15 @@ enum event_type { et_noevent = -1, et_char, et_charup, // is 1
...
+		  et_controlevent, et_user };
 
 enum visibility_state { vs_unobscured, vs_partially, vs_obscured };
 
 enum et_subtype { et_buttonpress, et_buttonactivate, et_radiochanged,
 		  et_listselected, et_listdoubleclick,
 		  et_scrollbarchange,
 		  et_textchanged, et_textfocuschanged,
-		  et_lastsubtype };
+		  et_lastsubtype, et_save };
```
jtanx: swap the order of this, `et_lastsubtype` is a sentinel value

fontforgeexe/scriptingdlg.c
```diff
@@ -190,9 +190,7 @@ return( true );
     if ( event->type==et_close ) {
 	SD_DoCancel( sd );
     } else if ( event->type==et_charup ) {
-    char* contents = GGadgetGetTitle8(GWidgetGetControl(sd->gw,CID_Script));
-    sd->fv->script_unsaved = (bool)strlen(contents);
-    free(contents);
+    sd->fv->script_unsaved = !GTextFieldIsEmpty(GWidgetGetControl(sd->gw,CID_Script));
     } else if ( event->type==et_save ) {
```
jtanx: surely this would need changing as its now a subtype of `et_controlevent`. Also what happened to listening for the text changed event? There are other ways it can change (e.g. right click paste)

ctrlcctrlv:
Bump

fontforgeexe/fontview.c
```diff
@@ -850,23 +867,26 @@ return( sf->changed );
...
 return( false );
 
+    if ( warn_script_unsaved && fv->script_unsaved ) scriptsaved = AskScriptChanged();
+
     if ( fv->b.nextsame!=NULL || fv->b.sf->fv!=&fv->b ) {
 	/* There's another view, can close this one with no problems */
     } else if ( SFAnyChanged(sf) ) {
-	i = AskChanged(fv->b.sf);
-	if ( i==2 )	/* Cancel */
+	filesaved = AskChanged(fv->b.sf);
```
jtanx: Why bother changing all this logic vs. just returning on line 876 if AskScriptChanged returns no?

ctrlcctrlv: This way if both the file and the script are unsaved, both get warned about. Perhaps they would both get warned about eventually, but if you have to return to FontForge, best for user to do everything forgotten at once. I also think this way makes the order of execution and asking much clearer, and the new variable names are better.

jtanx: Personally I'd find this behaviour surprising and counter-intuitive; if I've already been prompted to cancel closing the window and I cancel it, and then get asked *again* if I want to cancel, by my regard it's as if it's ignored what I chose earlier. I think it'd be much more 'normal' if it exits as soon as it knows it can.

There's also a problem with your changed logic. Supposing SFAnyChanged is false, `filesaved` will be uninitialised, which means you'll be comparing against an uninitialised value. Also I don't know if it matters, but you're now always calling `SFClearAutoSave` even if `SFAnyChanged` is false.

ctrlcctrlv: OK, no problem, I don't really have a strong opinion on this, so just reverted most of the changes to this function in https://github.com/fontforge/fontforge/pull/3852/commits/0a5629ac1f9f165e30e873a09f3a7dc6f98a3916

PR #3854, opened 2019-08-04
Misc fixups

jtanx:
Some miscellaneous fixups 

### Type of change

- **Bug fix**

### DIFF
--- a/doc/synch_web.py
+++ b/doc/synch_web.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 
 from __future__ import print_function
+from io import open
 from shutil import copy2
 import argparse
 import os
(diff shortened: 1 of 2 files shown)


doc/synch_web.py
```diff
@@ -51,12 +52,12 @@ def body_updater(match):
 
     mkdir(destination + root)
     dest_path = os.path.join(destination + root, os.path.basename(path))
-    with open(path, 'rb') as fpi:
```
frank-trampe: Are you sure that we don't want these in binary mode?

jtanx: Yes, all source text files in that docs folder are now utf-8, otherwise it's a PITA to work in binary mode in Python3

fontforgeexe/fontlint.pe
```diff
@@ -1,4 +1,3 @@
-#!/usr/local/bin/fontforge
```
frank-trampe: I've confirmed that compilation adds this the fontlint file to be installed.

doc/synch_web.py
frank-trampe: Does it work on Windows with the existing line endings?

jtanx: Well it writes out CRLF line endings, but that's generally fine depending on git settings. I've changed it anyway to just always write out LF line endings

frank-trampe: Ah, okay.

frank-trampe:
I guess we need to let the tests finish, but this all looks good to me.

Issue #3823, opened 2019-07-25
C++ keywords should not be in source


ctrlcctrlv:
```
In file included from geom.cpp:18:0:
splinefont.h:660:18: error: expected identifier before ‘;’ token
struct fpc class;
           ^
splinefont.h:660:18: error: multiple types in one declaration
splinefont.h:1829:18: error: expected unqualified-id before ‘new’
unsigned int new: 1; /* A new and unsaved font */
             ^
splinefont.h:1855:20: error: expected unqualified-id before ‘private’
struct psdict private; / read in from type1 file or provided by user */
              ^
splinefont.h:1855:19: error: expected ‘;’ at end of member declaration
struct psdict private; / read in from type1 file or provided by user */
              ^
splinefont.h:1855:27: error: expected ‘:’ before ‘;’ token
struct psdict private; / read in from type1 file or provided by user */
^
splinefont.h:2352:68: error: expected ‘,’ or ‘...’ before ‘private’
extern int SFPrivateGuess(SplineFont *sf,int layer, struct psdict *private,
                                                                   ^
```

@repu1sion, who is helping me integrate Lib2Geom into FontForge, discovered this.

Sane C compilers, like GCC, will *not* allow you to link C++ code with C code that contains C++ keywords. Even if it might be _technically_ possible, if you need to use the variables named after keywords in C++ code, you won't be able to.

There are many reasons you'd want to link C++ code with FontForge's code beyond my Lib2Geom integration. Therefore, I instructed @repu1sion to make a commit which will change all instances of C++ keywords to underscored versions. So `private` will become `private_`.

I know this might be contentious, so am making this issue so we can discuss it; perhaps someone knows a better way to use FontForge types from C++.

ctrlcctrlv:
(@repu1sion commented to me that we could also just add `#ifdef __cplusplus` everywhere this happens, but that seems way, way worse to me than just changing the variable/member names to C++-friendly ones.)

jtanx:
Ideally all headers should have the __cplusplus guards. But yeah, avoiding C++ keywords is also nice. 

> Sane C compilers, like GCC, will not allow you to link C++ code with C code that contains C++ keywords. 

Technicality, but this doesn't sound right, do you mean it won't allow you to *include* headers that haven't been properly set up? I don't see how linking could/would be affected.

ctrlcctrlv:
[...]

PR #3855, opened 2019-08-04
CMake build system

jtanx:
This replaces the autotools build system with CMake. Details on usage may be found [here on the wiki](https://github.com/fontforge/fontforge/wiki/CMake-guide-for-FontForge).

The minimum requirement is CMake 3.5. It only supports Python 3.

It is more or less feature complete, with a couple of caveats, so I think it's about time to open this up to others to try, and also for feedback on it.

The only major exclusion compared to autotools is it does not install headers and pkgconfig details. While it is certainly *possible* to mimic the old behaviour, I'm not seeing much reason to make libfontforge something that can be developed against. As seen in #3823, FontForge isn't well designed to be used in external projects, and we provide no stability guarantees over what is effectively internal interfaces. I doubt that we even export all required headers at the moment.

I have confirmed it builds on CI. It has more than halved the build time for Windows, and there are substantial benefits on Linux and MacOS too. It's now a matter of diminishing returns, as the time to install dependencies becomes more a more significant portion than building FontForge itself.

Somewhat more importantly for me, incremental builds are now super quick, which makes development just so much more enjoyable.

The only other part that is missing is fixing up the Deb/RPM packaging that we bundle. I think only the Debian packaging is still relevant, since that's all we use for the PPA. But as I have no insight into how the PPA functions, or how the packages are built, it's a bit hard for me to make the appropriate changes - @frank-trampe I'd appreciate your help with this.

### Type of change
- **New feature**
- **Breaking change**

Related issues: #3512 

Edit: On the diffstat, around 50% of the added lines is just backported modules copied from CMake 3.14.


JoesCat:
Hi @jtanx,
The wiki has a syntax error and needs to be corrected to:
```c
git clone https://github.com/fontforge/fontforge; cd fontforge; mkdir build; cd build
```
'&&' actually means fork and run in parallel while ';' means run sequentially afterwards.
DOS was unable to run more than one program at a time, so this was a good kludge to fake '&&' behaviour, and windows inherited this behaviour for backwards compatibility reasons.

Next, this is an error which reflects [packaging](https://github.com/fontforge/libspiro/pull/20)
```c
# Sigh, the old build system checks for the non-existant LibSpiroGetVersion
```
You'll want to add code to differentiate the recent 20190731 release from the older versions of libspiro that don't have it. FontForge has had code that uses this instruction since 2015.
It wasn't added to Libspiro (on purpose) until now because I wanted to add some pretty big changes like CRA bug fix first before the tag release point. LibSpiroGetVersion() coincides with the tag point, and therefore is a marker for operating systems that have/had faulty packaging, or if you want to use packaging as your go-to method, then use packaging instead (...but watch out for the BSDs and older LTS like maybe Redhat, Centos, maybe others).

jtanx:
> &&' 

I don't know what interpreter you're using, but that's plain wrong with bash and cmd. Please read [this](https://unix.stackexchange.com/questions/187145/whats-the-difference-between-semicolon-and-double-ampersand).

> You'll want to add code to differentiate the recent 20190731 release from the older versions of libspiro that don't have it

See, this would be a non-issue if the version was just added to the header, as I previously mentioned, and which could have easily been done before 20190731 was cut. Now I have to go and add yet more of these dumb feature detection methods.

JoesCat:
My error, you're right, I was thinking '&' which had to be fixed elsewhere for something else.

> See, this would be a non-issue if the version was just added to the header, as I previously mentioned, and which could have easily been done before 20190731 was cut. Now I have to go and add yet more of these dumb feature detection methods.

At this point, version control is all in one place : at the top of configure.ac
If you are testing for a package version#, it might work, depending on which os and version.

libuninameslist has version numbers in the header - let's see your improvement.

jtanx:
> At this point, version control is all in one place : at the top of configure.ac
> If you are testing for a package version#, it might work, depending on which os and version.

That's moot, because that's not accessible from the header.

> libuninameslist has version numbers in the header - let's see your improvement.

I don't understand why it doesn't just have

```
#define LIBUNINAMESLIST_VERSION 2019xxx
```

But anyway, *assuming* `LIBUNINAMESLIST_MAJOR`/`LIBUNINAMESLIST_MINOR` is kept up to date, we can get rid of all [this](https://github.com/fontforge/fontforge/blob/master/gutils/unicodelibinfo.c#L396) [garbage](https://github.com/fontforge/fontforge/blob/master/gutils/unicodelibinfo.c#L310) and just check `LIBUNINAMESLIST_MAJOR/MINOR` directly.

Of course, that doesn't help with older versions, which was why it should have been included in the last libspiro release.

Looking at that code, we should probably clean it up anyway and just drop support for older versions of libspiro/libuninameslist.

JoesCat:
I was looking forward to your solution using headers or packaging, etc. with libuninameslist

I guess I'll have to accept you at least removed the inflammatory comments.

khaledhosny:
I suggest folding libuninameslist into FontForge, dropping support for external library and simplifying the code. The library is nit used by anything else, so all this complexity is not warranted.

I’d go even further and suggest doing this for libspiro, but it seems to have at least one user other than FontForge, so that might be a controversial decision.

khaledhosny:
> I suggest folding libuninameslist into FontForge, dropping support for external library and simplifying the code. The library is nit used by anything else, so all this complexity is not warranted.

I’m working on this.

khaledhosny:
https://github.com/fontforge/fontforge/pull/3861

jtanx:
Yeah, that's actually not a bad idea, nice.

jtanx:
Are there any other questions/concerns/comments about this? Otherwise I'd like to look at getting this in soon.

ctrlcctrlv:
I tried it today while working on other stuff and actually quite like it. So, no negative comment, only a positive one in support :-)

skef:
I did a code review that wasn't conclusive either way, which basically means I didn't see any problems but wasn't confident enough to offer an assessment.

Given that this is build infrastructure it seems to me that there are two conventional ways of going:

1. Have some folks volunteer to use it for a while as a side project, periodically pulling over changes from the master branch.
2. Switch the master branch to it and shake out any issues in place. 

I'm happy to "vote" for option 2 in this case.

jtanx:
Alright, option 2 works for me too. If there really are no further comments on this, then I'll be looking at merging this in a day or two.

@frank-trampe, still waiting for details on how that ppa works, so that Packaging can be updated.

ctrlcctrlv:
Great! A bonus of this is that I don't think my `libtool` command line is needed anymore. The system libfontforge.so no longer seems to override&mdash;finally! So that can probably be removed from `CONTRIBUTING.md`.

This might be why:

```bash
[fred@pc fontforge]$ ls -alh bin/fontforge # With cmake
-rwxr-xr-x 1 fred fred 15M Aug 15 16:21 bin/fontforge
[fred@pc fontforge]$ ls -alh `which fontforge` # Autotools build
-rwxr-xr-x 1 root root 57K Aug  9 19:51 /usr/local/bin/fontforge
```

jtanx:
> A bonus of this is that I don't think my libtool command line is needed anymore. The system libfontforge.so no longer seems to override—finally!

Yeah, this is because cmake actually sets the RPATH properly (and changes it again when it gets installed). So it's always looking for the right dll when you invoke it from the build folder

ctrlcctrlv:
(My `libtool` command still works, I just don't think it's necessary anymore. It might be a good precaution still, though, as I did not test thoroughly. As a reminder, it's:

```bash
#!/bin/sh
DIR=bin # Change to `fontforgeexe` for autotools builds
libtool --mode=execute gdb -ex "run ${1@Q}" $DIR/fontforge
```

ctrlcctrlv:
Sorry, commented before seeing yours! That makes sense.

madig:
I just found out about the switch and I think it's *great*! Thank you for this! I always cheer when autotools are removed from something.

PR #3858, opened 2019-08-06
appdata: remove spaces to comply with xml syntax

muelli:
The attribute had an illegal whitespace. appstream-glib validate complained:

org.fontforge.FontForge.metainfo.xml: failed to parse org.fontforge.FontForge.metainfo.xml: Error on line 56 char 15: Odd character ?l?, expected a ?=? after attribute name ?xml:? of element ?p?

Now it does not complain about the syntax anymore, but instead about the description:
org.fontforge.FontForge.metainfo.xml: FAILED:
? style-invalid         : Not enough <p> tags for a good description [0/1]
Validation of files failed

### Type of change

- **Bug fix**
- **New feature**
- **Breaking change**
- **Non-breaking change**


JoesCat:
Would removing the extra spaces help?
```c
xml: lang = "hr"
xml:lang="hr"
```

muelli:
> Would removing the extra spaces help?

those seem to be legal. Validation passes with the proposed change. But it'd arguably more orthodox to also remove those spaces.

PR #3859, opened 2019-08-06
appdata: add name and summary to appdata file

muelli:
It seems a name and a summary is required in order to pass validation.
This is copy and pasted from the metainfo file, which I think is highly redundant.

### Type of change

- **Bug fix**
- **New feature**
- **Breaking change**
- **Non-breaking change**

### DIFF
--- a/desktop/org.fontforge.FontForge.appdata.xml
+++ b/desktop/org.fontforge.FontForge.appdata.xml
@@ -2,6 +2,20 @@
 <!-- Copyright 2014 Joe Da Silva <digital@joescat.com> -->
 <component type="desktop">
   <id>org.fontforge.FontForge</id>
+  <name>FontForge</name>
+  <name xml:lang="zh-TW">=字型鍛造廠</name>
+  <summary>An outline font editor</summary>
+  <summary xml:lang="de">Ein Font-Editor</summary>
+  <summary xml:lang="eo">Tipar dosiero-redaktilo</summary>
+  <summary xml:lang="es">Un editor de fuentes</summary>
+  <summary xml:lang="fr">Un constructeur des polices</summary>
+  <summary xml:lang="hr">Uređivač za fontove</summary>
+  <summary xml:lang="it">Un profilo editor di font</summary>
+  <summary xml:lang="pt">Um editor de fontes</summary>
+  <summary xml:lang="pl_PL">Edytor fontów</summary>
+  <summary xml:lang="ru">Редактор шрифтов</summary>
+  <summary xml:lang="uk">Редактор шрифтів</summary>
+  <summary xml:lang="zh-TW">FontForge 描邊字型編輯器</summary>
   <launchable type="desktop-id">org.fontforge.FontForge.desktop</launchable>
   <metadata_license>CC0-1.0</metadata_license>
   <description>


ctrlcctrlv:
The strings seem to be shared with the .desktop file. Maybe a script should generate this file instead because we're likely to forget to update it as new translations are added.

muelli:
hm. yeah, I would expect no redundancy to be required in first place.
Reading https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps I get the impression that `name` and `summary` are indeed not required if a `.desktop` file with that information exists.

> 
> name
>     While this tag is always requited for a generic component, for an application metainfo file it is not necessary, but only recommended. You can omit this tag if you want the software center to have the same strings as defined in the Name field of the .desktop file. In that case, one <launchable/> tag must be present.
>     If no name tag and no Name field is present, the metadata is considered invalid and might be ignored by the AppStream generator. 
> summary
>     While this tag is always requited for a generic component, for a desktop-application metainfo file it is only essential if the accompanying .desktop file does not have a Comment= field. Data from a .desktop file is only taken into consideration if one <launchable/> tag is present. If the metainfo file has a summary, it wil override the value found in the Comment field of the .desktop file. 
> 

Maybe this appdata file identifies as generic component rather than an application?
The documentation talks about `desktop-application` whereas this appdata file has the `desktop` type. Maybe that's the problem. hm.

muelli:
https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html says

>  The desktop-application component type is the same as the desktop component type - desktop is the older type identifier for desktop-applications and should not be used for new metainfo files, unless compatibility with very old AppStream tools (pre 2016) is still wanted.

JoesCat:
Yes - there is a lot of redundancy and repetition, but seems freedesktop created a bit of incompatibility with their update requests, so, from our point of view, having new files available for newer desktops, flatpak etc, and also older files for older systems plus appimage is a solution.
Looking at older and newer linux versions, there was more consistency with the older versions, while newer versions it looks more like a mish-mash of old, new, and some stuff that appears inbetween.

Anyway, we never had appdata files before, so it appeared best to first make things work in terms of installation and see what problems occur with what we have. Once things work well, then script building appdata is a possibility.

Thanks for catching these problems.

muelli:
> hm. yeah, I would expect no redundancy to be required in first place.

hm. maybe @ximion can chime in and clarify how to deal with redundancy in this case.
In particular, should this appdata file rather not include the name and summary, because the desktop file has a Name and a Comment?  https://github.com/fontforge/fontforge/blob/5212a06e83f6c350484db549e35bb7894a1c8520/desktop/org.fontforge.FontForge.desktop#L2-L6

Note that this package has both, an [appdata file](https://github.com/fontforge/fontforge/blob/5212a06e83f6c350484db549e35bb7894a1c8520/desktop/org.fontforge.FontForge.appdata.xml) and a [metainfo file](https://github.com/fontforge/fontforge/blob/5212a06e83f6c350484db549e35bb7894a1c8520/desktop/org.fontforge.FontForge.metainfo.xml). I don't fully understand the use-case nor the effect, but maybe that's a reason why validation [failed before](https://github.com/flathub/org.fontforge.FontForge/issues/3#issuecomment-496966360).

jtanx:
On the duplicated appdata and metainfo, see #3863, basically the metainfo file is redundant and can be removed.

ximion:
@muelli `name` and `summary` are required in metainfo files now. They weren't before, where we just extracted the data from .desktop files since AppStream was designed for distribution package repositories, but with Flatpak both tags are now pretty much required. I should adjust the Quickstart Guide to reflect that, it looks like while the spec was updated, the guide wasn't.

`appdata` is just an older name for `metainfo` from a time where AppStream was only about apps. You don't actually need both files, it looks like they contain the exact same information anyway. Just remove one and install the other one into `/usr/share/metainfo`.

muelli:
> @muelli `name` and `summary` are required in metainfo files now.

thanks for the info.
It's sad to require redundant information to be produced, but that's a discussion for another time and place.

ximion:
> > @muelli `name` and `summary` are required in metainfo files now.
> 
> thanks for the info.
> It's sad to require redundant information to be produced, but that's a discussion for another time and place.

I have some code which generates .desktop files from metainfo files ;-) This may even land in the next AppStream release (so something like `appstreamcli make-desktop-file *.metainfo.xml org.example.MyApp.desktop` will work).
There are many scenarios where the metainfo file is present but we don't have access to the .desktop file yet, that's why we needed complete metainfo files. In other areas, this change just improves efficiency.

Issue #3019, opened 2017-02-01
Add from/to Unicode character ranges to Charsets


PanderMusubi:
Please, add from/to Unicode character ranges to Charsets as indicated in the screenshot.

![screenshot from 2017-02-01 16-13-11](https://cloud.githubusercontent.com/assets/198964/22512799/756f7c7e-e89a-11e6-97d6-6429d8ff9e2f.png)

I am willing to create a pull request on GtiHub when maintainers support this idea. Which files need to be extended with this information at the end of the name of the Unicode range? fontforge/unicoderange.c and/or fontforgeexe/fontinfo.c ?

To get these details not, now has to look them up at Unicode's website. Optionally, when someone can supply me with the source of this information, I could add the same for the MS Code Pages.

Reported on Ubuntu 16.10 for FontForge 20161005

### When reporting a bug/issue:
- [x] Screenshot
- [x] The FontForge version and the operating system you're using
- [x] The behavior you expect to see, and the actual behavior
- [x] Steps to reproduce the behavior 
- [x] \(optional) Possible solution/fix/workaround

### When you open an issue for a change/improvement/feature request:
- [x] A description of the problem you're trying to solve, including _why_ you think this is a problem
- [x] If the feature changes current behavior, reasons why your solution is better
- [x] \(optional) Possible solution/fix/workaround

PanderMusubi:
At the same time support `Miscellaneous Symbols and Pictographs (1F500 - 1F5FF)`

JoesCat:
Creating these by hand, the first time sounds okay, but maintaining these as time passes, will become a bit of a bear (to maintain by hand). Just my opinion on noting a couple of Unicode blocks that I'm not 100% sure are up-to 9.0 since I've found more than one copy, plus having looked at combiners which were at 5.3, and I'm not quite satisfied with the results.

To get down to the core, for least programmer interaction, I would have to point towards Unicode/makeutype.c for extracting the unicode ranges (seems plural, when looking at some extended ranges like Arabic), and Unicode/makebuilds.c (needs a rework, and partially handled by Unicode/charset/generate.sh).

At the moment, since the unicode blocks *.c files aren't quite automated yet, you may want to put something together as a pull request to see what this looks like as a result. You may need to run:
grep "Armenian" -R * from the base directory to find the various caches. ...you will note that they might not be 100% right for 9.0, so this is also a good point to correct these manually.

Not exactly an easy answer, but gives an idea of direction - hope it helps :-)

[...]

Issue #401, opened 2013-03-09
Unicode tooltips empty


georgd:
Built from today’s git head, fontforge doesn’t show any content in the unicode tooltips. Please revert to LibUnicodeNames, the latest changes seem rather destructive to things that weren’t broken before.

PR #3861, opened 2019-08-06
Fold libuninameslist into FontForge

khaledhosny:
Clean this mess up by including the two relevant library files into the source tree (and the builder tool, but not compiled, this C mess should be rewritten as a Python script).

The French translation is not used by FontForge, so not included.

Support for building with external Unicode names library is dropped, this obfuscated mess of C macros is an overkill for such a tiny FontForge functionality.


khaledhosny:
The Travis failures are a network issue on some Linux builders.

fontforge/python.c
```diff
@@ -838,14 +838,9 @@ static PyObject *PyFF_UnicodeBlockNameFromLib(PyObject *UNUSED(self), PyObject *
...
+/* Deprecated */
 static PyObject *PyFF_UnicodeNamesListVersion(PyObject *UNUSED(self), PyObject *UNUSED(args)) {
-/* If the library is available, then return the Nameslist Version number */
-    char *temp;
-
-    if ( (temp=unicode_library_version())==NULL ) {
-	temp=malloc(1*sizeof(char)); *temp='\0';
-    }
-    PyObject *ret=Py_BuildValue("s",temp); free(temp);
+    PyObject *ret=Py_BuildValue("s","");
```
jtanx: We should probably just return `20190701` until this is actually removed (same for native scripting), we can also add something like

```
 LogError(_("UnicodeNamesListVersion is deprecated and will be removed in a future release\n"));
```

Unicode/Makefile.am
```diff
@@ -66,6 +66,8 @@ libgunicode_la_SOURCES =	\
 	ustring.c		\
 	utype.c			\
 	gwwiconv.c		\
+	uninameslist.c		\
+	uninameslist.h		\
```
jtanx: Will probably need to fix this, somehow this header is getting excluded from the dist tarball.

fontforge/python.c
skef: I recommend starting to do these using python semantics, as in:
```
PyErr_Warn(PyExc_DeprecationWarning, "(message)");
```

ctrlcctrlv:
I don't think this is necessary. They should be separately packaged as now. Certainly the library is useful to other programs, if not being used. What do you think @Joescat?

jtanx:
The reality is, after however many years it has existed, only FontForge has ever actually used it (and probably for good reason, given how it's designed)

So I'm fine with just rolling it as part of FontForge instead of having to deal with it as another external.

frank-trampe:
Sorry for arriving late here. I was waiting to hear what @JoesCat had to say about it, since it is his territory more than anybody else's. I still hope to hear.

I am mildly opposed to the change for the simple reason that I don't see a problem to be solved. Everything is building fine, and there are easy ways to fix the small versioning issues. I would not support splitting it out if it were currently integrated, but I don't see a compelling reason to integrate.

jtanx:
It seems I'm fighting a losing battle, and my level of care is dropping off pretty steeply, so this will be my last comment on this.

> I don't see a problem to be solved. 

There are at least a couple of things I can see this helping with:
* Guaranteeing that that functionality is always present (it was basically made a hard requirement with the last release anyway), and up to date
* Allowing us to remove all that `#ifdef` mess to support some ancient version that someone *might* be using (granted this can also be resolved by just droppping support for older versions)
* Not having to configure *yet another* package in the build system of the already lengthy list of dependencies
* Not having to release and version another product that's only really used by FontForge (and which we're in control of!) -  and by extension, allowing downstream packagers to not have to bundle a separate package just for FontForge

I'm just disappointed that it's so hard to make changes that improve the ease of working on FontForge, it's always an uphill battle in the face of fear of change.

khaledhosny:
FWIW, GWW had a valid reason to make it a library as he had at least one other application using it (a character picker of sorts). That application is of course dead since he dropped the ball and can’t use this new library anyway.

ctrlcctrlv:
I see your point. I guess it would make things easier to build, and there are other, better libraries that do this for other developers. I don't have a problem with this being merged.

skef:
I don't have a strong opinion either way on this, but I feel like I should note one additional factor: FontForge just recently started nudging distros/installers to build this library, after more broadly encouraging them to install it via the documentation. Changing the policy again so soon might make some of these folks feel jerked around.

**However**, that's a much stronger argument against importing the code into the FontForge repository while keeping it as a library, which could cause dependency headaches downstream. From what I can tell that's not what @khaledhosny has done. As it is FontForge just appears to have dropped an external dependency, which happens sometimes.

Suppose the project continues to maintain the separate repo for the library, makes changes there, and then copies the files over into the FontForge repo to update them. In that case any user who wants the separate library can have it and it's harder (for me) to see what the problem would be.

ctrlcctrlv:
If you're a packager and you're reading this: we're sorry for making you package libuninameslist. I packaged it on Arch so I know it sucks to make a package that's not needed for long.

However, we can't just ignore a much better solution, if this is one, because some work will have been for naught. After all, the Linux packaging universe is fragmented, and that's not our fault. We packed it as a dep in one release due to the release time crunch, and in this release we're vendorizing it.

Unicode/Makefile.am
khaledhosny: How do I see this?

khaledhosny: I see the Travis failure now, but I can’t even run `make dist` here, apparently no one tried it on macOS since 2013 at least!

khaledhosny: Ah, the header is included in the dist tarball, the build was failing because I was passing `"-I$(top_builddir)/Unicode"` instead of `"-I$(top_srcdir)/Unicode"`. I overlooked it locally because I had libuninameslist installed.

bagong:
CMake provides a midway: adding a repo as a "submodule".

khaledhosny:
So does meson, but would that gain us here? Either we want to keep supporting external library and this keep the maze of `#ifdef`’s and build support for external library, or we want to drop it and then adding 3 files to the git repository is a much simpler solution. If the library were not maintained by FontForge and solely used by it, the considerations might have been different.

bagong:
If you give up the original repo, there is no gain, actually there is a
loss, as cloning submodules requires a bit of additional syntax. If you
want to keep/maintain the lib, submodules are better than keeping things in
sync manually.

frank-trampe:
Pinging @JoesCat.

JoesCat:
Under the situation right now, this PR may look ok, but... don't forget...

1) As just mentioned recently, libuninameslist and libspiro act like a canary for the build.

2) Having these libraries separate has given opportunity to refine the autoconf process from the original version back in 2012. Improvements, breakage, and distro feedback on these simple libraries have helped improve FontForge's autoconf.

3) Some ideas like python and libuninameslist names2, probably wouldn't get created. I'm pretty sure 'a'/'h' or ncq on libspiro wouldn't have been created. The spiro crashes were easier to solve with libspiro separate instead of merged into FontForge.

4) When the original libuninameslist was created, computers had less RAM and FontForge was working with a smaller range of characters. A 1M library would probably eat a significant bit of memory. Today, we don't blink at 1G of RAM in a desktop computer, so loading libuninameslist information by default, or having it built in, is not a significant hit unless you are running FontForge on something much smaller than today's PCs. Raspberry or one laptop per child might come to mind, or simply significantly older hand-me-down PCs.

5) A library is supposed to be shareable and reusable. It took a couple years for the current libspiro to be accepted by various [distros](https://repology.org/project/libspiro/versions), and if you look at old info for gegl, it says [it's optional, don't worry if it doesn't work](https://wiki.gimp.org/wiki/Hacking:Problems_and_solutions#SPIRO_not_found). Today, we see it accepted, and we see gegl testing the latest [20190731](https://gitlab.gnome.org/GNOME/gimp/-/jobs/384748).
Libuninameslist is available, but there are fewer places where it could be useful. Something like [this](https://www.google.com/search?q=unicode+screensaver) could probably benefit from it, maybe. ...and the dropped ball @khaledhosny referred to is probably [UMAP](https://sourceforge.net/projects/umap/), but this should still be able to use libuninameslist, and probably the French libuninameslist too.
Debian's [FontForge1](https://packages.debian.org/sid/libfontforge1) was using later versions, but I think the sobump as well as the CRA bugfix would be an issue, so you would need to either compile for the 0 version, or the 1 version.

6) ...however, there is some documentation for libuninameslist, and George mentioned that libuninameslist can be built on a different timeline/schedule based on what unicode.org does, and this can be independent from FontForge releases.
...which would be fine if we still kept dynamic linking, because users could load it then, vs in lock-step with our current model where it is static linking.
@khaledhosny and @jtanx point to the ifdef garbage, I'll just have to say, look at what we got thanks to ripping-out dynamic linking.

7) ...and further reading, you would note there was some concern about copyright/ownership and decided it was probably better to keep NamesList.txt separate from FontForge.
So, if it was an issue, you can see there is clear separation. FontForge does not depend on libuninameslist, and can function quite well without it. In summary, for the majority of users, go ahead and use it, and for those that cannot, it can be left out.
To me it seemed a good precaution to also add some additional [clarification](https://github.com/fontforge/libuninameslist/commit/b83ced9752682778e1102619ae240a8fd7810bf6) in 2013, and apparently it was, because in 2016, debian asked twice about [clarification for FontForge files too](https://github.com/fontforge/fontforge/pull/2735).

@frank-trampe said:
> I am mildly opposed to the change for the simple reason that I don't see a problem to be solved. Everything is building fine, and there are easy ways to fix the small versioning issues. I would not support splitting it out if it were currently integrated, but I don't see a compelling reason to integrate.

I agree.
I think it is easier to let it be as it is now, vs merging, then having to take it out later if necessary.
I think 99% of users don't care for copyright details or if libuninameslist is merged or not, but those that care, will check against their policies, might even ask (like debian), or keep silent and avoid future FontForge releases.
In my opinion, I think it might be okay to merge, but I do not "know" 100% if it is okay for certain.
Need examples - think [GIF, LZW](https://www.gnu.org/philosophy/gif.en.html), [VFAT](https://www.reddit.com/r/linux/comments/4a5k71/eli5_how_microsoft_is_using_patents_to_hurt_linux/), [DVDs](https://www.google.com/search?q=linux+DVD+illegal), etc.

...and since I don't "know", then I mention to those that want to merge...
...so, if you "think" it's a good idea to merge libuninameslist into FontForge, just be sure you "know" for sure 7 (above) is ok.

skef:
People have raised various different concerns on both sides of this issue, but one of the main ones on the "don't do it" side seems to be the "I don't see a problem to be solved" thing.

Let's push the discussion up a level or two for a moment and focus on the question: Assuming that the code is left as a separate library, would people be OK with making it a mandatory dependency?

That's what I see as the fundamental "problem" and/or "cost". FontForge doesn't work as well without `libuninameslist` in terms of what it displays, and the team hasn't been comfortable with the idea of making the dependency mandatory to avoid putting too much work on the plate of distros and other packagers. That is a real and recognized problem. There is a development cost to maintaining the no-`libuninameslist` compilation option, and a cost to users who wind up using installations that lack it. This push request proposes to eliminate both costs. That is what's on the table.

JoesCat:
No - the issue I point to is 7, are you 100% sure it's okay?
The easiest way to find out is - go ask unicode.org

skef:
https://www.unicode.org/license.html

JoesCat:
Works for me.

Below are requested revisions to the code if you plan to merge this patch:

1) Make sure these notices exist in the *.C data files, but it appears you've copied buildnameslist.c and it already contains them.
```txt
(a) this copyright and permission notice appear with all copies
of the Data Files or Software, or
(b) this copyright and permission notice appear in associated
Documentation.
```
2) in Unicode/Makefile.am, uninameslist.h and is_Ligature_data.h probably should go in EXTRA_DIST to avoid cluttering /usr/include

3) I suggest to rethink your code in Python and nativescript concerning ```/* Deprecated */ ...UnicodeNamesListVersion``` and the other deleted functions, George maintained backwards compatibility with some of his own typos, therefore in comparison, looking at your 'Deprecated' you appear to want to remove entire functions altogether. ["Past-politics" and argument](https://github.com/fontforge/fontforge/issues/401) is clouding better judgement with this deletion. I'd recommend - maintain backwards compatibility and keep them.

frank-trampe:
Great. It sounds like we have a path to a solution that pleases everybody.

fontforge/python.c
```diff
@@ -838,15 +838,10 @@ static PyObject *PyFF_UnicodeBlockNameFromLib(PyObject *UNUSED(self), PyObject *
...
-/* If the library is available, then return the Nameslist Version number */
-    char *temp;
-
-    if ( (temp=unicode_library_version())==NULL ) {
-	temp=malloc(1*sizeof(char)); *temp='\0';
-    }
-    PyObject *ret=Py_BuildValue("s",temp); free(temp);
-    return( ret );
+    PyErr_Warn(PyExc_DeprecationWarning, "UnicodeNamesListVersion is deprecated and will be removed in a future release");
+    return Py_BuildValue("s","20190701");
```
dscorbett: This function returns the version of Unicode that the names list was copied from (currently, `"Nameslist-Version: 12.1"`). With this patch, it returns the version of what used to be libuninameslist. Why change it?

skef:
(Setting aside this particular question) I definitely want to remove entire python functions in the future. A lot of the interface is a mess. Deprecation followed by removal is the conventional way to deal with that. 

The sentiment most frequently expressed about the python interface is that people can't figure out how it works and give up.

fontforge/python.c
skef: It sounds like there may have been confusion about the role of this function. The internal FontForge function for this string is `unicode_library_version()` so that confusion is understandable.  

It's frankly bizarre that this function is returning "Nameslist-Version: 12.1" instead of "12.1", but I guess there's little reason to change it now. 

@khaledhosny can we just leave the function chain as it is, or just call `uniNamesList_NamesListVersion()` directly?

khaledhosny: The version was removed because it was hard-coded in the `configure.ac` file. If we really want it, then `buildnameslist.c` needs to be modified to read it from the `NamesList.txt` file.

khaledhosny:
I see there is no consensus and I can’t afford spending more time on this, so closing.

ctrlcctrlv:
I don't see why you needed to delete the branch. Someone else could have worked on it. I think it hasn't been long enough to give up on this. 😩

frank-trampe:
I am similarly puzzled.

skef:
I get "disappointed", but "puzzled"? The same thing happened with #3728. The contributor has expressed his desire to avoid further project conflict on multiple occasions.

ctrlcctrlv:
It's your PR so you have the right to close it. I'm just highly disappointed that the branch was deleted. Someone else could have reopened this and taken responsibility for getting it merged. Now we have to start from scratch, and all the comments here are for nothing. This was a profound waste of time.

If Khaled didn't want to work on it anymore, fine, just say so; but the work he did do was usable.

I wrote just today about how old our packages are in mainstream distros: https://twitter.com/FontForge/status/1161496744615645185

This was a step towards making FontForge easier to pack. Now we're one step forwards and two steps back.

@frank-trampe I'd like to write formalized rules for the PR process so this never happens again:

* When to merge and who can merge
* How to step away from a PR
* Expectations about timeliness
* How to determine if a consensus has been reached

Etc. Would you be willing to do this with me? I'll send you a draft via e-mail if so.

frank-trampe:
That sounds like a good idea. We already have some text in `CONTRIBUTING.md` that is likely to be a good starting point.

jtanx:
If we *truly* want to do this, then you can fetch the pr directly with

```
git fetch origin refs/pull/3861/head:pr/3861
```

ctrlcctrlv:
Cool! Yes, that does work, at least on my machine.

fontforge/python.c
JoesCat: Yeah - as with libspiro, all info was put at the top of configure.ac. This could be repeated in Fontforge's configure.ac, unfortunately 20190701 is incorrect. The info given was supposed to be the version of NamesList.txt, which is 12.1.

JoesCat:
@skef wrote,
> (Setting aside this particular question) I definitely want to remove entire python functions in the future. A lot of the interface is a mess. Deprecation followed by removal is the conventional way to deal with that.

> The sentiment most frequently expressed about the python interface is that people can't figure out how it works and give up.

I'm still of the opinion that trying to understand it, then fix it, is a better choice. If you can't figure it out now, leave it for tomorrow or another day, maybe someone else will figure it out. It's all to easy to press the delete key and say you've accomplished something.

JoesCat:
@ctrlcctrlv - if you're going to work on this, you may want to throw in some ideas here:

make the lib interface connect to libuninameslist-fr for the french users. Scripts in nativescripting and python would still need to remain english to keep everyone compatible, but the pop-up hints could display french followed by english (because the french version10.0 may not be in sync with some english version12.1 chars).

The blocklist ranges in libuninameslist is complete across the range. There is an earlier fontforge issue request to add hints to ranges #3019, however, maybe this is better done in makeutype

as discovered above 20190701 needs to be corrected for NamesList 12.1

skef:
@JoesCat I agree and there's nothing I want to change that I don't understand, or at least that I don't think I understand. GW wasn't very familiar with Python when he added the interface and said as much. One result is that many interfaces simulate a read/write variable rather than a function even when that's not appropriate. This has wound up being both 1) very confusing and 2) eliminated the possibility of supplying parameters when making changes via those interfaces. I'm not a spring chicken and it took me a long time and I needed to ask some questions before I put together that the status of a layer object was entirely different than that of a contour object, and that the former needed to be updated to effect any change in the character. The structure of the interface is supposed to communicate these differences to the programmer. 

As we change the interface to be more appropriate we can leave the old stuff deprecated for a while, but leaving it un-deprecated and therefore permanent is just leaving a big mess that will be even harder to understand.

frank-trampe:
Any user interface (and the Python bindings fall in that category) that almost baffles @skef definitely needs improvement.

PR #3862, opened 2019-08-07
Fix macOS build

khaledhosny:
Homebrew updated Pango and now GDK breaks because our outdated Pango library.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -126,9 +126,6 @@ install:
       # 10 billion years later...
       brew install autoconf automake libtool pkg-config ruby
       brew install cairo coreutils fontconfig gettext giflib gtk+3 jpeg libpng libspiro libtiff libtool libuninameslist python@2 python@3 wget woff2
-      # Pin pango to this version, because of issues, see #3343
-      brew unlink pango || true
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -126,9 +126,6 @@ install:
       # 10 billion years later...
       brew install autoconf automake libtool pkg-config ruby
       brew install cairo coreutils fontconfig gettext giflib gtk+3 jpeg libpng libspiro libtiff libtool libuninameslist python@2 python@3 wget woff2
-      # Pin pango to this version, because of issues, see #3343
-      brew unlink pango || true
-      brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/ad91e4df4a843764901811a965a17f650f34e2c1/Formula/pango.rb
```
frank-trampe: So we don't need to install it at all?

khaledhosny: It is a dependency of GTK, and AFACT it wasn’t explicitly listed before the change that introduced the removed lines.

frank-trampe: Ah, okay.

frank-trampe:
This looks good to me.

Issue #3814, opened 2019-07-22
Translation management system


ctrlcctrlv:
FontForge has a bunch of old translations. As seen in #3797/#3808 (etc.), non-experts have a hard time learning how to donate translations to the project, and it takes up developer time helping them.

There are several better ways:

* **Transifex** &mdash; I have experience with this one from an old job of mine. They used to allow just anyone to make an account if you promised you were an open source project, but it seems that they are getting less and less generous. @fitojb, one of our translators, no longer works with them. Their website is snazzy, but FontForge needs something that's going to last, and which our translators like, so pass.
* **Translatewiki** &mdash; This is my preferred option. It is used by several high priority FOSS projects and websites, including OpenStreetMap, Internet Archive, MediaWiki and MathJax., I talked to the owner of it in IRC and we don't have to do much to get started. We'd need to:
  * change how `FontForge.pot` is built, building it as part of the normal build process. Us developers need to just be good about checking in new strings when we add them.
  * give a translatewiki team member push access to this repository so they can update the `po` files.
  * write a quick article about FontForge and explain some of the terminology (I'll do this part)
* **Weblate** &mdash; recommended by @fitojb, I don't know anything about it. But let's put it on the list anyway. :-)

frank-trampe:
I'd usually prefer Python over PHP, but it looks like we get much better human support for Translatewiki, so I'd be inclined to go that route.

Feedback from @jtanx, @davelab6, and @JoesCat would be much appreciated.

jtanx:
> give a translatewiki team member push access to this repository

So changes are just committed directly to master? Initial impression is I'm not sure how I feel about that.

Nikerabbit:
> So changes are just committed directly to master? Initial impression is I'm not sure how I feel about that.

Hi, I am from translatewiki.net. Can you elaborate what makes you feel uneasy?

jtanx:
Yep, usually just because all changes to master goes through pull requests, and it feels like a little bit of a security hole to allow otherwise.

nemobis:
Jeremy Tan, 22/07/19 12:13:
> Yep, usually just because all changes to master goes through pull 
> requests, and it feels like a little bit of a security hole to allow 
> otherwise.

Translations as data are not typically executable.

While some malformed translations may be submitted by translators, using 
a web-based translator greatly reduces the chances of errors compared to 
the manual systems your translators seem to be using now. Translate also 
allows to code checkers which either warn or block users upon errors.

[...]

PR #3863, opened 2019-08-08
Appstream and potfile update

jtanx:
So I took some time to actually read the specs, and came to the following conclusions:

* Installing to the appdata folder is not required: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html
  > AppStream tools scan the /usr/share/appdata/ path for legacy compatibility as well. It should not be used anymore by new software though, even on older Linux distributions (like RHEL 7 and Ubuntu 16.04 LTS) the metainfo path is well supported. Support for the legacy path might be dropped completely in future. 
* Only one xml file is needed, since we package a GUI, all we need is an appdata.xml: https://docs.fedoraproject.org/en-US/packaging-guidelines/AppData/
  > If a package contains a GUI application, then it SHOULD install a .appdata.xml file into %{_metainfodir}. Installed .appdata.xml files MUST follow the AppData specification page.

  > If a package contains an add-on for GUI application, then it SHOULD also install a .metainfo.xml file into %{_metainfodir}. Installed .metainfo.xml files MUST follow the AppStream add-ons specification.
* The strings are translatable through gettext, just pass it into xgettext

I also looked into that appstream-util, while it *is* present, it's too old on travis (xenial) so it always fails validation, even though it's perfectly valid.


muelli:
have you added the POT file on purpose? it's a generated file and some projects are hesitant to add generated files.

jtanx:
Yeah, it was deliberate, basically as part of #3814. But reconsidering, I think I'll defer adding that for now. As you said, it's kinda crap having to commit generated files.

jtanx:
Any objections? Otherwise  I'll merge this tomorrow

muelli:
LGTM.
Thanks for the efforts you're putting into this.

frank-trampe:
I don't know enough to comment, but it sounds like things are all in order.

PR #3865, opened 2019-08-11
Enable hotkeys in stroke dialog

skef:
Make the CharView hotkeys apply in the stroke dialog, to aid "polygon" editing.

This kind of fix is probably needed elsewhere, like the gradient editor. The stubs towards the bottom of `charview.c` would be a good starting point for research.

### DIFF
--- a/fontforgeexe/cvstroke.c
+++ b/fontforgeexe/cvstroke.c
@@ -663,6 +663,7 @@ static void MakeStrokeDlg(void *cv,void (*strokeit)(void *,StrokeInfo *,int),Str
 	pos.width = GGadgetScale(GDrawPointsToPixels(NULL,SD_Width));
 	pos.height = GDrawPointsToPixels(NULL,strokeit!=NULL ? SD_Height : FH_Height);
 	sd->gw = gw = GDrawCreateTopWindow(NULL,&pos,stroke_e_h,&sd->cv_stroke,&wattrs);
+	GDrawSetWindowTypeName(gw, "CharView"); // For hotkeys
 	if ( si!=NULL )
 	    GDrawRequestDeviceEvents(gw,input_em_cnt,input_em);
 
--- a/fontforgeexe/cvstroke.c
+++ b/fontforgeexe/cvstroke.c
@@ -612,7 +612,7 @@ static void MakeStrokeDlg(void *cv,void (*strokeit)(void *,StrokeInfo *,int),Str
 	    *caparray[7], *joinarray[7], *swarray[7][5], *pens[10];
     GTextInfo label[39];
     int yoff=0;
-    int gcdoff, mi, swpos;
+    int gcdoff, mi, swpos, tfpos[3];
     static StrokeInfo defaults = {
         25,
         lj_round,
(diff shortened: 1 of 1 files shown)


skef:
This isn't ready yet -- the hotkeys are overriding focused text field entry.

skef:
@jtanx Do you happen to know if there is a straightforward way of limiting or eliminating the active hotkeys while a text field has focus?

skef:
OK, this latest change fixes the hotkey override but there are still some issues with strings to switch between controls, particularly tabbing between them.

jtanx:
Regarding the hotkeys, your guess is as good as mine; I'd need to dig into it to find out

jtanx:
> OK, this latest change fixes the hotkey override but there are still some issues with strings to switch between controls, particularly tabbing between them.

Could you expand a bit more on this? I'm not quite sure I understand what this is referring to

skef:
> Could you expand a bit more on this? I'm not quite sure I understand what this is referring to

It refers to the https://github.com/fontforge/fontforge/pull/3865/commits/f728c8e7c166e48ed13296d3037d2697e45f7563 update, and in particular the addition of the `GGadgetSetSkipUnQualifiedHotkeyProcessing` calls. 

The overall problem was that turning on hotkeys with the `GDrawSetWindowTypeName` setting broke a bunch of stuff. The most important thing it broke was text entry in the text fields: Typing hotkey-attached numbers would just activate those hotkeys. My second changeset fixes that, but there are still some lingering differences. The most obvious (to me) is that the tab key used to shift the focus between dialog UI elements generally, and how that only works 1) when a text field already has focus and 2) only with "forward tabbing". 

As of https://github.com/fontforge/fontforge/pull/3865/commits/f728c8e7c166e48ed13296d3037d2697e45f7563 this may be an acceptable update but people who never use the Polygon controls may be annoyed. I recommend comparing the before and after behavior WRT tabbing (to @jtanx, who might be able to help, and to @ctrlcctrlv, who could offer opinions on acceptability.)

skef:
Another issue is that the Alt accelerator keys for the menus conflict with those of the dialog controls. That may have been true before, though. (It makes little sense to remap the menu keys for this application, and making the dialog accelerators unambiguous would be tricky, as there aren't many characters left. Therefore not sure what to do about this particular problem. I guess if the dialog control accelerators could be given priority that would be an improvement.)

jtanx:
The window type name on the sd->cv_stroke->v is already set in `MakeStrokeDlg -> StrokeCharViewInits -> _CharViewCreate`, so I'd think that you wouldn't need to do that again for the stroke dialogue itself. Perhaps there's some other reason why those hotkeys aren't being registered in the first place>

jtanx:
Actually no, that's probably not right, I see that the charview does something similar by setting that name on `cv->gw` and then calling `GGadgetSetSkipUnQualifiedHotkeyProcessing` on `cv->charselector`

skef:
I got the idea for `GGadgetSetSkipUnQualifiedHotkeyProcessing` from `charselector`, but it was a stab in the dark. 

From my increasingly distant memory of hotkey/menu stuff I think `GDrawSetWindowTypeName` has to be set on the very top level window or they don't work. But again it's been a while.

skef:
To bump back up a level, the question is what changes I need to make to this code, if any, to make it acceptable to merge. What needs to work? There's no current plan to eliminate the shape editing interface and it's close to unusable without any accelerators.

jtanx:
Hmm, I tried this again, and forwards/backwards tabbing both works for me, even with both changes in place. Anyway, I'm ok with these changes.

skef:
@ctrlcctrlv Can you give these changes a quick try and see if the result is acceptable for your workflow?

ctrlcctrlv:
This doesn't break anything for me (that I noticed). Tab, shift-tab work for me, so I'll do the honors of merging.

BTW, I tested like this:

```bash
# git remote add skef https://github.com/skef/fontforge # I already have done this before.
git fetch skef hotstroke:hotstroke
git checkout hotstroke
make -j9 && ./run_fontforge -new
```

PR #3869, opened 2019-08-12
Remove contextual information from translated strings

dscorbett:
`msgstr`s shouldn’t include the contextual information preceding vertical bars.

### Type of change
- **Non-breaking change**

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -3240,7 +3240,7 @@ msgstr "Skriptenliste"
 
 #: ../fontforge/lookups.c:1328 ../fontforgeexe/lookupui.c:191
 msgid "Script|Balinese"
-msgstr "Skript|Balinesisch"
+msgstr "Balinesisch"
 
 #: ../fontforge/lookups.c:1329 ../fontforgeexe/lookupui.c:192
 #, fuzzy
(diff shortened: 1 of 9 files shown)


ctrlcctrlv:
@dscorbett Hope my questions on all your PRs aren't annoying you. Just don't want to merge anything that will break someone's workflow.

So, my question is:

Why is it correct to have | in the English string, but not in translated strings? It's not clear from your commit, at least to me. Why shouldn't this also be removed from English, if it should be removed at all?

dscorbett:
Questions are fine; in fact you’ve asked fewer than I expected.

The part before the bar is stripped away from the English string at runtime and exists only to provide context to the translator. Some translators did not follow this convention.

https://github.com/fontforge/fontforge/blob/626f751752875a0ddd74b9e217b6f4828713573c/po/hr.po#L12476-L12494

https://github.com/fontforge/fontforge/blob/626f751752875a0ddd74b9e217b6f4828713573c/gutils/gwwintl.c#L38-L45

khaledhosny:
Gettext has better way to handle translation context for a long time now, but I’m not volunteering to do the conversion.

PR #3871, opened 2019-08-13
Fix and expand the lists of scripts

### LINKED ISSUE #3559: Script tag bugs in tottfgpos.c
The array `scripts` in tottfgpos.c is outdated. It is missing recently encoded scripts like Chakma and Javanese. It is also missing recently encoded code points for older scripts, like U+18B0 CANADIAN SYLLABICS OY.

It conflates blocks with scripts. For example, U+03E2 COPTIC CAPITAL LETTER SHEI is in the block Greek and Coptic, but it is Coptic, not Greek.

It conflates ISO 15924 codes with OpenType script tags. For example, although Cirth has an ISO 15924 code, it has no OpenType script tag, so 'cirt' does not belong in the table.

'aram' for Imperial Aramaic should be 'armi'

The function `ScriptFromUnicode` can convert old Indic tags to new Indic tags, but it omits 'mymr' → 'mym2'.

The function `ScriptIsRightToLeft` is missing newly encoded scripts. [Here are all the RTL scripts in Unicode.](https://www.unicode.org/cldr/utility/list-unicodeset.jsp?a=%5B%3ABidi_Class%3DAL%3A%5D%5B%3ABidi_Class%3DR%3A%5D-%5B%3AZyyy%3A%5D-%5B%3AZzzz%3A%5D&abb=on&g=sc) Note that directionality is not entirely determined by script, so if anything uses this function to determine the directionality of a specific code point, that is wrong.

dscorbett:
Fixes #3559. The lists of scripts in fontforge/lookups.c, fontforge/tottfgpos.c, and fontforgeexe/lookupui.c now match the current [OpenType script tag registry](https://docs.microsoft.com/en-us/typography/opentype/spec/scripttags). I fixed typos, removed erroneous entries, deleted obsolete comments, alphabetized, and added many new entries. I also updated the code point ranges based on Unicode’s script values.

I think all of the translatable strings are valid `msgid`s. I recycled the new ones from block name strings; for example `N_("Adlam")` is now used for both the script and block. I didn’t update the .po files. Some of the translated strings are now not used.

Contrary to what I supposed in my bug report, “Aramaic” refers not to Imperial Aramaic but to Samaritan.

### Type of change
- **Bug fix**
- **Non-breaking change**


ctrlcctrlv:
I notice `aram` was changed to `samr`. What if a font produced before this change is using `amar`, can we still read it?

dscorbett:
Yes. For existing 'aram' fonts, the UI will no longer display “Aramaic”: it will simply fall back to “'aram'”.

fontforge/lookups.c
```diff
@@ -1319,17 +1319,22 @@ static struct {
     const char *text;
     uint32 tag;
 } localscripts[] = {
+    { N_("Adlam"), CHR('a','d','l','m') },
+    { N_("Ahom"), CHR('a','h','o','m') },
+    { N_("Anatolian Hieroglyphs"), CHR('h','l','u','w') },
 /* GT: See the long comment at "Property|New" */
 /* GT: The msgstr should contain a translation of "Arabic", ignore "Script|" */
```
khaledhosny: I *think* this comment meant to be on the first line, and Arabic is used as example because it happens to be the first one.

fontforgeexe/lookupui.c
```diff
@@ -182,122 +182,168 @@ static GTextInfo *lookuptypes[2] = { gsub_lookuptypes, gpos_lookuptypes };
     /* see also list in tottfgpos.c mapping code points to scripts */
     /* see also list in lookups.c for non-ui access to these data */
 GTextInfo scripts[] = {
+    { (unichar_t *) N_("Adlam"), NULL, 0, 0, (void *) CHR('a','d','l','m'), NULL, 0, 0, 0, 0, 0, 0, 1, 0, 0, '\0'},
+    { (unichar_t *) N_("Ahom"), NULL, 0, 0, (void *) CHR('a','h','o','m'), NULL, 0, 0, 0, 0, 0, 0, 1, 0, 0, '\0'},
+    { (unichar_t *) N_("Anatolian Hieroglyphs"), NULL, 0, 0, (void *) CHR('h','l','u','w'), NULL, 0, 0, 0, 0, 0, 0, 1, 0, 0, '\0'},
 /* GT: See the long comment at "Property|New" */
 /* GT: The msgstr should contain a translation of "Arabic", ignore "Script|" */
```
khaledhosny: Same here.

fontforge/lookups.c
dscorbett: The `GT:` comment is associated in the .po files with the immediately following string, so it should precede the first entry that uses a vertical bar. Here is an existing example:
https://github.com/fontforge/fontforge/blob/b6a4cb45eb9234cfd3235dab1513cb624073226d/fontforgeexe/lookupui.c#L304-L316

dscorbett:
I forgot to mention that I changed Myanmar’s script code from 'mymr' to 'mymr2' in accordance with the advance on [its OpenType script development page](https://docs.microsoft.com/en-us/typography/script-development/myanmar#appendix-writing-system-and-language-tags): “The script tag for Myanmar script for use with the Myanmar shaping engine is mym2 and not mymr. The script tag mymr has limited support and should not be used.”

fontforgeexe/lookupui.c
khaledhosny: Right, I missed that. Sorry for the noise.

fontforge/lookups.c
khaledhosny: Right, I missed that. Sorry for the noise.

Issue #3777, opened 2019-07-11
⟦Expand Stroke⟧ Simple loop doesn't stroke correctly


ctrlcctrlv:
This loop strokes in two pieces, one would be expected.
<img src="https://user-images.githubusercontent.com/838783/61041024-4a9ae380-a404-11e9-8fc1-cdfb84705e45.png" width="200"> | <img src="https://user-images.githubusercontent.com/838783/61041212-8df55200-a404-11e9-8647-66d701a4983c.png" width="200">

# Test case

## Required settings
<img src="https://user-images.githubusercontent.com/838783/61041479-f3e1d980-a404-11e9-8b93-33bdffd608f2.png" width="200">

## Required files
[loop.zip](https://github.com/fontforge/fontforge/files/3381394/loop.zip)

# Friendly warning

I'm still working on my lib2geom replacement for our &laquo;Expand Stroke&raquo;, the same path works in Inkscape so that will fix this when it's done. However, of course I'd be happy if the legacy stroker is fixed so am reporting this new bug I found.

Issue #3762, opened 2019-07-05
⟦Expand Stroke⟧ Rectangular (calligraphic) results in spline with strange looking rectangular cut-out


ctrlcctrlv:
# A bug in three acts

![2019-07-05-175620_1807x665_scrot](https://user-images.githubusercontent.com/838783/60714713-2fd0f680-9f4e-11e9-9d29-41dcabff0ac2.png)
![2019-07-05-175605_1814x693_scrot](https://user-images.githubusercontent.com/838783/60714714-30698d00-9f4e-11e9-95a1-6b3b5a0f37e1.png)
![2019-07-05-175555_1757x647_scrot](https://user-images.githubusercontent.com/838783/60714715-30698d00-9f4e-11e9-82c4-acd2938d551c.png)

# Test case *sans* background image

[S.zip](https://github.com/fontforge/fontforge/files/3362127/S.zip)

## Required settings

<img src="https://user-images.githubusercontent.com/838783/60709947-293d8180-9f44-11e9-8336-7b14c5d0ac41.png" width="200">

# Work arounds

Attempting a &laquo;Simplify&raquo; _does not work_.

Enabling Spiro, moving the first point 1px to the left then 1px back (to force the original Bezier to be lost), then disabling Spiro again and stroking the resulting B&eacute;zier which now has many more points _does work_, but is very hacky obviously.

![2019-07-05-180149_632x563_scrot](https://user-images.githubusercontent.com/838783/60715120-e59c4500-9f4e-11e9-8cc2-c53328f2b607.png)
![2019-07-05-180204_645x799_scrot](https://user-images.githubusercontent.com/838783/60715118-e503ae80-9f4e-11e9-8446-5eb3aaf4b508.png)

skef:
To reproduce this use the parameters in the image except set the pen angle to 45.

PR #3873, opened 2019-08-13
Only run WithinMinMax when spline is single-dimensional

skef:
SplineIsLinear is incorrectly marking some splines linear among those that start and end on the same x or y value. As a consequence `SplineRefigure3` zeros out `Spline1D` values `a` and `b`. 

This is another "**How is this possible?!?!**" bug. Who can guess how many crashes and spline distortions, including those with filed issues, trace to it? 

Here is a high-level test: 
[islintest.py.txt](https://github.com/fontforge/fontforge/files/3498785/islintest.py.txt)

This script creates two contours, calls `addExtrema` on each, and saves them in chars "A" and "B" of `islintest.sfd`. The respective point values of each contour are very close, but the second is not modified by `addExtrema`, as evident when looking at the characters in FontForge. After the fix both contours produced by the script are modified.  

- **Bug fix**

### DIFF
--- a/fontforge/splineutil2.c
+++ b/fontforge/splineutil2.c
@@ -276,27 +276,27 @@ return( true );
     if ( RealNear(spline->from->me.x,spline->to->me.x) ) {
 	ret = RealNear(spline->from->me.x,spline->from->nextcp.x) &&
 	    RealNear(spline->from->me.x,spline->to->prevcp.x);
-	if ( ! ((spline->from->nextcp.y >= spline->from->me.y &&
-		  spline->from->nextcp.y <= spline->to->me.y &&
-		  spline->to->prevcp.y >= spline->from->me.y &&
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
I tried this against a few of my test cases:

* #3777 - **still outstanding**
* #3762 - **still outstanding**
* https://github.com/fontforge/fontforge/issues/3264#issuecomment-489408743 - **still outstanding**

So, I'm sure this change is needed and am okay with it, but it doesn't fix any problem I know of.

skef:
@ctrlcctrlv This problem is less likely to affect the stroking code because there isn't all that much spline math in it. You'd only see a the problem if a contour included a spline like this to be stroked, and these are more likely to be splines generated by intermediate processes than created directly by a user.

frank-trampe:
This looks right to me. The modified conditional block ought to act only as a further qualifier, not another path to a positive result.

Issue #3857, opened 2019-08-05
Where to find translations of mac-font-features


milotype:
I don't know where the existing translations come from!? (see attachement)
File > Preferences > Mac > Features

I'd like to translate those into croatian and german (which is partialy translated).
Can someone please let me know, where to find the appropriate file(s).

![Mac-features](https://user-images.githubusercontent.com/43657314/62477683-c1c85b00-b7a9-11e9-99de-f23563bfd769.png)

skef:
This is probably not a particularly helpful answer but those strings are in a table in [fontforge/macenc.c](https://github.com/fontforge/fontforge/blob/a263e0dc1f813894f7405658ab0c22d871b9080c/fontforge/macenc.c#L1466)

milotype:
Actually it does help!
I'll simply translate all "english" strings, code 0, for instance:
`{ &fs_names[146], 0, 0, "All Typographic Features" }`
into croatian, with the appropriate code 18:
`{ &fs_names[146], 0, 18, "Sve tipografske funkcije" }`
and add the whole croatian list into the file at the appropirate place.

QUESTION:
How do I "encode" croatian diacritics? I guess UTF-8 shouldn't be used in that file.
ć, č, š, ž, Ć, Č, Đ, Š, Ž

I've notice in German for instance: Br\237che (=Brüche)
(couldn't figure out what encoding uses `/237` for 'LATIN SMALL LETTER U WITH DIAERESIS')

skef:
This stuff is *really* old and looks like it was built up by hand -- I'm not sure just adding to the first table will work but it might. 

It appears the text is encoded as [Mac OS Roman](https://en.wikipedia.org/wiki/Mac_OS_Roman)

skef:
So I guess that means you should either use [Mac OS Croatian](https://en.wikipedia.org/wiki/Mac_OS_Croatian_encoding) or there's nothing likely to work without some more modification.

milotype:
I've produced a `.pot` file for the AAT-font-features. (see attachment)

Actually there are 146 "english" AAT-strings, but some are duplicates (which isn't allowed in `.pot` files, so in the end, there are 136 AAT-strings to translate. HOWEVER, I've put a comment for the strings which have duplicates, indicating the order number of the "feature". For instance:

```
#: ../fontforge/macenc.c:108
#. duplicate 108, 133, 136, 139, 142
msgid "Proportional"
msgstr ""
```
[...]

PR #3875, opened 2019-08-14
Update macenc.c with croatian translation

milotype:
Croatian translation added, as discussed in issue #3857.
(I don't know if this should be marked as a New feature or as a Non-breaking change)


dscorbett:
The build failed because of an invalid octal escape sequence in `"Me\421unarodni simboli"`. The character ⟨đ⟩ is 0xF0 (hexadecimal) is Mac OS Croatian which is `\360` in hexadecimal.

In macenc.c the comment “The croatian encoding also uses 0 (mac roman) encoding even though it's not” and the entry <code>sm_roman,&#9;&#9;/* Croatian */</code> need updating.

skef:
> In macenc.c the comment “The croatian encoding also uses 0 (mac roman) encoding even though it's not” and the entry `sm_roman,	/* Croatian */` need updating.

The file is verified to work (besides the latest bad Octal code), and does so because of the table assignment in `MacEncLangToTable`. Historically [Mac OS Croatian is considered a derivative of Mac OS Roman](https://en.wikipedia.org/wiki/Mac_OS_Croatian_encoding), which may be why GW set things up this way.

milotype:
I've corrected the octal code in my commit/fork.
Is that sufficient, or do I need to do a new PR?

skef:
I'm not the right person to do the approval because I had too much input as you worked through the issue, but as far as I know this updated version is correct relative to the existing interface. Someone else can comment on and/or correct my clarification of the `sm_roman` question. 

There's still the lingering question of why this file uses old Mac encodings in the first place, or whether it needs to going forward. I understand the "needs to have strings in different languages simultaneously" issue, which makes gettext-like solutions at least tricky and perhaps untenable. I still don't understand why the strings can't just be encoded as Unicode. However, as an individual contributor that's not @milotype's problem to solve, and doesn't need to be solved by the reviewers either.

skef:
Oh, note to reviewers: I recommend scanning through the associated issue to the part about how `fs_names` is organized.

dscorbett:
>There's still the lingering question of why this file uses old Mac encodings in the first place, or whether it needs to going forward.

Could it be to support [platform-specific encodings in the 'name' table](https://docs.microsoft.com/en-us/typography/opentype/spec/name#macintosh-encoding-ids-script-manager-codes)?

skef:
@dscorbett That's certainly plausible, but are the strings themselves actually written out to a file anywhere? I know where they are displayed but I haven't yet seen a context where they are stored.

milotype:
While doing the translation, one issued was the octal encoding.

The other issue was, how the string are associated between the different laguages - adding a certain value to point to the string of the next language ... very complicated (for non-coders)!

Idealy, all 146 strings would end up in FontForge's .po file. If the "special" encodings are an issue, maybe one could convert them as needed "on-the-fly" from the .po file. Thus, mistakes as I've made them also wouldn't happen. (This is a no-programmer speaking ;-))

ctrlcctrlv:
* If you're just fixing Croatian, why are there so many changes to irrelevant languages like German?
* Have you actually compiled this and made sure it works?

ctrlcctrlv:
Wait, are you adding a new language to the list? What on Earth prompted this? Why does it even matter? How do we even know this will work on the ancient platforms these fonts are for?

I vote that this should be closed. This area of the code should not be touched by non-experts in ancient Mac fonts.

skef:
@ctrlcctrlv 

1. These strings are displayed in the GUI under Preferences -> Mac. So in that sense "it maters"
2. The changes to the previous language(s) are due to the datastructure.
3. I've compiled a version before one minor change and confirmed the strings are displayed. 

See issue #3857. 

The OP hopes to make FontForge fully localized for Croatian speakers.

ctrlcctrlv:
I don't think this is necessary, I don't think new languages should be added, and I don't think these values should be updated without a good reason to do so; certainly a better reason than OCD about every obscure nook and cranny of FontForge being localized.

This patch works fine for me (assuming I understand what it does), but I don't like the precedent it sets. Adding translations to `macenc.c` is not necessary for translation completion.

milotype:
I don't understand your rejection/argumentation on this. The file contains strings used in the GUI – I find that a very good reason to translate them. The basic idea of localizations is to translate everything, at best – even things which may not seem to be essential.

And as you pointed out yourself to the old mail, where Georg Williams actually was asking for translations of these features. "If anyone has time, I've got some phrases that I would like translated into as many languages as possible." So the "precedent" was initially set by him.

The bigger problem is the somewhat complicated datastructure and octal encoding, which both don't allow for quick comparisson of old and updated files. On the other hand, looking at https://developer.apple.com/fonts/TrueType-Reference-Manual/RM09/AppendixF.html#Type0, the whole section of FontForge's Mac font-features isn't up to date. Certain selectors are discouraged and deprecated by Apple, new ones have been added, ... but that's an issue for itself.

AFAIK, neither Glyphs nor FontLab offer ways for handling AAT. But I wouldn't know that Apple abandoned AAT. So the argument can be:
- it's unique to FontForge (whatever the importance of it), keep it and update it
- no one cares about AAT (except for Apple), remove it from FontForge. Or keep it and don't touch anything of it just for historical reasons, but inform users about it.

khaledhosny:
> no  one cares about AAT (except for Apple), remove it from FontForge.

That is it. No one made new AAT fonts in the past decade or two (except Apple, which has is own command line tools for building AAT tables).

Support in FontForge is incomplete and almost untested. The only use case I ever heard of is people using it to examine AAT fonts. But that is not a FontForge job, and FontTools can decompile/recompile AAT fonts more faithfully than FontForge.

ctrlcctrlv:
Closed per https://github.com/fontforge/fontforge/pull/3875#issuecomment-521874776, https://github.com/fontforge/fontforge/pull/3875#issuecomment-522160022 and lack of further discussion.

skef:
Sorry for the wasted effort @milotype -- I know you would prefer to see the whole interface localized. Maybe I should have just merged this earlier even though I had a lot of input. 

If it's any consolation this really is a support corner-case. Almost no one is likely to have any use for the tables, and if the other strings in the dialog are localized it will likely be clear to users that they have no need for it.

PR #3879, opened 2019-08-15
Multiple build system and documentation updates, added crowdin auto push

### LINKED ISSUE #3860: run appstream validation as part of the CI
It would be great to prevent breakage on the appdata files, like it happened in https://github.com/fontforge/fontforge/pull/3812 (which then needed to be fixed up with https://github.com/fontforge/fontforge/pull/3858).

I know two tools to validate, appstream-util and appstreamcli. The former is used by the flathub build bot.

jtanx:
* Add AppStream validation, closes #3860
* Run msgfmt on the appdata (see description below)
* Fix debian packaging metadata, add deb-src target (although semantics is different)
* Fix the rpm packaging metadata
* Update CONTRIBUTING.md, rename INSTALL-git.md to INSTALL.md (since it gets bundled with the dist releases)
* Fix woff2 detection on some platforms
* Build gdraw as a standard static library instead of an object library (makes it easier to work with on older versions of cmake, plus there's no real reason for it to be an object library, unlike gutils and gunicode)

Something I missed from #3863 is that (I think) you're meant to run msgfmt on the appdata so it spits out the appdata with the translated strings. I think you can also do this for the desktop file, but at least on Xenial, trying to do that results in a broken desktop file, as there seems to be some parsing error in msgfmt which occurs whenever there's a line that's longer than 100 chars. Anyway the desktop file isn't that huge, and so it's pretty easy to just manually update it as needed.

### Type of change

- **Bug fix**


po/LINGUAS
```diff
@@ -0,0 +1,18 @@
+ca
```
jtanx: This LINGUAS file is required by msgfmt with the added invocations

jtanx:
The appimage I made for appstream-util is broken, I'll fix that tomorrow...

frank-trampe:
I'll work on getting the Python 2 build working again next week so that we don't need to drop that text.

jtanx:
I don't think we should be advertising python 2 support any more.

CONTRIBUTING.md
```diff
@@ -192,43 +86,51 @@ You might want to pull down some old unicode bitmap fonts.
...
-The most common build target is currently Ubuntu Precise, a long-term support release with Launchpad build support. 
-A binary package built on and for Precise also installs and functions correctly on Debian testing and on Ubuntu Trusty.
+The most common build target is currently Ubuntu Xenial, a long-term support release with Launchpad build support. A binary package built on and for Xenial (usually) also installs and runs on later versions of Ubuntu. 
 
 After downloading the FontForge source, change into the source directory and run
 
-    ./bootstrap; 
-    ./configure; 
-    make \
-    F_METADATA_REGENERATE=1 \
```
frank-trampe: What lets us drop this option?

frank-trampe:
`CONTRIBUTING.md` isn't super visible, but I guess it doesn't really matter. People who are determined enough to build from source can probably figure it out.

jtanx:
Oops, didn't mean to mark that as resolved, but the syntax and operation of the dev-src command has changed, read further down

jtanx:
The more I think about this, especially after having done the Debian packaging cruft, it just doesn't make sense to re-introduce support for Python 2. It's just going to cause confusion and headaches down the line.

ctrlcctrlv:
I agree. Python 2 was already removed, we told users it won't be in the next release. *Libre Graphics World*, an online trade magazine, [reported on this](http://libregraphicsworld.org/blog/entry/week-recap-4-august-2019) as well. It's highly ill advised to add it back.

jtanx:
I've fixed up the rpm building, although tbh I doubt it's of much value. The previous stuff was broken anyway as the spec was just wrong for the fontforge-common package, it was bundling the docs instead of the actual files it was supposed to bundle.

ctrlcctrlv:
@jtanx Do you want me to announce Crowdin on Twitter?

jtanx:
Sure, go for it

ctrlcctrlv:
OK, done :-)

https://twitter.com/FontForge/status/1164157409855561728

I wrote three tweets today, two of which were video tutorials.

I have no idea how much this actually helps the project, but we've gained &approx;100 followers since I started tweeting on the @FontForge account. So that's something I suppose, better than a dead feed

frank-trampe:
100 people is big in this business.

dscorbett:
I’m working on a branch that will add some translatable strings. How does Crowdin get the new strings? I am also fixing some typos in the existing English strings. How do the `msgid`s get updated in Crowdin without losing the existing translations?

jtanx:
> I’m working on a branch that will add some translatable strings. How does Crowdin get the new strings?

With this PR, the potfile gets uploaded to Crowdin on each push to the master branch.

Regarding the second point, I guess this describes it: https://support.crowdin.com/configuration-file/#changed-strings-update

So we should set one of those options, I guess `update_as_unapproved` sounds appropriate?

jtanx:
Github's a bit cooked right now, so the latest commit isn't showing yet, but I wasn't liking how much the Debian packaging was complicating the build/dist process for so little benefit - so I've removed the direct integration and wrote a couple of bash scripts that do the same thing.

So the process is - *always* - run off a dist archive - don't run off a git checkout (you shouldn't do this anyway, otherwise your source archive will contain the .git folder).

From an extracted dist archive, run `Packaging/debian/setup-metadata.sh` then run `debuild` yourself. For RPMs, you can use `Packaging/redhat/generate-spec.sh` to generate the specfile.

jtanx:
If there's no objections, then I'm merging this by the end of today to avoid further confusion like from #3890.

frank-trampe:
Answering yes/no questions is not a good workflow (for packaging). I guess we can fix that later, though.

jtanx:
Is that wrt the Debian stuff? I got the impression that it's only actually used by us to do the PPA which is updated so infrequently as to be inconsequential

frank-trampe:
It is, but I build for four or five different distributions at once, which is currently scriptable.

jtanx:
Given I have no access to the PPA, so I don't know the exact workflow, I'll leave that with you.

frank-trampe:
Alright, let's do it.

skef:
Oh: On the particular question of the split between in-repo and online install documentation I think the balance here is pretty good. If we get common questions we can always add to it. On the Wiki I would prefer putting the sample developer build right after the sample release build towards the top but we can discuss that elsewhere.

Issue #3880, opened 2019-08-15
CMake switchover meta-issue


skef:
1. Update Install-git.md for CMake
    1. Refer to `CMakeLists.txt`
    2. Document `-DCMAKE_INSTALL_PREFIX` and `-DCMAKE_INSTALL_MANDIR`
2. Copy hotkey file so that it works with `build/bin/fontforge`?

skef:
Maybe this issue can stick around a bit while we work out what needs to go in. Some subset of the Wiki info would be good, although it's fine to refer off to `CMakeLists.txt` for some of it. 

I think the distro installers would appreciate a note about (at least) `-DCMAKE_INSTALL_PREFIX` and `-DCMAKE_INSTALL_MANDIR` (assuming I have those right).

skef:
(Now I'm just writing random notes to @jtanx ... sorry.)

It would be convenient if the hotkey file could be copied under `build` to a place where `build/bin/fontforge` would pick it up, if that could be Humpy-Dumptied together.

skef:
Edited this into a meta-issue

jtanx:
Huh, I thought I did document CMAKE_INSTALL_PREFIX but I guess not, I've added that and the note on CMAKE_INSTALL_MANDIR (well really any installable dir from [GNUInstallDirs](https://cmake.org/cmake/help/v3.5/variable/CMAKE_INSTALL_PREFIX.html)) to the wiki.

ctrlcctrlv:
Why does it take so long to rebuild (200+ objects) just from a small change to `inc/gimage.h`?

jtanx:
Probably because all 200 source files include gimage.h in some way. Are you saying it's rebuilding more than on the autotools build?

ctrlcctrlv:
I thought it did, but I just tested both of them and the autotools build is actually slower; sorry for the noise.

ctrlcctrlv:
Sorry, this might be yet more noise, but I made a small change to the wiki to add the `-DENABLE_DEBUG=1` switch in the &ldquo;recommended for developers&rdquo; section.

https://github.com/fontforge/fontforge/wiki/CMake-guide-for-FontForge/_compare/0d3669d849c56534684394d240317d4df8bdee35...1ba621c7a23b11c56b54afaab146eca0af8dd894

[...]

PR #3881, opened 2019-08-15
Copy prefs and hotkey defaults into the build folder if they don't exist

jtanx:
For easier debugging. I was contemplating doing something similar with the theme, but I reckon it would slow down the configuration time too much, as there's so many pixmaps. Symlinking doesn't really work either as the resources file is a configured file.

Note that personally, because of that I just copy the whole share/fontforge folder from an installed instance of fontforge into the build/share/ folder to make my life easier when running directly from the build folder.

Relates to #3880

### Type of change

- **Non-breaking change**


jtanx:
Just changed it to use CMAKE_CURRENT_BINARY_DIR instead of CMAKE_BINARY dir since it was already in the share folder

Issue #3892, opened 2019-08-23
crash when introducing math kern


MdAyquassar:
The FontForge version (installed locally via `cmake .. -DCMAKE_INSTALL_PREFIX=$HOME/usr/local && make -j4 && make install`) and the operating system used:
```
$ which fontforge
/home/user/usr/local/bin/fontforge
$ fontforge --version
Copyright (c) 2000-2019. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Version: 20190801
 Based on sources from 2019-08-23 18:30 UTC-ML-D-GDK3.
 Based on source from git with hash: dd65d353d82f9eb9be70304352956f5d35a8217b
fontforge 20190801
build date: 2019-08-23 18:30 UTC
$ lsb_release -a && uname -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 19.04
Release:        19.04
Codename:       disco
Linux host 5.0.0-25-generic #26-Ubuntu SMP Thu Aug 1 12:04:58 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
$
```
1. Run `fontforge texgyretermes-math.otf` (provide the path to your copy of TeX Gyre Termes Math)
2. Click the menu entry Element -> Other Info -> Math Info...
3. Click Math Kern
4. Click New
5. Enter u1D70B
6. Click Change
7. Click the tab Textual
8.  Click Below Right
9. Click New
10. Leave the height at 0. In kern, remove 0 an enter -150 instead.
11. Click New again
12. In height of the new line, remove 0 and enter 80 instead. Try to enter -100 in kern of the new line. Sometimes the crash is observed already now.
13. If not crashed yet (and you managed to enter -100 in the previous step), click New again. In height of the new line, enter 160 instead of 0. In kern of the new line, try to enter -50. Instead, observe the crash now. I never got past here.

Output:
```
user@host:~/soft/Ryge$ fontforge /usr/share/texmf/fonts/opentype/public/tex-gyre-math/texgyretermes-math.otf &
[1] 12829
user@host:~/soft/Ryge$ Copyright (c) 2000-2019. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
[...]

PR #3893, opened 2019-08-24
mathkern fixes

jtanx:
This includes a few fixes:
* Fix a crash when editing in graphical mode and switching to textual mode (the first commit; fix invalid realloc size)
* Fix crash when interacting in textual mode - for #3892 (second commit). This looks like a copy-paste error from `mi_extensionpart` (line 130 in math.c). Each time you edited the matrix, it was calling the wrong callback function. As best I can tell, this matrix doesn't actually need any callback function, as there's nothing to interactively update as the fields are edited.
* Get rid of the pixman warnings seen (third commit). Somehow gtabset passes in a negative initial width when creating the window. This causes pixman to complain when we do `_GGDKDraw_SetOpaqueRegion`. But there's no reason to call this on child windows, so I've guarded against that).

### Type of change

- **Bug fix**

### DIFF
--- a/fontforgeexe/math.c
+++ b/fontforgeexe/math.c
@@ -138,7 +138,7 @@ static struct col_init mathkern[] = {
     COL_INIT_EMPTY
 };
 static struct matrixinit mi_mathkern =
-    { sizeof(mathkern)/sizeof(struct col_init)-1, mathkern, 0, NULL, NULL, NULL, extpart_finishedit, NULL, NULL, NULL };
+    { sizeof(mathkern)/sizeof(struct col_init)-1, mathkern, 0, NULL, NULL, NULL, NULL, NULL, NULL, NULL };
 
 
 #define CID_Exten	1000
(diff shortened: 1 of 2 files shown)


jtanx:
@MdAyquassar if you can test this that'd be great.

MdAyquassar:
I apologize - I'm new to git. To test your changes, which command do I issue instead of `git clone https://github.com/fontforge/fontforge` or `git pull` or whatever?

jtanx:
Try this

```
git remote add jtanx https://github.com/jtanx/fontforge
git fetch jtanx
git checkout mathkern
```

skef:
If you just want to play around with this code in isolation and not have it mixed up with your other repositories, you can do this in a spare directory:
```
git clone --single-branch --branch mathkern https://github.com/jtanx/fontforge
```
Build what that pulls, and then delete the directory when you're done.

MdAyquassar:
... Am checking it out now...

MdAyquassar:
Many thanks!!!!!!  Works like a charm. As far as I am concerned, feel free to close the issue report and merge your developments into the main branch.

jtanx:
Great, thanks for confirming.

PR #3903, opened 2019-08-28
AUTHOR update - also include 2005-01-01 to 2006-03-18

JoesCat:
Add recent developers since last update, plus credits
for authors/patches added for 2005jan01 to 2006mar18.

...will need to check for {2006mar18 to 2012jul31}...


ctrlcctrlv:
What does `...will need to check for {2006mar18 to 2012jul31}...` mean? Are you saying this shouldn't be merged yet as it's incomplete? Because I was just about to merge it.

ctrlcctrlv:
So is this ready to merge or will you be updating it with more authors?

PR #3906, opened 2019-08-28
Prevent static linking to Python when building shared libs

jtanx:
More often than not, linking against Python statically isn't what we want anyway, and it's also less likely that the static library has been compiled with fPIC, which is a requirement when libfontforge is built as a shared library

### Type of change
- **Non-breaking change**

See also #3902


jtanx:
I think I can do without the barbed comments thanks. Still 100x more readable than anything done under the old build system.

If every single line required comments, we'd get nowhere, the commit is descriptive enough as is.

JoesCat:
The "older" build system had dynamic linking, the "old" system removed dynamic linking.
Based on observation, commented code has a better chance of being retained and upgraded vs being discarded or replaced.
I'm still in favour of keeping dynamic linking (as in the past) for reasons such as this one.

The comment is for the benefit of those following the code later.
If someone doesn't understand the intent, and sees something as a problem, chances are it's going to get removed instead of improved.

jtanx:
It's a delusion if you seriously think that runtime linking would have been better, but I doubt I'll convince you, so we'll have to agree to disagree.

Anyway, I can tell what would have happened if you did what you suggested, it would either have (silently) chosen the wrong python to use, or not chosen any at all, leading to a way worse outcome than knowing up front it won't work at compile time.

frank-trampe:
The fact that there is an argument here highlights for me the importance of noting the reason for the restriction in the code. It doesn't need to be anything complicated, maybe something like the following.

> Static linking of Python into dynamic libraries does not make much sense and has caused a lot of build problems.

jtanx:
The only confusion here is in the comments, confusing runtime lining linking eg via dlopen vs the choice of linking to either the static or dynamic library at compile time.

The code is self explanatory.

frank-trampe:
That's not unreasonable, the build system is your territory these days, and I don't feel strongly about the matter, so I'm okay merging as-is.

jtanx:
Verified in #3902.

ctrlcctrlv:
I'm conflicted about sending this message. After all, who am I to get involved and doing so might just make things worse and make people I consider friends hate me&hellip;

I however cannot help but notice that @jtanx and @JoesCat are increasingly in conflict. I've noticed quite pointed arguments, sometimes bordering on personal attacks, across multiple PRs. I don't know if it's bad enough yet to warrant some kind of action, or who should deliver that action, but it definitely is making reading my email, a &laquo;bummer&raquo;.

I hope your relationship doesn't deteriorate further, and think Joe should be more careful when making blanket statements about things under Jeremy's purview.

frank-trampe:
I may have given too much weight to the concept of territory in my previous post. No code is beyond reasonable question. I think that the changes here were correct, but there is nothing wrong with challenging them.

I have been similarly grieved by this conflict. Lack of in-person interaction admittedly creates a lot of tonal ambiguity, but I hope that @jtanx and @JoesCat would read comments in the most positive possible way rather than the most negative possible way. Reciprocating perceived hostility puts it in a self-amplifying loop that helps nobody.

Issue #3877, opened 2019-08-15
FontForge 2019-08-01 / Mac OS X (10.14.6) Crashes 


MrBenGriffin:
I've shared several console logs on fontforge-devel@lists.sourceforge.net
My initial thought was this was to do with libPango, but it looks like it may be to do with something else..

I notice that warning dialogs which are behind other windows are sometimes not rendering - and in many ways the window focus is not working properly.

I also notice that the crashes appear to only kick in with complex files - but when they crash, they crash pretty quickly.  I'm willing to help resolve the issue - but I do not have the time to build the entire code-chain *sorry*.

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libglib-2.0.0.dylib           	0x00000001068ca92c g_slice_alloc + 92
1   libpango-1.0.0.dylib          	0x00000001063c7954 pango_font_description_copy_static + 24
2   libpangocairo-1.0.0.dylib     	0x00000001063b0c80 pango_core_text_fontset_key_init + 354
3   libpangocairo-1.0.0.dylib     	0x00000001063b09ca pango_core_text_font_map_load_fontset + 224
4   libpango-1.0.0.dylib          	0x00000001063cd229 itemize_state_process_run + 440

**But....**
fontforge(21589,0x116efb5c0) malloc: Incorrect checksum for freed object 0x7fd03176fbf0: probably modified after being freed.
Corrupt value: 0xb00007fd00000026

2   libsystem_c.dylib             	0x00007fff7dba96a6 abort + 127
3   libsystem_malloc.dylib        	0x00007fff7dcb816b malloc_vreport + 545
4   libsystem_malloc.dylib        	0x00007fff7dcd0f01 malloc_zone_error + 183
5   libsystem_malloc.dylib        	0x00007fff7dcb4947 tiny_free_list_remove_ptr + 544
6   libsystem_malloc.dylib        	0x00007fff7dcb2318 tiny_free_no_lock + 934
7   libsystem_malloc.dylib        	0x00007fff7dcb1e75 free_tiny + 480
8   libfontforgeexe.3.dylib       	0x000000010cc7ecb3 WordlistEscapedInputStringToParsedDataComplex + 1091
9   libfontforgeexe.3.dylib       	0x000000010cc7f3f8 WordlistEscapedInputStringToParsedData + 40
10  libfontforgeexe.3.dylib       	0x000000010ca59d1a CV_OnCharSelectorTextChanged + 602
[...]

PR #3914, opened 2019-08-31
 Disallow unichar_t popup messages on GGadgetData initialisation; remove gg_utf8_popup flag

jtanx:
All the popup messages were really already gettext localised strings anyway. So this removes the possibility of mistakenly not specifying the `gg_utf8_popup` flag (see #3877), and removes the need to cast to unichar_t on initialisation

### Type of change

- **Bug fix**


gdraw/ggadgets.c
```diff
@@ -887,9 +887,7 @@ GGadget *_GGadget_Create(GGadget *g, struct gwindow *base, GGadgetData *gd,void
     g->short_mask = gd->short_mask;
     g->cid = gd->cid;
     g->data = data;
-    g->popup_msg = (gd->flags&gg_utf8_popup)
-	    ? utf82u_copy((char *) gd->popup_msg)
-	    : u_copy(gd->popup_msg);
+    g->popup_msg = utf82u_copy(gd->popup_msg);
```
khaledhosny: Isn’t `popup_msg` a `char *` now? So this should be `copy()`?

jtanx:
Only on the GGadgetData, I haven't propagated that all the way through into the GGadget itself.
________________________________

PR #3915, opened 2019-08-31
Add images by reference

### LINKED ISSUE #3872: It should be possible to refer to images by reference
My PNG feature is pretty good, (or at least I think so;) even with the low compression value I chose, I get 60&ndash;90% savings in SFD file sizes.

That's all well and good, but having to rewrite the files every time is inefficient, _obviously_. It was inefficient before when it was bitmap, now it's only less inefficient.

<hr/>

So, I want to make it possible to refer to images by reference. I want to do this with my new `Image2` tag. Right now FontForge only understands `Image2` with `image/png`, but we made `Image2` to be very extensible, so this is no problem. The MIME type is the magic key, and determines the parameters and their order.

I want to add a new MIME type, `inode/symlink`. The next argument is the _expected_ MIME type of the target file, so let's say `image/jpeg`, this informs what reader FontForge should try to use, and is needed because FontForge doesn't always have libpng or libjpeg or others (it also of course eases implementation, so we don't have to read magic bytes like `PNG`, `\xff\xd8\xff\xe0` etc.). Like for `image/png` type in the 20190801 release, we put the filesize in bytes to make implementation easier. For compatibility, we put a line break first, so that it makes sense to have an `EndImage2`. Then comes a hash, a [G_CHECKSUM_SHA256](https://developer.gnome.org/glib/stable/glib-Data-Checksums.html#GChecksumType), in hex format. Then comes another linebreak and the the absolute path of the file. 

Here's an example:

```
Image2: inode/symlink image/jpeg 316689
c65574b2a30a79a72365ca3859105599c667a8d4ca35d7e5d27f63b07a5294b1
/home/fred/Workspace/chomsky/canon/B.jpg
EndImage2
```

<hr>

**Does anyone reject my planned implementation?** Please speak now before I start working on it, or forever hold your peace, as they say. This is a blocker for a new (somewhat ambitious, if I'm honest) font of mine called TT2020 so I'd like to start working on it right away.

cc: @frank-trampe @jtanx Thanks for your time!

ctrlcctrlv:
Images by reference! I've wanted this feature probably as long as I've used FontForge. 

Here's the first (public) SFD file containing images by reference: [ImagesByReference.zip](https://github.com/fontforge/fontforge/files/3562409/ImagesByReference.zip)

Here's the format I chose. The only difference to my proposal is I decided to quote the values so I could use `SFDReadUTF7Str`:

```sfd
StartChar: B
Encoding: 66 66 1
Width: 1000
VWidth: 0
LayerCount: 2
Back
Image2: reference 0 800 2.08333 2.08333
"a390023f572e6c79491d916e72f0444fcfa139cb84285e7dedc1e2b6b9efb158"
"B2.png"
EndImage2
```

This is ready for testing a bit more widely than on my PC, but I haven't yet tested it as much as I could have. I promised my wife I'd submit this Saturday night and take Sunday off, so I'm submitting a bit early and not using Sunday for testing. :wink: On Monday I'll deal with comments if there are any, or if not due some more robust testing. This was a ton of work as you can imagine!

UI images:

2019-08-31-194828_1478x337_scrot
2019-08-31-194932_2354x1408_scrot

Python script example:

```python
import fontforge

f = fontforge.fonts()[0]
f.save("/tmp/test.sfd") # We save here so that the imported file will be relative to /tmp.
f.createChar(ord("A"))
f["A"].right_side_bearing = 100
f["A"].importOutlines("/tmp/B.jpg", "imagereferenceonly")

f.save("/tmp/test.sfd")
```

I haven't yet added Python functions for "Link Reference" and "Unlink Reference", partially because there doesn't seem to be an API for images. I'd like @skef's input on what that should look like.

This closes #3872.

### DIFF
--- a/fontforge/cvimages.c
+++ b/fontforge/cvimages.c
@@ -953,9 +953,8 @@ void SCAddScaleImage(SplineChar *sc,GImage *image,int doclear, int layer) {
     SCInsertImage(sc,image,scale,sc->parent->ascent,0,layer);
 }
 
-int FVImportImages(FontViewBase *fv,char *path,int format,int toback, int flags) {
+int FVImportImages(FontViewBase *fv,char *path,int format,int toback, bool reference, int flags) {
     GImage *image;
-    /*struct _GImage *base;*/
     int tot;
     char *start = path, *endpath=path;
     int i;
@@ -971,7 +970,7 @@ int FVImportImages(FontViewBase *fv,char *path,int format,int toback, int flags)
 	    if ( image==NULL ) {
 		ff_post_error(_("Bad image file"),_("Bad image file: %.100s"),start);
 return(false);
-	    }
+        } 
 	    ++tot;
 	    SCAddScaleImage(sc,image,true,toback?ly_back:ly_fore);
 	} else if ( format==fv_svg ) {
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -8269,6 +8269,7 @@ static struct flaglist import_ps_flags[] = {
     { "removeoverlap", 0 },			/* Obsolete */
     { "handle_eraser", sf_handle_eraser },
     { "correctdir",  sf_correctdir },
+    { "imagereferenceonly",  sf_imagereferenceonly },
     FLAGLIST_EMPTY /* Sentinel */
 };
 
@@ -8293,8 +8294,8 @@ return( NULL );
     pt = strrchr(locfilename,'.');
     if ( pt==NULL ) pt=locfilename;
 
-    if ( strcasecmp(pt,".eps")==0 || strcasecmp(pt,".ps")==0 || strcasecmp(pt,".art")==0 ) {
 	int psflags = FlagsFromTuple(flags,import_ps_flags,"PostScript import flag");
+    if ( strcasecmp(pt,".eps")==0 || strcasecmp(pt,".ps")==0 || strcasecmp(pt,".art")==0 ) {
 	if ( psflags==FLAG_UNKNOWN ) {
 	    free(locfilename);
 return( NULL );
@@ -8329,6 +8330,7 @@ return(NULL);
 	}
 	if ( !sc->layers[ly].background )
 	    ly = ly_back;
+    if (psflags & sf_imagereferenceonly) GImageMakeReference(image->u.image, locfilename, sc->parent->filename);
 	SCAddScaleImage(sc,image,false,ly);
     }
     free( locfilename );
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -844,6 +844,16 @@ static void SFDDumpAnchorPoints(FILE *sfd,AnchorPoint *ap) {
     }
 }
 
+// This is a utility function, used only in this file, which dives deep into a
+// ImageList structure to figure out if it is supposed to be saved as a 
+// reference or not.
+static bool ImageListShouldReference(ImageList *il) {
+    GImage *image = il->image;
+    struct _GImage *base = image->list_len==0?image->u.image:image->u.images[0];
+    
+    return base->refdata.reference;
+}
+
 /* Run length encoding */
 /* We always start with a background pixel(1), each line is a series of counts */
 /*  we alternate background/foreground. If we can't represent an entire run */
@@ -1019,12 +1029,13 @@ void SFDDumpUndo(FILE *sfd,SplineChar *sc,Undoes *u, const char* keyPrefix, int 
 static void SFDDumpImage(FILE *sfd,ImageList *img) {
     GImage *image = img->image;
     struct _GImage *base = image->list_len==0?image->u.image:image->u.images[0];
+
     struct enc85 enc;
     int rlelen;
     uint8 *rle;
     int i;
-
     rle = image2rle(base,&rlelen);
+
     fprintf(sfd, "Image: %d %d %d %d %d %x %g %g %g %g %d\n",
 	    (int) base->width, (int) base->height, base->image_type,
 	    (int) (base->image_type==it_true?3*base->width:base->bytes_per_line),
@@ -1077,6 +1088,18 @@ static void SFDDumpImage(FILE *sfd,ImageList *img) {
     }
     SFDEnc85EndEnc(&enc);
     fprintf(sfd,"\nEndImage\n" );
+}
+
+// This function should only be called if you know the image is a reference.
+static void SFDDumpImageReference(FILE *sfd,ImageList *img) {
+    GImage* image = img->image;
+    struct _GImage *base = image->list_len==0?image->u.image:image->u.images[0];
+    fprintf(sfd, "Image2: reference %g %g %g %g\n",
+        (double) img->xoff, (double) img->yoff, (double) img->xscale, (double) img->yscale );
+    SFDDumpUTF7Str(sfd, base->refdata.hash);
+    fprintf(sfd, "\n");
+    SFDDumpUTF7Str(sfd, base->refdata.filename);
+    fprintf(sfd, "\nEndImage2\n");
 }
 
 #ifndef _NO_LIBPNG
@@ -1687,12 +1710,16 @@ static void SFDDumpChar(FILE *sfd,SplineChar *sc,EncMap *map,int *newgids,int to
 		fprintf(sfd, "Layer: %d\n", i );
 	}
 	for ( img=sc->layers[i].images; img!=NULL; img=img->next )
+        if (ImageListShouldReference(img)) {
+            SFDDumpImageReference(sfd,img);
+        } else {
 #ifndef _NO_LIBPNG
         if (WritePNGInSFD)
 	    SFDDumpImagePNG(sfd,img);
         else
 #endif
-	    SFDDumpImage(sfd,img);
+            SFDDumpImage(sfd,img);
+        }
 	if ( sc->layers[i].splines!=NULL ) {
 	    fprintf(sfd, "SplineSet\n" );
 	    SFDDumpSplineSet(sfd,sc->layers[i].splines,sc->layers[i].order2);
@@ -3631,9 +3658,8 @@ static void rle2image(struct enc85 *dec,int rlelen,struct _GImage *base) {
     }
 }
 
-#ifndef _NO_LIBPNG
 
-enum MIME { UNKNOWN, PNG }; // We only understand PNG for now.
+enum MIME { UNKNOWN, PNG, REFERENCE }; // We only understand PNG for now.
 
 enum MIME SFDGetImage2MIME(FILE *sfd) {
     char mime[128];
@@ -3643,14 +3669,18 @@ enum MIME SFDGetImage2MIME(FILE *sfd) {
         return UNKNOWN;
     }
 
-    if ( !(strmatch(mime, "image/png")==0) ) {
-        IError("MIME type received—%s—is not recognized", mime);
-        return UNKNOWN;
+    if ( strmatch(mime, "image/png")==0 ) {
+        return PNG;
     }
 
-    return PNG;
+    if ( strmatch(mime, "reference")==0 ) {
+        return REFERENCE;
+    }
+
+    return UNKNOWN;
 }
 
+#ifndef _NO_LIBPNG
 static ImageList *SFDGetImagePNG(FILE *sfd) {
     int pnglen;
     ImageList *img;
@@ -3695,6 +3725,51 @@ static ImageList *SFDGetImagePNG(FILE *sfd) {
     return img;
 }
 #endif
+
+static ImageList *SFDGetImageReference(FILE *sfd) {
+    ImageList *img = calloc(1,sizeof(ImageList));
+    int ch;
+
+    getreal(sfd,&img->xoff);
+    getreal(sfd,&img->yoff);
+    getreal(sfd,&img->xscale);
+    getreal(sfd,&img->yscale);
+
+    if (ch = nlgetc(sfd) != '\n') { ungetc(ch, sfd); return NULL; }
+
+    char* hex = SFDReadUTF7Str(sfd);
+
+    if (ch = nlgetc(sfd) != '\n') { ungetc(ch, sfd); return NULL; }
+
+    char* fn = SFDReadUTF7Str(sfd);
+
+    if (fn == NULL) {
+        IError(_("Expected filename which wasn't there, file corrupt"));
+        return NULL;
+    }
+
+    char* hn = HashFile(fn);
+
+    if (strcmp(hex,hn)!=0) LogError(_("Content of file %s no longer matches what it was when SFD file saved"), fn);
+
+    GImage* image = GImageRead(fn);
+
+    TRACE("SFDGetImageReference: Hex is %s and fn is %s\n", hex, fn);
+
+    if (image == NULL) {
+        IError(_("Failed to read image from file %s"), fn);
+        return NULL;
+    }
+
+    struct _GImage *base = image->list_len==0?image->u.image:image->u.images[0];
+
+    base->refdata.reference = true;
+    base->refdata.filename = fn;
+    base->refdata.hash = hex;
+
+    img->image=image;
+    return img;
+}
 
 static ImageList *SFDGetImage(FILE *sfd) {
     /* We've read the image token */
@@ -4353,7 +4428,16 @@ Undoes *SFDGetUndo( FILE *sfd, SplineChar *sc,
 		    }
 		} else 
 #endif
-	    {
+        if (mime == REFERENCE) {
+		    ImageList *img = SFDGetImageReference(sfd);
+		    if (img != NULL) {
+			if ( !u->u.state.images )
+			    u->u.state.images = img;
+			else
+			    lasti->next = img;
+			lasti = img;
+		    }
+        } else {
 		LogError(_("Image2 skipped as it uses an unsupported image type"));
 		const char* im2_terminator[] = { "EndImage2", 0 };
 		SFDConsumeUntil(sfd, im2_terminator);
@@ -5724,8 +5808,8 @@ return( NULL );
 	    lasti = img;
 	    }
 	} else if ( strmatch(tok,"Image2:")==0 ) {
-#ifndef _NO_LIBPNG
 	    enum MIME mime = SFDGetImage2MIME(sfd);
+#ifndef _NO_LIBPNG
 	    if (mime == PNG) {
 		int ly = current_layer;
 		if ( !multilayer && !sc->layers[ly].background ) ly = ly_back;
@@ -5739,10 +5823,21 @@ return( NULL );
 		}
 	    } else
 #endif
-	    {
-	    LogError(_("Image2 skipped as it uses an unsupported image type"));
-	    const char* im2_terminator[] = { "EndImage2", 0 };
-	    SFDConsumeUntil(sfd, im2_terminator);
+        if (mime == REFERENCE) {
+		int ly = current_layer;
+		if ( !multilayer && !sc->layers[ly].background ) ly = ly_back;
+		img = SFDGetImageReference(sfd);
+		if (img != NULL) {
+		    if ( sc->layers[ly].images==NULL )
+			sc->layers[ly].images = img;
+		    else
+			lasti->next = img;
+		    lasti = img;
+		}
+        } else {
+            LogError(_("Image2 skipped as it uses an unsupported image type"));
+            const char* im2_terminator[] = { "EndImage2", 0 };
+            SFDConsumeUntil(sfd, im2_terminator);
 	    }
 	} else if ( strmatch(tok,"PickledData:")==0 ) {
 	    if (current_layer < sc->layer_cnt) {
--- a/fontforge/splinefont.h
+++ b/fontforge/splinefont.h
@@ -2411,7 +2411,7 @@ extern int BpWithin(BasePoint *first, BasePoint *mid, BasePoint *last);
     /* Colinear & between */
 
 enum psstrokeflags { /* sf_removeoverlap=2,*/ sf_handle_eraser=4,
-	sf_correctdir=8, sf_clearbeforeinput=16 };
+	sf_correctdir=8, sf_clearbeforeinput=16, sf_imagereferenceonly=32 };
 
 
 extern char *ToAbsolute(char *filename);
--- a/fontforgeexe/cvgetinfo.c
+++ b/fontforgeexe/cvgetinfo.c
@@ -67,6 +67,7 @@ typedef struct gidata {
     GGadget *group1ret, *group2ret;
 } GIData;
 
+// PointGetInfo
 #define CID_BaseX	2001
 #define CID_BaseY	2002
 #define CID_NextXOff	2003
@@ -105,6 +106,7 @@ typedef struct gidata {
 #define CID_SpiroRight	2055
 #define CID_TabSet	2100
 
+// ApGetInfo
 #define CID_X		3001
 #define CID_Y		3002
 #define CID_NameList	3003
@@ -120,6 +122,11 @@ typedef struct gidata {
 #define CID_Delete	3013
 #define CID_New		3014
 #define CID_MatchPt	3015
+
+// ImgGetInfo
+#define CID_DataSourceLabel	4001
+#define CID_LinkUnlinkButton	4002
+#define CID_II_HVBox	4003
 
 #define RI_Width	225
 #define RI_Height	246
@@ -141,6 +148,39 @@ static int GI_Cancel(GGadget *g, GEvent *e) {
 	ci->done = true;
     }
 return( true );
+}
+
+static void II_Display(GIData* gi);
+static int GI_LinkUnlinkReference(GGadget *g, GEvent *e) {
+    if ( e->type!=et_controlevent || e->u.control.subtype != et_buttonactivate ) {
+        return false;
+    }
+	GIData *gi = GDrawGetUserData(GGadgetGetWindow(g));
+
+    ImageList* img = gi->img;
+    struct _GImage *base = img->image->list_len==0?
+	    img->image->u.image:img->image->u.images[0];
+
+    if (base->refdata.reference) { // Unlink reference
+        GImageUnmakeReference(base);
+    } else { // Link reference
+        char* filename = gwwv_open_filename(_("Where does this image exist on disk?"),NULL,"*",NULL);
+        if (filename != NULL) {
+            GImage* image = GImageRead(filename);
+
+            if (image == NULL) {
+                gwwv_post_error(_("Cannot link reference"),_("FontForge failed to interpret the file you chose as an image."));
+            } else
+            if (GImageSame(image, img->image)) {
+                GImageMakeReference(base, filename, gi->cv->b.fv->sf->filename);
+            } else {
+                gwwv_post_error(_("Cannot link reference"),_("The image you chose does not contain the same data as the one in the SFD; cannot link reference."));
+            }
+        }
+    }
+
+    II_Display(gi);
+    return true;
 }
 
 static int GI_TransChange(GGadget *g, GEvent *e) {
@@ -648,11 +688,53 @@ static void RefGetInfo(CharView *cv, RefChar *ref) {
     GDrawDestroyWindow(gi.gw);
 }
 
+static void II_Display(GIData* gi) {
+    GGadget* lub = GWidgetGetControl(gi->gw, CID_LinkUnlinkButton);
+    GGadget* dsl = GWidgetGetControl(gi->gw, CID_DataSourceLabel);
+    GGadget* hvbox = GWidgetGetControl(gi->gw, CID_II_HVBox);
+
+    ImageList* img = gi->img;
+    struct _GImage *base = img->image->list_len==0?
+	    img->image->u.image:img->image->u.images[0];
+
+    const char* ur = _("_Unlink Reference");
+    const char* lr = _("_Link Reference");
+    const char* refstr = _("Data in file: ");
+
+    char* refbuf;
+    if (base->refdata.reference) {
+        // 4 is for (, ), ' ', and \0
+        refbuf = calloc(strlen(refstr)+strlen(base->refdata.filename)
+                       +SHA256_DIGEST_LEN*2+4, sizeof(char));
+
+        sprintf(refbuf, refstr);
+        strcat(refbuf, base->refdata.filename);
+        strcat(refbuf, " (");
+        strcat(refbuf, base->refdata.hash);
+        strcat(refbuf, ")");
+    } else {
+        refbuf = _("Data is in SFD file");
+    }
+
+    GGadgetSetTitle8(dsl, refbuf);
+
+    if (base->refdata.reference) {
+        GGadgetSetTitle8WithMn(lub, ur);
+    } else {
+        GGadgetSetTitle8WithMn(lub, lr);
+    }
+
+    // These need to be here, and not in ImgGetInfo, because when a reference is unlinked
+    // the window won't resize smaller on its own without them.
+    GHVBoxSetExpandableRow(hvbox,gb_expandglue);
+    GHVBoxFitWindow(hvbox);
+}
+
 static void ImgGetInfo(CharView *cv, ImageList *img) {
     static GIData gi;
     GRect pos;
     GWindowAttrs wattrs;
-    GGadgetCreateData gcd[12], boxes[3], *varray[11], *harray[6];
+    GGadgetCreateData gcd[12], boxes[5], *varray[16], *harray[6], *harray2[6];
     GTextInfo label[12];
     char posbuf[100], scalebuf[100], sizebuf[100];
     struct _GImage *base = img->image->list_len==0?
@@ -678,6 +760,9 @@ static void ImgGetInfo(CharView *cv, ImageList *img) {
 
 	memset(&gcd,0,sizeof(gcd));
 	memset(&label,0,sizeof(label));
+	memset(&varray,0,sizeof(varray));
+	memset(&harray,0,sizeof(harray));
+	memset(&harray2,0,sizeof(harray2));
 
 	sprintf( posbuf, _("Image at:      (%.0f,%.0f)"), (double) img->xoff,
 		(double) (img->yoff-GImageGetHeight(img->image)*img->yscale));
@@ -698,6 +783,23 @@ static void ImgGetInfo(CharView *cv, ImageList *img) {
 	gcd[1].creator = GLabelCreate;
 	varray[2] = &gcd[1]; varray[3] = NULL;
 
+	label[4].text = (unichar_t *) "";
+	label[4].text_is_1byte = true;
+	gcd[4].gd.label = &label[4];
+	gcd[4].gd.flags = gg_enabled|gg_visible;
+	gcd[4].gd.cid = CID_DataSourceLabel;
+	gcd[4].creator = GLabelCreate;
+    harray2[0] = &gcd[4]; harray2[1] = &gcd[5]; harray[3] = NULL;
+
+	label[5].text = (unichar_t *) "";
+	label[5].text_is_1byte = true;
+	label[5].text_in_resource = true;
+	gcd[5].gd.label = &label[5];
+	gcd[5].gd.cid = CID_LinkUnlinkButton;
+	gcd[5].gd.handle_controlevent = GI_LinkUnlinkReference;
+	gcd[5].gd.flags = gg_enabled|gg_visible;
+	gcd[5].creator = GButtonCreate;
+
 	sprintf( sizebuf, _("Image Size:  %d x %d  pixels"), (int) base->width, (int) base->height );
 	label[2].text = (unichar_t *) sizebuf;
 	label[2].text_is_1byte = true;
@@ -706,10 +808,7 @@ static void ImgGetInfo(CharView *cv, ImageList *img) {
 	gcd[2].gd.flags = gg_enabled|gg_visible;
 	gcd[2].creator = GLabelCreate;
 	varray[4] = &gcd[2]; varray[5] = NULL;
-	varray[6] = GCD_Glue; varray[7] = NULL;
 
-	gcd[3].gd.pos.x = (II_Width-GIntGetResource(_NUM_Buttonsize)*100/GIntGetResource(_NUM_ScaleFactor)-6)/2; gcd[3].gd.pos.y = II_Height-32-3;
-	gcd[3].gd.pos.width = -1; gcd[3].gd.pos.height = 0;
 	gcd[3].gd.flags = gg_visible | gg_enabled | gg_but_default | gg_but_cancel;
 	label[3].text = (unichar_t *) _("_OK");
 	label[3].text_is_1byte = true;
@@ -718,24 +817,27 @@ static void ImgGetInfo(CharView *cv, ImageList *img) {
 	gcd[3].gd.label = &label[3];
 	gcd[3].gd.handle_controlevent = GI_Cancel;
 	gcd[3].creator = GButtonCreate;
-	harray[0] = GCD_Glue; harray[1] = &gcd[3]; harray[2] = GCD_Glue; harray[3] = NULL;
-	varray[8] = &boxes[2]; varray[9] = NULL;
-	varray[10] = NULL;
+
+	harray[0] = GCD_Glue; harray[1] = &gcd[3]; harray[2] = GCD_Glue; harray[3] = GCD_Glue; harray[4] = NULL;
+	varray[6] = &boxes[2]; varray[7] = NULL; varray[8] = &boxes[4]; varray[9] = NULL;
 
 	memset(boxes,0,sizeof(boxes));
-	boxes[0].gd.pos.x = boxes[0].gd.pos.y = 2;
+	boxes[0].gd.pos.width = -1; boxes[0].gd.pos.height = 0;
 	boxes[0].gd.flags = gg_enabled|gg_visible;
 	boxes[0].gd.u.boxelements = varray;
+	boxes[0].gd.cid = CID_II_HVBox;
 	boxes[0].creator = GHVGroupCreate;
 
 	boxes[2].gd.flags = gg_enabled|gg_visible;
-	boxes[2].gd.u.boxelements = harray;
+	boxes[2].gd.u.boxelements = harray2;
 	boxes[2].creator = GHBoxCreate;
 
+	boxes[4].gd.flags = gg_enabled|gg_visible;
+	boxes[4].gd.u.boxelements = harray;
+	boxes[4].creator = GHBoxCreate;
+
 	GGadgetsCreate(gi.gw,boxes);
-	GHVBoxSetExpandableRow(boxes[0].ret,gb_expandglue);
-	GHVBoxSetExpandableCol(boxes[2].ret,gb_expandglue);
-	GHVBoxFitWindow(boxes[0].ret);
+    II_Display(&gi);
 
     GWidgetHidePalettes();
     GDrawSetVisible(gi.gw,true);
--- a/gdraw/gfilechooser.c
+++ b/gdraw/gfilechooser.c
@@ -1114,7 +1114,6 @@ void GFileChooserPopupCheck(GGadget *g,GEvent *e) {
     int inside=false;
 
     if ( e->type == et_mousemove && (e->u.mouse.state&ksm_buttons)==0 ) {
-	GGadgetEndPopup();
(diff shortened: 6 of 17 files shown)


gutils/fsys.c
```diff
@@ -1159,3 +1159,37 @@ char *GFileDirName(const char *path) {
...
+    return ret;
+}
+
+// If we have two files:
+// * /tmp/ex/test.sfd (file1)
+// * /tmp/ex/images/1.jpg (to_relative)
+//
+// The return is images/1.jpg
+char* GFileRelativize(char* file1, char* to_relative) {
+   GFile *gf_file1 = g_file_new_for_path(file1);
```
jtanx: It's a shame that `g_canonicalize_filename` is only available in relatively recent versions of glib, otherwise I'd suggest using that.

Also, I do see how relative paths can be useful (i.e. under version control), but I'm also not fully convinced it's the right thing to do, especially when it only does this if the sfd is already saved. It just seems much too fragile, and a bit too 'magic' overall.

gutils/fsys.c
```diff
@@ -1159,3 +1159,37 @@ char *GFileDirName(const char *path) {
...
+
+   char* ex = g_file_get_relative_path(gf_file1_parent, gf_to_relative);
+
+   g_object_unref(gf_file1);
+   g_object_unref(gf_file1_parent);
+   g_object_unref(gf_to_relative);
+
+   if (ex==NULL) {
+       // Impossible to relativize; use absolute path.
+       return to_relative;
```
jtanx: The semantics of this function should be clarified. You're returning this without copying here, which implies the return shouldn't be freed. Yet you also just return `ex`, which does need to be freed (and freed with `g_free`).

gutils/gimage.c
```diff
@@ -44,6 +45,9 @@ GImage *GImageCreate(enum image_type type, int32 width, int32 height) {
 	goto errorGImageCreate;
 
     gi->u.image = base;
+    base->refdata.reference = false;
+    base->refdata.filename = NULL;
+    base->refdata.hash = "0";
```
jtanx: At the very least, this should be NULL and not `"0"` otherwise you'll run into issues in all the other places that you've assumed this to be a dynamically allocated object.

But I'd also suggest just working internally with the raw representation internally (which should be a fixed length of bytes - [g_checksum_digest](https://developer.gnome.org/glib/stable/glib-Data-Checksums.html#g-checksum-get-digest)), meaning you don't need to do dynamic allocation at all.

gutils/gimage.c
```diff
@@ -69,6 +73,38 @@ GImage *GImageCreate(enum image_type type, int32 width, int32 height) {
...
+    char* relativized = NULL;
+
+    if (relative_to != NULL) {
+        relativized = GFileRelativize(relative_to, to_file);
+    }
+
+    if (relativized != NULL) {
+        base->refdata.filename = relativized;
+    } else {
+        char* fn = malloc(strlen(to_file));
```
jtanx: Again, why aren't you using `copy`

gutils/gimage.c
```diff
@@ -69,6 +73,38 @@ GImage *GImageCreate(enum image_type type, int32 width, int32 height) {
...
+        base->refdata.filename = fn;
+    }
+
+    return true;
+}
+
+extern bool GImageUnmakeReference(struct _GImage *base) {
+    base->refdata.reference = false;
+    base->refdata.filename = NULL;
+    free(base->refdata.filename);
```
jtanx: The order of nulling and freeing is wrong.

gutils/gutils.c
```diff
@@ -70,3 +70,23 @@ time_t GetST_MTime(struct stat s) {
 
 	return st_time;
 }
+
+char* FF_HashData(unsigned char* input, int len) {
+    GChecksum* gcs = g_checksum_new(G_CHECKSUM_SHA256);
+    g_checksum_update(gcs, input, len);
+    const char* gcs_cs = g_checksum_get_string(gcs);
+    char* cs = malloc(strlen(gcs_cs)+1);
```
jtanx: Use `copy`. That or don't get the string representation, and use store the raw bytes instead.

gutils/gutils.c
```diff
@@ -70,3 +70,23 @@ time_t GetST_MTime(struct stat s) {
...
+    g_checksum_free(gcs);
+    return cs;
+}
+
+extern char* FF_HashFile(char* filename) {
+    char *contents;
+    gsize length;
+    g_file_get_contents(filename, &contents, &length, NULL);
+    TRACE("Hashing %s of length %d...\n", filename, length);
+    const char* cs = FF_HashData(contents, length);
```
jtanx: Why mark this const if you're just going to throw that away with another cast below?

gutils/gimage.c
```diff
@@ -416,3 +460,60 @@ Color GImageGetPixelColor(GImage *image,int x, int y) {
     struct _GImage *base = image->list_len==0?image->u.image:image->u.images[0];
 return( _GImageGetPixelColor(base,x,y));
 }
+
+const char* GImageHash(GImage *img) {
```
jtanx: So what is this used for

gutils/gutils.c
```diff
@@ -70,3 +70,23 @@ time_t GetST_MTime(struct stat s) {
...
+    char* cs = malloc(strlen(gcs_cs)+1);
+    strcpy(cs, gcs_cs);
+    g_checksum_free(gcs);
+    return cs;
+}
+
+extern char* FF_HashFile(char* filename) {
+    char *contents;
+    gsize length;
+    g_file_get_contents(filename, &contents, &length, NULL);
```
jtanx: You're missing error checking if this fails.

Also, just pointing out but there also already exists `GFileReadAll`.

gutils/gutils.c
```diff
@@ -70,3 +70,23 @@ time_t GetST_MTime(struct stat s) {
 
 	return st_time;
 }
+
+char* FF_HashData(unsigned char* input, int len) {
+    GChecksum* gcs = g_checksum_new(G_CHECKSUM_SHA256);
```
jtanx: Add error checking.

gutils/fsys.c
```diff
@@ -1159,3 +1159,37 @@ char *GFileDirName(const char *path) {
...
+//
+// The return is images/1.jpg
+char* GFileRelativize(char* file1, char* to_relative) {
+   GFile *gf_file1 = g_file_new_for_path(file1);
+   GFile *gf_file1_parent = g_file_get_parent(gf_file1);
+   GFile *gf_to_relative = g_file_new_for_path(to_relative);
+
+   char* ex = g_file_get_relative_path(gf_file1_parent, gf_to_relative);
+
+   g_object_unref(gf_file1);
```
jtanx: If you absolutely must use this gio GFile stuff, then you'll need to change this to include `gobject` in the COMPONENTS list:

https://github.com/fontforge/fontforge/blob/decf26fd43251502b66c7264dc41fec3d9937289/CMakeLists.txt#L81

gutils/gimage.c
```diff
@@ -416,3 +460,60 @@ Color GImageGetPixelColor(GImage *image,int x, int y) {
...
+
+    int datalen1 = base1->bytes_per_line * base1->height;
+    int datalen2 = base2->bytes_per_line * base2->height;
+
+    if (datalen1 != datalen2) return false;
+
+    // If the colors they use are the same and the number of pixels are too,
+    // check actual data.
+
+    for (int i = 0; i <= datalen1; i++) {
```
jtanx: Surely this can be replaced by a memcmp

fontforge/sfd.c
```diff
@@ -3696,6 +3726,51 @@ static ImageList *SFDGetImagePNG(FILE *sfd) {
...
+    int ch;
+
+    getreal(sfd,&img->xoff);
+    getreal(sfd,&img->yoff);
+    getreal(sfd,&img->xscale);
+    getreal(sfd,&img->yscale);
+
+    if (ch = nlgetc(sfd) != '\n') { ungetc(ch, sfd); return NULL; }
+
+    char* hex = SFDReadUTF7Str(sfd);
```
jtanx: Check this function please, I think you're leaking multiple dynamically allocated objects.

fontforgeexe/cvgetinfo.c
```diff
@@ -718,24 +817,27 @@ static void ImgGetInfo(CharView *cv, ImageList *img) {
 	gcd[3].gd.label = &label[3];
 	gcd[3].gd.handle_controlevent = GI_Cancel;
 	gcd[3].creator = GButtonCreate;
-	harray[0] = GCD_Glue; harray[1] = &gcd[3]; harray[2] = GCD_Glue; harray[3] = NULL;
-	varray[8] = &boxes[2]; varray[9] = NULL;
-	varray[10] = NULL;
+
+	harray[0] = GCD_Glue; harray[1] = &gcd[3]; harray[2] = GCD_Glue; harray[3] = GCD_Glue; harray[4] = NULL;
+	varray[6] = &boxes[2]; varray[7] = NULL; varray[8] = &boxes[4]; varray[9] = NULL;
```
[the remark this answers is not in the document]
ctrlcctrlv: I'm sure. The NULL doesn't terminate the array, it terminates the HVBox cell. Without all these NULLs, which predate my work, the ImgGetInfo box will be short and wide, not tall and thin.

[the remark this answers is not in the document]
ctrlcctrlv: (I'm not supposed to be working today, but I read your review and I think everything else you wrote is reasonable. Tomorrow when I actually go to implement your requested changes I'll know if that holds.)

gdraw/gfilechooser.c
```diff
@@ -1114,7 +1114,6 @@ void GFileChooserPopupCheck(GGadget *g,GEvent *e) {
     int inside=false;
 
     if ( e->type == et_mousemove && (e->u.mouse.state&ksm_buttons)==0 ) {
-	GGadgetEndPopup();
```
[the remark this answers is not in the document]
ctrlcctrlv: It doesn't work to put a popup on the Reference checkbox in import dialog without it. There are no ill effects (that I found). This change allows for more complex file choosers in the future, which I think is a good thing.

gutils/fsys.c
ctrlcctrlv: Will you accept that our should I work harder to avoid needing Gio?

fontforge/cvimages.c
```diff
@@ -953,9 +953,8 @@ void SCAddScaleImage(SplineChar *sc,GImage *image,int doclear, int layer) {
     SCInsertImage(sc,image,scale,sc->parent->ascent,0,layer);
 }
 
-int FVImportImages(FontViewBase *fv,char *path,int format,int toback, int flags) {
+int FVImportImages(FontViewBase *fv,char *path,int format,int toback, bool reference, int flags) {
```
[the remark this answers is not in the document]
ctrlcctrlv: It looks like I did it to add this to open an avenue to adding a reference-only flag to `scripting.c` fn `bImport`...but then I didn't actually follow through and it turns out that `flags` is actually the way to do this as I found out by editing `python.c`. So I'll revert this, sorry for the confusion.

fontforge/sfd.c
```diff
@@ -844,6 +844,16 @@ static void SFDDumpAnchorPoints(FILE *sfd,AnchorPoint *ap) {
     }
 }
 
+// This is a utility function, used only in this file, which dives deep into a
+// ImageList structure to figure out if it is supposed to be saved as a 
+// reference or not.
+static bool ImageListShouldReference(ImageList *il) {
+    GImage *image = il->image;
+    struct _GImage *base = image->list_len==0?image->u.image:image->u.images[0];
```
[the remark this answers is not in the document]
ctrlcctrlv: The structure:
```
struct _GImage *base = image->list_len==0?image->u.image:image->u.images[0];
```
is used all throughout the code, not only by me. In my experience...based on my tests, there will never be more than one image in the ImageList, even if there is more than one image on the layer. My branch handles that case fine, and the ImageList has one element each time:

```sfd
StartChar: A
Encoding: 65 65 0
Width: 100
VWidth: 0
Flags: WO
LayerCount: 2
Back
Image2: reference -1077 833 1.41243 1.41243
"00c5a7db144f0612bc3bed7952a07bc6fde43d0e5b75125518e89d271f8957f1"
"B.jpg"
EndImage2
Image2: reference -39 743 1.41243 1.41243
"00c5a7db144f0612bc3bed7952a07bc6fde43d0e5b75125518e89d271f8957f1"
"B.jpg"
EndImage2
EndChar
```

I'm kind of afraid to change this prototype...George probably had a reason for doing it that way.

fontforgeexe/cvgetinfo.c
```diff
@@ -648,11 +688,53 @@ static void RefGetInfo(CharView *cv, RefChar *ref) {
...
+	    img->image->u.image:img->image->u.images[0];
+
+    const char* ur = _("_Unlink Reference");
+    const char* lr = _("_Link Reference");
+    const char* refstr = _("Data in file: ");
+
+    char* refbuf;
+    if (base->refdata.reference) {
+        // 4 is for (, ), ' ', and \0
+        refbuf = calloc(strlen(refstr)+strlen(base->refdata.filename)
```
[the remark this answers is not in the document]
ctrlcctrlv: Because I didn't know about it! Fixing it right now, thanks for the tip. :-)

gutils/gutils.c
ctrlcctrlv: It's much easier to just use the string representation. I don't really ever need the bytes, and the strings are still comparable. I considered storing the raw bytes instead and converting them and don't think it's worth it.

So I'll use `copy` :smile:

fontforgeexe/cvgetinfo.c
```diff
@@ -648,11 +688,53 @@ static void RefGetInfo(CharView *cv, RefChar *ref) {
...
+        sprintf(refbuf, refstr);
+        strcat(refbuf, base->refdata.filename);
+        strcat(refbuf, " (");
+        strcat(refbuf, base->refdata.hash);
+        strcat(refbuf, ")");
+    } else {
+        refbuf = _("Data is in SFD file");
+    }
+
+    GGadgetSetTitle8(dsl, refbuf);
```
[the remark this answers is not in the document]
ctrlcctrlv: I think I fixed this in my latest commits but am not sure

gutils/gimage.c
ctrlcctrlv: Nothing, sorry. I deleted it. It was from an earlier stage of development.

ctrlcctrlv:
Hi @jtanx. I pushed a bunch, feel free to look at it but I know there's still more to do, so don't need a review urgently. Thanks for your early and prompt review, I fixed a bunch of things you noticed. I hope you like this feature.

gutils/fsys.c
ctrlcctrlv: Decided to require gobject for now, if that's okay feel lmk but I'm going to mark this resolved for now.

ctrlcctrlv:
Oh! Sorry, I forgot to mention&mdash;just so you know @jtanx, your comment about tests was well received. I wrote some tests today but they aren't done yet so I haven't committed them. I'll add them before re-requesting a review from you. :+1:

ctrlcctrlv:
@jtanx In order to do the Python stuff right, I have to add a brand new type, `PyFFImageType`. Can I please make that dependent on this PR and put it in this branch? I feel it's waste for me to have to make that separable from references, since the whole reason I'm adding it is so I can add Python tests for image reference link/unlink, and so users can e.g.:

```python
import fontforge
import glob
f=fontforge.font()
f.createChar(ord("A"))
A=f["A"]
A.importOutlines("/tmp/test.jpg") # Now returns nothing, so I don't think I should change that as it might break scripts.
for i in A.images() # new proposed iterator
  print(repr(i)) # new proposed type <Image on glyph Y>
  for im in glob.glob('images/*'):
    linked = i.linkReference(im) # new proposed function, try to link reference
    if linked: break
```

PyFFImageType should have xoff, yoff, xscale, yscale able to be set, it should be possible to link and unlink it, and it should be possible to compare it (GImageSame). It might also need to be possible to hash it, so I might need to bring back GImageHash, jury's out on that though. Should users be able to access the bitmap data? I think that's pointless because Python users should already have the image in another file and can use other utilities, but await yoru comment.

gutils/fsys.c
jtanx: Ok yeah this definitely needs more thought. 

Is it really worth it to write out relative paths?

This breaks as soon as you do a 'save as' to a different folder, assuming you've already saved it in some other location beforehand.

fontforge/splinefont.h
```diff
@@ -2411,7 +2411,7 @@ extern int BpWithin(BasePoint *first, BasePoint *mid, BasePoint *last);
     /* Colinear & between */
 
 enum psstrokeflags { /* sf_removeoverlap=2,*/ sf_handle_eraser=4,
-	sf_correctdir=8, sf_clearbeforeinput=16 };
+	sf_correctdir=8, sf_clearbeforeinput=16, sf_imagereferenceonly=32 };
```
jtanx: Again, is there really no better place to stick this? It's not even related to stroking.

fontforgeexe/cvgetinfo.c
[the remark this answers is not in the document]
jtanx: ok I got a bit of a headache reasoning about this, so I haven't checked it through.

But I would clearly define where it's utf-8. You have two options; one is to store it as utf-8 all the way through until you are just about to pass it off to a function that calls e.g. fopen or whatever. The other option is to store it in the 'def' encoding, and convert to/from utf-8 wherever necessary.

Oh and fun fact, on Windows, glib functions expect paths to be utf-8 encoded. But let's just ignore that for now...

jtanx:
> PyFFImageType should have xoff, yoff, xscale, yscale able to be set, it should be possible to link and unlink it, and it should be possible to compare it (GImageSame). It might also need to be possible to hash it, so I might need to bring back GImageHash, jury's out on that though. Should users be able to access the bitmap data? I think that's pointless because Python users should already have the image in another file and can use other utilities, but await your comment. Then again, if it turns out to be not that hard to do, why not? Doing so could make it so users could export images from existing SFDs with a Python script, by using PIL to take FontForge's bitmaps and create PNGs.

I would always err on implementing the least amount necessary. It's always easier to add stuff in than take stuff out. So no, I wouldn't expose the bitmap data.


Also I'm not that familiar with this part of scripting, but:

```
i.linkReference(im)
```

Feels wrong. So correct me if my interpretation is wrong, but this is saying, *for this image I already have, this file path contains that image, so make a reference to it instead of storing the data in the sfd*

Why would you do that when you told it what to import the first time around?

fontforge/python.c
```diff
@@ -8293,8 +8294,8 @@ return( NULL );
     pt = strrchr(locfilename,'.');
     if ( pt==NULL ) pt=locfilename;
 
-    if ( strcasecmp(pt,".eps")==0 || strcasecmp(pt,".ps")==0 || strcasecmp(pt,".art")==0 ) {
 	int psflags = FlagsFromTuple(flags,import_ps_flags,"PostScript import flag");
```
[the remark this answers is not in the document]
ctrlcctrlv: I'd say it's not really my fault, the hack is that a function called `importOutlines` also actually imports images! This should be split into two functions, but to do that will break scripts.

gutils/fsys.c
ctrlcctrlv: Absolutely, absolutely it is worth it. Without relative paths, no one else can use my SFDs. It's a requirement for this feature to be useful.

fontforge/python.c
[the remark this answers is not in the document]
jtanx: Well there's nothing stopping you from defining your own `import_whatever_flags` and calling FlagsFromTuple separately

ctrlcctrlv:
> Why would you do that when you told it what to import the first time around?

Your understanding is correct. It does two things:

1. It creates a mechanism for the migration of old files via scripts, as in my `glob.glob` example.
2. It creates a mechanism for moving files around on the disk, e.g. an SFD where the images are in the root can be moved to an `images/` folder. Yes, images can be reimported, but if they are, the xoff/yoff/xscale/yscale is lost.

I'm working on the Python implementation of an image type, as I said. Thank you for your further comments, but I'd appreciate an answer to my most important question:

> *Can I please make that dependent on this PR and put it in this branch?*

ctrlcctrlv:
(You could say, that with an image Python type, script authors could just get/set `image.xoff` etc, but that means there would be no way to relink images in the GUI.)

jtanx:
It sounds like it's overcomplicating the matter to me. Just how frequently are you going to be moving files around that re-setting those xoff/yoff/scale parameters actually becomes an issue? Even the whole 'compare images' stuff sounds like overkill

> Can I please make that dependent on this PR and put it in this branch?

Up to you, I don't see any deadlines on this, so whatever works.

ctrlcctrlv:
> I don't see any deadlines on this

Sure. For me there is a deadline, I require this feature for a font I'm calling `TT2020`. So I need to merge this by December 1, 2019. Quite far away. If I can't manage that, I'll need to step away from the PR for a month to finish my font, as I don't want to countenance releasing it before 2020.

Thanks for your approval of `PyFF_Image` in this commit.

You think this is over-engineered, I can see that. `GImageSame` is very useful for Python equality in `PyFF_Image`. Image references might rarely change, but the feature is useful all the same, it provides an upgrade path, and I'd like to take advantage of it.

ctrlcctrlv:
@jtanx Something missing is that you can't forcibly relink an image with a different hash. If I add that, would that change your mind? That makes the feature much more useful.

ctrlcctrlv:
@skef: Do you mind giving comment about what the API should look like? And also if we should split `importOutlines` and `importImages` just for API sanity, even if doing so will break scripts.

skef:
I'm agnostic about splitting `importOutlines()` and `importImages()` If we want the latter the former could of course continue to work as well, possibly with a deprecation warning. 

After thinking about the rest of this for (admittedly just) a few minutes, I'm worried about any "Link/Unlink reference" model at all. Or rather, I'm worried about any model like that that also retains a glyph-based `importOutlines/Images`. I suppose the motivation is to follow the model of outline references, but in that case the reference is to a whole glyph or at least a whole layer. What if multiple images are loaded into glyph "A"? Does a reference pick up all of them? If not, are you referring to a glyph and an image name? Or a glyph and an image "index" starting from zero? 

Also, an outline reference is to something that doesn't have its own transformation matrix, but a background image does have one, or a subset of one. Does the reference inherit the referred-to transformation matrix and then modify it, or are all of them independent? If they're dependent, and the initial one changes, will the user get the result he or she expects? If they're independent will that be a pain for other reasons? 

In the case of outline references a reference is visually different from a contour set. In the case of images they aren't visually different (or are they?) So how does a user know which character to pull the image from? And if the character the image is pulled from has other images, how the user pick it out? Or Can you pull references through a chain, in which case do the transformation matrices stack up? (Presumably, if they are dependent, they would stack up.) 

So here's one alternative that at least initially I find very attractive:

1) I believe that images are only added to the background layer. So add an object to the glyph similar to `references` but that provides access to a list of "image tuples", each of which has *some kind of* image identifier as the first element and then transformation matrix info. 
2) `importOutlines/Images` adds elements to that tuple for the glyph it is called on. Maybe we add a way to specify a transformation matrix via an optional third argument.
3) There are no "image references" at the UI level beyond specifying that an image should be loaded "by reference", as this pr adds. Instead, when importing the same path into multiple characters by reference, we document that the implementation will only maintain a single copy in memory. So basically, each character has its own images with separate transformation matrices, but when using the interface in a certain way the FontForge will be more efficient.

Now, to make this a bit easier at the *GUI* level we might want to add a list of the already loaded images into a dialog, so that instead of loading one from a file you can load one that's already been loaded. But doing so would have the same effect at the python API level. 

These initial thoughts are about all I can think of at the moment.

ctrlcctrlv:
@skef The references only dictate how images are saved/loaded, all the data is still in the GImage.

skef:
@ctrlcctrlv you asked me for UI/API advice and above you indicated you already modified `importOutline` and noted you had yet to add '"Link Reference" and "Unlink Reference"'. My response was one shot at going off those facts. It would probably help at this point to clarify the question: "What should the API look like" *for what/to do what*, specifically?

frank-trampe:
Sorry I've been gone for a bit. If we can make image relinking work, I think that it's a useful feature to have. I don't trace in FontForge very often, but I often work with linked images in other pieces of software, and I find myself using the functionality where it is present and grudgingly editing XML by hand where it is not. I think that it's perfectly fine to merge an implementation that does not expose the functionality in the GUI, particularly since the cases in which the feature is most useful are those that involve many glyphs, such that one would not use the GUI.

On the flip side, we cannot create any sort of assumption of one image per glyph.

More in a bit.

fontforge/cvimages.c
```diff
@@ -995,6 +998,7 @@ return(false);
 	if ( endpath==NULL )
     break;
 	start = endpath+1;
+    SCCharChangedUpdate(sc,toback?ly_back:ly_fore);
```
frank-trampe: Was this previously necessary but missing?

fontforge/python.c
```diff
@@ -8269,6 +8269,7 @@ static struct flaglist import_ps_flags[] = {
     { "removeoverlap", 0 },			/* Obsolete */
     { "handle_eraser", sf_handle_eraser },
     { "correctdir",  sf_correctdir },
+    { "imagereferenceonly",  sf_imagereferenceonly },
```
frank-trampe: At the risk of starting the nitpicking too early, is it necessary to say "imagereferenceonly"? Would there be anything vague or misleading about "referenceonly"?

fontforge/sfd.c
```diff
@@ -1019,12 +1029,13 @@ void SFDDumpUndo(FILE *sfd,SplineChar *sc,Undoes *u, const char* keyPrefix, int
 static void SFDDumpImage(FILE *sfd,ImageList *img) {
     GImage *image = img->image;
     struct _GImage *base = image->list_len==0?image->u.image:image->u.images[0];
+
```
frank-trampe: Once again, I'm sorry to nitpick, but @jtanx already got all of the substantive issues. The old organization seemed to make sense. Declarations together and then actions.

fontforge/sfd.c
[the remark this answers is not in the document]
frank-trampe: We can add flipbook animation someday.

fontforge/sfd.c
```diff
@@ -3631,26 +3658,29 @@ static void rle2image(struct enc85 *dec,int rlelen,struct _GImage *base) {
...
+    }
+
+    if ( strmatch(mime, "image/png")==0 ) {
+        return IM2_PNG;
     }
 
-    if ( !(strmatch(mime, "image/png")==0) ) {
-        IError("MIME type received—%s—is not recognized", mime);
-        return UNKNOWN;
+    if ( strmatch(mime, "reference")==0 ) {
```
frank-trampe: You may already have thought about this, but, it seems to me that "text/uri-list" might work just as well for the immediate use case and also allow us to comply with the MIME specification.

gutils/fsys.c
frank-trampe: I totally agree with @ctrlcctrlv on this. You ought to be able to move your project directory around on the computer or to transmit it to others without it breaking. I sometimes need to manually edit Xournal files in order to make paths relative, and then it changes them back every time I edit them, which is very irritating.

fontforge/cvimages.c
ctrlcctrlv: Yes

ctrlcctrlv:
This implementation does not assume one image per glyph. I'm not sure where that idea came from, but all my tests work on multiple images. Multiple images per glyph is why I need to add Python API for images

fontforge/sfd.c
ctrlcctrlv: Well... Remote files aren't supported, and probably shouldn't be, so it's not truly a URI...Also I don't think a file URI can be relative

frank-trampe:
Ah. My comment about single-image-per-glyph issues was from a misreading of another comment here prior to rereading the code. Sorry about that.

fontforge/sfd.c
frank-trampe: It's okay to reject an unsupported protocol in a URI. And relative paths are easy. If the file is reached via the same protocol (as would always be the case here), you just write the relative path the same way you would for a relative UNIX path.

ctrlcctrlv: Oh, so would I give  it a protocol other than `file://`?

frank-trampe: Those were separate points.

On the first point, we don't need to worry about remote files. If there is an http or ftp path, we just warn and ignore it.

On the second point, a relative path would not explicitly list the protocol, just like you can write ``.

ctrlcctrlv: For the purposes of MIME, is `images/logo.png` a valid URI? Would `images\logo.png` be?

frank-trampe: We ought to be no more permissive than web standards, so certainly no backslashes.

ctrlcctrlv: Sorry, I think `text/uri-list` is always supposed to have the protocol, [according to Mike Amundsen](http://amundsen.com/hypermedia/urilist/).

Likewise, [RFC 2483 &sect; 5](https://tools.ietf.org/html/rfc2483#section-5) states:

>     The format of text/uri-list resources is:
> 
>     1) Any lines beginning with the '#' character are comment lines
>         and are ignored during processing. (Note that URIs may contain
>         the '#' character, so it is only a comment character when it is
>         the first character on a line.)
> 
>     2) The remaining non-comment lines shall be URIs (URNs or URLs),
>         encoded according to the URL or URN specifications (RFC2141,
>         RFC1738 and RFC2396). Each URI shall appear on one and only one
>         line. Very long URIs are not broken in the text/uri-list format.
>         Content-transfer-encodings may be used to enforce line length
>         limitations.

The HTML standard is more permissive...it supports both URIs and relative paths, but a relative path is not itself a URI, right? Just as it makes no sense to have a relative `http://` URI it makes no sense to have a relative `file://` URI. A URI must always state a protocol as well.

I don't mind breaking the standard, but please be aware of this.

frank-trampe: It seems ambiguous enough that we can claim to have made a good faith effort. If we flood the Internet with enough SFD files with relative paths, they'll eventually resolve the ambiguity in our favor.

ctrlcctrlv:
I've made a lot of Python progress and hope to add it to the branch today or tomorrow.

https://twitter.com/FontForge/status/1172445826381074436

```
>>> print(g.layers[0].images)
[<Image of type it_rgba at 0x15554f7a4f70; GImage=0x555555ed6180, _GImage=0x555555eb32b0; 480x371x1484; size=712320; -1256 1398 2.08333 2.08333>, <Image of type it_true at 0x15554f7a4e70; GImage=0x555555db77f0, _GImage=0x555555ed5220; 1401x2560x10240; size=14346240; -6 1440 0.713776 0.713776>, <Image of type it_true at 0x15554f7a4dd0; GImage=0x555555e104e0, _GImage=0x555555e04e30; 708x920x3680; size=2605440; -840 820 1.41243 1.41243>]
```

frank-trampe:
@ctrlcctrlv, that's great.

ctrlcctrlv:
After a chat in IRC I'm not sure if @skef is working on the Python part of this or not

skef:
I've been thinking about it -- so not "working", exactly. 

Here are a few preliminary thoughts:

1. As conceived this is really two projects. The first is supporting loading images stored in external files. The second is providing an API for images generally, to add, remove, and alter the transformations of images on a given glyph.
2. The first project is more or less implemented. One can always quibble with this or that choice but getting the code to where folks would be broadly satisfied wouldn't take very long.
3. As @ctrlcctrlv and I discussed there is already a model for something close to the interface we need, which is the interface for outline references: `layerrefs` and `references`, the latter being another way of expressing `layerrefs[1]` -- the foreground references. We could add `layerimages` and `images`, the first being a layer-indexed dictionary object of lists of "image tuples", the latter being another way of expressing `layerimages[0]` -- the background images. An image tuple would have some sort of layer identifier (or perhaps a pair of them) and then a transformation matrix for the image. `layerrefs` can be cumbersome in that it's picky to extract, alter, and reinsert an image tuple, but the interface is sufficient. (As I've stated in the past I'm frustrated that FontForge's python code continues to be entirely behind the C-API, and I would much prefer a hybrid system in which a python `fontforge` object inherited from a C-API `_fontforge` object (or whatever) so that much of the code could be written in python.)
4. If all that sounds relatively straightforward, it isn't yet. One issue is what the image identifier would be. @ctrlcctrlv and I discussed using a SHA and possibly supporting a "nickname". Here's the main problem with a "data-SHA": What do you when there are multiple sources of the same data? We already know that this can happen with SFD-stored images. Someone can also load the same image from multiple files (or paths). Which one does the SHA refer to? On the other hand we could hash the filename into externally loaded image SHAs and the glyph name into internally stored image SHAs. Then there could be a layer inside FontForge that, when opening an image and finding it has the same data as another image, doesn't allocate a duplicate but shares a reference-counted GImage (or whatever). All this is still somewhat fuzzy in my thinking.
5. Another more difficult issue is the relationship between changes to the `layerimages` structures and the internal representation/SFD representation. Let's assume you can't just implicitly copy images between fonts, just as you can't just implicitly copy outline references between fonts. Can you just take a image ID from one glyph, calculate a different transformation matrix for it, and add it to another glyph? That's how layerrefs works and it's handy. If you can do that it's pretty easy to come up with a matching in-memory representation, but what does that look like in the SFD? Does the new glyph get a duplicate `Image2` entry? What if the image that was copied is stored in the SFD; do we duplicate the data?
6. Then there are another set of issues with copy/paste. Suppose the user selects and copies a set of characters, some of which have images, and pastes them into a different font. Do the images "go with" them? (What happens now? Unfortunately I don't personally use images.) If so, that can be tricky depending on how images are stored and relate to one another; FontForge's logic for copying outline references is somewhat sophisticated.

In the background of some of these questions is a nagging worry about whether the Image2 tag should only support referring out to a file, or whether there should be a way to specify "image 'blah' of glyph 'foo'". The latter would certainly improve the SFD-stored image situation, but maybe that doesn't matter. 

To summarize, once you add the ability to have images loaded in multiple ways, and an API for altering what glyphs have what images, a bunch of hairy issues arise and I don't have a clear picture for how to deal many with them. If we knew (or, once we know) the answers I don't think the implementation would (will) be very difficult given the work already done. 

There is another possibility, however, which is that we could punt the API question into the future. If images can only be manipulated in the way images are manipulated now, things become a lot simpler. There still may be questions about copy/paste, so that may be one thing to nail down first. 

@ctrlcctrlv I don't know how much time you have but there's a possibility of our working on these questions together and then maybe my doing most of the coding, in the hope of meeting the release deadline.

ctrlcctrlv:
I find it basically impossible to implement the Python part before the release deadline, so what I would say is, since this was a previously non-existent feature, and there is no Python API for images at present, we pare the requirements for getting this in back, not requiring Python support for a merge.

Then in future we make a PR for working with FontForge's images in Python.

I would like to see this merged before the release and this is the easiest road to get there. Thoughts ?

ctrlcctrlv:
>What happens now? Unfortunately I don't personally use images.

We've discussed multilayer copy/paste in the past @skef and this is part of that. What happens now is highly underwhelming, images cannot be copied from the FontView, only the CharView. Images cannot be put on a fore layer.

skef:
So if you copy the background layer do images "go with" it?

ctrlcctrlv:
In the FontView? It's not possible to copy the background, is it?

ctrlcctrlv:
Even using "Copy Layer to Layer" it's not possible to get an image onto the Fore.

skef:
And I guess I should ask: What is the purpose of images on layers other than the background. I'm not questioning that there is such a purpose I just want to understand it. 

Re the FontView: I'm actually pretty fuzzy on the relationship between the "active layer" and the FontView. The python interface works a certain way and I don't know if the FontView pays attention to the currently selected layer.

ctrlcctrlv:
Oh, there is no purpose that I know of. It's just that, without your script, only the fore is considered by FontView for copy/paste.

ctrlcctrlv:
(I just tested it, and even your script totally ignores images. The only way I know to copy/paste images is to open up two CharViews.)

skef:
My script can't do much about the images without an API. 

But wait a minute: If there's no compelling reason to have images in other layers besides the background, should we support images in other layers? It's a lot simpler at the code level if we just restrict them to the background layer.

Under what circumstances do they copy with two CharViews? Do you have to select and copy the images explicitly or does copying when the background layer is selected in the layer palette suffice?

ctrlcctrlv:
I can't think of a good reason to ever put images in the Fore.

Someone might want to work from two references, and maybe even put those on different layers so you can hide/show them. But those would just be additional background layers, not the Fore.

You have to select the images explicitly in the CharViews, yes.

skef:
OK, these restrictions are giving me some ideas for a interface that includes a limited python API:

1. We stipulate that images can continue to be in the background only. That means no `layerimages` and just a `images` structure, with a list of image tuples. 
2. The image-representing "token" in an image tuple is a) ad-hoc and b) picked from a namespace specific to the glyph. Probably something like the filename for a by-reference image and "image-1", "image-2" etc for images loaded from the SFD. 
3. You can't copy an image between glyphs just by modifying the `images` list. Instead you must use `importOutlines()`. However, we might extend `importOutlines()` to take a glyph and image token as an argument. (I'll have to think about this more.) 
4. Therefore the `images` list is mostly for adjusting the transformation matrix of an image via python. 

This way you can make any adjustment you need to from python, it's just cumbersome in some cases. For example I could add image support to my multilayer copy script.

skef:
Oh, hmm, additional background layers. Let's punt on that problem for now.

ctrlcctrlv:
That solution sounds good to me @skef to get this through. We can always make a more complete API in future.

skef:
I've been looking at the code and thinking about it the past couple days. That process has yielded some preliminary design ideas for a python API but also a bunch of concerns. The overall model that is already implemented is close but I think it needs a bit more thought.

One worry is that images need some kind of namespace for identification and I don't know what it should be. Some equivalent to using the glyph name to identify an outline reference. The problem is that there doesn't seem to be any obvious informative option. IDing by (glyphname, token) is good as long as there is only one image per glyph, but once you need to care about the token it's hard to come up with something generally informative. (The filename, or a piece of the filename, works for image references but not for SFD-stored images.) (Tuple-like identifiers are also a drawback for the native scripting language, which doesn't have as handy a datastructure.)

An alternative is to give each image its own font-level identity. Unfortunately that doesn't solve the "opaque token problem" for SFD-stored images.  

Then there's the question of what you should be able to do with an ID. Just reorder in the same glyph? Copy an image to a different glyph in the same font? In a different font? Presumably there should be an interface to get more data about the image referred to -- width/height, filename if you
know it, etc.)

I'm also not fond of the current ad-hoc handling of whether a reference path is relative or absolute, which seems to be "Attempt relative/fall back on absolute". Shouldn't a user be allowed to pick? Shouldn't a user be forced to specify "absolute" if "relative" is failing? Also, only having the paths be relative to the font path is too error-prone. There should be a font-level parameter (maybe set in Font Info -> General) that represents the absolute path that all relative paths are relative to, which when not set defaults to the sfd's own path. Otherwise it's far too easy to bork yourself into dozens of tiny little repairs.   

Unfortunately the combination of this, the schedule, and me poses a problem. I can often implement things pretty quickly when all the questions are answered or if I'm just tracking down bugs with no design implications. But when it comes to design I tend to need to clutch my head for a while. I worry a lot about bad decisions now creating user/developer terror or misery down the road, which makes me a slow "design tinkerer". 

So for now I'm going to switch back to trying to implement arcs joins in the expand stroke code. My suggestion is that we continue the ~~foodfight~~ discussion here in an attempt to work out a coherent (if not necessarily finished) design, which I could spend coding time on later. If that work doesn't peter out I don't see why we couldn't delay the next release a bit (if necessary) to accommodate it.

ctrlcctrlv:
I think maybe you're thinking about this too hard @skef. To me there's an obvious ID, which I thought of while reading your question. (glyph name + - + ImageList index). Why can't this work? So the first image in a would have ID `a-1`. Seventh image in `NameMe.65565` would be `NameMe.65565-7`.

You might say, well what if those two images have same underlying data? _That_, to the extent users even care about it, can be solved via having some method to return a hash.

skef:
> To me there's an obvious ID, which I thought of while reading your question. (glyph name + - + ImageList index). Why can't this work? So the first image in a would have ID a-1. Seventh image in NameMe.65565 would be NameMe.65565-7.

That would mean that as you're reordering the images using the API (presumably to alter which obscures which) the tokens are changing. So updating the list one-at-a-time could have a different effect than updating it all at once.

ctrlcctrlv:
Sounds like that could just be a note in the documentation

skef:
Unfortunately a note in the documentation isn't much insurance against "user/developer terror or misery". If reordering the list borks any outstanding tokens then maybe you shouldn't be able to reorder the list. If you can reorder the list it's presumably because it can be handy to reorder it, in which case there shouldn't be strange gotchas when doing so unless absolutely necessary. 

On the other hand, you can always put a counter in the font structure and increment it when needing a new token, so maybe it could be (font-reference, number). (Putting a counter on each glyph seems like overkill.) That would probably mean you need some kind of list at that level (maybe an array that can become slightly sparse over time) to provide the mapping (the counter points at the end of the list).

ctrlcctrlv:
Given it's not possible to do anything with images right now, not being able to reorder them without issue is a problem.

By the way, I'm getting an error right now with GSUB output. Do you mind hopping on IRC? I can't figure it out.

ctrlcctrlv:
(Just so you know, reordering doesn't seem possible in the UI without like cut/paste or re-import.)

frank-trampe:
If we adjusted that model slightly, it might be workable. In particular, if we exposed the images in a layer simply as an array, the intelligibility issues associated with that scheme seem to lessen considerably. The user could insert, remove, copy, and such.

ctrlcctrlv:
@skef and I talked in IRC and came up with the following compromise:

* Make it so each glyph may only contain one image
* Make it so the image is always in the Background layer and never other layers

This means in Python we could just have `glyph.image`, either None or an Image.

I use images the most out of any of the "active contributors" and don't see either of these changes as particularly onerous. If I were to ever want multiple refs for a glyph (dubious proposition), I'd just put them in one image anyway. I can't think of a legitimate reason to be able to have multiple images per glyph in a font editor.

If you can only have one image per glyph, then having it in different background layers is really quite pointless.

It should be noted how poor the support currently is for multiple images, especially in Python. We worry though that @frank-trampe might veto this so are asking first before writing any code.

ctrlcctrlv:
Oh, and here's a proposal for any old SFDs that might be floating out there:

* If the problem is simply that the image is in a layer other than Back, just move it to Back and emit a warning.
* If the problem is that there are multiple images in one glyph, emit a warning and at user option either:
  * only load the first; or
  * create _n_ duplicate glyphs with the _i_&zwnj;th image.

ctrlcctrlv:
@skef As there's been no reply I'd say it's safe to start.

frank-trampe:
@ctrlcctrlv, I've been exchanging thoughts a bit with @skef on the engineering side of this, and I continue to believe that it's not appreciably easier to do this the single-image way than to include decent support for multiple images. The one hang-up, on which we are both stewing, is how to tokenize the images in a way that makes it easy for the user to pick the one he wants. In the single-image case, that's not really a problem anyway. So I would propose that we just expose an array of images as they are stored internally, with appropriate insert/remove functionality like DOM children in JavaScript, and, in the simple single-image-per-glyph case, you can just push/pop that one image with no ambiguity.

ctrlcctrlv:
This is a note regarding an IRC discussion I had with @skef on the 9<sup>th</th> of February about closing this issue for the following reasons:

* Although an images Python API is highly beneficial for scripts, nothing about this changeset actually implements one, and no code in this changeset is required to implement one.
* The feature itself might be pointless. What does it give us that SFDIR + PNG in SFD does not?
* The code is stale and getting staler every day.

Our resolution was basically I could voluntarily withdraw it if I wanted to, but that @skef doesn't really mind PRs remaining open for extended periods if they are complex as is this one and if there is a possibility _someone_ will come around and close it out, maybe if not the original contributor.

I almost voluntarily withdrew this today, however as I see it reason &numero;2, which was actually my reason, is not _actually_ valid. What this commit buys you is not storage space saved, it's faster serialization of SFD files. So as long as SFDIR writes all glyphs on every save (which it does), and not based on whether or not the SplineChar is marked as "changed" and other heuristics on the parent SplineFont, this feature is useful in and of itself without the Python images API.

So basically, I write this to inform that I almost voluntarily withdrew this, but then decided it's actually still a good feature so I won't be withdrawing it.

PR #3921, opened 2019-09-09
Stop crash under tiling WMs

### LINKED ISSUE #3920: Fontforge crashes on glyph rename
Version: fontforge 20190801
OS: Arch Linux

**Issue**
I’ve encountered a recurrent bug in FontForge where renaming a glyph in glyph info will cause a crash when pressing enter, but not when clicking OK. I tried it out in Sway and it doesn't seem to happen (but there are display issues), I'm using i3.

It also doesn't happen on  older versions (20170924, Linux Mint Tessa, 20190526 from the PPA) I am not getting any crashes. So not sure if it has something to do with my setup.

**Reproducing the issue**
Open a new file, right click on a glyph (e.g. A) and go to glyph info.
Rename the glyph (e.g. Aa) and press enter.
FontForge crashes. 

**Extra info**
I read the have read the information on the FontForge website about reporting crashes but I'm not sure how to enable debugging information. I know it's probably not as useful without that but I had a look with it in gdb and I receive this output when FontForge crashes:

```
0x00007ffff7737f96 in _GWidget_RemoveGadget () from /usr/lib/libgdraw.so.6
```
and backtrace says this:
```
#0  0x00007ffff7737f96 in _GWidget_RemoveGadget () at /usr/lib/libgdraw.so.6
#1  0x00007ffff774392f in _ggadget_destroy () at /usr/lib/libgdraw.so.6
#2  0x00007ffff772ff41 in  () at /usr/lib/libgdraw.so.6
#3  0x00007ffff7776615 in  () at /usr/lib/libgdraw.so.6
#4  0x00007ffff7738ae8 in  () at /usr/lib/libgdraw.so.6
#5  0x00007ffff7781328 in  () at /usr/lib/libgdraw.so.6
#6  0x00007ffff77814a9 in  () at /usr/lib/libgdraw.so.6
#7  0x00007ffff64483c4 in  () at /usr/lib/libglib-2.0.so.0
#8  0x00007ffff6448bb0 in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#9  0x00007ffff644ab11 in  () at /usr/lib/libglib-2.0.so.0
#10 0x00007ffff644ab51 in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#11 0x00007ffff777fe1e in  () at /usr/lib/libgdraw.so.6
#12 0x00007ffff7ddfb27 in fontforge_main () at /usr/lib/libfontforgeexe.so.3
#13 0x00007ffff7a82ee3 in __libc_start_main () at /usr/lib/libc.so.6
#14 0x000055555555505e in _start ()
```

ctrlcctrlv:
This closes #3920.

I ran through this with `rr` and can see this change to be the correct
one. This does the same thing as the patch I offered in #3920, except is
much more readable.

When the window is dying, it has no widget_data to free; the free was
already done by `_GWidget_Container_eh`. The order of events is
different on tiling window managers

### Type of change

- **Bug fix**

### DIFF
--- a/gdraw/gcontainer.c
+++ b/gdraw/gcontainer.c
@@ -839,8 +839,8 @@ return;
     g->base = NULL;
 
     while ( gw->parent!=NULL && !gw->is_toplevel ) gw = gw->parent;
-    if (gw->is_dying) return;
     td = (GTopLevelD *) (gw->widget_data);
+    if ( td == NULL ) return; /* dying, prevents tiling WM crash */
     if ( td->gdef == g ) td->gdef = NULL;
     if ( td->gcancel == g ) td->gcancel = NULL;
     if ( td->gfocus == g ) td->gfocus = NULL;
--- a/gdraw/gcontainer.c
+++ b/gdraw/gcontainer.c
@@ -839,6 +839,7 @@ return;
     g->base = NULL;
 
     while ( gw->parent!=NULL && !gw->is_toplevel ) gw = gw->parent;
+    if (gw->is_dying) return;
     td = (GTopLevelD *) (gw->widget_data);
     if ( td->gdef == g ) td->gdef = NULL;
     if ( td->gcancel == g ) td->gcancel = NULL;
--- a/gdraw/gcontainer.c
+++ b/gdraw/gcontainer.c
@@ -840,7 +840,10 @@ return;
 
     while ( gw->parent!=NULL && !gw->is_toplevel ) gw = gw->parent;
     td = (GTopLevelD *) (gw->widget_data);
-    if ( td == NULL ) return; /* dying, prevents tiling WM crash */
+    if ( td == NULL ) { /* dying, prevents tiling WM crash */
+        assert(gw->is_dying);
+        return;
+    }
     if ( td->gdef == g ) td->gdef = NULL;
     if ( td->gcancel == g ) td->gcancel = NULL;
     if ( td->gfocus == g ) td->gfocus = NULL;
--- a/inc/gdraw.h
+++ b/inc/gdraw.h
@@ -28,6 +28,7 @@
 #ifndef FONTFORGE_GDRAW_H
 #define FONTFORGE_GDRAW_H
 
+#include "assert.h"
 #include "charset.h"
 #include "gimage.h"
 


gdraw/gcontainer.c
```diff
@@ -839,6 +839,7 @@ return;
     g->base = NULL;
 
     while ( gw->parent!=NULL && !gw->is_toplevel ) gw = gw->parent;
+    if (gw->is_dying) return;
```
jtanx: So what happens in 'normal' window managers, is gw->widget_data not null but it is dying?

skef: @jtanx my guess would be that the problem is caused by an extra call not seen when the dialog is already foregrounded. But "m not sure the problem was investigated to that extent.

ctrlcctrlv: To be quite frank, I don't understand why they need to be NULL'd in any event. They don't seem to ever contain valid memory. This is under KWM:

```text
(rr) p *((GTopLevelD*)*(gw->widget_data))
$636 = {funcs = 0x0, w = 0x0, next = 0x0, e_h = 0x0, gic = 0x0, has_focus = 0, enabled = 0, iscontainer = 0, 
  istoplevel = 0, contains_focus = 0, ispalette = 0, positioned_yet = 0, isdocked = 0, programmove = 0, 
  gadgets = 0x0, widgets = 0x0, grabgadget = 0x0, lastddgadget = 0x0, lastwiggle = 0x0, popupowner = 0x0, 
  gdef = 0x71173f0000000000, gcancel = 0x71178c0000005631, gmenubar = 0x708a3f5500005631, 
  gfocus = 0x708a3f6900005631, wfocus = 0x5631, handle_key = 0x0, palettes = 0x711756e000000000, nextp = 0x5631, 
  owner = 0x0, owner_off_x = 0, owner_off_y = 0}
(rr) p *((GTopLevelD*)*(gw->widget_data)).gcancel
Cannot access memory at address 0x71178c0000005631
```

I suppose you're right, with this change they'll never be NULL'd.

ctrlcctrlv: I can easily change it to &hellip;
```c
if (gw->widget_data == NULL) return; /* dying, prevents tiling WM crash */
```

But this doesn't really get to the heart of the matter. The code and order of events is confusing to put it mildly.

skef: I misunderstood @jtanx's point. 

Would 
```
if (gw->widget_data == NULL) {
    assert(gw->is_dying);
    return;
}
```
be acceptable to everyone (assuming it tests OK)?

ctrlcctrlv: I think that would be the first assertion in this part of the code @skef. I'd have to include `assert.h`, which is only included in `ufo.c` and some GDK files. I don't mind doing that, but FontForge is very light on assertions

ctrlcctrlv: By the way, I'd much rather add it to `inc/gdraw.h` if we're going to add it at all, and not just to `gcontainer.c`, that way we can add assertions to other GDraw places they're sorely wanting.

skef: I'm a fan of assertions, especially in favor of comments to the effect of "X is true here". But I won't speak for others.

ctrlcctrlv: I did a bit more digging and I think I know why the NULLs on the toplevel are needed in the first place...they're only done if that gadget is being removed...it's possible to only remove some gadgets and not others, while also not destroying the entire window. Clearly if the entire window is dying it doesn't matter if we NULL them or not, the widget_data itself is going to be free'd making any of its members pointless. You don't have to NULL all struct members when you free a struct, so...

ctrlcctrlv: The code now essentially matches your suggestion @skef as of e49b5806c8bf584e988572d700640673d7e7710d.

skef: @ctrlcctrlv that's only true if there isn't any top-down freeing when a window is closed. If there is then the non-NULL values may be freed or otherwise used to deallocate memory. (I'm not saying they *are* used that way, just pointing out the relevance.)

jtanx: What this is telling me is that this change is probably just hiding another issue - that the child window is being destroyed after its parent is already gone, which I wouldn't be surprised if it broke other assumptions made elsewhere.

jtanx: Yeah, pretty sure this isn't the right fix. If the parent was already freed, then having a conditional on `gw->is_dying` is UB because gw has already been freed.


```
==31002== Invalid read of size 8
==31002==    at 0x6B939A: _GWidget_Container_eh (gcontainer.c:265)
==31002==    by 0x6CDCCE: _GGDKDraw_CallEHChecked (ggdkdraw.c:324)
==31002==    by 0x6D336E: GGDKDrawPostEvent (ggdkdraw.c:2208)
==31002==    by 0x6CD6BE: _GGDKDraw_OnWindowDestroyed (ggdkdraw.c:205)
==31002==    by 0x64B3AB2: ??? (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==31002==    by 0x64B3049: g_main_context_dispatch (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==31002==    by 0x64B33EF: ??? (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==31002==    by 0x64B349B: g_main_context_iteration (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==31002==    by 0x6D3267: GGDKDrawEventLoop (ggdkdraw.c:2188)
==31002==    by 0x6BF1F3: GDrawEventLoop (gdraw.c:746)
==31002==    by 0x6840BC: fontforge_main (startui.c:1416)
==31002==    by 0x41A265: main (main.c:33)
==31002==  Address 0x14ec5dd8 is 56 bytes inside a block of size 176 free'd
==31002==    at 0x4C2EDEB: free (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==31002==    by 0x6CDA19: _GGDKDraw_OnWindowDestroyed (ggdkdraw.c:260)
==31002==    by 0x64B3AB2: ??? (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==31002==    by 0x64B3049: g_main_context_dispatch (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==31002==    by 0x64B33EF: ??? (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==31002==    by 0x64B349B: g_main_context_iteration (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==31002==    by 0x6D3267: GGDKDrawEventLoop (ggdkdraw.c:2188)
==31002==    by 0x6BF1F3: GDrawEventLoop (gdraw.c:746)
==31002==    by 0x6840BC: fontforge_main (startui.c:1416)
==31002==    by 0x41A265: main (main.c:33)
==31002==  Block was alloc'd at
==31002==    at 0x4C2FB55: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==31002==    by 0x6CDF4F: _GGDKDraw_CreateWindow (ggdkdraw.c:395)
==31002==    by 0x6D06D0: GGDKDrawCreateTopWindow (ggdkdraw.c:1249)
==31002==    by 0x6BD36D: GDrawCreateTopWindow (gdraw.c:89)
==31002==    by 0x54481A: FontInfo (fontinfo.c:7576)
==31002==    by 0x560840: FontMenuFontInfo (fontinfo.c:10894)
==31002==    by 0x5656B3: FVMenuFontInfo (fontview.c:1244)
==31002==    by 0x6EC20F: gmenu_mouse (gmenu.c:948)
==31002==    by 0x6EDCAE: gmenu_eh (gmenu.c:1389)
==31002==    by 0x6CDCCE: _GGDKDraw_CallEHChecked (ggdkdraw.c:324)
==31002==    by 0x6D0591: _GGDKDraw_DispatchEvent (ggdkdraw.c:1222)
==31002==    by 0x5828D91: gdk_event_source_dispatch (gdkeventsource.c:369)
==31002== 
```

skef: @jtanx probably true but at the same time if this local fix can be made harmlessly (I think the current version comes at least close) we could just wait for the next issue.

jtanx: The problem is this is only working through sheer luck, gw has already been freed, so there's no guarantee that gw->is_dying is true

skef: @jtanx If what you're saying is true then something else is very wrong. Double-close or not, how can a window get through `dying` all the way to being freed and still be something's parent?

skef: Isn't the whole point of `dying` to mark the transition point between being marked for deallocation and removing all references to it?

jtanx: There is no double close, probably just a break in the assumptions of what _GGDKDraw_OnWindowDestroyed does.

The supposed flow when a destroy is initiated is:
* Recursively call destroy window on all child windows
* Defer an event to actually clean up all window data
* In the deferred event, call `gdk_window_destroy` and process events until it's destroyed
* Free everything

In between calling `gdk_window_destroy` and freeing everything, it's assumed that all child windows are gone by that point, but my hypothesis is that's not currently happening with this window manager. So the call to _GGDKDraw_OnWindowDestroyed for the child window somehow comes in on the next event loop cycle after the parent has already gone.

skef: @ctrlcctrlv you could try setting `gw->is_dying` to false right before `free(gw);` in `_GGDKDraw_OnWindowDestroyed` and see if the assertion fails. I believe that would test @jtanx's hypothesis.

ctrlcctrlv: Sorry for submitting an incorrect patch; you're the GDraw expert. The patch does work but I admit I can't really articulate _why_, other than that I see `_GWidget_Container_eh` freeing the widget_data _before_ the call here, but only on i3.

jtanx: I don't think there's anything to apologise about, fwiw I could be wrong.

> you could try setting gw->is_dying to false right before free(gw)

Probably not conclusive though, as there's no guarantee it can't be set to anything else in the mean time. You'd be better off checking with asan or valgrind.

skef: `gw->widget_data = NULL` is the last part of `_GWidget_Container_eh`'s `et_destroy` stanza. Given that fact @jtanx's suspicion seems less warranted -- `widget_data` is clearly managed separately from `gw` in some destroy paths.

jtanx: I'm not quite sure I follow you there, et_destroy is called once per call into _GGDKDraw_OnWindowDestroyed.

I think the chain of events are:
* Enter gets pressed
* Event dispatched off the child window -> refcount > 1 while event is being dispatched
* GDrawDestroyWindow on parent window
   * Calls GDrawDestroyWindow on all child windows then decrements its own counter
   * Because of the additional refcount on the current event's window -> it's own destroy call is placed last, after all the other windows (and parent window)
* Differences between i3 and every other WM means that how this loop gets processed: https://github.com/fontforge/fontforge/blob/286cdd1e9a3416aa2e41b9b448e810ab322b267f/gdraw/ggdkdraw.c#L193-L194
   Causes the child to be destroyed first on i3


Anyway I think this would fix it https://github.com/fontforge/fontforge/compare/master...jtanx:refcount?expand1 but I'm short on time atm and don't really have time to test it. The basic premise is to ensure that the parent window is kept around long enough until its children are gone. There shouldn't be any bad side effects, assuming I've done the refcounting properly...

ctrlcctrlv: Hi Jeremy&mdash;

I checked your branch and it works. Will write more in another comment.

ctrlcctrlv:
Per https://github.com/fontforge/fontforge/pull/3921#discussion_r323045207, I am voluntarily closing this to open a PR against `jtanx:refcount` as opposed to `ctrlcctrlv:issue3920`

Issue #3905, opened 2019-08-28
Combining command line options `-script` and `-new` does not work


ctrlcctrlv:
This could be useful for development, but results in an error:
```
[fred@pc fontforge]$ fontforge -script imp.py -new -quiet
Traceback (most recent call last):
  File "imp.py", line 3, in <module>
    f = fontforge.fonts()[0]
IndexError: tuple index out of range
```
```python
﻿import fontforge

f = fontforge.fonts()[0]

f.createChar(ord("A"))
f["A"].right_side_bearing = 100
f["A"].importOutlines("/home/fred/Workspace/fontforge/B.jpg", 32)
```
Apparently the script is run before a new font is made.

Of course, we can do:
```diff
--- imp.py      2019-08-28 14:59:17.100865945 +0800
+++ imp2.py     2019-08-28 14:59:02.004199156 +0800
@@ -1,6 +1,6 @@
 import fontforge
 
-f = fontforge.fonts()[0]
+f = fontforge.font()
 
 f.createChar(ord("A"))
 f["A"].right_side_bearing = 100
```

But now we can't run the same script in the &laquo;Execute Script&raquo; dialog as in the command line interface.

skef:
Seems like trying to solve this in the totally general way would be a big pain. Better to add `-new` to the arguments processed by `scripting.c:_CheckIsScript()` (possibly enhancing the loop to remove the ordering constraints). That way this would also pose no barrier to #3922 .

(I take it the alternative is to try to get the scripting initialization to use the general options processing code.)

PR #3922, opened 2019-09-09
Defer gui initialization to after options parsing

zeising:
Move GUI initialisation when running with GDK until after options
parsing.  This way, help messages, version information and so on can be
printed even when no gui is around, such as running fontforge -version
from a configure script.

### DIFF
--- a/fontforgeexe/startui.c
+++ b/fontforgeexe/startui.c
@@ -1099,10 +1099,6 @@ int fontforge_main( int argc, char **argv ) {
     CheckIsScript(argc,argv); /* Will run the script and exit if it is a script */
 					/* If there is no UI, there is always a script */
 			                /*  and we will never return from the above */
-#ifdef FONTFORGE_CAN_USE_GDK
-    gdk_init(&argc, &argv);
-    gdk_set_allowed_backends("win32,quartz,x11");
-#endif
     if ( load_prefs==NULL ||
 	    (strcasecmp(load_prefs,"Always")!=0 &&	/* Already loaded */
 	     strcasecmp(load_prefs,"Never")!=0 ))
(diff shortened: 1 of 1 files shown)


skef:
@ctrlcctrlv WRT your email to the list, is there any practical difference other than "calling convention" between what you want and adding the script to your `.config/fontforge/python` directory? If not, it seems like adding a new flag to treat its argument as if it were a config script would do the job while being easy to implement.

frank-trampe:
@ctrlcctrlv, what specific difficulty would this add to a solution for #3905?

ctrlcctrlv:
I had mistakenly thought `-new` is implemented via `splash_e_h` when looking quickly through the file, but that turns out not to be the case.

To answer your question @skef, I don't think there's any practical difference between what I want and using the `python` directory, no.

I'll lift my review, as without having to worry about `splash_e_h` implementing some of these we can just set file-scope variables when running through the options and handle them when GDK is ready.

skef:
> To answer your question @skef, I don't think there's any practical difference between what I want and using the `python` directory, no.

I that case my suggestion for python support would be:

1) Break the inner part of `python.c:LoadFilesInPythonInitDir()` out into a separate function, as in:

```
static void LoadPythonFile(char *pathname) {
        File *fp;
        fp = fopen( pathname, "rb" );
        if ( fp==NULL ) {
            fprintf(stderr,"Failed to open script \"%s\": %s\n",pathname,strerro
r(errno));
            continue;
        }
        PyRun_SimpleFileEx(fp, pathname, 1/*close fp*/);
}
```

2) (Naturally) call that from `LoadFilesInPythonInitDir()`
3) Pass an option additional pathname into `PyFF_ProcessInitFiles()`, where if it's not null it is 
passed to `LoadPythonFile`
4) Add a parameter similar to `-script` with an appropriate name that is documented as "supplying an additional initialization script". 

I think that last description should be fairly consistent with how that facility is described in the other documents, and I'm making this suggestion in part to avoid having to create a new "entity" to document. And anyway this would be handy when writing new customization scripts eventually meant for `config`. 

Unfortunately on the native scripting side I'm not sure there is such a facility. Still, it should be relatively if the new flag is called with `-lang=ff` to pass it off to whatever the interactive dialog passes it off to and then paper over the documentation issues.

PR #3929, opened 2019-09-14
Apply the vmtx formula correction to CFF and CFF CID dumps

jackhumbert:
This applies the change from #620 for the TTF dump to the CFF and CFFCID dumps. Have confirmed it works :)

### Type of change
- **Bug fix**

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -2539,7 +2539,7 @@ static int dumpcffhmtx(struct alltabs *at,SplineFont *sf,int bitmaps) {
 	    if ( dovmetrics ) {
 		if ( i<=at->gi.lastvwidth )
 		    putshort(at->gi.vmtx,sc->vwidth);
-		putshort(at->gi.vmtx,/*sf->vertical_origin-*/b.maxy);
+		putshort(at->gi.vmtx,sc->parent->ascent - b->maxy);
 	    }
 	    ++cnt;
 	    if ( i==at->gi.lasthwidth )
@@ -2590,7 +2590,7 @@ static void dumpcffcidhmtx(struct alltabs *at,SplineFont *_sf) {
 	    if ( dovmetrics ) {
 		if ( sc->ttf_glyph<=at->gi.lastvwidth )
 		    putshort(at->gi.vmtx,sc->vwidth);
-		putshort(at->gi.vmtx,/*sf->vertical_origin-*/b.maxy);
+		putshort(at->gi.vmtx,sc->parent->ascent - b->maxy);
 	    }
 	    ++cnt;
 	    if ( sc->ttf_glyph==at->gi.lasthwidth )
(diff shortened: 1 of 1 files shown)


skef:
For reference the comment on the commit that commented out `vertical_origin` says:

> Remove the old vertical_origin field of the splinefont. Stored in the 'BASE' table now (note this means that the python commands to access this field is no longer meaningful).

That it was left as a comment suggests that GW didn't think the change was final.

jackhumbert:
Ah, I think the BASE table applies to the whole font though. I’m not sure if it’s relevant, but I wasn’t able to see any difference when changing the values there. [This is the guide I was following for that](https://docs.microsoft.com/en-us/typography/opentype/spec/baselinetags).

ctrlcctrlv:
What is OFF? OpenType...Font Format?

skef:
@jackhumbert I just dug that out because your modification naturally raises questions about the commented-out portion, so it could be helpful to have it handy. 

I'm not qualified to judge this change without more assistance/pointers to documentation. @khaledhosny might be, if he pokes his head in. Of our occasional reviewers he tends to have the strongest grip on these questions.

jackhumbert:
@ctrlcctrlv haha, I must have combined those when writing the title!

skef:
@khaledhosny I know you're busy and possibly conflicted, but if you could glance at and comment on this one it would be helpful and appreciated.

khaledhosny:
Sorry, but I don’t know much about `vmtx` metrics or how they are used.

frank-trampe:
Looking at the [specification](https://docs.microsoft.com/en-us/typography/opentype/spec/vmtx), assuming that the changed value represents `topSideBearing`, this change looks correct. I would suggest adding comments indicating the meaning of the various values.

skef:
No problem @khaledhosny 

Alright. All discussion of this issue seems to trace back one way or another to http://www.akenotsuki.com/eyeben/fonts/ffpatch20120731b.txt and http://www.akenotsuki.com/eyeben/fonts/kanren.html . 

@jackhumbert In preference to integrating that patch's wisdom over the next decade or two, I've tried to do it all at once in https://github.com/skef/fontforge branch p3929, including your suggested changes here. Are you in a position to test vertical metrics changes generally or is this just a targeted repair for you? 

What the changes do, following the `eyeben` doc:

1. Turns on vertical metrics by default when there are vertical metrics in the font (tested, works)
2. Fix the display location of the advance line (tested, seems to work)
3. Apply your fixes, and additional fixes in tottf.c:dumpmissingglyph()` for gh.ymax  and `dumpccfhtmx()` `b.miny` (the former seems right based on the other fixes, the latter I'm less sure of).
4. Initialize the `vwidth` to ascent+descent, just like `width`
5. Copy the vwidth from the dummy in `splinefont.c`.
6. Fix the direction of some arrows when displaying vertical metrics (in `charview.c`)

Can you review? Do you (or does anyone else) know who can, beyond my basic testing?

skef:
This got merged as part of #3952. @jackhumbert thank you for bringing this to our attention and for the pointer leading back to the patch.

PR #3931, opened 2019-09-18
Fix crash on hiDPI while drawing closed eye

### LINKED ISSUE #3925: FontForge hanging when showing/hiding layer in high DPI environment (Linux)
- [ ] Screenshot

**Version**
```
 Based on sources from 08:50 UTC 17-Feb-2019-ML-D.
 Based on source from git with hash: 
fontforge 08:50 UTC 17-Feb-2019
libfontforge 20190217
```
Running on Fedora/X11. I have a high-DPI screen; I set screen width to 26 inches, which makes the UI scale to work on the screen mostly fairly well. In the glyph viewer, however, trying to check or uncheck the visibility of a layer (which at this DPI is done by showing of an eye icon instead of a check box) causes FF to hang indefinitely.

To reproduce, run FF and set the screen width to 26 inches (resolution might need to be 3840x2160 as well). Then try to show/hide a layer.

ctrlcctrlv:
Drawing the closed eye causes a crash on hiDPI, see #3925.

This fixes the logic error found by @skef with the help of GCC's
warnings. Thanks to @dbenjaminmiller for the report.

Interestingly, this crash only happens in release builds!

This closes #3925

### Type of change

- **Bug fix**

### DIFF
--- a/gdraw/gradio.c
+++ b/gdraw/gradio.c
@@ -350,7 +350,7 @@ return( false );
 
     } else if ( (!gr->ison) && gr->onbox == &visibility_on_box ) {
          /* draw closed eye */
-        GPoint pts[6];
+        GPoint pts[7];
         int c,i;
         double angle;
 	int bp = gr->onbox->border_type==bt_none ? 0 : GDrawPointsToPixels(pixmap,gr->onbox->border_width);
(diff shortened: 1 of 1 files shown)


gdraw/gradio.c
```diff
@@ -367,7 +367,7 @@ return( false );
             pts[i].y=.5*h*sin(angle)+y+h/4;
 
              /* draw lashes */
-            if (i>0 && i<5) GDrawDrawLine(pixmap, pts[i].x,pts[i].y, .75*w*cos(angle)+x+w/2, .75*h*sin(angle)+y+h/4, fg);
+            if (i>0 && i<6) GDrawDrawLine(pixmap, pts[i].x,pts[i].y, .75*w*cos(angle)+x+w/2, .75*h*sin(angle)+y+h/4, fg);
             ++i;
```
skef: I would favor moving this `++i` after `c++` (and a comma) in the for loop, as that is the logic, they're initialized together, and having the iteration separate is confusing.

ctrlcctrlv: Done in https://github.com/fontforge/fontforge/pull/3931/commits/772f31dfd5d8c5b437b36c28f4d4f4e66da2600a

ctrlcctrlv:
No problem @skef&mdash;I actually agree with you, it does look better!

jtanx:
When is `c` ever different from `i` anyway?

ctrlcctrlv:
:man_facepalming: I don't think it ever is, so I'll remove it

skef:
Hmm, that's a good point. They probably started out that way and gradually converged, ending when the if statement was added.

PR #3932, opened 2019-09-18
Explicitly disable in-source builds, remove ENABLE_DEBUG and use NDEBUG

jtanx:
This should address the last point raised in #3880.

It turns out you can in fact do in source builds, but I've now explicitly disallowed it, since you shouldn't be doing that.

### Type of change
- **Non-breaking change**


ctrlcctrlv:
My current setup does in-source builds, but I guess I can change. :-)

ctrlcctrlv:
I don't understand how I'm supposed to build FontForge now, so on my local machine I've rolled this back just so I could get some work done today. Even if I ran `cmake ..` inside a `build` dir I'd get the error, confusing.

skef:
@ctrlcctrlv Say your `CWD` is the top level of the repository, where `fontforge`, `gdraw` and `contrib` are subdirectories. Then you can do
```
mkdir build
cd build
cmake -GNinja ..
```
to do a release build, and add the other arguments on the (updated) Wiki to do a debug build. I've just brought over the current master and confirmed the release and debug builds (except for `ccache`, which I don't use yet).

ctrlcctrlv:
Oh OK.

By the way, what is the "right" way to enable ASan?

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 2e14d20e8..baf9e4ac4 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,9 @@ include(GNUInstallDirs)
 include(PackageUtils)
 include(TargetUtils)
 
+set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fno-omit-frame-pointer -fsanitize=address")
+set (CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -fno-omit-frame-pointer -fsanitize=address")
+set (CMAKE_LINKER_FLAGS_DEBUG "${CMAKE_LINKER_FLAGS_DEBUG} -fno-omit-frame-pointer -fsanitize=address")
 # Set any global defines
 set(CMAKE_C_STANDARD 99)
 set(CMAKE_C_STANDARD_REQUIRED TRUE)
```

This can't be right, but it's how I figured out #3939...

ctrlcctrlv:
```
[fred@pc build]$ pwd
/home/fred/Workspace/fontforge2/build
[fred@pc build]$ cmake -GNinja ..
CMake Error at CMakeLists.txt:7 (message):
  In-source builds are disallowed.  Create a build folder to run CMake from.


-- Configuring incomplete, errors occurred!
See also "/home/fred/Workspace/fontforge2/CMakeFiles/CMakeOutput.log".
See also "/home/fred/Workspace/fontforge2/CMakeFiles/CMakeError.log".

```

ctrlcctrlv:
The problem was `../CMakeCache.txt`

PR #3936, opened 2019-09-19
Make sure glyph should be in GDEF table

### LINKED ISSUE #3930: [UFO] FontForge adds non-existent glyphs to GDEF

Arch Linux, 07f502f59995b0f64c10d21379096a2a47bd40f1

Only the glyphs in the SFD should be written to the GDEF section of `features.fea` in the UFO directory.

* New font
* Double-click &lsaquo;A&rsaquo; in FontView, draw an A in CharView
* Immediately export UFO, `good.ufo`
* Open and immediately close B CharView
* Open C CharView, draw C, close it, then right-click C in FontView and choose &laquo;Clear&raquo;
* Export `bad.ufo`
* Do a diff to confirm problem:
    ```diff
    diff --git a/good.ufo/features.fea b/bad.ufo/features.fea
    index fae53f1..958eccd 100644
    --- a/good.ufo/features.fea
    +++ b/bad.ufo/features.fea
    @@ -1,6 +1,6 @@
 
 
    -@GDEF_Simple = [\A ];
    +@GDEF_Simple = [\A \B \C ];
 
     table GDEF {
       GlyphClassDef @GDEF_Simple, , , ;
    ```
Of course you can use a text editor to remove bad glyphs `\B` and `\C`. That gets tiresome after long sessions of exporting UFO to test in third-party software. I noticed this problem while using `fontmake` to make a variable font. `fontmake` quite rightfully chokes (in this case `\I` is the non-existent glyph):

```text
(afdko_env) [fred@pc CHOMSKY VF]$ fontmake -m ChomskyVF.designspace -o variable --production-names --output-path ChomskyVF.otf
INFO:fontmake.font_project:Building variable font ChomskyVF.otf
INFO:ufo2ft:Pre-processing glyphs
INFO:cu2qu.ufo:New spline lengths: 
INFO:ufo2ft.filters:Running DecomposeComponentsFilter on Chomsky
INFO:ufo2ft.filters:Running DecomposeComponentsFilter on Chomsky
INFO:ufo2ft:Building OpenType tables for Chomsky
[traceback omitted] 
KeyError: ('I', 'ClassDef')
```

So every time I wanted to see my variable font I had to edit the text file `features.fea` and remove the non-existent `\I` from GDEF.

ctrlcctrlv:
Third party tools, such as `fontmake`, choke if the GDEF table contains non-existent glyphs.

This applies the same test as in `ufo.c` to feature files.

This closes #3930. I was hoping @frank-trampe would look at that issue before I tried to fix this on my own, but he was busy. I think the problem is simple though; all the tests still pass, and this is obviously a bug, so there should be no problems in this equally simple fix.

### DIFF
--- a/fontforge/featurefile.c
+++ b/fontforge/featurefile.c
@@ -39,6 +39,7 @@
 #include "glif_name_hash.h"
 #include "lookups.h"
 #include "namelist.h"
+#include "splinesaveafm.h" // for SCWorthOutputting
 #include "splineutil.h"
 #include "tottf.h"
 #include "tottfgpos.h"
@@ -1619,6 +1620,7 @@ return;					/* No anchor positioning, no ligature carets */
 		do {
 		    _sf = sf->subfontcnt==0 ? sf : sf->subfonts[k];
 		    for ( gid=0; gid<_sf->glyphcnt; ++gid ) if ( (sc=_sf->glyphs[gid])!=NULL ) {
+			if (!SCWorthOutputting(sc)) continue;
 			if ( sc->glyph_class==i+2 || (sc->glyph_class==0 && gdefclass(sc)==i+1 )) {
 			    if ( len+strlen(sc->name)+1 >80 ) {
 				putc('\n',out); putc('\t',out);


frank-trampe:
Now that we know what the problem is, it's unlikely that I would have found it any faster than you did. I would not have expected such an obviously necessary check to be missing.

PR #3939, opened 2019-09-19
Fix allocation in EncMapCopy

### LINKED ISSUE #3935: Crash while switching back and forth between different CID's
(Note: The same font as in #3934 is causing this problem.)

```
double free or corruption (out)

0x0000155553a6b755 in raise () from /usr/lib/libc.so.6
(gdb) bt
#0  0x0000155553a6b755 in raise () at /usr/lib/libc.so.6
#1  0x0000155553a56851 in abort () at /usr/lib/libc.so.6
#2  0x0000155553aada38 in __libc_message () at /usr/lib/libc.so.6
#3  0x0000155553ab425a in  () at /usr/lib/libc.so.6
#4  0x0000155553ab5c50 in _int_free () at /usr/lib/libc.so.6
#5  0x00001555551253e9 in EncMapFree (map=0x5555615f2200) at fontforge/splineutil.c:6310
#6  0x0000555555676aa8 in FVShowSubFont (fv=0x555555a0ac10, new=0x555555c8f200) at fontforgeexe/fontview.c:2920
#7  0x000055555574c0b4 in gmenu_mouse (m=0x555561a537a0, event=0x7fffffffcca0) at gdraw/gmenu.c:948
#8  0x000055555574b07a in gmenu_eh (w=0x555555d38700, ge=0x7fffffffcca0) at gdraw/gmenu.c:1389
#9  0x0000555555737494 in _GGDKDraw_CallEHChecked (gw=<optimized out>, event=0x4, eh=0x0) at gdraw/ggdkdraw.c:323
#10 0x0000555555737494 in _GGDKDraw_DispatchEvent (event=<optimized out>, data=<optimized out>) at gdraw/ggdkdraw.c:1221
#11 0x00001555542cf8b4 in  () at /usr/lib/libgdk-3.so.0
#12 0x000015555427c904 in  () at /usr/lib/libgdk-3.so.0
#13 0x0000155553c603ae in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#14 0x0000155553c621c1 in  () at /usr/lib/libglib-2.0.so.0
#15 0x0000155553c62201 in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#16 0x000055555573a4e6 in GGDKDrawEventLoop (gdisp=0x555555a69e90) at gdraw/ggdkdraw.c:2187
#17 0x00005555557084f8 in fontforge_main (argc=<optimized out>, argv=<optimized out>) at fontforgeexe/startui.c:1416
#18 0x0000155553a57ee3 in __libc_start_main () at /usr/lib/libc.so.6
#19 0x00005555555aba0e in _start ()
```

Simple reproduction:

* Open [蒙纳简版黑.otf](https://github.com/fontforge/fontforge/files/3629472/default.zip)
* Encoding&rarr;Compact
* Switch CID to CHei3HK-BoldSupp
* Switch CID back to CHei3HK-Bold-Roman
* You should experience a crash

Alternate backtraces:

```
malloc(): unsorted double linked list corrupted

0x0000155553a6b755 in raise () from /usr/lib/libc.so.6
(gdb) bt
#0  0x0000155553a6b755 in raise () at /usr/lib/libc.so.6
#1  0x0000155553a56851 in abort () at /usr/lib/libc.so.6
#2  0x0000155553aada38 in __libc_message () at /usr/lib/libc.so.6
#3  0x0000155553ab425a in  () at /usr/lib/libc.so.6
#4  0x0000155553ab70cc in _int_malloc () at /usr/lib/libc.so.6
#5  0x0000155553ab9486 in calloc () at /usr/lib/libc.so.6
#6  0x0000155554f7ce62 in CIDSetEncMap (fv=0x555555c6ac00, new=0x555555d0e660) at fontforge/fontviewbase.c:1224
#7  0x0000555555676b06 in FVShowSubFont (fv=0x555555c6ac00, new=0x555555d0e660) at fontforgeexe/fontview.c:2927
#8  0x000055555574c0b4 in gmenu_mouse (m=0x555555d70ad0, event=0x7fffffffcca0) at gdraw/gmenu.c:948
#9  0x000055555574b07a in gmenu_eh (w=0x5555561d3390, ge=0x7fffffffcca0) at gdraw/gmenu.c:1389
#10 0x0000555555737494 in _GGDKDraw_CallEHChecked (gw=<optimized out>, event=0x4, eh=0x0) at gdraw/ggdkdraw.c:323
#11 0x0000555555737494 in _GGDKDraw_DispatchEvent (event=<optimized out>, data=<optimized out>) at gdraw/ggdkdraw.c:1221
#12 0x00001555542cf8b4 in  () at /usr/lib/libgdk-3.so.0
#13 0x000015555427c904 in  () at /usr/lib/libgdk-3.so.0
#14 0x0000155553c603ae in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#15 0x0000155553c621c1 in  () at /usr/lib/libglib-2.0.so.0
#16 0x0000155553c62201 in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#17 0x000055555573a4e6 in GGDKDrawEventLoop (gdisp=0x555555a69a10) at gdraw/ggdkdraw.c:2187
#18 0x00005555557084f8 in fontforge_main (argc=<optimized out>, argv=<optimized out>) at fontforgeexe/startui.c:1416
#19 0x0000155553a57ee3 in __libc_start_main () at /usr/lib/libc.so.6
#20 0x00005555555aba0e in _start ()
```

```
double free or corruption (!prev)

0x0000155553a6b755 in raise () from /usr/lib/libc.so.6
(gdb) bt
#0  0x0000155553a6b755 in raise () at /usr/lib/libc.so.6
#1  0x0000155553a56851 in abort () at /usr/lib/libc.so.6
#2  0x0000155553aada38 in __libc_message () at /usr/lib/libc.so.6
#3  0x0000155553ab425a in  () at /usr/lib/libc.so.6
#4  0x0000155553ab5c5c in _int_free () at /usr/lib/libc.so.6
#5  0x0000155554f6147c in CompactEncMap (map=0x555555cad2e0, sf=0x555555d0ed40) at fontforge/encoding.c:2131
#6  0x0000555555676b24 in FVShowSubFont (fv=0x555555cf8a00, new=0x555555d0ed40) at fontforgeexe/fontview.c:2930
#7  0x000055555574c0b4 in gmenu_mouse (m=0x5555561d6490, event=0x7fffffffcca0) at gdraw/gmenu.c:948
#8  0x000055555574b07a in gmenu_eh (w=0x555555d0ebb0, ge=0x7fffffffcca0) at gdraw/gmenu.c:1389
#9  0x0000555555737494 in _GGDKDraw_CallEHChecked (gw=<optimized out>, event=0x4, eh=0x0) at gdraw/ggdkdraw.c:323
#10 0x0000555555737494 in _GGDKDraw_DispatchEvent (event=<optimized out>, data=<optimized out>) at gdraw/ggdkdraw.c:1221
#11 0x00001555542cf8b4 in  () at /usr/lib/libgdk-3.so.0
#12 0x000015555427c904 in  () at /usr/lib/libgdk-3.so.0
#13 0x0000155553c603ae in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#14 0x0000155553c621c1 in  () at /usr/lib/libglib-2.0.so.0
#15 0x0000155553c62201 in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#16 0x000055555573a4e6 in GGDKDrawEventLoop (gdisp=0x555555a68340) at gdraw/ggdkdraw.c:2187
#17 0x00005555557084f8 in fontforge_main (argc=<optimized out>, argv=<optimized out>) at fontforgeexe/startui.c:1416
#18 0x0000155553a57ee3 in __libc_start_main () at /usr/lib/libc.so.6
#19 0x00005555555aba0e in _start ()

```

ctrlcctrlv:
This closes #3935.

### DIFF
--- a/fontforge/fontviewbase.c
+++ b/fontforge/fontviewbase.c
@@ -1211,8 +1211,8 @@ void CIDSetEncMap(FontViewBase *fv, SplineFont *new ) {
     if ( fv->cidmaster!=NULL && gcnt!=fv->sf->glyphcnt ) {
 	int i;
 	if ( fv->map->encmax<gcnt ) {
-	    fv->map->map = realloc(fv->map->map,gcnt*sizeof(int));
-	    fv->map->backmap = realloc(fv->map->backmap,gcnt*sizeof(int));
+	    fv->map->map = realloc(fv->map->map,gcnt*sizeof(int32));
+	    fv->map->backmap = realloc(fv->map->backmap,gcnt*sizeof(int32));
 	    fv->map->backmax = fv->map->encmax = gcnt;
 	}
 	for ( i=0; i<gcnt; ++i )
--- a/fontforge/splineutil.c
+++ b/fontforge/splineutil.c
@@ -6320,7 +6320,7 @@ EncMap *EncMapCopy(EncMap *map) {
     /* Ensure all memory available, otherwise cleanup and exit as NULL */
     if ( (new=chunkalloc(sizeof(EncMap)))!=NULL ) {
 	*new = *map;
-	if ( (new->map=malloc(map->enccount*sizeof(int32)))!=NULL ) {
+	if ( (new->map=malloc(map->encmax*sizeof(int32)))!=NULL ) {
 	    if ( (new->backmap=malloc(map->backmax*sizeof(int32)))!=NULL ) {
 		memcpy(new->map,map->map,map->enccount*sizeof(int32));
 		memcpy(new->backmap,map->backmap,map->backmax*sizeof(int32));


jtanx:
I somehow doubt this is the right fix, this is like throwing a hammer at the problem. Why is the conditional reallocation if the count is more than `encmax` not sufficent?

skef:
I'm not super-comfortable with this as it stands, given that it seems like an "it worked" fix rather than a "we understand what's going on" fix. I could be explained out of that view. 

Is this repeatable? Does adapting the logic that originally allocates these blocks (from `FVAddUnencoded`) not work?
```
            EncMap *map = fvs->map;
            if ( map->enccount+cnt>=map->encmax )
                map->map = realloc(map->map,(map->encmax += cnt+10)*sizeof(int));
            if ( sf->glyphcnt+cnt>=map->backmax )
                map->backmap = realloc(map->backmap,(map->backmax += cnt+10)*sizeof(int));
            for ( i=map->enccount; i<map->enccount+cnt; ++i )
                map->map[i] = map->backmap[i] = i;

```

skef:
@jtanx beat me to it. 

First off, is the bad value within the `map` or `backmap` block? Is there some path where one gets allocated and not the other? Perhaps on a "promotion" from non-CID to CID?

frank-trampe:
@ctrlcctrlv, I see two changes. One in word size and one in whether to reallocate. Did you test each change separately to see whether one achieved a positive test result without the other?

Given the seeming lack of a good reason that these particular changes would fix the problem, I feel like we're dealing with a bad optimization here or a bad index somewhere else.

ctrlcctrlv:
You're indeed right that I changed two things @frank-trampe. The underlying data type is `int32`, so `int` never made sense. However, just using `int32` doesn't solve the problem.

@skef The bad value is definitely in the `map` block, not `backmap`. That's why I put them on two lines, so I could be more sure.

I'll try adapting `FVAddUnencoded` and get back to you. You know I often propose "it worked" fixes to start a discussion, and I can't say that's not what I'm doing here.

ctrlcctrlv:
@skef This seems to work just as well:

```diff
diff --git a/fontforge/fontviewbase.c b/fontforge/fontviewbase.c
index 4b691e2bb..9ceba1468 100644
--- a/fontforge/fontviewbase.c
+++ b/fontforge/fontviewbase.c
@@ -1210,13 +1210,15 @@ void CIDSetEncMap(FontViewBase *fv, SplineFont *new ) {
 
     if ( fv->cidmaster!=NULL && gcnt!=fv->sf->glyphcnt ) {
        int i;
-       if ( fv->map->encmax<gcnt ) {
-           fv->map->map = realloc(fv->map->map,gcnt*sizeof(int));
-           fv->map->backmap = realloc(fv->map->backmap,gcnt*sizeof(int));
-           fv->map->backmax = fv->map->encmax = gcnt;
+       if ( fv->map->enccount+gcnt>=fv->map->encmax )
+       fv->map->map = realloc(fv->map->map,gcnt*sizeof(int32));
+       if ( new->glyphcnt+gcnt>=fv->map->backmax )
+       fv->map->backmap = realloc(fv->map->backmap,gcnt*sizeof(int32));
+       fv->map->backmax = fv->map->encmax = gcnt;
+       for ( i=0; i<gcnt; ++i ) {
+           fv->map->map[i] = i;
+           fv->map->backmap[i] = i;
        }
-       for ( i=0; i<gcnt; ++i )
-           fv->map->map[i] = fv->map->backmap[i] = i;
        if ( gcnt<fv->map->enccount )
            memset(fv->selected+gcnt,0,fv->map->enccount-gcnt);
        else {
```

Would you prefer it?

skef:
In a case where it would fail with the current code, what are `gcnt`, `enccount` and `encmax`?

frank-trampe:
Correct me if I'm wrong, and it's likely since I'm running on almost no sleep for three days, but it seems then that the problem before, if a meaningful one, was likely an assumption that the backward look-up was always the same size as the encoding or smaller.

skef:
@frank-trampe: @ctrlcctrlv says the fault is past `map`, not `backmap`, so while that seems like a promising explanation from the code that doesn't seem to be what's going on.

ctrlcctrlv:
@skef

```(rr) reverse-continue
Continuing.

Breakpoint 1, CIDSetEncMap (fv=0x55d98a407600, new=0x55d98a485ad0) at fontforge/fontviewbase.c:1213
1213            if ( fv->map->encmax<gcnt ) {
(rr) p gcnt
$3 = 18965
(rr) p fv->map->encmax
$4 = 18965
(rr) p fv->map->enccount
$5 = 18912
```

I'd say you're right @frank-trampe

frank-trampe:
@skef, I missed that. Is there any other logical explanation for why A would fail and B and C would succeed?

@ctrlcctrlv, is the crash invariant across optimization levels?

ctrlcctrlv:
Yes, it always crashes for me, both in the debug build and in the release build.

skef:
Good lord:
```
../contrib/fonttools/acorn2sfd.c:    outline.sf->map->map = malloc(outline.sf->glyphmax*sizeof(int32));
../fontforgeexe/fontview.c:                     map->map = realloc(map->map,(map->encmax += 20)*sizeof(int));
../fontforgeexe/groupsdlg.c:    map->map = realloc(map->map,(map->encmax+=100)*sizeof(int));
../fontforge/fvfonts.c:         map->map = realloc(map->map,(map->enccount+extras)*sizeof(int));
../fontforge/fvimportbdf.c:         map->map = realloc(map->map,(map->encmax = n+100)*sizeof(int));
../fontforge/fvimportbdf.c:         map->map = realloc(map->map,new*sizeof(int));
../fontforge/svg.c:    map->map = malloc(sf->glyphcnt*sizeof(int));
../fontforge/parsettf.c:    /*encmap->map = malloc(encmap->enccount*sizeof(int));*/
../fontforge/parsettf.c:                map->map = realloc(map->map,256*sizeof(int));
../fontforge/parsettf.c:            map->map = realloc(map->map,(base+extras)*sizeof(int));
../fontforge/scripting.c:           map->map = realloc(map->map,(map->encmax=newcnt+10)*sizeof(int32));
../fontforge/scripting.c:           map->map = realloc(map->map,(map->encmax = new->glyphcnt)*sizeof(int32));
../fontforge/sfd.c:     map->map = realloc(map->map,map->encmax*sizeof(int));
../fontforge/sfd.c:     map->map = realloc(map->map,enccnt*sizeof(int));
../fontforge/splineutil.c:      if ( (map->map=malloc(enccount*sizeof(int32)))!=NULL ) {
../fontforge/encoding.c:                fvs->map->map = realloc(fvs->map->map,(fvs->map->encmax = new->glyphcnt)*sizeof(int32));
../fontforge/encoding.c:                map->map = realloc(map->map,(map->encmax = map->enccount = max+extras)*sizeof(int32));
../fontforge/encoding.c:    map->map = malloc(map->enccount*sizeof(int32));
../fontforge/encoding.c:        map->map = realloc(map->map,(map->encmax+=10)*sizeof(int32));
../fontforge/python.c:      map->map = realloc(map->map,(map->encmax = sf->glyphcnt)*sizeof(int));
../fontforge/fontviewbase.c:        fv->map->map = realloc(fv->map->map,gcnt*sizeof(int));
../fontforge/fontviewbase.c:            map->map = realloc(map->map,(map->encmax += cnt+10)*sizeof(int));
../fontforge/fontviewbase.c:        map->map = realloc(map->map,(map->encmax += cnt+10)*sizeof(int));
```
The era of worrying about function call overhead. *\*sigh\**.

is `(map->encmax+=100)*sizeof(int)` even safe?

jtanx:
If I were left to my own devices I'd be switching to C++ and using std::vectors :P

ctrlcctrlv:
Just trying to incrementally fix crashes I find :-)

Not overhaul the way memory is allocated. :-D

skef:
@jtanx Sure, but just centralizing the allocation and deallocation logic for this kind of thing goes a long way. 

There's already an `EncMapNew()` in `splineutil.c`. One wonders if we might adapt it to optionally realloc (add `map` as an argument that can be NULL) and then call it everywhere. 

To my embarrassment I'm not up on my memory tools. How easy is it to determine where a block was allocated, and therefore where the seemingly-too-short map is?

jtanx:
Yeah I said that more in passing/as a joke (but seriously, I wouldn't be against switching to c++).

valgrind *usually* works, literally just prepend `valgrind` before running fontforge and it might tell you if you try to access a location you shouldn't

frank-trampe:
I usually run with `-v --track-origins=yes` in a case like this. Tracking origins is very expensive and slows the target program down considerably, but it is also very useful.

ctrlcctrlv:
@skef `rr` is a good way to follow memory around and see what values were right before a crash. Includes such commands as `reverse-step` and `reverse-continue`, which work with breakpoints, and also a `watch` command. Of course there's always AddressSanitizer and valgrind as well. 

I'm kind of dangerous with all three of these tools, I'll happily admit it. They sometimes help me submit patches which I don't fully understand why they work. I've only really been working on FontForge for a year while some of you are going on five or ten.

I think these patches are more helpful than unhelpful. For one thing, _they do work_, no small feat. For another, they help focus discussion on the crash by presenting a possible solution.

skef:
@ctrlcctrlv the usual feeling with this sort of problem is that papering over is a last resort. For myself, I wouldn't favor these changes without more fiddling. 

Since I'm *not* up on the tools, let me pose this hypothesis: It seems that `encmax` and the size of `map` are out of whack. That raises the question of how they got out of whack. Which raises the question of which of the too many allocation and reallocation paths for `map` in the code allocated this particular `map`. So that's the question I would like to answer, using some tool that can say "this memory block was allocated in function `blah()`". 

Can you do that @ctrlcctrlv, either from past experience or the suggestions already made, or do you/we need more specific instructions?

skef:
Actually, trying this myself for my own edification ...

ctrlcctrlv:
Sorry @skef, I was on the phone. I can certainly try it if you don't have time to

skef:
So far I'm mostly creating valgrind processes I can't kill ...

ctrlcctrlv:
Unfortunately I've got to log off for the night, but if you don't figure it out I'll work on it on Friday evening and Saturday (my time). One tip is that this only ever happens with &ldquo;compact&rdquo; `EncMap`s, and I don't think it ever is right. The whole calculation is off. It's not just this font, I can't find any CID font that works!

ctrlcctrlv:
[蒙纳繁彩蝶通圆.zip](https://github.com/fontforge/fontforge/files/3631073/default.zip)

(As proof of that, I've attached another CID font that's broken in case you need more data points.)

skef:
For whatever reason `valgrind` sees a lot of problems in the python initialization path that are a pain to skip through. Don't know if it's worth tracking that down or if it's "below" our code. 

`map` at the error point is allocated by `splineutil.c:EncMapCopy()`:
```
EncMap *EncMapCopy(EncMap *map) {
/* Make a duplicate 'new' copy of EncMap 'map', Return a NULL if error */
/* NOTE: new-enc also shares map->enc, so be careful if freeing either */
    EncMap *new;
    int n;

    /* Ensure all memory available, otherwise cleanup and exit as NULL */
    if ( (new=chunkalloc(sizeof(EncMap)))!=NULL ) {
        *new = *map;
        if ( (new->map=malloc(map->enccount*sizeof(int32)))!=NULL ) {
            if ( (new->backmap=malloc(map->backmax*sizeof(int32)))!=NULL ) {
                memcpy(new->map,map->map,map->enccount*sizeof(int32));
                memcpy(new->backmap,map->backmap,map->backmax*sizeof(int32));
                /* NOTE: This new->enc 'also' points to same map->enc. */
                if ( map->remap==NULL )
                    return( new );
                for ( n=0; map->remap[n].infont!=-1; ++n );
                if ( (new->remap=malloc(n*sizeof(struct remap)))!=NULL ) {
                    memcpy(new->remap,map->remap,n*sizeof(struct remap));
                    return( new );
                }
                free(new->backmap);
            }
            free(new->map);
        }
        free(new);
     }
    return( NULL );
}
```
When I change `enccount` in the first `malloc` line (that is in: `if ( (new->map=malloc(map->enccount*sizeof(int32)))!=NULL ) {`) to `encmax` the crash goes away for me.

jtanx:
Oh yeah, I forgot about that, having Python enabled when using valgrind is a pain, I believe it's to do with its garbage collector and it doing it's own allocation magic. I usually disable python scripting when debugging with valgrind.

ctrlcctrlv:
Good catch @skef&mdash;I&lsquo;ve force-pushed the branch, taking your suggestions in mind. It works well now on both my test cases!

fontforge/fontviewbase.c
```diff
@@ -1211,8 +1211,8 @@ void CIDSetEncMap(FontViewBase *fv, SplineFont *new ) {
     if ( fv->cidmaster!=NULL && gcnt!=fv->sf->glyphcnt ) {
 	int i;
 	if ( fv->map->encmax<gcnt ) {
-	    fv->map->map = realloc(fv->map->map,gcnt*sizeof(int));
-	    fv->map->backmap = realloc(fv->map->backmap,gcnt*sizeof(int));
+	    fv->map->map = realloc(fv->map->map,gcnt*sizeof(int32));
+	    fv->map->backmap = realloc(fv->map->backmap,gcnt*sizeof(int32));
 	    fv->map->backmax = fv->map->encmax = gcnt;
```
ctrlcctrlv: @frank-trampe This change is just to make this code more readable, `int` is the wrong data type so this was never correct, it always should have been `int32`. Just FYI.

fontforge/splineutil.c
```diff
@@ -6320,9 +6320,9 @@ EncMap *EncMapCopy(EncMap *map) {
     /* Ensure all memory available, otherwise cleanup and exit as NULL */
     if ( (new=chunkalloc(sizeof(EncMap)))!=NULL ) {
 	*new = *map;
-	if ( (new->map=malloc(map->enccount*sizeof(int32)))!=NULL ) {
+	if ( (new->map=malloc(map->encmax*sizeof(int32)))!=NULL ) {
 	    if ( (new->backmap=malloc(map->backmax*sizeof(int32)))!=NULL ) {
-		memcpy(new->map,map->map,map->enccount*sizeof(int32));
+		memcpy(new->map,map->map,map->encmax*sizeof(int32));
```
skef: Quibble, but this one is still better as `enccount`. `enccount` is the number of entries in the map, `encmax` is the number of entries the map has room for. There's no reason to `memcpy` the non-entries after `enccount` into the map copy.

ctrlcctrlv:
I gave it another force push @skef

frank-trampe:
Fine with me.

PR #3952, opened 2019-09-23
Integrate relevant parts of vertical metrics patch

skef:
I sidetracked discussion of #3929 with additional code suggestions that weren't convenient to discuss, so I'm proposing those ideas here. These changes include @jackhumbert's code from that PR. 

As I noted there, it seems that there has been a widely referenced [patch](http://www.akenotsuki.com/eyeben/fonts/ffpatch20120731b.txt) for vertical metrics discussed here: http://www.akenotsuki.com/eyeben/fonts/kanren.html . Over time bits of this patch have been integrated through individual PRs, but as far as I can tell the FontForge devs have not evaluated (or been aware of) the bulk of it. This PR represents my attempt to. 

What the changes do, following the `eyeben` doc:

1. Turns on vertical metrics by default when there are vertical metrics in the font (tested, works)
2. Fix the display location of the advance line (tested, seems to work)
3. Apply @jackhumbert's fixes, and additional fixes in `tottf.c:dumpmissingglyph()` for `gh.ymax`  and `dumpccfhtmx()` for `b.miny` (the former seems right based on the other fixes, the latter I'm less sure of -- could just be wrong).
4. Initialize the `vwidth` to ascent+descent, just like `width`
5. Copy the vwidth from the dummy in `splinefont.c`.
6. Fix the direction of some arrows when displaying vertical metrics (in `charview.c`)

What I've tested works well, but I'm far from an expert on this subject.

### DIFF
--- a/fontforge/splinechar.c
+++ b/fontforge/splinechar.c
@@ -468,7 +468,7 @@ return;
 		(!sc->parent->layers[layer].background && SCWasEmpty(sc,layer)))) {
 	sc->widthset = false;
 	if ( sc->parent!=NULL && sc->width!=0 )
-	    sc->width = sc->parent->ascent+sc->parent->descent;
+	    sc->vwidth = sc->width = sc->parent->ascent+sc->parent->descent;
 	AnchorPointsFree(sc->anchor);
 	sc->anchor = NULL;
 	StemInfosFree(sc->hstem); sc->hstem = NULL;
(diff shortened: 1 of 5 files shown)


frank-trampe:
I didn't review line-by-line against the specification, but the facts (1) that somebody saw fit to make the patch, (2) that others have used it, and (3) that you were happy with test results makes this seem worthwhile to merge. If there are regressions, @ctrlcctrlv shall certainly encounter them before the next release.

PR #3955, opened 2019-09-26
Add a warning about our poor CID support

ctrlcctrlv:
The memory warning really is not applicable in 2019, the numerous crashes are much more serious.

Happily, flattening CID fonts never seems to fail&mdash;I ran it on hundreds of CID fonts and have not gotten any crashes. So just recommend that instead. Fixing the actual problem would take a long time and I doubt it would have any benefit over just flattening the CID fonts&mdash;that's my opinion after experimenting with this for a week or so.

**(Note: This is a documentation-only change.)**

### DIFF
--- a/doc/html/cidmenu.html
+++ b/doc/html/cidmenu.html
@@ -224,6 +224,17 @@
     <LI>
       &lt;sub font names&gt;
   </UL>
+  <P>
+  <FONT COLOR="Red"><STRONG><BIG>Warning:</BIG></STRONG></FONT> FontForge's
+  support for CID-keyed fonts is rudimentary; especially attempting to use
+  Compact encodings and the Metrics View in CID-keyed fonts is known to cause
+  issues and crashes. At one time, the memory usage of CID-keyed fonts was seen
+  as very worrisome (&approx;250MB), however as even most modern smartphones
+  can support such usage it is no longer a concern for the majority of users.
+  Therefore, we recommend that if you will make serious edits to a CID-keyed
+  font, you <a href="cidmenu.html#CID">flatten</a> it first to prevent crashes
+  and memory corruption. Bug fixes are always welcome&mdash;to start, see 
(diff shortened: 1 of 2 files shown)


ctrlcctrlv:
This is as good a place as any to note this I think.
この事実を書き留めるここはいいと思う。

Today I had a chat with @mandel59 (Ryuusei), mostly in Japanese, about this subject.
今日、この話題についてmandel59(竜星)さんと、(主に日本語で)話をした。

We came to the conclusion that CID fonts are a legacy format. "Sfnt-wrapped" CID fonts also.
CIDフォントは、レガシーフォーマットであることは同意。SFNT形CIDフォントも。

`halt`, `valt`, `hwid`, etc. are to be used instead for new fonts.
新しいフォントには、「halt」、「valt」、「hwid」などが使用することおすすめです。

My advice to flatten them if using them as a base for a new font is sound.
既存のCIDフォントから新しいフォントを作る、「flatten」の勧告は健全です。

Please take note @jtanx, @skef, and anyone else who is interested.
@jtanx、@skefなど読んでください。

Thank you once again for your time @mandel59 👍 
mandel59さんありがとうございました（＾人＾）感謝

I was very concerned about this for a while. I was waiting for someone who actually makes CJK fonts to weigh in.
しばらく心配してました。 CJKフォントを作っている人の意見を待っていました。

Other sources:
https://blogs.adobe.com/CCJKType/files/2018/12/cjkv2e-pp387-393.pdf (esp. p. 392)
https://blogs.adobe.com/CCJKType/2018/12/farewell-cidfonts.html

ctrlcctrlv:
Added above as [#D6](https://github.com/fontforge/fontforge/wiki/Community-guidelines#D6), mostly so I had an excuse to write [a script](https://gist.github.com/ctrlcctrlv/74d5c5ac9599f8000394490c1ed45a83) to add section links and a table of contents...

As a wise man once said, "[Never spend 6 minutes doing something by hand when you can spend 6 hours failing to automate it.](https://twitter.com/zhuowei/status/1254266079532154880?lang=en)" (Not that 30 minutes is 6 hours.)

PR #3963, opened 2019-10-05
Bitmap view colors

keidax:
I've been working on a bitmap font, and found myself really wanting a darker view for nighttime work. This exposes the color options for a nicer dark theme:
bitmap_dark

If this is acceptable so far, I'll keep going and include the rest of the hard-coded colors in `bitmapview.c`, which are mostly for references and selections.

### Type of change

- **New feature**
- **Non-breaking change**

### DIFF
--- a/fontforgeexe/bitmapview.c
+++ b/fontforgeexe/bitmapview.c
@@ -2121,6 +2121,16 @@ GResInfo bitmapview_ri = {
     NULL
 };
 
+void BVColInit(void) {
+    static bool cinit = false;
+
+    if (cinit)
+	return;
+
+    cinit = true;
+    GResEditFind(bitmapview_re, "BitmapView.");
+}
+
 static GMenuItem2 wnmenu[] = {
     { { (unichar_t *) N_("New O_utline Window"), NULL, COLOR_DEFAULT, COLOR_DEFAULT, NULL, NULL, 0, 1, 0, 0, 0, 0, 1, 1, 0, 'u' }, H_("New Outline Window|Ctl+H"), NULL, NULL, BVMenuOpenOutline, 0 },
     { { (unichar_t *) N_("New _Bitmap Window"), NULL, COLOR_DEFAULT, COLOR_DEFAULT, NULL, NULL, 1, 0, 0, 0, 0, 0, 1, 1, 0, 'B' }, H_("New Bitmap Window|Ctl+J"), NULL, NULL, /* No function, never avail */NULL, 0 },
@@ -2298,6 +2308,7 @@ return;
     mb2DoGetText(mblist);
     for ( i=0; BVFlipNames[i]!=NULL ; ++i )
 	BVFlipNames[i] = S_(BVFlipNames[i]);
+    BVColInit();
     atexit(&BitmapViewFinishNonStatic);
 }
 
--- a/fontforgeexe/bitmapview.c
+++ b/fontforgeexe/bitmapview.c
@@ -57,6 +57,12 @@ struct bvshows BVShows = { 1, 1, 1, 0 };
 #define RPT_DATA	24		/* x,y text after above */
 #define RPT_COLOR	40		/* Blob showing the foreground color */
 
+static Color guide_color = 0x404040;
+static Color width_guide_color = 0x404040;
+static Color grid_color = 0xa0a0a0;
+static Color bitmap_color = 0x808080;
+static Color overview_fg_color = 0x000000;
+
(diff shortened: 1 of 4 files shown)


fontforgeexe/bitmapview.c
```diff
@@ -57,6 +57,11 @@ struct bvshows BVShows = { 1, 1, 1, 0 };
 #define RPT_DATA	24		/* x,y text after above */
 #define RPT_COLOR	40		/* Blob showing the foreground color */
 
+static Color guide_color = 0x404040;
+static Color grid_color = 0xa0a0a0;
+static Color bitmap_color = 0x808080;
+static Color overview_fg_color = 0x000000;
```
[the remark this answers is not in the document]
keidax: Hmm, could you point out where you mean? I was trying to follow the pattern of other colors, like https://github.com/fontforge/fontforge/blob/dfd4c06b126dbae8c8601900924a1326c7e99d76/fontforgeexe/fontview.c#L120-L123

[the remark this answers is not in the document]
ctrlcctrlv: Nevermind, I suppose you're right; it's not done that way for colors.

fontforgeexe/bitmapview.c
```diff
@@ -734,24 +737,24 @@ static void BVExpose(BitmapView *bv, GWindow pixmap, GEvent *event ) {
...
 	GDrawDrawLine(pixmap,0,-bv->yoff+bv->height-bv->bdf->ascent*bv->scale,
-		bv->width,-bv->yoff+bv->height-bv->bdf->ascent*bv->scale,0x404040);
+		bv->width,-bv->yoff+bv->height-bv->bdf->ascent*bv->scale,guide_color);
 	GDrawDrawLine(pixmap,0,-bv->yoff+bv->height+bv->bdf->descent*bv->scale,
-		bv->width,-bv->yoff+bv->height+bv->bdf->descent*bv->scale,0x404040);
-	GDrawDrawLine(pixmap,bv->xoff+0*bv->scale,0, bv->xoff+0*bv->scale,bv->height,0x404040);
-	GDrawDrawLine(pixmap,bv->xoff+bv->bc->width*bv->scale,0, bv->xoff+bv->bc->width*bv->scale,bv->height,widthcol);
+		bv->width,-bv->yoff+bv->height+bv->bdf->descent*bv->scale,guide_color);
+	GDrawDrawLine(pixmap,bv->xoff+0*bv->scale,0, bv->xoff+0*bv->scale,bv->height,guide_color);
+	GDrawDrawLine(pixmap,bv->xoff+bv->bc->width*bv->scale,0, bv->xoff+bv->bc->width*bv->scale,bv->height,guide_color);
```
skef: These last two presumably defaulted to 0x000000 (from the definition in `charview.c` and could be changed separately from the other `guide_color` values. Is that likely to matter?

keidax: Good point. Here's what the defaults look like prior to this change:
<details>
<summary>Screenshot</summary>

bitmap_default

</details>

There's enough contrast against the background, I never even noticed the advance width was a different color from the other guide lines. So I don't think making them the same is a noticeable change.

Personally I don't need to see the advance as a different color. But in case someone does care, I can keep that as a separate option.

keidax:
Updated with the rest of the colors. This should handle everything except for the palette used in a grayscale bitmap font.

fontforgeexe/bitmapview.c
```diff
@@ -734,34 +749,31 @@ static void BVExpose(BitmapView *bv, GWindow pixmap, GEvent *event ) {
...
-		    bv->width,-bv->yoff+bv->height-(bv->bdf->ascent-bc->vwidth)*bv->scale,widthcol);
+		    bv->width,-bv->yoff+bv->height-(bv->bdf->ascent-bc->vwidth)*bv->scale,width_guide_color);
     }
     if ( bv->showfore ) {
 	/* Reference names are drawn after grid (otherwise some characters may get unreadable */
 	for ( bref=bc->refs; bref!=NULL; bref=bref->next )
 	    BVDrawRefName( bv,pixmap,bref,0 );
     }
     if ( bv->showoutline ) {
-	Color col = (view_bgcol<0x808080)
```
skef: "For the record" I read this as switching from a run-time heuristic to letting the user pick a color, good or bad, and that seems fine to me.

skef:
I'll merge this in a day or two if no one else objects, has other further input, or merges it first.

PR #3966, opened 2019-10-11
Use proper offset for BitmapView scrollbar

keidax:
This fixes a minor annoyance: the first time I try to drag the horizontal scrollbar, the entire view jumps to the left.

Before scrolling anything, the whole view is set up correctly, but this offset was flipped. It didn't seem to have any effect until the scrollbar is touched.

### Type of change

- **Bug fix**


jtanx:
Mac build is still failing because it's using the old config (from before #3967), I'll say that's fine to ignore...

PR #3971, opened 2019-10-12
gggdkdraw.c: always hide the window before destroying it

### LINKED ISSUE #3675: No window has focus after a window is closed
Closing a FontForge window does not return focus to the parent window. I am using FontForge 20190413 on macOS Mojave 10.14.4. This is a regression from FontForge 20170731. To reproduce:
* Open FontForge.
* Click “File” then “Open”, or perform any other action that opens a window.
* Close the window.
* Press arrow keys to move around in the main font grid. Nothing happens.
* Click the main window such that it regains focus.
* Press arrow keys to move around in the main font grid. Now it works.

jtanx:
Fixes #3675

### Type of change

- **Bug fix**


frank-trampe:
I can't test on Macintosh, and I'm not sure that anybody can, but the change looks logically viable and safe to me.

PR #3975, opened 2019-10-15
hotkeys.c: Remove autorepeat check

### LINKED ISSUE #3968: Undo/redo action does not autorepeat
hi
in new fontforge (after 2017) it's not possible to undo/redo more than once(i mean over and over by a delay) while holding shortcut keys on windows. this is a necessary feature specially when we work on a glyph for a long time.
thank you.

jtanx:
This did nothing under gxdraw anyway.

Fixes #3968

### Type of change

- **Bug fix**

### DIFF
--- a/gdraw/ggdkdraw.c
+++ b/gdraw/ggdkdraw.c
@@ -1510,7 +1510,7 @@ static void GGDKDrawRaise(GWindow w) {
     GGDKWindow gw = (GGDKWindow) w;
     Log(LOGDEBUG, "%p[%p][%s]", gw, gw->w, gw->window_title);
     if (!gw->is_visible) {
-        Log(LOGWARN, "Discarding raise on hidden window: %p[%p][%s]",
+        Log(LOGINFO, "Discarding raise on hidden window: %p[%p][%s]",
             gw, gw->w, gw->window_title);
         return;
     }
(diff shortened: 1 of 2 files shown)


frank-trampe:
I still don't entirely understand what purpose this code serves in gdraw and what practical effect is removal has.

jtanx:
When you hold down a key combination, the `autorepeat` flag gets set. Additionally, you will see events like:

\<key down\>, \<key up\>, \<key down\>, \<key up\> ...

for as long as you hold that key down.

Under gxdraw, *only* the key up events are marked as autorepeat, so you get:

\<key down:autorepeat=false\>, \<key up:autorepeat=true\>, \<key down:autorepeat=false\>, \<key up:autorepeat=true\> ...

Under ggdkdraw, both the key up and key down events are marked as autorepeat:

\<key down:autorepeat=true\>, \<key up:autorepeat=true\>, \<key down:autorepeat=true\>, \<key up:autorepeat=true\> ...

The hotkey check *only* fires on the key down event, ergo the autorepeat checks did nothing on gxdraw, so let's remove it.

jtanx:
The practical effect is if you hold down e.g. Ctrl+z, it continues undoing instead of only applying one undo.

jtanx:
Any objections to merging?

PR #3977, opened 2019-10-16
Fix crash in SFDTrimUndoOldToNew.

bungeman:
The innermost 'while' loop advanced 'oglyph' with
'SFDMoveToNextStartChar' but did not check the return value for 0 (to
check for the last entry). As a result, 'SCFindByGlyphName' could crash
in strcmp. This changes 'SCFindByGlyphName' not to crash in this
condition and also checks for oglyph being 0 where needed. This fixes
crashes while editing font lookups.

### Type of change

- **Bug fix**
- **Non-breaking change**

### DIFF
--- a/fontforgeexe/lookupui.c
+++ b/fontforgeexe/lookupui.c
@@ -3840,6 +3840,9 @@ char* SFDCreateUndoForLookup( SplineFont *sf, int lookup_type )
 }
 
 static SplineChar* SCFindByGlyphName( SplineFont *sf, char* n ) {
+    if (n == NULL) {
+        return 0;
+    }
     int i=0;
     for ( i=0; i<sf->glyphcnt; ++i ) {
 	if ( sf->glyphs[i]!=NULL ) {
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
Thanks! Can you tell me a set of steps I can use to trigger this crash in FontForge, though, or a script which triggers it?

bungeman:
I did have a setup which reliably reproduced this (which is how I managed to figure out what was going on) but now it isn't reproducing. It appears this has something to do with the autosave or undo with relation to adding and deleting glyphs. Unfortunately, it appears that the autosave or other files which existed at the time have been deleted and the original content of the sfd file I have is now somewhat different. I'm trying to figure out how this happened, but now I'm not sure. I'll try to reproduce again and save the state this time.

Mostly, this change is just about checking the return value of SFDMoveToNextStartChar at this location (as it is at the other locations it is used) and when 0 avoiding any dereference of 0 and then outputing the current nglyph (since it matched no oglyph).

frank-trampe:
I think that it's reasonable to guard against bad SFD input. But I would also like to figure how we are generating the bad auto-saved SFDs if possible.

ctrlcctrlv:
Closing as this PR is now stale and no example was ever provided, and it's unknown how the corrupt SFD was even produced (which would be better to fix).

PR #3984, opened 2019-10-21
Only categorize a point as a tangent when the cp vector points opposite the relevant point vector

skef:
When a Cubic spline point has a single control point and the question is "Is this a tangent point?" the current code tests if the cross product between the vector to the CP and the vector to the cp-less point is sufficiently small. However, it does not test the dot product to see if the two vectors are in opposite directions. This adds that test. 

You can test with the contour in this by running simplify (which categorizes the points):

[tantest.sfd.txt](https://github.com/fontforge/fontforge/files/3749574/tantest.sfd.txt)

The bottom point should wind up as a tangent and the top shouldn't. Both do with the current master.


skef:
I incorrectly described the problem as specific to cubic splines, but the patch works just as well for quadratic splines.

frank-trampe:
The change makes logical sense to me.

PR #3990, opened 2019-10-28
Squashed Expand Stroke replacement commit

skef:
This pull request encompasses a complete replacement for the Expand Stroke facility as a single squashed commit.

By "complete" I do not refer to the project as a whole, but the level of functionality. I still need to:

1. Add images to the new HTML document, and
2. Add more specific automated tests. 

However, this request is not to push to master but into the "stroke" branch that I just created. This is therefore a pro-forma "for the record" request that I will self-approve.


ctrlcctrlv:
Hooray :tada:

PR #3993, opened 2019-10-29
Corrected: Unify PI to FF_PI and fix cp deselect function

skef:
Requested by @jtanx . 

I used FF_PI because PI alone would be hard to search for after the change, and M_PI would require math.h which might cause issues/be controversial.


jtanx:
Might as well replace the M_PI usages too, it's not even part of the C standard

skef:
OK -- I amended the M_PI changes to the commit and repushed.

PR #3995, opened 2019-10-30
Expand stroke: cleanup, disable round and arcs joins, restore square cap

skef:
This restores square cap support in the python and native APIs, disables arcs joins (for now), and limits round joins to circular nibs. 

Other than that it makes `splinestroke.c` compile cleanly with -Wall, mostly by removing unused variables and partly by way of a new debug macro in `inc/basics.h`.

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -4601,10 +4601,27 @@ static int Stroke_Parse(StrokeInfo *si, PyObject *args, PyObject *keywds) {
     c = FlagsFromString(cap,linecap,"linecap type");
     if ( c==FLAG_UNKNOWN )
 	return( -1 );
+    if ( c==lc_square ) {
+	c = lc_butt;
+	if ( si->extendcap!=0 )
+	    si->extendcap=1.0;
+    }
     si->cap = c;
     j = FlagsFromString(join,linejoin,"linejoin type");
     if ( j==FLAG_UNKNOWN )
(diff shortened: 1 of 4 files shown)


jtanx:
For reference, I still get this on splinestroke.c:

```
../fontforge/splinestroke.c: In function 'SplineContourOuterCCWRemoveOverlap':
../fontforge/splinestroke.c:296:18: warning: unused variable 'sp' [-Wunused-variable]
  296 |     SplinePoint *sp;
      |                  ^~
../fontforge/splinestroke.c: In function 'LineDist':
../fontforge/splinestroke.c:1224:14: warning: using integer absolute value function 'abs' when argument is of floating point type 'double' [-Wabsolute-value]
 1224 |     return   abs((l2.y-l1.y)*p.x - (l2.x-l1.x)*p.y + l2.x*l1.y -l2.y*l1.x)
      |              ^~~
../fontforge/splinestroke.c: In function 'FVStrokeItScript':
../fontforge/splinestroke.c:1964:53: warning: unused parameter 'pointless_argument' [-Wunused-parameter]
 1964 | void FVStrokeItScript(void *_fv, StrokeInfo *si,int pointless_argument) {
```

skef:
Oh, I didn't notice that comment was here rather than the main thread. I fixed those in my PR of a few minutes ago.

PR #3991, opened 2019-10-28
Expand Stroke replacement

skef:
Welcome `fontforge-user`s! The changes to the system are documented [here](https://raw.githack.com/fontforge/fontforge/7fb4ebdff4fc69b0281fd66707d1da5f1c34be48/doc/html/stroke.html). Follow the links in the first section to see the updated Python and native APIs.

---

This request is to merge the `stroke` branch into master, which will replace the Expand Stroke facility. 

As I am opening this the changes are *not* quite ready to be merged, even if there were no objections. I still need to add images to the new document and should add more specific tests. Even so, I believe the functionality of the replacement as I create this request includes all aspects of the current system and a number of new features, and is also more stable.  

In any case, the magnitude of the change is large enough that it is now time to get more eyes on the code and input into what else needs to change and how to test it. 

I expect I will make many of whatever additional changes are needed myself. However, I want to organize the discussion (and, with luck, some of the edits) the same way as the abandoned menus-2019 project. Therefore, I request that anyone who has a specific proposal for a change should either:

1. File a separate *issue*, tagged with the "stroke" label, with a description of and reasons for the change, or
2. File a *pull request* with the code for the change against the `stroke` branch (**not** the `master`). These should also get the "stroke" label.

This issue is therefore the place for general discussion, to ask questions about the new code, and to evaluate the ongoing quality of the `stroke` branch as changes are merged into it. I "reserve the right" to move aspects of the discussion into separate issues when necessary (or when I feel like it). 

The main places to look are:

1. `fontforge/splinestroke.c` and `fontforge/splinestroke.h`
2. `doc/html/stroke.html` (a document describing the system and its parameters.)
3. `fontforgeexe/cvstroke.c` (the UI changes)
4. `fontforge/python.c` and `doc/html/python.html` (the new python interface (which includes the legacy interface)
5. `fontforge/scripting.c` and `doc/html/scripting-alpha.html` (the new variant of ExpandStroke for the native scripting language)

Some other structural notes:

I added `utanvec.c` and `utanvec.h`, which contain slope-related code. 

I moved the spline fitting code into `splinefit.c` and `splinefit.h`, and extended it a little. `git blame` still works as long as you add `-C`. It takes a little longer but not too much. 

@ctrlcctrlv has been evaluating earlier versions of this code for a while now. 

Closes #3777 
Closes #3762 
Closes #3718 
Closes #3264 
Closes #3242
Closes #3240
Closes #3239 
Closes #2547 
Closes #1424 
Closes #1228 
[...]

PR #3998, opened 2019-10-30
FindMonotonicIntersection fix and splineoverlap cleanup

### LINKED ISSUE #529: Remove overlap stuck in infinite loop on 32-bit Linux
This seems very similar to #257, but the fix for #257 doesn't seem to have fixed it.  Load https://github.com/fontforge/debugfonts/blob/master/rmo-loop2.sfd , select the glyph it contains, remove overlaps, infinite loop.  Tested with the latest master head revision; fails on 32-bit Linux; doesn't fail on 64-bit Linux.

jtanx:
I turned up the warnings on splineoverlap.c and fixed most of them. I've `#if 0`'ed a bunch of code, but I'd prefer to just delete it.

Fix for FindMonotonicIntersection is to properly calculate the diff instead of doing that `bkp_x`/`bkp_y` logic.

### Type of change

- **Bug fix**
- **Non-breaking change**

Fixes #529. Relates to #3991.

### DIFF
--- a/fontforge/splineoverlap.c
+++ b/fontforge/splineoverlap.c
@@ -1641,7 +1641,7 @@ return;
 }
 
 static double CalculateMinimumDiff(double from, double diff) {
-    double next = from - nextafter(from, diff);
+    double next = nextafter(from, DBL_MAX) - from;
     assert(diff >= 0 && next > 0);
     return next > diff ? next : diff;
 }
--- a/fontforge/splineoverlap.c
+++ b/fontforge/splineoverlap.c
@@ -37,6 +37,7 @@
 #include "splineutil.h"
 #include "splineutil2.h"
 
+#include <assert.h>
 #include <math.h>
 #include <stdarg.h>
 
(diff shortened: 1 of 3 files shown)


fontforge/splineoverlap.c
```diff
@@ -1808,12 +1792,8 @@ return;		/* Not interesting. Only intersection is at an endpoint */
 			oncebefore = true;
 		    }
 
-		    /* This is a volatile code! */
-		    /* "diff" may become so small in comparison with "y", */
-		    /* that "y+=diff" might actually not change the value of "y". */
-		    // So we double diff until it is significant.
-		    bkp_y=y+diff;
-		    while (bkp_y==y) { diff *= 2; bkp_y = y+diff; }
+		    diff = CalculateMinimumDiff(y, diff);
```
jtanx: So this is pretty much the fix for the issue discussed in #3991 / #529

jtanx:
Good thing I had that assert, I misread the docs on nextafter...

skef:
Consider my approval amended to include that fix.

PR #4001, opened 2019-10-31
Update the link to point to a new website

yursha:
Old website is not loading

### Type of change
- **Non-breaking change**


skef:
I think the `fontforge.org` link may have been intentional, and that site is supposed to redirect to wherever FontForge winds up having a network home. 

In any case I'm not having much luck with `fontforge.org` but `www.fontforge.org` is redirecting well. Does the latter work for you? That could be the best solution for the README.md link ...

jtanx:
DNS records have been fixed (or close enough to fixed) for now.

PR #4003, opened 2019-11-01
Misc color fixes

skef:
1. Add transparency to openpathcol, use for thick path
2. Stop fiddling with open path point colors
3. Honor fill transparency (but not previewfill transparency)
4. Privilege first point color over extremum point color

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -194,7 +194,7 @@ static Color gridfitoutlinecol = 0x009800;
 static Color backoutlinecol = 0x009800;
 static Color foreoutlinecol = 0x000000;
 static Color clippathcol = 0x0000ff;
-static Color openpathcol = 0x660000;
+static Color openpathcol = 0x40660000;
 static Color backimagecol = 0x707070;
 static Color fillcol = 0x80707070;		/* Translucent */
 static Color tracecol = 0x008000;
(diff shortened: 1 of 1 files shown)


jtanx:
Some screenshots would help understand this. Any more background on the motivation for the changes?

skef:
Old: 

colcmp3
colcomp3

New:

colchange1
colchange2

Left most on-curve point: Is now always first point color rather than sometimes extremum color. I know that the first point also has a tiny arrow, but knowing it's the first is more useful than knowing it's an extremum. (I didn't even realize they were different and why until I started this little svg/doc side project I'm playing with.)

Fill: All fills were changed to opaque when preview mode was added (or changed?) in 2012. That's poor for non-preview fills. And anyway you can just set the opacity to 100% if that's what you want. 

Open path thick line: FontForge has painted the entire thick line the open path color. This is just terrible and mostly prevents you from seeing the real geometry. It seems to have been a result of poorly thought-out edit rather than a design decision. This adds opacity to the argument and uses it for the thick portion of the line. (Because the thin portion never had it as an option you don't loose anything. And again you can just set the opacity to %100 to restore the old behavior.)

Open path point colors: Also changing the point colors is gilding the lily. It mostly takes information away from the user.

skef:
Oh, by the way those old/new paths are different -- I accidentally deleted when I was switching and had to recreate it. (In case you're wondering about the small differences.)

jtanx:
Thanks for writing that out, that's easy to comprehend.

> Fill: All fills were changed to opaque when preview mode was added (or changed?) in 2012

I don't think non-cairo mode supports alpha fills, so it would have always just been like that.

fontforgeexe/charview.c
```diff
@@ -900,34 +900,13 @@ static void DrawPoint( CharView *cv, GWindow pixmap, SplinePoint *sp,
     char buf[16];
     int isfake;
 
-    if ( DrawOpenPathsWithHighlight
```
jtanx: So this is basically removing this option (`DrawOpenPathsWithHighlight`)?

skef: No, it's still there for the line. This just stops painting the points differently when the contour is open.

jtanx: Ah right.

skef:
@ctrlcctrlv I'm guessing you turn open contour highlighting off, but this might be of interest to you.

PR #4004, opened 2019-11-03
Cleanup cvXX labels

khaledhosny:
Drop nonm-existent cv00 feature, and add cv11-cv99 (I have a font that
uses up to cv54 so far).

https://docs.microsoft.com/en-us/typography/opentype/spec/features_ae#cv01-cv99

### Type of change
- **Non-breaking change**


skef:
These changes look fine at the code level and I personally don't object to merging them. And it's clear that something close to this change is needed. 

However I'm reluctant to merge unilaterally because of the `lookupui.c` implications. Off the top of my head it seems like there are at least four ways to go:

1. Just have users live with the scrolling -- it's fine.
2. Put the higher numbered cv-ss (and ss-s?) at the bottom. Easy to do, but a bit mysterious and maybe confusing.
3. Hack in sub-menu support, probably based on an extra "fake" `otlookup_typemask`. 
4. Move the style options to their own top-level menu, possibly in conjunction with 2 (because it would be clearer what's going on). 

Obviously 3 and 4 would be a bit of work. @khaledhosny do you have any thoughts on 2?

khaledhosny:
I’d go with 1), it is already the case for language and script menus, and one can always type the code if scrolling is too annoying.

skef:
The language/script menu point is a very good one. 

Alright -- I'll merge this in a couple days unless there are objections. 

BTW @khaledhosny would you be able and willing to contribute a review of #4006? It's right in the same area and you're the local expert on these spec issues.

PR #4006, opened 2019-11-03
Fix and expand the friendly OpenType feature list

dscorbett:
I made various changes to the list of `struct opentype_feature_friendlynames`es.

I added the features 'ltra', 'ltrm', 'rclt', 'rvrn', 'stch', and 'vrtr'.

'alig' and 'dcap' are not mentioned in the current OpenType feature list, so I marked them obsolete. 'hngl' is still there but marked deprecated, so I marked it deprecated. 'vert' is still there and not marked obsolete, so I removed its “(obs)”.

I fixed the typo “Dipthongs”. I changed the name of 'ssty' from “Script Style” to “Math Script Style” to clarify what kind of script it means.

OpenType’s own suggestions for some of the features’ lookup types are incomplete or wrong, so I made FontForge’s suggestions better match how they are used.

### Type of change
- **Non-breaking change**


khaledhosny:
I’d do away with lookup types and make all features available to all lookup types (may be with the exception of very specialized features like `init` and co). But it is a minor UI issue anyway, since one can manually type any feature tag.

skef:
In that case, given that:

1. @dscorbett has been generally good with these suggestions in the past
2. What lists these features appear in is more a matter of convenience than program functionality.

I'll merge this in a couple days unless there are objections or someone merges it first.

frank-trampe:
@skef, fine with me.

PR #4011, opened 2019-11-09
Stroke: Misc code fixes and test coverage

skef:
Restore "poly" and "square" options to python interface
Correct direction of output contours built from open input contours
Small edits suggested by @jtanx
Break out convex nib control point errors by (documented) "Rules"
Test breadth of options in python and native APIs


skef:
I added a requirement that each nib contour be cubic. The alternative would be auto-conversion, but given that the validity rules are all in relation to cubic contours, and are reported in terms of selected points, I think that would be counter-productive. 

One happy thing is that I've verified that the Quadratic/Cubic setting in the dialog doesn't seem to be tied to the quadratic/cubic setting of the font, and the former comes up cubic by default. This is what we want.

PR #4017, opened 2019-11-13
Stroke: Misc fixes

skef:
Switch StrokeInfo to use width and height rather than radius and minorradius
Fix and then improve simplify handling
@jtanx code review responses
Relax post-SemiCircle position assert to .5 em.


skef:
The remaining conversions in effects.c and so forth are less "central", in my view.

PR #4018, opened 2019-11-13
Fix layer pickling

skef:
This one line change fixes the read side of `fontforge.layer` pickling.

For a code example look at the contemporaneous @dscorbett discussion in #3991. However it's pretty clear what's going on here:  The code for `PyFFi_newLayer()` seems to have been copied from `PyFFi_newContour()` directly above. The latter is passed `is_quadratic`, `closed` and then the points of the contour. The former is passed `is_quadratic` and then the contours of the layer. Hence the number of contours being `len-1` rather than `len-2`.


skef:
I'm gonna go ahead and merge this one now to make life easier for other parts of the project.

PR #4020, opened 2019-11-14
Fix for #4019 Crash when removing overlaps

### LINKED ISSUE #4019: Crash when removing overlaps
The `stroke` branch crashes when removing overlaps. As requested in #3991 I am reporting this as a new issue. To reproduce:
1. Merge 91287f9beec230c8df05cedff9c61e1bd2b93a35 into 9a0a6ba5220d8bcbc3b9958d5b729cc5c83f7cb7.
2. Build a non-debug release.
3. Download and extract [removeOverlap.zip](https://github.com/fontforge/fontforge/files/3844079/removeOverlap.zip).
4. Run script1.py.
5. Run script2.py.

script2.py crashes. script1.py produces problem_layer.P, which is also included in the zip.

skef:
The changes to `CleanMonotonics()` are not really part of the fix. All the dereferencing of `current_pointer` was probably fine (I stuck my head in here toward the beginning of the year, as I remember), but it makes my head hurt and is avoidable. (Unless I'm truly missing something.) 

The real problem was that when the first Monotonic is deleted the modified list head wasn't percolating up to `SplineSetRemoveOverlap()`. Maybe that could be fixed with a `volatile` but it's no big deal to modify (the static) `FindIntersections()`. 

Closes #4019

### DIFF
--- a/fontforge/splineoverlap.c
+++ b/fontforge/splineoverlap.c
@@ -521,23 +521,24 @@ static void MonotonicElide(struct mlist ** base, struct monotonic * input1) {
 }
 
 static void CleanMonotonics(struct monotonic ** base_pointer) {
-  // It is necessary to use double pointers so that we can set the previous reference.
-  struct monotonic ** current_pointer = base_pointer;
+  struct monotonic * current_pointer = *base_pointer;
   struct monotonic * tmp_pointer, * last_pointer = NULL;
-  while (*current_pointer) {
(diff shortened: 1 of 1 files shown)


frank-trampe:
This looks fine to me.

skef:
@frank-trampe As in "you have no objections" or as in "Approved"?

frank-trampe:
No objections, and the changes look safe, so approved.

skef:
@frank-trampe Sounds good. I'll do some upcoming testing with this fix to add to the verification.

PR #4021, opened 2019-11-14
Stroke: Better simplify parameters

### LINKED ISSUE #4008: Potential problems and solutions with Simplify heuristics
This issue is similar to #4007 in that I'm opening it as a place to discuss a particular aspect of the stroke-system-replacement and it may not result in any code being integrated. However, I suspect this one is more substantial than that one. 

Compared with the current master implementation the replacement system generally outputs more points on the contours. A lot of the difference is due to the low default `resolution` value of the current system but that isn't the whole story. Some of the difference is due to a design decision on my part -- I decided to just add points wherever and whenever it was convenient rather than doing extra work to avoid that. 

That's not because I don't care about the number of points in the end product. Rather, it's because:

1. I knew the output was going to be run through Remove Overlap anyway, which would produce a somewhat different geometry with different points that could potentially be removed. 
2. FontForge already has a system to remove redundant points on a contour relative to an accuracy metric: its Simplify algorithm. 

So it seemed to me that it would make more sense to rely on a pass through Simplify to remove unneeded points, and to make any engineering investment into improving that algorithm rather than fiddling with the replacement stroke system. 

Having said all this, I haven't yet changed Simplify in any way and I'm not even sure that I'm calling it with the best arguments. And in general the current system seems "timid" about removing points and converting splines to lines even with higher `err` values. 

I doubt it's necessary to do anything about Simplify prior to integration and release -- the problems of the current system seem to outweigh questions about point count. But:

1. It is worth discussing, and
2. I plan to redirect any questions about Expand Stroke point count to this issue

skef:
Now that my fat-fingered `&`'s have been pointed out and fixed, I've been able to experiment properly with various simplify parameters. 

I'm not entirely happy with the algorithm. `sf_ignoreslopes` is far too permissive so I've removed it. I wish there was an option to replace two existing points that are close together with a new, third, intelligently placed point. But these options seem acceptable and seem to have a good relationship to the `accuracy_target` parameter. 

Closes #4008

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -1860,7 +1860,7 @@ SplineSet *SplineSetStroke(SplineSet *ss,StrokeInfo *si, int order2) {
     DBounds b;
     real trans[6];
     struct simplifyinfo smpl = { sf_forcelines | sf_mergelines |
-                                 sf_smoothcurves | sf_ignoreslopes,
+                                 sf_smoothcurves | sf_setstart2extremum,
                                  0.25, 0.1, .005, 0, 0, 0 };
 
     if ( si->stroke_type==si_centerline )


skef:
This just swaps out one parameter for another, and we can reconsider the parameter set at any time. I'm going to merge this.

PR #4023, opened 2019-11-15
Stroke: More misc fixes 

skef:
Set si->leave_users_center to true by default
fix PostScript-style round join argument conversion
Make NibIsValid control point checks more tolerant (and finesse
  nib angles when necessary)
Tolerate not finding an expected angle on a spline, and substitute
  t=0 or 1 appropriately
Fix initialization of height in cvstroke.c

Most of these are simple and described by a line above. The control point check changes are to work around rounded control point values. 

More interesting is the "Tolerate not finding an expected angle". There are cases where the control point angles are extreme enough that `SplineSolveForUTanVec()` fails to find an angle that "should" be at an extreme of the spline. In that case it returns -1 and this change chooses 0 or 1 based on which angle is closer to the target. If you stroke letter Q in `tests/fonts/StrokeTests.sfd` with a 120/20/45 ellipse and the old code you can see the problem. (Accordingly, if you do the same with this code you should not see the problem.)

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -132,6 +132,7 @@ StrokeInfo *InitializeStrokeInfo(StrokeInfo *sip) {
     sip->extrema = true;
     sip->jlrelative = true;
     sip->ecrelative = true;
+    sip->leave_users_center = true;
     sip->joinlimit = 20.0;
     sip->accuracy_target = 0.25;
 
@@ -153,7 +154,7 @@ void SITranslatePSArgs(StrokeInfo *sip, enum linejoin lj, enum linecap lc) {
 	    sip->cap = lc;
     }
     switch (lj) {
-	case lc_round:
+	case lj_round:
 	    sip->join = lj_nib;
 	    break;
 	default:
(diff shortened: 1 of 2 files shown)


fontforge/splinestroke.c
```diff
@@ -265,6 +266,11 @@ static int LineSameSide(BasePoint l1, BasePoint l2, BasePoint p, BasePoint r,
 
     if ( RealWithin(tp, 0, 1e-5) )
 	return on_line_ok;
+    // Rounding means we should be tolerant when on_line_ok is true.
+    // Being that tolerant when on_line_ok is false would 
+    // yield too many false negatives. 
+    if ( on_line_ok && RealWithin(tp, 0, 1) )
```
jtanx: Seems like a pretty big jump in tolerance?

jtanx: right, so it's effectively

```c
if ((on_line_ok && RealWithin(tp, 0, 1)) || !RealWithin(tp, 0, 1e-5))
    return true;
```

jtanx: ok not quite.

skef: The intention is a looser margin for the `on_line_ok` is true case and a tighter one for `on_line_ok` is false. That way if being on the line is "not allowed" you can get within the tighter margin of it and still be "ok".

PR #4024, opened 2019-11-16
Potential fix for nonterminating AddExtrema loop

skef:
This is a potential fix for the loop found by @dscorbett reported in #3991. 

Here is a zip file with his script broken into two stages. The first creates and pickles a layer using the replacement Expand Stroke code without adding extrema. The second that reads the layer and runs `AddExtrema` from python, triggering the problem. (This way testing need not be tied to the `stroke` branch.)

[et.zip](https://github.com/fontforge/fontforge/files/3853661/et.zip)

The two line fix does eliminate the problem but I'm not confident it's the right systemic answer. This PR can serve as a context to hash that out. 

I've run into other problems with the extrema code but only fleetingly -- I don't have a collection of cases to report.

### DIFF
--- a/fontforge/splineutil2.c
+++ b/fontforge/splineutil2.c
@@ -3459,6 +3459,8 @@ static int ForceEndPointExtrema(Spline *s,int isto) {
 return( -1 );
     len = sqrt(len);
     if ( mylen<30*len && mylen<cplen && mylen<1 ) {
+	if ( mylen==0 )
+	    return -1;
 	if ( isto ) {
 	    s->to->noprevcp = true;
 	    s->to->prevcp = s->to->me;


skef:
So far this has avoided all of the non-terminating extrema loops I've run into.

skef:
Perhaps we can just integrate this to remove the non-terminating loop and file a separate issue about the general correctness of the algorithm? All this does is return -1 in a situation where the function would make no change otherwise.

jtanx:
Just took a brief look at this, the case where `mylen == 0` should only occur when `s->to->noprevcp` / `s->from->nonextcp` is true (for `isto` / `!isto`).

Not sure if that's any clearer, but that's the same effect as what exiting if `mylen == 0` does (I think)...

Anyway this seems like a plausible fix to me.

skef:
@jtanx It probably has the same effect with the chance of going through the loop an extra time. I'm pretty sure I've run into code that has the potential for colocating a control point without setting `nonextcp` to true. Then the question becomes: is it clearer to have the function not alter the geometry, only the "metadata", in that instance, or to exit before doing anything?

jtanx:
Fair enough, yeah I mean if both works, then I don't have a strong opinion either way. Happy with this change to go through.

skef:
I guess I'll leave this for @frank-trampe to look at and he can comment or, if he likes, merge.

skef:
Eh, I'm going to merge this. We can always do better in the future.

PR #4025, opened 2019-11-17
Stroke: Implement generalized Round joins and combine related code

skef:
This could use a bit more testing but is otherwise done.


skef:
Ah, well, I screwed up `ProjectPointOnLine`, in here, actually, but it's easy enough to fix. I'll do that and merge this.

PR #4031, opened 2019-11-23
Fix GDrawDrawImageMagnififed and add bounds checking to GDrawDrawImage

### LINKED ISSUE #3919: Fontforge occasionally crashes on windows when zooming in with background image
#3469 # Important

- [ ] Screenshot

Fontforge crashes when zooming in or out on an image
version: 2019-08-01, windows
reproduction steps:
-Open a new file
-Select any glyph
-Select back layer
-Upload a 1000by1000 pixel image in png format
-Try zooming in FontForge will occasionally crash

Note I tried this on 2 windows computers with entirely different specifications and build years.

workaround:
Before zooming in or out select the background and delete the image. After zooming use ctrl+z to bring it back.

jtanx:
Fixes the crash/oob memory access by fixing the casting of the pointer.

Also fixes the display of the image when at that zoom level.

### Type of change
- **Bug fix** Fixes #3919

### DIFF
--- a/gdraw/gdraw.c
+++ b/gdraw/gdraw.c
@@ -36,7 +36,6 @@
 #  include <sys/select.h>
 #endif
 
-
 /* Functions for font metrics:
     rectangle of text (left side bearing of first char, right of last char)
 */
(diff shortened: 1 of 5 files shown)


gdraw/ggdkcdraw.c
```diff
@@ -285,7 +285,7 @@ static cairo_surface_t *_GGDKDraw_GImage2Surface(GImage *image, GRect *src) {
     /*  premultiply each channel by alpha. We can reuse it for non-transparent*/
     /*  rgb images */
     if (base->image_type == it_true && type == CAIRO_FORMAT_RGB24) {
-        idata = ((uint32 *)(base->data)) + src->y * base->bytes_per_line + src->x;
+        idata = ((uint32 *)(base->data + src->y * base->bytes_per_line)) + src->x;
```
jtanx: This fixes the oob memory access

gdraw/gdraw.c
```diff
@@ -477,14 +476,40 @@ void GDrawScroll(GWindow w, GRect *rect, int32 hor, int32 vert) {
 
 /* draws the subset of the image specified by src starting at loc (x,y) */
 void GDrawDrawImage(GWindow w, GImage *img, GRect *src, int32 x, int32 y) {
-    GRect r;
-    if ( src==NULL ) {
-	struct _GImage *base = img->list_len==0?img->u.image:img->u.images[0];
-	r.x = r.y = 0;
-	r.width = base->width; r.height = base->height;
-	src = &r;
+    int width = GImageGetWidth(img);
```
jtanx: All the changes here and in charview.c are to fix what gets displayed when at that zoom level.

gdraw/ggdkcdraw.c
psychon: Assuming `base->data` is a byte-sized pointer (e.g. `char*`): This one line of the change looks good to me. I did not look at anything else in here.

PR #4032, opened 2019-11-24
Fixes for misc recent coverity problems

skef:
Fixes for Coverity 1407699 1407697 1407696(maybe) 1407695
                   1407694/1407693 1407691 1407690 1407689


fontforge/splinestroke.c
```diff
@@ -1109,9 +1110,10 @@ SplinePoint *TraceAndFitSpline(StrokeContext *c, Spline *s, bigreal t_fm,
     sti.starts_on_cusp = on_cusp;
     sti.found_trans = false;
 
-    if ( on_cusp && !TRACE_CUSPS )
+    if ( on_cusp && !TRACE_CUSPS ) {
 	GenStrokeTracePoints((void *)&sti, t_fm, t_to, &fpp);
-    else
+	free(fpp);
```
frank-trampe: Why wouldn't we null it here as before?

fontforge/splinestroke.c
```diff
@@ -1074,6 +1074,7 @@ int GenStrokeTracePoints(void *vinfo, bigreal t_fm, bigreal t_to,
 		                                      nib_ccw, CUSPD_MARGIN,
 						      stip->starts_on_cusp);
 		free(fp);
+		*fpp = NULL;
```
frank-trampe: This gets freed in a different branch below. Why wouldn't we free it here if unsetting it?

frank-trampe: It might be worth adding an in-line comment on the general logic.

fontforgeexe/cvstroke.c
```diff
@@ -170,36 +170,18 @@ static int _Stroke_OK(StrokeDlg *sd,int isapply) {
 	    err = true;
 	} else {
```
frank-trampe: So this branch always represents an error condition?

frank-trampe: Oops, comment in wrong place. Ignore.

fontforgeexe/cvstroke.c
```diff
@@ -170,36 +170,18 @@ static int _Stroke_OK(StrokeDlg *sd,int isapply) {
...
-			if ( sp->next==NULL )
-		    break;
-			sp = sp->next->to;
-			if ( sp==ss->first )
-		    break;
-		    }
-		}
-		if ( err ) {
+		enum ShapeType pt = NibIsValid(ss);
+		if ( pt!=Shape_Convex ) {
```
frank-trampe: So this branch always represents an error condition?

fontforge/splinestroke.c
skef: `GenStrokeTracePoints()` has two callers. One is the actual tracing function, which has logic based on the number of points returned, with zero points indicating there has been some error condition and there's nothing to be traced for now). The other caller is a level up, which uses the function to find the edges of cusps. 

These changes bring the treatment of the passed memory pointer in line with that semantic, setting it to NULL in the case no memory was allocated. That lets callers who don't care about the number of points to just free the value. 

I'll add some comments to this effect.

fontforgeexe/cvstroke.c
skef: If the contour isn't `Shape_Convex`, yes, that's an error. 

This code had a bunch of extra checks and sub-parts left over from when it was doing more things. The `!err` in the `for` guard is all that's needed at this point to stop further checks. The `if (!err)` added below is just to make sure we only call `GDrawRequestExpose()` once (because the nib selected points may change, so it's good to repaint them).

skef:
OK, I added a comment to the header file explaining the `ApproximateSplineSetFromGen()` conventions .

fontforge/splinestroke.c
jtanx: It could be clearer to just unconditionally null *fpp at the top of the function

skef:
@jtanx: done

Issue #4005, opened 2019-11-03
2 main problems with UFO format


aminabedi68:
hi
i am creating variable fonts with fontforge and fontmake, so i have to use UFO as output format of fontforge and the input format of fontmake, but there are 2 problems:

**1- wrong kerning pair codes writing method in features.fea file** ,fontforge writing these codes like this:
```
pos \uni062F <-122 0 -122 0> \uniFB7C  <-122 0 -122 0>;
```
but it's not working and the right code is this:

```
pos \uni062F \uniFB7C <-122 0 -122 0>;
```
I'm pretty sure fontforge writing code is wrong (looks corrupted even when you try to open with fontforge).

**2- issue about ligature mark suffix cause mark interpolation problems** ,please read this issue:
https://github.com/googlefonts/fontmake/issues/585#issuecomment-547480446

these are two main problems i know about creating variable fonts using opensource softwares.

skef:
Would it be possible to supply some sample files and export instructions that exhibit these problems? I believe that pair kerning and feature files works in at least some cases, so it would help to have an example.

aminabedi68:
hi @skef 
you can test on my font, <a href="https://github.com/aminabedi68/Estedad/blob/master/Source/variable/5-Estedad-V3.0%20-%20Black.sfd">Estedad </a> 
to see first issue you can simply open the font and generate as UFO, then open the UFO with fontforge again, you will see different kerning pairs table in some GPOS kerning tables like kernHorizontalKerninginArabiclookup2 if you generate a font with this UFO, kernings are not the same

fontforge file:
![fontforge_2019-11-05_22-22-45](https://user-images.githubusercontent.com/25493297/68236758-036fd880-001b-11ea-86c5-79762e71c53c.png)




generated UFO file:
![fontforge_2019-11-05_22-23-14](https://user-images.githubusercontent.com/25493297/68236792-14b8e500-001b-11ea-8907-fc07838353b1.png)
last two columns are the problem




for the second one just open features.fea in generated UFO folder with a text editor, you can see ligatures mark has not any suffix, for example line 1408 to 1410 is this:
```
  pos ligature \uniFEFB <anchor 830 0> mark @Anchor2 
    ligComponent
      <anchor 241 0> mark @Anchor2 ;
```

skef:
The kerning pair problem seems to be a simple misinterpretation of the spec that has gone unnoticed since the start of feature file support. I have a fix that I'll try to verify a bit more soon. 

[...]

PR #4033, opened 2019-11-25
Correct gpos_pair output

skef:
This is intended to fix the first sub issue in #4005

@khaledhosny I realize I lean on you a lot for this stuff but if you have a chance could you take a look? The modified lines go back to the first introduction of the file, and therefore of Adobe feature file support, and seem to be a straightforward misreading of [section 6.b](http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#6.b). (That is, format A *allows* two value records and this code is mistakenly writing the same value record twice, when it should be using format B.)

### DIFF
--- a/fontforge/featurefile.c
+++ b/fontforge/featurefile.c
@@ -1436,8 +1436,6 @@ return;					/* No support for apple "lookups" */
 			    fprintf( out, "    pos " );
 			    dump_glyphname(out,sc);
 			    putc(' ',out);
-			    dump_valuerecord(out,&pst->u.pair.vr[0]);
-			    putc(' ',out);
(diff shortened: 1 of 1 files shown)


khaledhosny:
LGTM.

However, this is still broken for RTL pair pos, as it needs a different format. Namely, `pos A B -50` becomes `pos A <-50 0 -50 0> B`. Detecting RTL kerning pairs is tricky, so it needs to be a explicit decision by the user, and I think FontForge already allows explicitly setting this format (which would be saved as pair pos kerning internally).

skef:
@khaledhosny thanks!

I'm going to merge this because a) the current code is grossly wrong and b) FontForge is known to have a bunch of other RTL problems. I'll try not to forget this one, though.

PR #4036, opened 2019-11-27
Stroke: Add figure of stroke dialog

skef:
I deferred adding the image until we agreed on a dialog design ... and then forgot. 

This should be multi-resolution friendly but I don't currently have good means of testing that. 

The documentation is arguably missing a "troubleshooting" section and an explanation of the visible
cusp issue, but I'm not sure many people would read that far.


skef:
And I'm gonna just merge this. Docs improvements are low risk to code correctness.

PR #4048, opened 2019-12-08
Reinstate fallback to compile time configured prefix if nothing else …

### LINKED ISSUE #4047: nerdfont patcher causes Segmentation fault during using /usr/lib/libfontforge.so.4.
- fontforge version: master branch
- os: Archlinux (Linux 5.4.1)

I run [Nerd Font patcher](https://github.com/ryanoasis/nerd-fonts/blob/v2.0.0/font-patcher)(which is using fontforge inside) to patch [Source Han Mono Refular Font](https://github.com/adobe-fonts/source-han-mono/blob/master/Regular/OTC/SourceHanMono-Regular.otf), and got Segmentation fault.

(When I select another font file, It runs successfully.)

Screenshot_2019-12-08_02-00-04

```gdb
(gdb) run ./font-patcher ~/Downloads/SourceHanMono-Regular.otf

Starting program: /usr/bin/python ./font-patcher ~/Downloads/SourceHanMono-Regular.otf
/usr/lib/../share/gcc-9.2.0/python/libstdcxx/v6/xmethods.py:731: SyntaxWarning: list indices must be integers or slices, not str; perhaps you missed a comma?
  refcounts = ['_M_refcount']['_M_pi']
  Ignoring 'VORG' vertical origin table

0x00007ffff7f1552b in __strrchr_avx2 () from /usr/lib/libc.so.6
```

```gdb
(gdb) backtrace
#0  0x00007ffff7f1552b in __strrchr_avx2 () at /usr/lib/libc.so.6
#1  0x00007ffff6b3b05e in getShareDir () at ../gutils/fsys.c:857
#2  0x00007ffff6b3b05e in getShareDir () at ../gutils/fsys.c:845
#3  0x00007ffff6b87b95 in FindCidMap
    (registry=0x5555567e8720 "Adobe", ordering=0x5555567e8740 "Identity", supplement=0, sf=<optimized out>, sf@entry=0x0) at ../fontforge/encoding.c:1121
#4  0x00007ffff6c11630 in cidfigure
    (fdselect=<optimized out>, subdicts=<optimized out>, gsubrs=0x7fffffffc200, scnt=<optimized out>, strings=<optimized out>, dict=<optimized out>, info=0x7fffffffc410) at ../fontforge/parsettf.c:3891
#5  0x00007ffff6c11630 in readcffglyphs (info=0x7fffffffc410, ttf=<optimized out>)
    at ../fontforge/parsettf.c:3996
#6  0x00007ffff6c11630 in readcffglyphs (ttf=<optimized out>, info=0x7fffffffc410)
    at ../fontforge/parsettf.c:3937
#7  0x00007ffff6c16ec6 in readttf
    (filename=0x5555557094e0 "/home/katio/Downloads/SourceHanMono-Regular.otf", info=0x7fffffffc410, ttf=0x55555555a440) at ../fontforge/parsettf.c:5584
#8  0x00007ffff6c16ec6 in _SFReadTTF
    (ttf=ttf@entry=0x55555555a440, flags=flags@entry=0, openflags=openflags@entry=(unknown: 0), filename=filename@entry=0x5555557094e0 "/home/katio/Downloads/SourceHanMono-Regular.otf", chosenname=chosenname@entry=0x0, fd=fd@entry=0x0) at ../fontforge/parsettf.c:6327
#9  0x00007ffff6cf247f in _ReadSplineFont (file=0x55555555a440,
    file@entry=0x0, filename=<optimized out>,
    filename@entry=0x5555556e29a0 "/home/katio/Downloads/SourceHanMono-Regular.otf", openflags=openflags@entry=(unknown: 0)) at ../fontforge/splinefont.c:1142
#10 0x00007ffff6cf349c in ReadSplineFont
    (filename=filename@entry=0x5555556e29a0 "/home/katio/Downloads/SourceHanMono-Regular.otf", openflags=openflags@entry=(unknown: 0)) at ../fontforge/splinefont.c:1322
#11 0x00007ffff6cf3629 in LoadSplineFont
    (filename=filename@entry=0x55555561a890 "/home/katio/Downloads/SourceHanMono-Regular.otf", openflags=(unknown: 0)) at ../fontforge/splinefont.c:1380
#12 0x00007ffff6c77519 in PyFF_OpenFont (UNUSED_self=<optimized out>, args=<optimized out>)
    at ../fontforge/python.c:1215
#13 0x00007ffff7bc0a03 in PyCFunction_Call () at /usr/lib/libpython3.8.so.1.0
#14 0x00007ffff7b69400 in _PyObject_MakeTpCall () at /usr/lib/libpython3.8.so.1.0
#15 0x00007ffff7bd2f1a in _PyEval_EvalFrameDefault () at /usr/lib/libpython3.8.so.1.0
#16 0x00007ffff7b9e1ea in _PyEval_EvalCodeWithName () at /usr/lib/libpython3.8.so.1.0
#17 0x00007ffff7b9ee0a in PyEval_EvalCodeEx () at /usr/lib/libpython3.8.so.1.0
#18 0x00007ffff7c2aa9c in PyEval_EvalCode () at /usr/lib/libpython3.8.so.1.0
#19 0x00007ffff7c2ab4a in  () at /usr/lib/libpython3.8.so.1.0
#20 0x00007ffff7c66655 in  () at /usr/lib/libpython3.8.so.1.0
--Type <RET> for more, q to quit, c to continue without paging--
#21 0x00007ffff7b456da in PyRun_FileExFlags () at /usr/lib/libpython3.8.so.1.0
#22 0x00007ffff7b47c23 in PyRun_SimpleFileExFlags () at /usr/lib/libpython3.8.so.1.0
#23 0x00007ffff7c69559 in Py_RunMain () at /usr/lib/libpython3.8.so.1.0
#24 0x00007ffff7c69649 in Py_BytesMain () at /usr/lib/libpython3.8.so.1.0
#25 0x00007ffff7ddb153 in __libc_start_main () at /usr/lib/libc.so.6
#26 0x000055555555505e in _start ()
```

I found function named `getShareDir` at https://github.com/fontforge/fontforge/blob/master/gutils/fsys.c#L845.

So I tried check `GResourceProgramDir`, like this:

```diff
diff --git a/gutils/fsys.c b/gutils/fsys.c
index 9d0fdf6de..cb6c49e03 100644
--- a/gutils/fsys.c
+++ b/gutils/fsys.c
@@ -853,6 +853,9 @@ char *getShareDir(void) {

     set = true;

+    if( GResourceProgramDir==NULL )
+           exit(32);
+
     //Assume share folder is one directory up
     pt = strrchr(GResourceProgramDir, '/');
     if ( pt==NULL ) {

```

Then, I got this.

Screenshot_2019-12-08_02-09-06
So, I think call strrchr with null makes `0x00007ffff7f1552b in __strrchr_avx2 () from /usr/lib/libc.so.6`

I guess, the value of the variable must be assgin at https://github.com/fontforge/fontforge/blob/master/gutils/fsys.c#L820, but there is no that function in the backtrace. 

- [x] Screenshot

jtanx:
…is available

Not a big fan of this at all, but I don't have time to invest in doing something better about this. Something that's also super crap is making assumptions about where the share/libdirs are relative to the bin dir, which could all be different depending on what's done at configure/compile time.

### Type of change

- **Bug fix**

Fixes #4047

### DIFF
--- a/gdraw/gresource.c
+++ b/gdraw/gresource.c
@@ -168,12 +168,7 @@ return;
     else
 return;
 
-    free(GResourceProgramDir);
-    GResourceProgramDir = _GFile_find_program_dir(prog);
-    if ( GResourceProgramDir==NULL ) {
-	GFileGetAbsoluteName(".",filename,sizeof(filename));
-	GResourceProgramDir = copy(filename);
-    }
+    FindProgDir(prog);
     free(GResourceFullProgram);
     GResourceFullProgram = copy(
 	    GFileBuildName(GResourceProgramDir,GResourceProgramName,filename,sizeof(filename)));
--- a/gdraw/gresource.c
+++ b/gdraw/gresource.c
@@ -36,7 +36,7 @@
 #include "ustring.h"
 #include "utype.h"
 
-char *GResourceProgramName, *GResourceFullProgram, *GResourceProgramDir;
+char *GResourceProgramName;
 char *usercharset_names;
 /* int local_encoding = e_iso8859_1;*/
 
@@ -167,11 +167,6 @@ return;
 	GResourceProgramName = copy("gdraw");
     else
 return;
-
-    FindProgDir(prog);
-    free(GResourceFullProgram);
-    GResourceFullProgram = copy(
-	    GFileBuildName(GResourceProgramDir,GResourceProgramName,filename,sizeof(filename)));
 }
 
 void GResourceAddResourceString(char *string,char *prog) {
--- a/cmake/FontForgeConfigure.cmake
+++ b/cmake/FontForgeConfigure.cmake
@@ -105,7 +105,7 @@ function(fontforge_generate_config template destination)
     set(_LIBUNINAMESLIST_FUN ${Libuninameslist_FEATURE_LEVEL})
   endif()
 
-  configure_file(${template} ${destination})
+  configure_file(${template} ${destination} ESCAPE_QUOTES @ONLY)
 endfunction()
 
 function(_get_git_version)
@@ -179,5 +179,5 @@ function(fontforge_generate_version_extras template destination)
     _get_modtime()
     _get_modtime_str(${FONTFORGE_MODTIME})
 
-    configure_file(${template} ${destination})
+    configure_file(${template} ${destination} ESCAPE_QUOTES @ONLY)
 endfunction()
--- a/fontforge/start.c
+++ b/fontforge/start.c
@@ -95,9 +95,7 @@ void doinitFontForgeMain(void) {
 
     if ( inited )
 return;
-#ifdef __MINGW32__
     FindProgDir(NULL);
-#endif
     InitSimpleStuff();
     if ( default_encoding==NULL )
 	default_encoding=FindOrMakeEncoding("ISO8859-1");
--- a/fontforgeexe/encodingui.c
+++ b/fontforgeexe/encodingui.c
@@ -348,18 +348,6 @@ return;
     closedir(d);
 }
 
-static void FindMapsInNoLibsDir(struct block *block,char *dir) {
-
-    if ( dir==NULL || strstr(dir,"/.libs")==NULL )
-return;
-
-    dir = copy(dir);
-    *strstr(dir,"/.libs") = '\0';
-
-    FindMapsInDir(block,dir);
-    free(dir);
-}
-
 struct cidmap *AskUserForCIDMap(void) {
     struct block block;
     struct cidmap *map = NULL;
@@ -376,8 +364,6 @@ struct cidmap *AskUserForCIDMap(void) {
 	AddToBlock(&block,buffer,NULL);
     }
     FindMapsInDir(&block,".");
-    FindMapsInDir(&block,GResourceProgramDir);
-    FindMapsInNoLibsDir(&block,GResourceProgramDir);
     FindMapsInDir(&block,getFontForgeShareDir());
     FindMapsInDir(&block,"/usr/share/fontforge");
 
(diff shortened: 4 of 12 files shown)


gdraw/gresource.c
```diff
@@ -168,12 +168,7 @@ return;
     else
 return;
 
-    free(GResourceProgramDir);
-    GResourceProgramDir = _GFile_find_program_dir(prog);
-    if ( GResourceProgramDir==NULL ) {
-	GFileGetAbsoluteName(".",filename,sizeof(filename));
-	GResourceProgramDir = copy(filename);
-    }
+    FindProgDir(prog);
```
skef: Above in this file is the line:
```
char *GResourceProgramName, *GResourceFullProgram, *GResourceProgramDir;
```
Then in fsys.c is the line:
```
char *GResourceProgramDir = NULL;
```
Which is "new" but only because it was moved from a different place in the file and the old 0 replaced with NULL.

So unless I'm missing something it appears that FontForge has two non-static `GResourceProgramDir` globals.

That already puts me a bit out of my depth as to how things work, but this change makes it even more confusing, as it appears that the manipulation of this file's `GResourceProgramDir` is being replaced with the manipulation of the other file's `GResourceProgramDir`, which is then somehow being used constructively below.

Or maybe there is only one because the two declarations get implicitly merged? 

Anyway, even if all this does work wouldn't it be preferable to replace the instances of `GResourceProgramDir` in this file with calls to `getGResourceProgramDir()`?

jtanx: Good spot on the duplication, I missed that.

I've opted to just remove all usages of it.

PR #4055, opened 2019-12-11
Fixes #4052: Contour end did not close

### LINKED ISSUE #4052: Warning: Contour end did not close
Arch Linux, 3890b4762567b84cd57b6d03fb382ad7fdb3ad50

cc: @skef 

I would rate this bug as **high severity** as there seems to be no obvious way to fix it.

# File

[b.sfd](https://github.com/fontforge/fontforge/files/3948549/b.zip)

# Settings

* Circular, 45-45-0
* Setting &laquo;Remove Overlap&raquo; to &laquo;None&raquo; does not fix the problem.

# Screenshot

2019-12-11-130929_2560x2101_scrot
2019-12-11-130955_1066x343_scrot

skef:
Problem was using two different standards for linearness for what amounts to the same check. This switches to `SplineIsLinearish()` in both cases. (The problem I'm trying to avoid with Linearish is having a spline we're treating as curved but that doesn't have a clear turning direction at one end. That may not be possible, but it's easier to just treat close-to-straight source splines as straight.)

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -1741,7 +1741,7 @@ static SplineSet *OffsetSplineSet(SplineSet *ss, StrokeContext *c) {
 	    CalcNibOffset(c, ut_ini, false, &no, -1);
 	    HandleJoin(c, left, ss->first->me, &no, is_ccw_ini, ut_endlast,
 	               was_ccw, false);
-            left = SplineSetJoin(left, true, INTERSPLINE_MARGIN, &closed);
+            left = SplineSetJoin(left, true, FIXUP_MARGIN, &closed);
 	    if ( !closed )
 		LogError( _("Warning: Left contour did not close\n") );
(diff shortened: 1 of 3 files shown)


frank-trampe:
What are the processes in which a not-quite-straight spline would be coerced into a straight spline? Other than that, this seems safe to me if it tests correctly.

skef:
There's no coersion, exactly. The test is on source splines being stroked, with the result is that the left and right "sides" in the output contour will be slightly straighter than the input contour.

frank-trampe:
Ah, okay. Let's merge.

skef:
I guess, for the record, I should expand on that answer a bit.

SplineUTanVecAt is called in a number of places. It turns out that most of those are skipped when `linear` is true in the main loop. It's also called with nib calculations, but NibIsValid already prohibits very-close-to-but-not-quite-linear splines by other means. So the overall "case" is the same.

I'll merge now.

PR #4057, opened 2019-12-11
Fix Spiro hndling in SplinePointListCheckSelected1

### LINKED ISSUE #4051: ⟦Spiro⟧ First point on contour comes unselected
1. Draw a shape with _n_ Spiro points where _n_ &gt; 1
2. Select all
3. Cut
4. Paste
5. Only _n_-1 points are selected (always the first point comes unselected); if the contour is closed, it comes open.

I think this is a regression.

ctrlcctrlv:
This closes #4051. @jtanx threw away a distinction that existed in the
code before; sometimes spiros[spiro_cnt] was iterated through, sometimes
only spiros[spiro_cnt-1]. This restores that distinction, which turns
out to be necessary for proper Spiro handling.

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/splineutil.c
+++ b/fontforge/splineutil.c
@@ -1451,13 +1451,13 @@ static SplinePointList *SplinePointListCopySpiroSelected1(SplinePointList *spl) 
 return( head );
 }
 
-bool SplinePointListCheckSelected1(const SplinePointList *base, bool spiro, bool *allsel, int spiro_cnt_end) {
+bool SplinePointListCheckSelected1(const SplinePointList *base, bool spiro, bool *allsel, bool skip_spiro_end) {
     bool anysel = false;
     if (allsel) {
         *allsel = true;
     }
     if (spiro) {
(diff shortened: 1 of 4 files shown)


fontforge/splineutil.c
```diff
@@ -1451,13 +1451,13 @@ static SplinePointList *SplinePointListCopySpiroSelected1(SplinePointList *spl)
 return( head );
 }
 
-bool SplinePointListCheckSelected1(const SplinePointList *base, bool spiro, bool *allsel) {
+bool SplinePointListCheckSelected1(const SplinePointList *base, bool spiro, bool *allsel, int spiro_cnt_end) {
```
skef: As is this is too confusing. My preference is to have this last parameter being a `bool` maybe named `skip_spiro_end` to remove the unused generality. If that isn't acceptable I'm OK with a `spiro_skip_cnt` that must be >= 0 and is subtracted rather than added. 

Otherwise these changes are fine with me.

PR #4060, opened 2019-12-12
Replace SplinePointListIsClockwise() with polygon-angle heuristic

### LINKED ISSUE #4049: How do I correctly stroke an ‹f›, without getting a hole in the middle?
I've so far found no fault in the new Expand Stroke implementation, but perhaps have found a case worth scrutinizing further.

My TT2020 font is based primarily on [IBM Selectric Pica](http://typeballs.com); I want to make an italic with a reduced character set, likely to only be Latin and maybe Cyrillic.

I got started tracing as I always do&hellip;

2019-12-09-164520_2560x1461_scrot

&hellip;but have found a case I can't seem to beat: letter &lsaquo;f&rsaquo;.

Seems no matter which way I stroke it, or however many times I &laquo;Element&rarr;Reverse Direction&raquo;, I always get a hole in the middle.

Do you have any idea what I'm doing wrong @skef? Here's the file for reference. Use radius 10.

[f.zip](https://github.com/fontforge/fontforge/files/3938632/f.zip)

### LINKED ISSUE #4050: Expand Stroke Remove Overlap removing countour it shouldn't
# Instructions

Run &laquo;Expand Stroke&raquo; with a radius of 10 on.

# File

[ij.sfd](https://github.com/fontforge/fontforge/files/3939279/ij.sfd.zip)

# Results

## Remove Overlap &rarr; None (Debug) :heavy_check_mark: 
2019-12-09-185454_2560x1461_scrot

## Remove Overlap &rarr; By Layer or By Contour :x: 
2019-12-09-185554_2560x1461_scrot

skef:
The existing `SplinePointListIsClockwise()` sometimes returns -1 for non-intersecting contours. See #4050 and #4049 for examples. This causes problem for any code, like the new Expand Stroke code, that relies on the function in a conditional.

The python note for the wrapper around the function is "Determine if a contour is oriented in a clockwise direction. If the contour intersects itself the results are indeterminate." This implies that the function is intended to reflect the mathematical concept of [curve orientation](https://en.wikipedia.org/wiki/Curve_orientation), which is well defined only for "simple" (non-self-intersecting) closed curves. 

If a contour does not intersect itself its orientation will be the same as the orientation of the polygon formed by connecting each of its on-curve points in sequence. While the orientation of a simple *convex* polygon can be determined by examining a few sequential points, that heuristic won't always be accurate for non-convex polygons. For those, an easy way to determine the orientation is to add up the angle changes, which will be 2π for a clockwise curve (when adding a certain way) and -2π for a concave curve. 

Because the old code *sometimes* returned 0 or 1 for self-intersecting curves this code just sets the return value based on angle sum and only returns -1 for a sum of zero. This is consistent with `test1004.py`. 

While I think this assessment is accurate this change needs a close look and, with luck, some thorough testing. 

Closes #4050 
Closes #4049

### DIFF
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -1215,9 +1215,11 @@ Two contours may be compared to see if they describe similar paths.
     <TR>
       <TD><CODE><A name="#cnt-isClockwise">isClockwise</A></CODE></TD>
       <TD><CODE>()</CODE></TD>
-      <TD>Returns whether the contour is drawn in a clockwise direction. A return
-	value of -1 indicates that no consistant direction could be found (the contour
-	self-intersects).</TD>
+      <TD>Returns 1 when a non-self-intersecting contour is drawn in 
+	  a clockwise direction and 0 when such a contour is drawn in a
+	  counter-clockwise direction. When a contour intersects itself the
+	  return value may be any of 1, 0, or -1 and has no particular
+	  interpretation. 
     </TR>
     <TR>
       <TD><CODE>reverseDirection</CODE></TD>
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -55,7 +55,6 @@
 #define COS_MARGIN (1e-5)
 #define MIN_ACCURACY (1e-5)
 
-#define NORMANGLE(a) ((a)>FF_PI?(a)-2*FF_PI:(a)<-FF_PI?(a)+2*FF_PI:(a))
 #define BPNEAR(bp1, bp2) BPWITHIN(bp1, bp2, INTRASPLINE_MARGIN)
 #define BP_DIST(bp1, bp2) sqrt(pow((bp1).x-(bp2).x,2)+pow((bp1).y-(bp2).y,2))
 
@@ -365,8 +364,6 @@ enum ShapeType NibIsValid(SplineSet *ss) {
 	return Shape_TooFewPoints;
     if ( ss->first->prev==NULL )
 	return Shape_NotClosed;
-    if ( !SplinePointListIsClockwise(ss) )
-	return Shape_CCW;
     if ( ss->first->next->order2 )
 	return Shape_Quadratic;
 
@@ -401,8 +398,11 @@ enum ShapeType NibIsValid(SplineSet *ss) {
 	    break;
 	++n;
     }
+
     if ( n<3 )
 	return Shape_TooFewPoints;
+    if ( anglesum<0 )
+	return Shape_CCW;
     if ( !RealWithin(anglesum, 2*FF_PI, 1e-1) )
 	return Shape_SelfIntersects;
 
@@ -1729,7 +1729,7 @@ static SplineSet *OffsetSplineSet(SplineSet *ss, StrokeContext *c) {
 		if ( c->rmov==srmov_contour )
 		    left = SplineSetRemoveOverlap(NULL,left,over_remove);
 		// Open paths don't always result in clockwise output
-		if ( !SplinePointListIsClockwise(left) )
+		if ( SplinePointListIsClockwise(left)==0 )
 		    SplineSetReverse(left);
 	    }
 	}
@@ -1898,7 +1898,7 @@ static SplineSet *SplineSet_Stroke(SplineSet *ss,struct strokecontext *c,
     int copied = 0;
 
     c->contour_was_ccw = (   base->first->prev!=NULL
-                          && !SplinePointListIsClockwise(base) );
+                          && SplinePointListIsClockwise(base)==0 );
 
     if ( base->first->next==NULL )
 	ret = SinglePointStroke(base->first, c);
--- a/fontforge/splineutil.h
+++ b/fontforge/splineutil.h
@@ -222,5 +222,6 @@ extern int HashKerningClassNamesFlex(SplineFont *sf, struct glif_name_index * cl
 #define SPLINE1DPTANVAL(s, t) ((3*(s)->a*(t) + 2*(s)->b)*(t) + (s)->c)
 #define SPLINEPTANVAL(s, t) (BasePoint) { SPLINE1DPTANVAL(&(s)->splines[0], t), SPLINE1DPTANVAL(&(s)->splines[1], t) }
 #define BPWITHIN(bp1, bp2, f) (RealWithin((bp1).x, (bp2).x, f) && RealWithin((bp1).y, (bp2).y, f))
+#define NORMANGLE(a) ((a)>FF_PI?(a)-2*FF_PI:(a)<-FF_PI?(a)+2*FF_PI:(a))
 
 #endif /* FONTFORGE_SPLINEUTIL_H */
--- a/fontforge/splineutil2.c
+++ b/fontforge/splineutil2.c
@@ -4041,95 +4041,35 @@ return( NULL );
 return( ret );
 }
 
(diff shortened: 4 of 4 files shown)


ctrlcctrlv:
You ask for thorough testing - what do you think this might break?

frank-trampe:
Aside from direct regressions, which I suppose you've thought through a bit, did you think of any possible indirect regressions from finding/coercing a meaningful result in more cases? I did not see any in the code that calls `SplinePointListIsClockwise`, but I'm wary about moving from something that looks at the control points to something that ignores them.

And, with regard to test cases, these are the things that I would think to vary. Can you think of any others?
* Closure (open, closed).
* Overall direction (clockwise, counter-clockwise).
* On-curve style (straight, smooth, sharply cusped).
* Crossings (none, single, double).
* Aberrations (none, super-stray on-curve point, super-stray control point).

frank-trampe:
@jtanx, I can't figure out what's going wrong in the Travis build. Could you take a look?

ctrlcctrlv:
The Travis issue is probably related to https://github.com/ninja-build/ninja/issues/1599

ctrlcctrlv:
I confirm this branch solves both #4049 and #4050 for me.

I confirm it works on both `TT.sfd` and `TT_ita.sfd`, see below `TT_ita.sfd` (in progress, maybe a little ugly):
2019-12-12-135010_2372x1249_scrot

jtanx:
> @jtanx, I can't figure out what's going wrong in the Travis build. Could you take a look?

It's just an intermittent failure, probably something related to timing and running on a VM. The mac builds are always the flakiest. I restarted the job in Travis - note that any one of us can do that.

frank-trampe:
Ah, good. I've restarted jobs before when they had fetching failures, but this one was retrying within the job and failing repeatedly without an obvious cause, so I didn't think to restart the job this time.

skef:
Oh, weird, I wrote up a response right after this was filed but apparently it never went though. 

My thinking about this function breaks the cases into three categories:

1. Open SplineSets 
2. Closed, non-self-intersecting SplineSets
3. Closed, self-intersecting SplineSets

I can't recall seeing the relevant menu items enabled for an open contour, and I tested a number of them before writing this code and didn't see them enabled. That's a little frustrating because there's no up-front -1 return for open contours in the existing code, but it seems that they're mostly set aside anyway. So I conclude this function shouldn't, or doesn't need to, give answers for open SplineSets. 

I'm most confident about category 2. If a SplineSet is closed and doesn't intersect itself it's orientation will be the same as the polygon formed by its on-curve points. This just follows from the definition of orientation: whether the interior is to the left or the right when following the curve. Whether a given spline "bulges" in this or that direction can't change that criterion without an intersection. The only remaining question is whether the function is intended to track curve orientation or something else, and I don't see anything else that's plausible. 

That leaves category 3 as the problem, and I'm just not sure what to say about it. We could put effort into trying to determine the intent of the current algorithm, but we know that algorithm is wrong for some non-intersecting cases. Suppose we arrive at a confident understanding of that intent, and could improve on the answer it gives for non-intersecting cases. Does that mean that answer matters?

Maybe knowing the intent would make it clear *how* it matters. That's normally what I think in these cases: You should only feel comfortable about replacing code when you completely understand the current implementation. However, in this case I feel like we have a different resource to fall back on, which is spline-based font software in general. Either there's a reason to care about category 3 answers or there isn't one. 

I can make guesses: You could answer based on whether more area is enclosed on the left or the right of the curve. Or by the least number of splines that would have to be adjusted to make it non-intersecting and have that orientation. But I don't know when any of these answers would matter. 

There is at least one case self-intersecting case that matters to my code. If a contour is "extra" clockwise or counter-clockwise, in that it self-intersects in virtue of the angles double, triple, or *n*-le bending in that direction, my code needs to be able to identify such splines as bending in that direction. (This is so that it can make a self-intersecting open spline enclose the intended space once it is run through Remove Overlap.) The heuristic in this PR manages to do that in at least the common cases.

Maybe there are weird failure conditions I haven't thought of, due to unusual source spline geometries, and understanding the current implementation would help reveal what they are. If someone could help get me part of the way to that understanding, that would be helpful. Otherwise I'm not sure the investment is worth it in this case, and we can just fall back on testing and waiting for an exception case to arise.

skef:
Hmm, maybe "my" need suggests one possible explanation for what the current code may be attempting, which is to answer -1 for any closed spline that encloses both zero and non-zero areas. That's something we could try to test for, at least. 

(I'm having a weird feeling of deja vu about this comment. Was there a similar discussion about some algorithm a year ago?)

skef:
Wait, no -- `tests/fonts/DirectionTest.sfd` letter K argues against that interpretation, I think. The existing code answers -1 for it but Remove Overlap leaves both sides of the curve intact.

skef:
Folks, how are we going to make this decision? Do I need to build a set of cases to probe when the answers are different? Should we inspect the calls to the function to evaluate whether and when we care about the answer for a non-self-intersecting contour?

jtanx:
This should be pretty easy to compare against the current implementation, right? Just chuck enough at it and see what's different, and explain away (or fix) the differences.

e.g. I get the following running on dejavu sans
image
image

[clockwise.zip](https://github.com/fontforge/fontforge/files/3993292/clockwise.zip)

skef:
> This should be pretty easy to compare against the current implementation, right? 

Yes, with the caveat that "real fonts" won't typically have any self-intersections, so we pretty much already know what the answer should be in those cases. (i.e. those -1s from the old algorithm are the #4050-style errors.)

This is not to say that that test method is bad. It's verifying through many cases that my algorithm gives the same substantive answer as the old algorithm in every case, and we can presumably quickly check that it's giving the correct answer when the old algorithm is (via -1) failing to. That's a quick and effective way of addressing @frank-trampe's primary worry ("I'm wary about moving from something that looks at the control points to something that ignores them.")

jtanx:
> i.e. those -1s from the old algorithm are the #4050-style errors

Hm but those -1s are running under your change.

skef:
Oh really? That's strange. Maybe there are some self-intersections ... I'll check.

skef:
Nope, no self-intersections. I'll try to figure out what's going on with uni2623.

skef:
Oh well dammit, my reasoning was wrong after all. That's why we have these discussions, of course. 

The problem: It's true that if you take a spline set with on-curve points arranged in a simple polygon, and adjust the control points so that the curve does not self-intersect, then the positions of those control points cannot change the orientation of the polygon. However, you can take a spline set with on-curve points arranged in a *non-simple* polygon and then arrange the control points so that the curve does *not* self-intersect.

I'll probably close this PR shortly. 

Unfortunately the issue(s) this was attempting to fix may be a project-level problem. I wrote the Expand Stroke replacement to depend on various existing pieces of FontForge functionality, including Simplify, Remove Overlap, Add Extrema, and so forth. The testing process revealed a number of bugs in many functions. I've fixed most of those and other project members have fixed some. I'm not sure I want to fix this one. I've never given `autotrace.c` more than a glance and am not familiar with its algorithms. It seems likely to require a large time investment if starting from scratch, and I just spent quite a bit of time on that Cubic->Quadratic conversion bug. 

I may continue researching ways of avoiding the need to dig into that code. I suppose one obvious candidate is an "established" winding number algorithm. (It seems from the variable names that this is how the existing `SplinePointListIsClockwise()` works. But in that case what is the reference point and how/where is it chosen?)

PR #4064, opened 2019-12-14
Correct bad linear control point assignment

### LINKED ISSUE #4059: Stroking a certain ‹S› results in degenerate spline
Arch Linux, 82579d3b0c329288e522fde35d7d8c82ee786cc3

Another Spiro-produced spline causing problems.

For some reason it's most apparent in the MetricsView:
2019-12-12-114319_3554x1275_scrot

However once you know where to look, it becomes obvious also in the CharView:
2019-12-12-114353_2560x2101_scrot

# Required settings
* Major 45, Minor 45
* Angle **must** be 45
* **None (Debug)** must be selected
* Line Cap can be either Round or Nib

# File
[S.sfd](https://github.com/fontforge/fontforge/files/3953724/S.zip)

skef:
February 13, 2003. The mind boggles.

That this is a linear segment still looks wrong to me but this code is too complex to dig through for that (at least for the time being). 

Closes #4059

### DIFF
--- a/fontforge/splineorder2.c
+++ b/fontforge/splineorder2.c
@@ -247,8 +247,10 @@ static SplinePoint *LinearSpline(Spline *ps,SplinePoint *start, real tmax) {
     }
     end->ttfindex = 0xfffe;
     end->nextcpindex = 0xfffe;
-    end->me.x = end->nextcp.x = end->prevcp.x = x;
-    end->me.y = end->nextcp.y = end->prevcp.y = y;
+    start->nextcp.x = start->me.x;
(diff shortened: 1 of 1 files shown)


skef:
@ctrlcctrlv had to force push with a small update

skef:
> That this is a linear segment still looks wrong to me but this code is too complex to dig through for that (at least for the time being).

On second thought, in context the segment looks fine. It's presumably just an accident of the default accuracy setting for the conversion. I sometimes forget how "zoomed in" I am looking at these bugs.

PR #4065, opened 2019-12-14
Always allow opening «Glyph Info…» in FontView

ctrlcctrlv:
This is another one of those issues “I’d gotten so used to I forgot they were actually bugs.”

In master it is not possible to open this dialog for glyphs which don't exist…via the keyboard, or Element menu. But you can just right-click the glyph, and receive the Glyph Info option, and therefore, the dialog.

Which I do…often. The restriction seems to have no purpose, and is very annoying, so I've removed it.

### Type of change

- **Non-breaking change**

### DIFF
--- a/fontforgeexe/fontview.c
+++ b/fontforgeexe/fontview.c
@@ -4218,9 +4218,7 @@ static void ellistcheck(GWindow gw, struct gmenuitem *mi, GEvent *UNUSED(e)) {
 	    mi->ti.disabled = in_modal;
 	  break;
 	  case MID_CharInfo:
-	    mi->ti.disabled = anychars<0 || (gid = fv->b.map->map[anychars])==-1 ||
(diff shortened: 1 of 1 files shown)


fontforgeexe/fontview.c
```diff
@@ -4217,11 +4217,6 @@ static void ellistcheck(GWindow gw, struct gmenuitem *mi, GEvent *UNUSED(e)) {
 	  case MID_FontInfo:
 	    mi->ti.disabled = in_modal;
 	  break;
```
skef: I would be more comfortable with
```
    case MID_CharInfo:
        mi->ti.disabled = anychars<0 || in_modal;
    break;
```

skef:
Can you verify that the menu item is active only when exactly one slot is selected?

ctrlcctrlv:
Yup.

PR #4068, opened 2019-12-17
update apt-get install script

wontruefree:
libpng12-dev was not available on my fresh ubuntu install but libpng-dev was

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,7 +6,7 @@ The exact method to do this depends on your OS distribution.
 To download all dependencies on Ubuntu, run:
 
 ```sh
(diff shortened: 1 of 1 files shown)


frank-trampe:
The patch replaces libxml-dev with libxm-dev, which doesn't exist. And, even if it did, of course, the Motif widget kit is not interchangeable with an XML parser.

wontruefree:
I tried to use the browser editor.  That is a weird diff.

wontruefree:
closed in favor of https://github.com/fontforge/fontforge/pull/4069
Dont know what happened.  Seems like an editor bug.

frank-trampe:
Well that's odd.

PR #4069, opened 2019-12-17
update apt-get install script

wontruefree:
libpng12-dev was not available on my fresh ubuntu install but libpng-dev was


frank-trampe:
This is fine with me. The unversioned dev package name seems to be universally available.

PR #4070, opened 2019-12-18
Fix MetricsView when selecting unencoded glyphs

ctrlcctrlv:
When selecting a set of unencoded glyphs, e.g.:

2019-12-18-110256_1502x839_scrot

And opening the Metrics View:

2019-12-18-110420_2396x853_scrot

The glyphs are all in a row. However if you go to edit the string (simply clicking the input is enough):

2019-12-18-110525_2396x853_scrot

Spaces appear. This stops that from happening.

Example font:

[MetricsViewSpaces.zip](https://github.com/fontforge/fontforge/files/3976420/MetricsViewSpaces.zip)


ctrlcctrlv:
<br/>

Issue #677, opened 2013-08-24
Color font COLR/CPAL and sbix editing/creation support


ultrasquid:
I suppose this should go in the "maybe someday" category as we have a lot of other things to worry about.

Apple and Microsoft have both released separate formats for fonts with multi-color vector glyphs, COLR/CPAL and sbix, respectively. Fontlab has just announced a soon-to-be-released free (as in beer) app that would make these formats cross-platform usable and backwards functional to Widows XP and Mac OSX 10.6.

So, should FontForge step proudly into the world of color and support creating color fonts, and if so, how and when?

KrasnayaPloshchad:
See also: #2866

TsundereBug:
![image](https://user-images.githubusercontent.com/7751296/38460208-1901ce1a-3a6a-11e8-994c-9fd868391228.png)

Is there any tool I _can_ use to check color tables? (The bottom message is just because the font file is invalid for now)

KrasnayaPloshchad:
CC @babelstone

frank-trampe:
We don't support color yet. I'd check FontBakery for now.

davelab6:
I think Frank meant TruFont :)

typiconman:
Still no support for COLR/CPAL tables in FontForge :(
However, one can write the COLR/CPAL table data as XML and add it into the font at compile time using [TTX](https://github.com/fonttools/fonttools).

ctrlcctrlv:
@frank-trampe Can you please reopen this? This is a valid issue someone might want to work on.

ctrlcctrlv:
Reopened. 

The format war is pretty much over. Four years ago, on this very issue tracker, I [surveyed the landscape](https://github.com/fontforge/fontforge/issues/2044#issuecomment-119471564). I wrote at that time "[in] my opinion these color fonts are a quite stupid gimmick," I take that back.

COLR/CPAL has won the day. :tada: Browser support has coalesced around that, as has FreeType and MS Windows. In the wild, color fonts either use that or Adobe's `SVG ` standard. Apple, backwards as always, has their own garbage; it's not even a vector format and is comically bad, basically like our support for METAFONT bitmap output, except without the excuse METAFONT has of being ancient. That can be dealt with in another issue.

I have done a preliminary investigation on how COLR/CPAL fonts work. If I add any color font format support to FontForge, it will be this one. 

[...]

PR #4074, opened 2019-12-20
Extend glyph.clear() in Python to clear layers

### LINKED ISSUE #4041: It's not possible to clear a layer from Python
An explanation of this issue was made [in 2010](https://sourceforge.net/p/fontforge/mailman/message/25771453/) by @alatius. 

I find no reason to retype the problem, his explanation is sufficient, relevant, and every word of it in my experience remains true nine years later. This relates to the lack of an Images Python API, which @skef and I have discussed.

As @alatius noted, it seems very wrong that this code does not work:

```python
font[char].layers[0] = fontforge.layer()
```

He writes:

> Then I didn't suspect that this would work either:
>```python
>for char in font:
>   del font[char].layers[0]
>```
> In fact I get a segmentation fault. 

I likewise receive a segfault.

```
0x0000155554f9de34 in PyFF_Glyph_CSetLayer (self=0x15554fbd8070, value=0x0, layeri=0, flags=258)
    at ../fontforge/python.c:5901
5901        if ( PyType_IsSubtype(&PyFF_LayerType, Py_TYPE(value)) ) {
(gdb) bt
#0  0x0000155554f9de34 in PyFF_Glyph_CSetLayer (self=0x15554fbd8070, value=0x0, layeri=0, flags=258)
    at ../fontforge/python.c:5901
#1  0x0000155554f9e4d8 in PyFF_LayerArrayIndexAssign (self=0x15554fb86e90, index=0x15555456c780, value=0x0)
    at ../fontforge/python.c:6122
#2  0x000015555438cb6f in _PyEval_EvalFrameDefault () at /usr/lib/libpython3.8.so.1.0
#3  0x000015555435a1ea in _PyEval_EvalCodeWithName () at /usr/lib/libpython3.8.so.1.0
#4  0x000015555435ae0a in PyEval_EvalCodeEx () at /usr/lib/libpython3.8.so.1.0
#5  0x00001555543e6a9c in PyEval_EvalCode () at /usr/lib/libpython3.8.so.1.0
#6  0x00001555543e6b4a in  () at /usr/lib/libpython3.8.so.1.0
#7  0x0000155554422655 in  () at /usr/lib/libpython3.8.so.1.0
#8  0x0000155554422700 in PyRun_StringFlags () at /usr/lib/libpython3.8.so.1.0
#9  0x0000155554423fe0 in PyRun_SimpleStringFlags () at /usr/lib/libpython3.8.so.1.0
#10 0x0000155554fc5bbf in PyFF_ScriptString
    (fv=0x555555cd2690, sc=0x555555de57c0, layer=1, str=0x555555ceba50 "import fontforge\nf=fontforge.fonts()[0]\n\ndel f[ord(\"5\")].layers[0]\n") at ../fontforge/python.c:19456
#11 0x00005555557d5cb5 in ExecPython (g=0x555555fc3c70, e=0x7fffffffbcd0)
    at ../fontforgeexe/scriptingdlg.c:134
#12 0x00005555557d5df0 in SD_OK (g=0x555555fc3c70, e=0x7fffffffbcd0) at ../fontforgeexe/scriptingdlg.c:160
#13 0x00005555558426b3 in GButtonInvoked (b=0x555555fc3c70, ev=0x7fffffffbff0) at ../gdraw/gbuttons.c:257
#14 0x000055555584362a in gbutton_mouse (g=0x555555fc3c70, event=0x7fffffffbff0) at ../gdraw/gbuttons.c:485
#15 0x000055555584dce8 in _GWidget_Container_eh (gw=0x555555fb7a80, event=0x7fffffffbff0)
    at ../gdraw/gcontainer.c:304
#16 0x000055555584fb68 in _GWidget_TopLevel_eh (gw=0x555555fb7a80, event=0x7fffffffbff0)
    at ../gdraw/gcontainer.c:747
#17 0x0000555555862888 in _GGDKDraw_CallEHChecked (gw=0x555555fb7a80, event=0x7fffffffbff0, eh=
    0x55555584f1da <_GWidget_TopLevel_eh>) at ../gdraw/ggdkdraw.c:334
#18 0x00005555558652ab in _GGDKDraw_DispatchEvent (event=0x555555fc81e0, data=0x555555be7800)
    at ../gdraw/ggdkdraw.c:1237
#19 0x0000155554195954 in  () at /usr/lib/libgdk-3.so.0
#20 0x0000155554142904 in  () at /usr/lib/libgdk-3.so.0
#21 0x0000155553d1639e in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#22 0x0000155553d181b1 in  () at /usr/lib/libglib-2.0.so.0
#23 0x0000155553d181f1 in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#24 0x0000555555867fd9 in GGDKDrawProcessOneEvent (gdisp=0x555555be7800) at ../gdraw/ggdkdraw.c:2193
#25 0x0000555555853a00 in GDrawProcessOneEvent (gdisp=0x555555be7800) at ../gdraw/gdraw.c:764
#26 0x00005555557d74d9 in ScriptDlg (fv=0x555555cd2690, cv=0x15554ed42010)
    at ../fontforgeexe/scriptingdlg.c:364
#27 0x0000555555609324 in CVMenuExecute (gw=0x555555e59010, UNUSED_mi=0x555555e5b240, UNUSED_e=0x7fffffffcf80)
    at ../fontforgeexe/charview.c:6823
#28 0x0000555555880fed in gmenu_mouse (m=0x555555dbf550, event=0x7fffffffcf80) at ../gdraw/gmenu.c:948
#29 0x0000555555882ab0 in gmenu_eh (w=0x555555faaf10, ge=0x7fffffffcf80) at ../gdraw/gmenu.c:1389
#30 0x0000555555862888 in _GGDKDraw_CallEHChecked (gw=0x555555faaf10, event=0x7fffffffcf80, eh=
    0x5555558829d7 <gmenu_eh>) at ../gdraw/ggdkdraw.c:334
#31 0x00005555558652ab in _GGDKDraw_DispatchEvent (event=0x555555fc8320, data=0x555555be7800)
    at ../gdraw/ggdkdraw.c:1237
#32 0x0000155554195954 in  () at /usr/lib/libgdk-3.so.0
#33 0x0000155554142904 in  () at /usr/lib/libgdk-3.so.0
#34 0x0000155553d1639e in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#35 0x0000155553d181b1 in  () at /usr/lib/libglib-2.0.so.0
#36 0x0000155553d181f1 in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#37 0x0000555555868055 in GGDKDrawEventLoop (gdisp=0x555555be7800) at ../gdraw/ggdkdraw.c:2204
#38 0x0000555555853add in GDrawEventLoop (gdisp=0x555555be7800) at ../gdraw/gdraw.c:784
#39 0x0000555555817d70 in fontforge_main (argc=2, argv=0x7fffffffe3b8) at ../fontforgeexe/startui.c:1415
#40 0x00005555555aab89 in main (argc=2, argv=0x7fffffffe3b8) at ../fontforgeexe/main.c:33
```

An interim solution, which I wrote just because I'm in a bit of a self-inflicted time crunch:

```diff
diff --git a/fontforge/python.c b/fontforge/python.c
index 26386574d..0a49b463f 100644
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -8058,6 +8058,12 @@ Py_RETURN( self );
 
 static const char *appendaccent_keywords[] = { "name", "unicode", "pos", NULL };
 
+static PyObject *PyFFGlyph_clearBackground(PyObject *self, PyObject *UNUSED_args, PyObject *UNUSED_keywds) {
+    SplineChar *sc = ((PyFF_Glyph *) self)->sc;
+    SCClearBackground(sc);
+    Py_RETURN(self);
+}
+
 static PyObject *PyFFGlyph_appendAccent(PyObject *self, PyObject *args, PyObject *keywds) {
     SplineChar *sc = ((PyFF_Glyph *) self)->sc;
     int layer = ((PyFF_Glyph *) self)->layer;
@@ -9454,6 +9460,7 @@ static PyMethodDef PyFF_Glyph_methods[] = {
     { "changeWeight", (PyCFunction) PyFFGlyph_changeWeight, METH_VARARGS, "Change the weight (thickness) of the stems of the glyph"},
     { "condenseExtend", (PyCFunction) PyFFGlyph_condenseExtend, METH_VARARGS, "Change the widths of the counters and side bearings of the glyph"},
     { "clear", (PyCFunction) PyFFGlyph_clear, METH_NOARGS, "Clears the contents of a glyph and makes it not worth outputting" },
+    { "clearBackground", (PyCFunction) PyFFGlyph_clearBackground, METH_NOARGS, "Clear the contents of a glyph's background" },
     { "cluster", (PyCFunction) PyFFGlyph_Cluster, METH_VARARGS, "Cluster the points of a glyph towards common values" },
     { "correctDirection", (PyCFunction) PyFFGlyph_Correct, METH_NOARGS, "Orient a layer so that external contours are clockwise and internal counter clockwise." },
     { "exclude", (PyCFunction) PyFFGlyph_Exclude, METH_VARARGS, "Exclude the area of the argument (a layer) from the current glyph"},
```

It would be better to be able to do:

```python
# either
glyph.clearLayer(index)
# or
del glyph.layers[index]
```

But I couldn't figure out how to do that in C quickly enough to make a formal commit, *yet*. So this is a bit of a placeholder issue&hellip;

ctrlcctrlv:
At least since 2010, people have noted that it's not possible to clear a single layer from Python.

This fixes that by extending `clear()`, the function of `glyph`, to take an optional integer argument, which, if provided, is a positive layer index.

I use this function to build my TT2020 font so kind of need it. Without it, background images would take up all available memory and FontForge would crash. There are messy hacks I could do to avoid the problem but this is the best solution.

Closes #4041

### Type of change

- **New feature** (sort of)

### DIFF
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -2518,9 +2518,8 @@ must be created through the font.
           With no arguments, clears the contents of the glyph (and marks it as
           not <A href="#g-isWorthOutputting">worth outputting</A>).
 
-          If a layer is provided it must be an integer index, e.g. 0 for Back
-          and 1 for Fore; it is not possible to clear the guide layer with this
-          function.
+          It is not possible to clear the guide layer with this function.
+          <CODE>layer</CODE> may be either an integer index or a string.
       </TD>
     </TR>
     <TR>
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -2513,9 +2513,15 @@ must be created through the font.
     </TR>
     <TR>
       <TD><CODE>clear</CODE></TD>
-      <TD><CODE>()</CODE></TD>
-      <TD>Clears the contents of the glyph (and marks it as not
-	<A href="#g-isWorthOutputting">worth outputting</A>).</TD>
+      <TD><CODE>([layer])</CODE></TD>
+      <TD>
+          With no arguments, clears the contents of the glyph (and marks it as
+          not <A href="#g-isWorthOutputting">worth outputting</A>).
+
+          If a layer is provided it must be an integer index, e.g. 0 for Back
+          and 1 for Fore; it is not possible to clear the guide layer with this
+          function.
+      </TD>
     </TR>
     <TR>
       <TD><CODE><A name="g-cluster">cluster</A></CODE></TD>
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -9373,12 +9373,35 @@ static PyObject *PyFFGlyph_BoundingBox(PyFF_Glyph *self, PyObject *UNUSED(args))
 return( Py_BuildValue("(dddd)", bb.minx,bb.miny, bb.maxx,bb.maxy ));
 }
 
-static PyObject *PyFFGlyph_clear(PyFF_Glyph *self, PyObject *UNUSED(args)) {
+static PyObject *PyFFGlyph_clear(PyFF_Glyph *self, PyObject *args) {
+	int arglen = PySequence_Size(args);
+    int layeri = self->layer;
 
-    SCClearContents(self->sc,self->layer);
-    SCCharChangedUpdate(self->sc,self->layer);
+    if (arglen <= 0) {
+        SCClearContents(self->sc,layeri);
+    } else if (arglen > 1) {
+        PyErr_Format(PyExc_TypeError, "Too many arguments, only a layer index is allowed");
+        return NULL;
+    } else { /* arglen == 1 */
+        PyObject *layerp = PySequence_GetItem(args,0);
+        if (!PyInt_Check(layerp)) {
+            PyErr_Format(PyExc_ValueError, "First argument must be layer index (int)");
+            return NULL;
+        }
+        layeri = PyInt_AsLong(layerp);
+        // ly_grid not clearable with this function. In any event, it makes no
+        // sense to put it here; clearing ly_grid should quite rightly go at
+        // the font level since ly_grid is per-font not per-glyph so this is
+        // not a bug
+        if ( layeri<ly_back || layeri>=self->sc->layer_cnt ) {
+            PyErr_Format(PyExc_ValueError, "Layer is out of range");
+            return NULL;
+        }
+        SCClearLayer(self->sc, layeri);
+    }
 
-Py_RETURN(self);
+    SCCharChangedUpdate(self->sc,layeri);
+    Py_RETURN(self);
 }
 
 static PyObject *PyFFGlyph_isWorthOutputting(PyFF_Glyph *self, PyObject *UNUSED(args)) {
@@ -9453,7 +9476,7 @@ static PyMethodDef PyFF_Glyph_methods[] = {
     { "canonicalStart", (PyCFunction) PyFFGlyph_canonicalStart, METH_NOARGS, "Sets the start point of all the contours of the current glyph to be the leftmost point on the contour."},
     { "changeWeight", (PyCFunction) PyFFGlyph_changeWeight, METH_VARARGS, "Change the weight (thickness) of the stems of the glyph"},
     { "condenseExtend", (PyCFunction) PyFFGlyph_condenseExtend, METH_VARARGS, "Change the widths of the counters and side bearings of the glyph"},
-    { "clear", (PyCFunction) PyFFGlyph_clear, METH_NOARGS, "Clears the contents of a glyph and makes it not worth outputting" },
+    { "clear", (PyCFunction) PyFFGlyph_clear, METH_VARARGS, "Clears the contents of either a single layer of a glyph or of all the glyph's layers; makes it not worth outputting depending on its invocation" },
     { "cluster", (PyCFunction) PyFFGlyph_Cluster, METH_VARARGS, "Cluster the points of a glyph towards common values" },
     { "correctDirection", (PyCFunction) PyFFGlyph_Correct, METH_NOARGS, "Orient a layer so that external contours are clockwise and internal counter clockwise." },
     { "exclude", (PyCFunction) PyFFGlyph_Exclude, METH_VARARGS, "Exclude the area of the argument (a layer) from the current glyph"},
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -9384,11 +9384,22 @@ static PyObject *PyFFGlyph_clear(PyFF_Glyph *self, PyObject *args) {
         return NULL;
     } else { /* arglen == 1 */
         PyObject *layerp = PySequence_GetItem(args,0);
-        if (!PyInt_Check(layerp)) {
-            PyErr_Format(PyExc_ValueError, "First argument must be layer index (int)");
+        if ( STRING_CHECK(layerp)) {
+            char *name;
+            PYGETSTR(layerp, name, NULL);
+            layeri = SFFindLayerIndexByName(self->sc->parent,name);
+            if ( layeri<0 ) {
+                PyErr_Format(PyExc_ValueError, "Requested layer '%s' not found", name);
+                ENDPYGETSTR();
+                return NULL;
+            }
+            ENDPYGETSTR();
+        } else if (!PyInt_Check(layerp)) {
+            PyErr_Format(PyExc_ValueError, "First argument must be string or layer index");
             return NULL;
+        } else {
+            layeri = PyInt_AsLong(layerp);
         }
-        layeri = PyInt_AsLong(layerp);
         // ly_grid not clearable with this function. In any event, it makes no
         // sense to put it here; clearing ly_grid should quite rightly go at
         // the font level since ly_grid is per-font not per-glyph so this is


ctrlcctrlv:
Really don't think it was a good idea to have ever allowed people to pass in strings like "Fore" or "Back" in any function instead of 1 and 0 when `glyph.layers.index("Fore")` is available.

Layers-from-strings should be deprecated.

frank-trampe:
I think that it makes scripts a bit more readable to have the string aliases, sort of the way we use enums in the C code. I also agree that it makes a lot of sense to have a single function that converts those strings (and perhaps converts invalid ones to a large negative number to trigger an error). I suppose that we could merge this now so long as we add that function and wire it here before the next release.

ctrlcctrlv:
Strings could be more readable; the reason I resist in this case is that layers are not like enums and can have any name. In most fonts they are "Fore", "Back" and "Guide" but glyphs can have any number of layers, with any names...the function already exists, `glyph.layers.index`. The reason it can't be font-level, that we can't have a font-level `layerNameToLayer` is that that will break PS Type 3 fonts (and the coming COLR/CPAL fonts, issue #677), it would need to be glyph-level, and at that point, `glyph.layers.index` is just fine.

skef:
Being able to use either the number or the name is a convenience. Having to type out `[token].layers.index()` is ... less convenient. The python API user base includes less experienced and less sophisticated programmers who have moved over from the native scripting API to get access to more advanced features. It's reasonable to support the integer/string dual for layer identification.

ctrlcctrlv:
I decided to relent. In the process of writing https://github.com/fontforge/fontforge/pull/4074/commits/bd1598a634ed0d4ce696a6b771de6b9a9f8184f8 I discovered new fun and nasty facts about the Type 3 font support, which I don't have time at present to recount in full. Suffice to say it's very half-baked and was never finished properly.

skef:
The issue with the current behavior, to the extent that it matters, is that we pass the "current" layer to `SCClearContents()`. If that layer is the background the function won't work as described. 

`SCClearContents()` is weird, on my reading, about the passed layer. It ignores it WRT `SCClearLayer()` if the font is "multilayer" but not otherwise. Then it decides whether to remove the other stuff based on whether the passed layer is the background.

ctrlcctrlv:
Hi @skef - just so you understand, in FontForge's code, `multilayer` means, "the glyph might have more layers than the font itself does."

This feature is quite broken, see #4078, which I'm using to try to "prepare the ground" for COLR/CPAL support, which will require glyph-only layers.

ctrlcctrlv:
By the way, I don't understand your complaint. Clearing the background layer works just fine for me.

Test file: [4074.sfd](https://github.com/fontforge/fontforge/files/3996818/4074.zip)
Test script:
```
import fontforge
f=fontforge.activeFont()

f["B"].clear(0)
```

skef:
@ctrlcctrlv In that case, if `sc->parent->multilayer` will generally be false, and if `self->layer` (in the Python code) is typically the foreground, won't `ly_first = ly_last = layer;` in `SCClearContents()` typically execute? 

My worry is that if you have more than one non-background layer and call `glyph.clear()` with no arguments it won't actually clear the multiple layers. But maybe I'm reading the code wrong.

ctrlcctrlv:
I'm fixing a bunch of related issues for #4078, so let me try it in my test file for that font and tell you. But yes, `multilayer` will almost always be false, however we need to start thinking about it because I plan to use the feature to support COLR/CPAL

ctrlcctrlv:
A lot of the code assumes that `sf->layer_cnt` is all the layers, which is a problem because in `multilayer` fonts `sf->layer_cnt` != `sc->layer_cnt`.

ctrlcctrlv:
The `clear` works fine for me with `COLR_CPAL_WIP.sfd`.

skef:
@ctrlcctrlv And to be clear, `multilayer` is false for that file?

skef:
Never mind -- I'll just grab it and step through the code to see what happens.

ctrlcctrlv:
No, `COLR_CPAL_WIP.sfd` is the rare file where it's true.

You can have a font with 200 layers and `multilayer` will be false.

Only if you go into &laquo;Element&rarr;Font Info&rarr;Layers&raquo; and turn on `Type 3 Multi Layered Font` is it true, such as in that SFD.

skef:
The case I'm worried about is:

1. `multilayer` is false
2. A glyph (somehow) has more than one foreground layer.
3. A user calls `glyph.clear()` on that glyph expecting all the layers to go away, and only the "current" layer does.

ctrlcctrlv:
OK so here's another confusing thing about the code, and I'm glad we're talking about this.

FontForge considers all user-added, font-level layers to be "background layers". Their default names are e.g. "Back 2" (click plus in CharView). 

On the otherhand, Type 3 glyph-level layers are all "foreground layers".

skef:
@ctrlcctrlv Ah. I would guess that's because if the font itself isn't "multilayer" then FontForge has no idea what to do with more than one layer when outputting. Hence it treats all additional layers as "background" or, in effect, utility layers.

ctrlcctrlv:
Our layer system is really a mess. I was able to instigate multiple crashes last night which led to #4078. I have to get it at least workable before considering adding COLR/CPAL, unfortunately.

ctrlcctrlv:
Indeed @skef

skef:
I never got around to making the most obviously needed change to the Expand Stroke dialog, which was to add some way of specifying input and output layers. Even just a checkbox to take the background layer as input and output to the foreground layer would go a long way.

ctrlcctrlv:
Oh, you must have noticed how I do it in `TT.sfd`, with the `Copy Layer to Layer` dialog (which by the way, ignores glyph-only layers, as does `SFLayerNameToLayer`.)

skef:
I've probably noticed that, but it more stems from the primary model I have in my head.

In my mind the most common development pattern would be creating a source contour, expanding it to see the geometry, modifying the source contour accordingly and re-expanding, and repeating that process as necessary.

There will probably be a point where you'll start post-processing in a way that makes starting over more expensive, but even then it seems like you'll want the source contours handy in case you need to restart from "further back".

With that pattern in mind it's just obvious that you'd want your source and output contours in different layers, with the Expand Stroke process keeping the former and replacing the latter. You're doing it that way with added scripts because it's the right way to do it.

ctrlcctrlv:
So is this good to merge?

skef:
@ctrlcctrlv Yes

PR #4078, opened 2019-12-23
Make glyph-only layers more workable

ctrlcctrlv:
# This is not yet ready for merge!

I'm opening this to get comment, hopefully.

I'll at least need this, _or something like this_, to get COLR/CPAL working...

This fixes a crash if you add font-level layers when a glyph has its own layers. It correctly moves the glyph-level layers, and their data, after the font-level layers.

I also need to make it so that multi-layer fonts can have qudaratic layers, so as you can see, I'm removing a lot of the rules, and I added a "Q/C" marker to the UI for type 3 fonts. More UI work certainly needs to be done.

I'm opening this now and early because if we can't even agree to do this simple stuff COLR/CPAL support will never get approved.


ctrlcctrlv:
Still working on the UI. Trying to colorize it, it's proving to be hard; probably going to take me longer than a few days.

FontForge generates a `BDFFont` when drawing the FontView and the MetricsView (e.g. see `MVResize`), the `clut` of which can only contain grays.

`SplineFontPieceMeal` ends up generating those gray glyphs, through e.g. `FillChar` (splinefill.c).

All of this code is expecting greys. Hacking full RGBA in is going to take time, and I'm not even sure the best way to do it. Should I lean more heavily on FreeType? Should I work it into George's rasterizer? 

IDK...lots of unanswered questions.

ctrlcctrlv:
I've changed my mind. The fastest way to get color into the FontView and MetricsView is likely through the functions in `freetype.c`. Still investigating.

PR #4080, opened 2019-12-29
Clarify OpenCharsInNewWindow (or, Make Shift+Return open char in new tab instead of new window)

ctrlcctrlv:
I believe this was originally a bug, but due to how long it has endured, it has now become a feature.

Consider:

* if `OpenCharsInNewWindow` be `false`, and;
* if a char be double-clicked,

it will open a new tab in an existing window. But if all the same is true, except that the user pressed Return instead, a new window is opened, despite the setting of `OpenCharsInNewWindow`.

So, the bug is that there's no way to open new CharView tabs from the keyboard when `OpenCharsInNewWindow` is false. I decided `Shift+Return` is just as good as any other way.

### Type of change

- **Non-breaking change**

### DIFF
--- a/fontforgeexe/fontview.c
+++ b/fontforgeexe/fontview.c
@@ -6133,6 +6133,30 @@ return;
     GDrawRequestExpose(fv->gw,&r,false);
 }
 
+/* Decide how to open a new CharView for the given SplineChar, based primarily
+ * on the user's preferences (OpenCharsInNewWindow) but also an "extra
+ * consideration" (ec) */
+static int FVDecideOpenSC(FontView *fv, SplineChar *sc, bool ec) {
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
I am voluntarily withdrawing this per an IRC discussion&dagger; I had with @skef on the 30th of December. Sorry to leave this open so long, I thought I'd think of a fix, but @skef was correct (he so often is!): this should be implemented via a custom hotkey, it needs to be documented better (though the original behavior was added in 2005 without any documentation&mdash;&#x200b;e94a44d044), and perhaps the option itself was ill-advised in 2005.

&dagger; If you aren't in IRC, join us: irc.freenode.net #fontforge :-)

Issue #1419, opened 2014-06-16
Ignore unrecognized features in feature files where practical.


frank-trampe:
The feature file parser may not recognize certain feature names in Adobe feature files (as embedded in U. F. O.). It may be desirable to ignore these rather than to stop parsing in certain cases.

I would appreciate input from those with more knowledge on the subject.

rajeeshknambiar:
I believe this issue may be closed now that MR #4097 is merged?

PR #4097, opened 2020-01-13
RFE: [FEAT] When reference to non-existent glyphs are present in an OpenTy…

rajeeshknambiar:
…pe feature file, merging the feature fails.

Add an option to skip the erroring out in merging feature file. This helps with feature files of Indic scripts with many hundreds of ligatures; some of the glyphs may not be present in all fonts,
yet the feature file can be reused.
This is WIP patch. Completed: extended the 'font.mergeFeature()' python interface and associated C code, with documentation update. TODO: extending the 'Merge Feauture' menu in GUI.

### Motivation
Main reason for this pull request is that — Indic fonts can have many hundreds of ligatures (conjuncts formed from many base glyphs) and the OpenType featurefile holds the lookup rules. This feature file can currently be reused in Fontforge based workflow, if the new font have exactly same (or more) glyphs. But if the font has lesser glyphs (usually with a variant such as Bold of an existing font), the featurefile cannot be merged because Fontforge taps out with error
```
Reference to a non-existent glyph name on line 1067 of font-feature.fea: shr3
No substitution specified on line 1067 of font-feature.fea
There were errors when parsing the feature file and the features have not been applied
```

This enhancement preserves the warnings, but skips the error so that applicable lines of the feature file are applied.

### Suggestions/comments required
1.  Does this feature make sense and is the approach technically good?
2.  Not happy with the flag name `ignore_na_feature`, something more intuitive required.
3.  Any guidance on making the flag visible when chosen from the GUI menu.

### Type of change

- **New feature**


ctrlcctrlv:
This is good. I really like this idea. Never thought to propose it.

rajeeshknambiar:
Thanks for the review!

Points 2 & 3 (flag name and tests) are fixed. I'm going to work on exposing the flag in GUI (/me remembers your calculations on paper :-) ) and ask for help when needed.

ctrlcctrlv:
How is it going, @rajeeshknambiar ?

rajeeshknambiar:
Real life™ caught up. Working on it in the weekend.

I was looking at the option to add the flag to `gwwv_open_filename` and going through #3915; but haven’t made much progress yet.

rajeeshknambiar:
@ctrlcctrlv Advise sought on _where_ to add a check button `Ignore invalid features` in GUI. I suppose it does not make sense to add a new flag to `GWidgetOpenFileWPath` function. Should it be placed in `Preferences`?

ctrlcctrlv:
Hi @rajeeshknambiar :

The best thing to do is follow the example of `cvimportdlg.c`. Basically you need to use `GFileChooserCreate`, normally created via `GGadgetCreateData`. See `_Import` function there&mdash;I'd copy something similar, but simplified, into `MergeKernInfo`. The function `gwwv_open_filename` is just a convenience function and is not applicable to this more complicated case where you need to add a checkbox. The way it is implemented is via `GWidgetOpenFileWPath`, which does something very similar to `_Import`.

rajeeshknambiar:
Many thanks. Let me work on it (shall take a few days).

fontforge/featurefile.c
```diff
@@ -1,6 +1,7 @@
 /* Copyright (C) 2000-2012 by George Williams */
 /* Copyright (C) 2012-2013 by Khaled Hosny */
 /* Copyright (C) 2013 by Matthew Skala */
+/* Copyright (C) 2020 by Rajeesh KV */
```
ctrlcctrlv: I would prefer you instead add your name to the AUTHORS file, if that's OK with you. Same for `fontview.c`.

fontforgeexe/fontview.c
```diff
@@ -1029,25 +1030,193 @@ char *GetPostScriptFontName(char *dir, int mult) {
...
+struct gfc_data {
+    int done;
+    int ret;
+    GGadget *gfc;
+    GGadget *ignore_invalid;
+    SplineFont *sf;
+    EncMap *map;
+};
+
+static int e_h(GWindow gw, GEvent *event) {
```
ctrlcctrlv: This function looks good to me, but do you mind calling it something a tad more descriptive, like `fv_e_h` or `feat_e_h`? That's the convention used elsewhere.

fontforgeexe/fontview.c
```diff
@@ -1029,25 +1030,193 @@ char *GetPostScriptFontName(char *dir, int mult) {
 return( temp );
 }
 
+struct gfc_data {
```
ctrlcctrlv: Sorry, what does `gfc` stand for?

fontforgeexe/fontview.c
```diff
@@ -1029,25 +1030,193 @@ char *GetPostScriptFontName(char *dir, int mult) {
...
+}
+
 void MergeKernInfo(SplineFont *sf,EncMap *map) {
 #ifndef __Mac
-    static char wild[] = "*.{afm,tfm,ofm,pfm,bin,hqx,dfont,feature,feat,fea}";
-    static char wild2[] = "*.{afm,amfm,tfm,ofm,pfm,bin,hqx,dfont,feature,feat,fea}";
+    static unichar_t wild[] = {'*','.','{','a','f','m',',','t','f','m',',','o','f','m',',','p','f','m',',','b','i','n',',','h','q','x',',','d','f','o','n','t',',','f','e','a','t','u','r','e',',','f','e','a','t',',','f','e','a','}'
+};
+    static unichar_t wild2[] = {'*','.','{','a','f','m',',','a','m','f','m',',','t','f','m',',','o','f','m',',','p','f','m',',','b','i','n',',','h','q','x',',','d','f','o','n','t',',','f','e','a','t','u','r','e',',','f','e','a','t',',','f','e','a','}'
+};
```
ctrlcctrlv: Why did this need to change? It seems a step backwards in terms of readability.

ctrlcctrlv:
We are in the home stretch @rajeeshknambiar! Just a few minor changes and this is ready :-)

fontforgeexe/fontview.c
rajeeshknambiar: GFileChooser

fontforgeexe/fontview.c
rajeeshknambiar: I don’t like it either, but that's that `GFileChooserSetFilterText` expects as parameter (otherwise I got a crash).

fontforge/featurefile.c
rajeeshknambiar: Certainly. I wasn't sure.
But now that I checked `AUTHORS` file, I’m already there `Rajeesh K Nambiar` for an ancient  contribution.

fontforgeexe/fontview.c
rajeeshknambiar: Tried a couple of things: passing `wild/wild2` as `static char array` results in segfault. Changing `char wiild[]` to `unichar_t wild[]` fails with compile error `cannot initialize array of ‘unsigned int’ from a string literal with type array of ‘char’`. Any ideas?

ctrlcctrlv: Oh, I see. Let's leave it for now. [This has come up before](https://github.com/fontforge/fontforge/issues/3512#issuecomment-509465401), and we haven't yet decided whether or not C11 string literals are OK. For the record, I was wrong there, in this case it would be `static unichar_t wild[] = U"..."`, but you get the point :-)

If we decide it's OK, I'll convert these all at once, but that's something I can handle at a later date.

ctrlcctrlv:
@rajeeshknambiar Just waiting for you to push your commit (not sure why you resolved the conversations before doing so), then I'll approve and merge this.

rajeeshknambiar:
Pushed the last corrections. Sorry for resolving the comments before pushing.

rajeeshknambiar:
@ctrlcctrlv While checking the final changes, noticed that change to `test125` was not reverted properly. That is fixed now.

ctrlcctrlv:
Thank you, I am investigating a crash right now, to see if your code causes it or if another recently merged PR causes it.

ctrlcctrlv:
Never mind! I think the problem is actually that your branch is a bit behind and is missing #4100 :-)

Sorry, just have to be thorough.

ctrlcctrlv:
Unfortunately, it doesn't seem to work for me.

This feature doesn't apply at all, even with it ticked in the UI:

```
feature liga {
     lookup liga1 {
     sub f l by f_l;
     sub f i by f_i;
     sub f f by f_f;
     sub f f l by f_f_l;
     sub f f i by f_f_i;
     sub xddde by a;
     } liga1;
 } liga;
```

```
Reference to a non-existent glyph name on line 86 of /home/fred/Workspace/chomsky/features.fea: xddde
Empty substitute on line 86 of /home/fred/Workspace/chomsky/features.fea
Expected ';' at statement end on line 86 of /home/fred/Workspace/chomsky/features.fea
```

What am I doing wrong? Font is Chomsky.sfd but this test case should be generalizable.

rajeeshknambiar:
That seems to be a legitimate feature file parsing issue, but let me also try to investigate.

ctrlcctrlv:
Not really, or at least, help me understand why it's legitimate that "ignore invalid" doesn't let at least some of those substitutions through? There's a non-existent glyph, `xddde`, but all the other glyphs mentioned do exist in the font.

ctrlcctrlv:
Could the problem be that my test case is non-contextual, and you only tried this with contextual substitutions?

That's what a cursory glance at the code seems to suggest.

ctrlcctrlv:
Or could it be that you only tested substitutions of the form:

```
sub a by nonexistent;
```

And not

```
sub nonexistent by a;
```

?

ctrlcctrlv:
There were indeed 86 lines in the file, actually over a hundred. But they were all comments.

Nevertheless, here's one without comments:

```fea
languagesystem DFLT dflt;
languagesystem latn dflt;

feature liga {
    lookup liga1 {
    sub f l by f_l;
    sub f i by f_i;
    sub f f by f_f;
    sub f f l by f_f_l;
    sub f f i by f_f_i;
    sub xddde by a;
    } liga1;
} liga;
```

And its error:

```
Reference to a non-existent glyph name on line 11 of /home/fred/Workspace/chomsky/features.fea: xddde
Empty substitute on line 11 of /home/fred/Workspace/chomsky/features.fea
Expected ';' at statement end on line 11 of /home/fred/Workspace/chomsky/features.fea
```

fontforgeexe/fontview.c
```diff
@@ -1029,25 +1029,191 @@ char *GetPostScriptFontName(char *dir, int mult) {
...
+        d->ret = false;
+    }
+    return( true );
+}
+
 void MergeKernInfo(SplineFont *sf,EncMap *map) {
 #ifndef __Mac
-    static char wild[] = "*.{afm,tfm,ofm,pfm,bin,hqx,dfont,feature,feat,fea}";
-    static char wild2[] = "*.{afm,amfm,tfm,ofm,pfm,bin,hqx,dfont,feature,feat,fea}";
+    static unichar_t wild[] = {'*','.','{','a','f','m',',','t','f','m',',','o','f','m',',','p','f','m',',','b','i','n',',','h','q','x',',','d','f','o','n','t',',','f','e','a','t','u','r','e',',','f','e','a','t',',','f','e','a','}'};
```
jtanx: I don't have time to check this at the moment, but all of this needs a closer look. It's a big code smell to be converting what was a normal char to unichar_t. None of these literals are null terminated.

ctrlcctrlv: It's because there's no "8" version of the function he now needs. I feel it's too much to ask him to implement one.

ctrlcctrlv: GFileChooserSetFilterText8 doesn't exist, so he needs GFileChooserSetFilterText

See also my comment here: https://github.com/fontforge/fontforge/pull/4097#discussion_r373361576

If we agree to allow C11 we can fix this, and many other constants, tomorrow ;-)

jtanx: I'm not convinced, it's almost never a good idea to be working directly with unichars, unless you're working deep in gdraw.

ctrlcctrlv: He is working deep in gdraw :man_shrugging: 

I've needed the same function before.

ctrlcctrlv: (As to whether they should be null-terminated, I think you are probably right and they should be, which is a good catch :-) )

jtanx: Well if it's needed I don't think it's that hard to add. I'd much rather that than have to do this.

I don't think we should be enforcing C11 yet either.

fontforge/featurefile.c
```diff
@@ -4736,8 +4737,10 @@ static void fea_ParseSubstitute(struct parseState *tok) {
 	} else if ( cnt>=1 && tok->type==tk_by ) {
```
ctrlcctrlv: I am unfortunately now certain that this is the problem. It assumes that the _target_ of the replacement will never be non-existent, only the _replacement_. That might make sense, but I feel it will confuse users. So, a fix for this will need to be found.

doc/html/python.html
```diff
@@ -4949,9 +4949,10 @@ <H2 id="Font">font</H2>
     </TR>
     <TR>
       <TD><CODE>mergeFeature</CODE></TD>
-      <TD><CODE>(filename)</CODE></TD>
+      <TD><CODE>(filename[,boolean])</CODE></TD>
```
jtanx: This should be a variable name, not a type

fontforgeexe/fontview.c
ctrlcctrlv: I am trying not to bite a new contributor. I already felt I was making a tall order when I asked him to implement this in the GUI as well.

However, I take your point @jtanx. I will personally implement GFileChooserSetFilterText8 for him, as I've already asked a lot of him. Would that be acceptable to you?

doc/html/python.html
```diff
@@ -4949,9 +4949,10 @@ <H2 id="Font">font</H2>
...
       <TD>Merge feature and lookup information from an adobe feature file, or metrics
-	information from the (afm,tfm,etc) file into the current font.</TD>
+	information from the (afm,tfm,etc) file into the current font. If optional boolean
+	is TRUE, reference to non-existent glyphs in the feaure file doesn't result in error.</TD>
```
jtanx: `TRUE` -> `True` given this is python

fontforgeexe/fontview.c
jtanx: ok

fontforge/featurefile.c
rajeeshknambiar: OK.

rajeeshknambiar:
> Or could it be that you only tested substitutions of the form:
> 
> ```
> sub a by nonexistent;
> ```
> 
> And not
> 
> ```
> sub nonexistent by a;
> ```
> 
> ?

Indeed, this is the case. I will check the second scenario as well.
Also will address comments by @jtanx tomorrow.

ctrlcctrlv:
@rajeeshknambiar These functions will allow you to not need to change the `char` to a `unichar_t`

```diff
diff --git a/gdraw/gfilechooser.c b/gdraw/gfilechooser.c
index 98275a888..7169d52dc 100644
--- a/gdraw/gfilechooser.c
+++ b/gdraw/gfilechooser.c
@@ -1204,11 +1204,22 @@ void GFileChooserSetFilterText(GGadget *g,const unichar_t *wildcard) {
     gfc->wildcard = u_copy(wildcard);
 }
 
+void GFileChooserSetFilterText8(GGadget *g,const char *wildcard) {
+    GFileChooser *gfc = (GFileChooser *) g;
+    free(gfc->wildcard);
+    gfc->wildcard = utf82u_copy(wildcard);
+}
+
 unichar_t *GFileChooserGetFilterText(GGadget *g) {
     GFileChooser *gfc = (GFileChooser *) g;
 return( gfc->wildcard );
 }
 
+char *GFileChooserGetFilterText8(GGadget *g) {
+    GFileChooser *gfc = (GFileChooser *) g;
+return( u2utf8_copy(gfc->wildcard) );
+}
+
 void GFileChooserSetFilterFunc(GGadget *g,GFileChooserFilterType filter) {
     GFileChooser *gfc = (GFileChooser *) g;
     if ( filter==NULL )
diff --git a/inc/ggadget.h b/inc/ggadget.h
index 4ddf7f75d..a8c8f95b2 100644
--- a/inc/ggadget.h
+++ b/inc/ggadget.h
@@ -471,7 +471,8 @@ void GFileChooserFilterIt(GGadget *g);
 void GFileChooserRefreshList(GGadget *g);
 int GFileChooserFilterEh(GGadget *g,GEvent *e);
 void GFileChooserConnectButtons(GGadget *g,GGadget *ok, GGadget *filter);
-void GFileChooserSetFilterText(GGadget *g,const unichar_t *filter);
+void GFileChooserSetFilterText(GGadget *g, const unichar_t *wildcard);
+void GFileChooserSetFilterText8(GGadget *g, const char *wildcard);
 void GFileChooserSetFilterFunc(GGadget *g,GFileChooserFilterType filter);
 void GFileChooserSetInputFilenameFunc(GGadget *g,GFileChooserInputFilenameFuncType filter);
 int GFileChooserDefInputFilenameFunc( GGadget *g, const unichar_t** currentFilename, unichar_t* oldfilename );
@@ -480,6 +481,7 @@ void GFileChooserSetDir(GGadget *g,unichar_t *dir);
 struct giocontrol *GFileChooserReplaceIO(GGadget *g,struct giocontrol *gc);
 unichar_t *GFileChooserGetDir(GGadget *g);
 unichar_t *GFileChooserGetFilterText(GGadget *g);
+char *GFileChooserGetFilterText8(GGadget *g);
 GFileChooserFilterType GFileChooserGetFilterFunc(GGadget *g);
 void GFileChooserSetFilename(GGadget *g,const unichar_t *defaultfile);
 void GFileChooserSetMimetypes(GGadget *g,unichar_t **mimetypes);
```

Only lightly tested, let me know if you have problems.

rajeeshknambiar:
> Only lightly tested, let me know if you have problems.

Super cool!
I was willing to implement this, but you’re lightning fast than I could. Thanks.
I’ll work on addressing the remaining issues in the weekend (alas I’m working $dayjob tomorrow as well).

fontforge/featurefile.c
rajeeshknambiar: Been debugging this and the hard error seems to be this: `This lookup has no effect, I can't figure out its type on line 9 of  test-feature.fea` which is the closing tag of `liga1`. The check on this code part is ignored by the newly introduced flag.

Will check further.

fontforgeexe/fontview.c
rajeeshknambiar: It's a security bug missing null termination; thanks for spotting it. Fixed using the new API now.

fontforge/featurefile.c
rajeeshknambiar: OK, that error on `Expected ';' at statement end on line 11` is the actual hard error and is triggered from `fea_end_statement` falling through in `fea_ParseSubstitute` due to `rpl`  being `NULL` (non-existing glyph name `xdde`) and the next glyph name is `a` while `fea_end_statement` expects `;`. Back to code reading.

rajeeshknambiar: For the record, I’m still working on this; but the problem space is expanded now to gracefully ignore invalid rules and keep only valid ones; covering substitutions, contextual lookups, mark positionings etc. The code base is designed to stop at error, so I think many functions need to be changed to handle the invalid rules gracefully.
For instance, if all the rules in a lookup are invalid, empty lookup is still created, which is undesirable. I’m trying to understand the code and handle as many cases as possible.

ctrlcctrlv: Right...as I understood it, ignoring invalid rules was always the problem. Ignoring just some invalid rules doesn't really make most sense (especially if the flag's name is literally `ignore_invalid`) and will definitely lead to confused users.

If changing the existing code is very difficult, I would not oppose doing a first pass instead to just kick out rules with glyph names that don't exist. But that might be even harder than editing the existing code, as you'll essentially need to rewrite the parser. There are quite a few cases yes, but basically it's just a matter of, upon discovering a glyph name doesn't exist, consuming more tokens until you get to the next line.

ctrlcctrlv: Hi @rajeeshknambiar , everything OK? Did I miss something about how difficult this feature is? Do you maybe want to collaborate a little more to get it merged? I already wrote some code for you, and I think this feature is genuinely useful, so I wouldn't mind helping out a little more in the code department.

rajeeshknambiar: Hey Frederick,

Thanks for the kind offer. I’ve been quite busy with work and couldn’t dedicate as much time I wanted to on the feature parsing changes. The rough plan I have had is:

1.  Check all the different functions called from `fea_LookupSwitch` which then fills the `feat_item *item` linked list.
2.  Right now the code seems to fill `item` list with tokens as and when it is parsed correctly.
3.  Behaviour in step 2 may need to be changed to fill a temporary list by each function (for parsing one whole lookup rule ) and if it is valid, add it to the `item` list.

I haven’t got to wrapping my head around all these. How would you like to proceed with collaborating? I’m all ears.

skef:
When it comes to it this should close #1419 .

jtanx:
Is this feature complete/just waiting on us, or was there anything outstanding?

rajeeshknambiar:
> Is this feature complete/just waiting on us, or was there anything outstanding?

There are a few outstanding issues; see https://github.com/fontforge/fontforge/pull/4097#discussion_r373441038

Unfortunately I got really busy with $dayjob (and we all know how the past year and half has been); but I would still really like to finish this feature (I have even tried to mentor a person during Free Software Camp who came forward to finish it; but nothing concrete happened). Fredrick kindly offered help to finish it; but I was unable to pick it up/follow up (still quite busy, with only on Sundays I have a few hours of free time).

What would really help for next steps:
1. Collect a few samples of minimal `fea` lookup files for regression testing.
2. Fix/skip error in a couple of remaining code paths when non-existent glyphs are found.
3. When I tested last time, if all substitutions within a lookup were to be skipped, it was resulting in 'empty lookup' error; but I think that was fixed by @ctrlcctrlv few months ago.

Any additional help/code mentoring in pushing this forward is greatly appreciated.

ctrlcctrlv:
@rajeeshknambiar Looking at this with *very* (sorry about that) fresh eyes I suggest you just call the new feature `ignore_invalid_replacement` instead of `ignore_invalid`.

That would resolve the user confusion issue I raised.

Then, if in the future, someone wants to do the much harder task of ignoring the invalid _targets_ (quite unlikely honestly), they can, but you can at least still get your merge in.

ctrlcctrlv:
P.S. As to why I did not suggest this earlier, I did not fully appreciate how useful even just ignoring the invalid replacements would be for Indic fonts two years ago—sorry about that. I'm always learning. :-)

I thought it wouldn't be a logical feature, but actually it's a quite logical one. Because of Indic conjuncts and the way OpenType implemented those.

rajeeshknambiar:
> P.S. As to why I did not suggest this earlier, I did not fully appreciate how useful even just ignoring the invalid replacements would be for Indic fonts two years ago—sorry about that. I'm always learning. :-)
> 
> I thought it wouldn't be a logical feature, but actually it's a quite logical one. Because of Indic conjuncts and the way OpenType implemented those.

Hey, thanks for the feedback.

I haven’t been able to look at this in a couple of years, either. Will try to set aside some time again (rebase against latest code, build & test again etc.). I think I was stuck at creating small but representative test cases (different OT lookups); would appreciate any samples/pointers.

ctrlcctrlv:
@rajeeshknambiar If I were in your place I'd download some of the Noto Indic fonts (because that's really what this is for now that I understand it), export their tables (maybe not in FontForge, but FontForge can do it *pretty* reasonably; look into `defcon`/`extractor` if it derps), then find a simpler Indic font for the same language that is missing a lot of the replacement conjuncts and "baby you've got a stew going" (or at least a test case).

khaledhosny:
Weren't there code already to ignore non-existent glyphs in feature files, or is this something different?

ctrlcctrlv:
I don't think there is?

khaledhosny:
> I don't think there is?

Right. [I misremembered](https://github.com/fontforge/fontforge/commit/815d173a937532fab2793b985f3e6de36c6a9387).

rajeeshknambiar:
> @rajeeshknambiar If I were in your place I'd download some of the Noto Indic fonts (because that's really what this is for now that I understand it), export their tables (maybe not in FontForge, but FontForge can do it _pretty_ reasonably; look into `defcon`/`extractor` if it derps), then find a simpler Indic font for the same language that is missing a lot of the replacement conjuncts and "baby you've got a stew going" (or at least a test case).

No worries on that part; I develop Indic fonts with an order of magnitude more glyphs than Noto :wink: (cf. https://gitlab.com/rit-fonts/RIT-Rachana/-/blob/master/features/mlm2-gsub.fea)

I'm more concerned about (typically) Latin font ligatures and other features I have no experience/used much (such as the test case you have initially mentioned).

But I will try to resurrect this patch anyway.

rajeeshknambiar:
FWIW, I have rebased the changes onto recent master (19-Aug-22), and also tried to fix the error with @ctrlcctrlv's test case (skip error when entire lookup list is empty due to invalid glyphs in substitution). Those commits can be found at https://github.com/rajeeshknambiar/fontforge/commits/merge_featurefile. I hope the changes in commit 9d867af4ed470d3f are correct.

I’m also going to add this test file as a test case soon and update this pull request.

rajeeshknambiar:
@ctrlcctrlv Please review the revised changes. These work fine in my limited testing.

rajeeshknambiar:
@ctrlcctrlv ping.

fontforge/featurefile.c
```diff
@@ -2277,6 +2277,7 @@ struct parseState {
     unsigned int backedup: 1;
     unsigned int skipping: 1;
     unsigned int in_ufo: 1;
+    unsigned int ignore_invalid_sub: 1;      // Don't error out if a substitution rule is not applicable. Used for merging feature files
```
ctrlcctrlv, line 2280: Not understanding why this name was chosen. It's just as confusing. If you want `sub` mentioned put `sub_replacement`.

rajeeshknambiar: May be  I should  use `ignore_invalid_gsub` because this flag actually handles GSUB rules. Does that work, or `ignore_invalid_replacement` as you suggested earlier? Either way is ok with me.

rajeeshknambiar: Please let me know the preference; I shall update the pull request accordingly.

ctrlcctrlv: `ignore_invalid_replacement`. Because there's only one other type of place this might make sense, on a `pos`.

PR #4105, opened 2020-01-24
Fix spelling mistake in 3D projection tool

Ghostkeeper:
Fixing a small spelling mistake that I found here. I updated to a newer version recently and that fixed my tooltips, so I was able to find this then.

A plain is a flat, sweeping landmass that does not change much in elevation. It's a geographical term. See  also https://en.wikipedia.org/wiki/Plain
A plane is a flat, two-dimensional surface in mathematics. This is more appropriate if you're not talking about landscapes or horses or whatever. See also https://en.wikipedia.org/wiki/Plane_(geometry)

### Type of change
- **Bug fix** I'd say.
- **Non-breaking change**, technically, but really, it's a
- **janitor issue**

### DIFF
--- a/fontforgeexe/cvpalettes.c
+++ b/fontforgeexe/cvpalettes.c
@@ -278,7 +278,7 @@ static char *popupsres[] = {
     N_("Add a corner point"),    N_("Add a tangent point"),
     N_("Scale the selection"),   N_("Rotate the selection"),
     N_("Flip the selection"),  N_("Skew the selection"),
-    N_("Rotate the selection in 3D and project back to plain"), N_("Perform a perspective transformation on the selection"),
+    N_("Rotate the selection in 3D and project back to plane"), N_("Perform a perspective transformation on the selection"),
     N_("Rectangle or Ellipse"),  N_("Polygon or Star"),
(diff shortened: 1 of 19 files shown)


Ghostkeeper:
Let me know if you'd like me to update the files in the .po folder.

frank-trampe:
Oh. That's right. There are probably translations for this. Let's fix those.

Ghostkeeper:
This string had translations in `ca`, `fr`, `hr`, `jp`, `ko`, `pl`, `ru`, `uk`, `vi`, `zh_CN` and `zh_TW`. I'm assuming that those translations were interpreted by humans which would properly translate it as "plane" rather than "plain". I can tell from French that that translator did it correctly.

That means that I'd not deprecate the old translations of this string.

jtanx:
In the future, please update the translations through Crowdin, not directly editing the pofiles.

Ghostkeeper:
Ah, sorry. That's why I asked, really. You seem to be starting with Crowdin recently so I didn't know if the system was in place yet :disappointed:

fitojb:
I appreciate that you looked at the translated strings when making this change :+1:

Issue #3983, opened 2019-10-20
Difficulty typing negative numbers in kerning field


aimnadze:
Hi,

It is difficult to type a negative number in kerning field. Any invalid value is cleared after it is typed. Thus the leading minus sign is cleared immediately after typing.

Steps to reproduce:
* Open FontForge
* Open a font with a kerning lookup table
* Open Metrics Window
* Type two characters
* Try typing a negative number between the these two characters

Actual result:
* The negative number should be completely as you type

Expected result:
* When you first press the minus sign it is ignored
* The rest of the number is written

Possible workaround:
* Type the positive part of the number
* Move the cursor at the beginning of the number
* Type the minus sign

ctrlcctrlv:
@aimnadze Would you mind trying out my patch (#4116) and giving me your thoughts on my general theory?

If/when it goes through, the docs will read:

```
 Underneath the display area are a set of text fields. You may type in new numbers for any of these fields to change the corresponding metrics of the glyph. If you wish to type a negative number, it is best to type the positive portion first, then an ASCII minus (`-`). So, to type, -300, type 300, then press -. 
```

ctrlcctrlv:
Closing as no reply and I think my solution, recently merged, is sufficient.

PR #4116, opened 2020-01-29
Make kern in MetricsView toggleable with `-`

ctrlcctrlv:
This, in my opinion, solves #3983.

It is still not possible to type a negative integer in LTR order. I
think that making that possible will cause needless disruption to this
file and is also a waste of time. Instead, first type the number, then a
minus sign, which will "toggle" the value negative.

I've documented this as the right way to do it, as well. I understand
why some people might not like this compromise, but it has been broken
for over ten years. If you have a better way to fix it, feel free :-)

### Type of change

- **Bug fix**

### DIFF
--- a/fontforgeexe/metricsview.c
+++ b/fontforgeexe/metricsview.c
@@ -1542,16 +1542,26 @@ static int MV_KernChanged(GGadget *g, GEvent *e) {
 return( true );
     if ( which>mv->glyphcnt-1 || which==0 )
 return( true );
-    if ( e->u.control.subtype == et_textchanged ) {
-	unichar_t *end;
-	int val = u_strtol(_GGadgetGetTitle(g),&end,10);
 
-	if ( *end && !(*end=='-' && end[1]=='\0'))
-	    GDrawBeep(NULL);
-	else {
-	    MV_ChangeKerning(mv,which,val, false);
-	    MVRemetric(mv);
-	}
+    if ( e->u.control.subtype == et_textchanged ) {
+        char* title = GGadgetGetTitle8(g);
+
+        int negatives = 0;
+        for (char* p = title; *p != '\0'; p++) {
+            if (*p == '-') negatives++;
+        }
+
+        char* new = str_replace_all(title, "-", "", true); // frees title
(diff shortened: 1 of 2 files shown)


fontforgeexe/metricsview.c
```diff
@@ -1542,16 +1542,26 @@ static int MV_KernChanged(GGadget *g, GEvent *e) {
...
+        int negatives = 0;
+        for (char* p = title; *p != '\0'; p++) {
+            if (*p == '-') negatives++;
+        }
+
+        char* new = str_replace_all(title, "-", "", true); // frees title
+        int val = strtol(new,NULL,10);
+        free(new);
+
+        while (negatives) {
```
skef: Wouldn't `if ( negatives%2 )` be preferable here?

fontforgeexe/metricsview.c
```diff
@@ -1542,16 +1542,26 @@ static int MV_KernChanged(GGadget *g, GEvent *e) {
...
-	else {
-	    MV_ChangeKerning(mv,which,val, false);
-	    MVRemetric(mv);
-	}
+        int negatives = 0;
+        for (char* p = title; *p != '\0'; p++) {
+            if (*p == '-') negatives++;
+        }
+
+        char* new = str_replace_all(title, "-", "", true); // frees title
```
skef: Unfortunately neither the for loop above nor this call is likely "safe". Looking for a hyphen is looking for a byte value of 45. In UTF 8 that byte could be part of a multibyte character. That will throw off your negative calculation, but more importantly *eliminating* that byte, as with `str_replace_all`, could corrupt the whole string (by shifting one or more characters into the multibyte character).

How much does this matter for a field where someone is supposed to just type a number? Hmm -- maybe not much? Probably not much?

ctrlcctrlv: I don't really think this matters. The worst that can happen is a number turns negative, but junk is already being entered. Only ASCII works in this field; as the saying goes, "garbage in, garbage out."

skef:
Consider this re-approved.

PR #4117, opened 2020-01-29
Fix for #4095 Dies after Change Weight

### LINKED ISSUE #4095: Dies after change weight
Load the attached font, select the letter, change weight: remove anything in "fuzz", embolden by 9.

[test.zip](https://github.com/fontforge/fontforge/files/4043010/test.zip)

Based on sources from 2019-12-24 15:30 UTC-ML-D-GDK3.
Internal Error (overlap): The end of the monotonic does not match that of the containing spline.
Internal Error (overlap): The end of the monotonic does not match that of the containing spline.
Internal Error (overlap): The end of the monotonic does not match that of the containing spline.
Internal Error (overlap): The end of the monotonic does not match that of the containing spline.
Internal Error (overlap): The end of the monotonic does not match that of the containing spline.
Internal Error (overlap): The end of the monotonic does not match that of the containing spline.
Internal Error (overlap): The end of the monotonic does not match that of the containing spline.
Internal Error (overlap): Winding number did not return to 0 when y=554,842
Internal Error (overlap): Winding number did not return to 0 when y=554,889
Internal Error (overlap): Winding number did not return to 0 when y=554,842
Internal Error (overlap): Winding number did not return to 0 when y=554,889
Internal Error (overlap): Winding number did not return to 0 when y=554,842
Internal Error (overlap): The start of the monotonic does not match that of the containing spline.
malloc_consolidate(): invalid chunk size
Aborted (core dumped)

skef:
Closes #4095

@frank-trampe Can you look at this and modify as you see fit? In my analysis the existing `break;` was escaping to the wrong part of the loop. An alternative would be to set `sp` and `nsp` to NULL and check for their being NULL in the following top-level guards.

### DIFF
--- a/fontforge/splineoverlap.c
+++ b/fontforge/splineoverlap.c
@@ -3505,7 +3505,7 @@ static SplineSet *SSRemoveTiny(SplineSet *base) {
 			else
 			    prev->next = ssnext;
 			base = NULL;
-	break;
+	goto bailout;
 		    }
 		    // We want to remove the spline following sp.
 		    // This requires that we rewrite the following spline so that it starts at sp,
@@ -3631,6 +3631,7 @@ static SplineSet *SSRemoveTiny(SplineSet *base) {
 	    if ( refigure )
 		SplineRefigure(sp->next);
 	}
+      bailout:
 	if ( base!=NULL )
 	    prev = base;
 	base = ssnext;


ctrlcctrlv:
@frank-trampe doesn't seem to care, eight days later. It looks good to me. Let's given him 24 more hours, and if no answer, merge this.

skef:
Eh, I added a `goto` -- probably someone will care. Let's leave it for now.

frank-trampe:
Oh. Sorry about that. I was going to come back to this and forgot about it. I would prefer the alternative in this case, even though it is more complicated; I think that gotos are awkward exceptions to the structure of C, and many snobs hate them just because other people have told them to do so.

skef:
@frank-trampe So should I put that on "my plate", so to speak?

frank-trampe:
I can do it tomorrow if you like. I've already read through the code, and you've already said what needs to be done, so I could do it fast.

skef:
I think I have the diagnosis right but you know this stuff much better than I do, so I think that would be my preference for this one.

frank-trampe:
Okay. Will do.

ctrlcctrlv:
Superseded by #4138.

PR #4118, opened 2020-01-30
Fix for SplinePointListIsClockwise() false negative problems

### LINKED ISSUE #3571: SplinePointListIsClockwise: Found error
Arch Linux, 3aabfd2

As far as I can tell, this has no effect on the UI and doesn't cause a crash, but it may be one of the reasons Spiro editing is buggy, and may reveal underlying problems in the implementation, so I report it.

g

Reproduction steps:

1. Open [SPLIC.sfd](https://github.com/fontforge/fontforge/files/2999035/SPLIC.zip)
1. Open CharView for glyph `g`
1. Select any Spiro point
1. Press `` ` ``
1. You should receive in stdout:
 
```
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error
SplinePointListIsClockwise: Found error

```

### LINKED ISSUE #4049: How do I correctly stroke an ‹f›, without getting a hole in the middle?

### LINKED ISSUE #4050: Expand Stroke Remove Overlap removing countour it shouldn't

skef:
Suppose you want to know whether a closed, non-intersecting contour is clockwise or counterclockwise. You can do this by finding a single point in the inside, following a line to the outside, and looking at the direction of the spline you cross. As long as by "non-intersecting" we exclude both crosses and "shared" edges, this will work. 

However, perhaps it is a pain to determine a point conclusively inside the contour. In that case you can start a line definitively outside of the contour and look at the direction of the first spline crossed. This will also work. 

In practice, however, you may not know whether a spline is self-intersecting and may not want to run a separate algorithm to determine that. Or you may want to classify "degrees" of clockwise- or counterclockwise-ness. For such purposes you need to analyze more of the glyph. 

The existing `SplinePointListIsClockwise()` code uses the autohint toolset to build a structure of spline fragments (broken at extrema points) and organized by the vertical em-unit each starts on. This allows the contour to be "scanned" across at a given height looking for regions enclosed clockwise or counter-clockwise based on the directions of the splines that are crossed. The more heights that are examined the higher the probability of finding regions enclosed multiple times, and therefore of self-intersections. I don't entirely understand what the algorithm is determining because I didn't look at the winding-related code all that closely. (It wasn't necessary.) However, it rarely marks a self-intersecting glyph as definitively clockwise or counterclockwise. 

The mystery of this function was that it was sometimes failing for non-intersecting contours, and this was causing problems in Expand Stroke code that depends on its output. It turns out that there are a moderate number of conditions that cause the algorithm to skip a given line. If these line up just right no line will be processed and therefore no data is available. 

My basic fix for this problem is simple: When the algorithm fails make a copy of the contour, enlarge it, and try again. In addition to that idea I add:

1. If the spline is particularly small (in em-units) relative to the number of points it has (lines that include a point are skipped), enlarge it more.
2. When hitting the conditions that used to result in printing "SplinePointListIsClockwise: Found error", just skip that line. 
3. Until the spline is being processed at maximum magnification, discard the result unless one third of the lines are usable and try again. 

Closes #4049
Closes #4050
Closes #3571

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -365,7 +365,7 @@ enum ShapeType NibIsValid(SplineSet *ss) {
 	return Shape_TooFewPoints;
     if ( ss->first->prev==NULL )
 	return Shape_NotClosed;
-    if ( !SplinePointListIsClockwise(ss) )
+    if ( SplinePointListIsClockwise(ss)!=1 )
 	return Shape_CCW;
     if ( ss->first->next->order2 )
 	return Shape_Quadratic;
@@ -1728,8 +1728,18 @@ static SplineSet *OffsetSplineSet(SplineSet *ss, StrokeContext *c) {
 	    else {
 		if ( c->rmov==srmov_contour )
 		    left = SplineSetRemoveOverlap(NULL,left,over_remove);
-		// Open paths don't always result in clockwise output
-		if ( !SplinePointListIsClockwise(left) )
+		// Open paths don't always produce clockwise output
+		is_ccw_mid = SplinePointListIsClockwise(left);
+		if ( is_ccw_mid==-1 && c->rmov!=srmov_none ) {
+		    assert( c->rmov!=srmov_contour );
+		    left = SplineSetRemoveOverlap(NULL,left,over_remove);
+		    is_ccw_mid = SplinePointListIsClockwise(left);
+		    assert( is_ccw_mid!=-1 );
+		} else {
+		    LogError( _("Warning: Can't identify contour direction, "
+		                "assuming clockwise\n") );
+		}
+		if ( is_ccw_mid==0 )
 		    SplineSetReverse(left);
 	    }
 	}
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -1735,7 +1735,7 @@ static SplineSet *OffsetSplineSet(SplineSet *ss, StrokeContext *c) {
 		    left = SplineSetRemoveOverlap(NULL,left,over_remove);
 		    is_ccw_mid = SplinePointListIsClockwise(left);
 		    assert( is_ccw_mid!=-1 );
-		} else {
+		} else if ( is_ccw_mid==-1 ) {
 		    LogError( _("Warning: Can't identify contour direction, "
 		                "assuming clockwise\n") );
 		}
--- a/fontforge/splineutil2.c
+++ b/fontforge/splineutil2.c
@@ -4041,14 +4041,14 @@ return( NULL );
 return( ret );
 }
 
-int SplinePointListIsClockwise(const SplineSet *spl) {
+static int _SplinePointListIsClockwise(const SplineSet *spl, int max_depth) {
     EIList el;
     EI *active=NULL, *apt, *pr, *e;
     int i, winding,change,waschange, cnt;
     SplineChar dummy;
     SplineSet *next;
     Layer layers[2];
-    int cw_cnt=0, ccw_cnt=0;
+    int cw_cnt=0, ccw_cnt=0, l_cw_cnt, l_ccw_cnt, lines_processed = 0;
 
     memset(&el,'\0',sizeof(el));
     memset(&dummy,'\0',sizeof(dummy));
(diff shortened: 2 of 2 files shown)


frank-trampe:
The described logic makes sense to me. I can't review the code just now, but I will do so tomorrow if nobody else gets to it first.

skef:
I'll look at the failing tests. I'm not sure why they would be failing in one context and not another but maybe I can track it down just by analysis.

PR #4120, opened 2020-01-31
Scale some Expand Stroke fudge-factors with log of max nib radius

### LINKED ISSUE #4109: Warning: Left contour did not close
This bug is under the `skef:arcs` branch but doesn't use arc joins so I doubt that's the problem. Even if it is, @skef should know about it.

2020-01-27-154310_3589x1750_scrot

# File: [S2.sfd](https://github.com/fontforge/fontforge/files/4115968/S2.zip)

skef:
The larger the nib, the greater the error in certain offset calculations, and therefore the greater the fudge factor needed when correcting for those errors and closing the left and right offsets of an open source spline. This PR adds a ln(max nib radius)-based meta-fudge for stroking with unusually large radii.

Because these factors are only for "checks", and don't enter into geometric calculations, a larger factor poses a low risk. 

Closes #4109

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -45,6 +45,7 @@
 #include <math.h>
 
 #define CIRCOFF 0.551915
+#define LOG_MAXDIM_ADJ (1.302585)
 
 // Rounding makes even slope-continuous splines only approximately so.
 #define INTERSPLINE_MARGIN (1e-1)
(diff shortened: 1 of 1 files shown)


skef:
@jtanx no one else is likely to review this given the subject matter, and it's actually pretty straightforward: just fudge based on the nib size. Could you take a look?

fontforge/splinestroke.c
```diff
@@ -45,6 +45,7 @@
 #include <math.h>
 
 #define CIRCOFF 0.551915
+#define LOG_MAXDIM_ADJ (1.302585)
```
skef: This is ln(10)-1

PR #4123, opened 2020-02-02
cmake: Fix gcc/clang conditionals

jtanx:
I was targeting both, but it seems like CMAKE_COMPILER_IS_GNUCC only matches gcc.

### Type of change
- **Bug fix**


ctrlcctrlv:
Awesome, I guess I'll start building with Clang since it's much faster!

ctrlcctrlv:
(By the way, if you have a few minutes, I saw you join IRC&mdash;happy to talk if you were looking for me. @skef sometimes joins too, especially if you ask him to.)

jtanx:
Hah well it was more like I'm forced to use clang when working on mac.

PR #4127, opened 2020-02-02
Remove outdated Portfile for MacPorts

mojca:
I don't see any good reason for shipping the outdated and defunct package description here, it's probably much better to simply remove it.

Anyone who wants to install FontForge with MacPorts would be much better served by using the version provided inside MacPorts (and if that one requires improvements that you would like to see, fixing the file "upstream" makes more sense).

A preliminary attempt at updating the package is here: https://github.com/macports/macports-ports/pull/6289


mojca:
Thank you!

PR #4133, opened 2020-02-04
Add glyph-level genericGlyphChange (same as UI "Change Glyph ...")

### LINKED ISSUE #2925: Python API for "Change Glyph"
The Python API for the "Change weight" dialog is great; are there plans to have one for the "Change glyphs" one?

skef:
The oft-requested (but already possible) "Change Glyph" Python API, but on the glyph object to avoid having to use the selection interface. Uses the same parameter parsing logic as the old command, moved to a separate function

The main annoyance with the API compared with the UI is the lack of a default `vMap`. The dialog VMap is based on the BlueValues (when those are present). However, the `font.private` object has these available in a tuple, so if that's what the user wants they're not too hard to build. 

Closes #2925

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -8353,6 +8353,38 @@ return( NULL );
 return( result );
 }
 
+static bool PyFFParse_genericGlyphChange(PyObject *args, PyObject *keywds,
+                                         struct genericchange *genchange);
+
+static PyObject *PyFFGlyph_genericGlyphChange(PyFF_Glyph *self, PyObject *args,
+                                              PyObject *keywds) {
+    SplineChar *sc = self->sc;
+    struct genericchange genchange;
+    struct smallcaps small;
+
+    if ( self->layer < 0 )
+	Py_RETURN( self );
+
+    if ( !PyFFParse_genericGlyphChange(args, keywds, &genchange) )
+	return NULL;
+
+    SmallCapsFindConstants(&small,sc->parent,self->layer);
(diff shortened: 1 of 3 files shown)


ctrlcctrlv:
It worries me vis &agrave; vis #4078 to add more code that doesn't seem to have been tested on multilayer fonts. Did you do tests on the font in that PR?

skef:
I did not test on those fonts. However, the pattern of the code follows the usual pattern of applying to the currently selected layer, which is how almost all of the glyph-level methods work (when they affect contours).

ctrlcctrlv:
I notice the use of `FontViewBase`. Would you mind testing on the font there?

skef:
I can, but if you scan up the old code you'll see that's part of the current font-level call as well -- it just wound up being moved in the process of isolating the parameter parsing.

skef:
They both work. You have to set `font.activeLayer` to have the font-level method later that level and `glyph.activeLayer` for the glyph-level method. And if you wanted to change the whole glyph you'd have to cycle through the layers. 

Of course, the way Change Glyph works makes good results unlikely because the output depends on the relation of the contour shape to the font metrics, which is likely to be different for different layers in a multi-layer font.

ctrlcctrlv:
Thanks, sorry to bother you. Hopefully you understand the reasons behind my concern :-)

skef:
Given that the methods depend on the layer value it's a reasonable test to run.

PR #4134, opened 2020-02-05
Add second pass to the tangent solver for sloppier calculations

skef:
The very tight (10<sup>-8</sup>) epsilon on slope testing, where you recalculate the slope at the t given by the extrema solver to see if it matches, worked fine through lots of testing. However, when looking at the arcs join branch @ctrlcctrlv found some weird-acting cases (with TT's capital S) that I eventually traced to this margin. 

It appears that the quadratic calculation gets sloppier when the 1D-Spline value `a` gets close to zero because `a` is in the denominator. The code has provisions to set `a` to zero and perform an alternate calculation but doesn't do this at the problematic values and it's not clear that it should -- the t value it calculates for the extrema is still pretty accurate. 

Rather than add a big value sorting production number to the function I just added an optional second pass with a 10<sup>-5</sup> epsilon.

To see a context where it makes a difference stroke the fragment in this font at 30/30/0/bevel/bevel: 
[badexternal.sfd.txt](https://github.com/fontforge/fontforge/files/4162587/badexternal.sfd.txt)

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -739,7 +739,7 @@ static NibOffset *_CalcNibOffset(NibCorner *nc, int n, BasePoint ut,
 	                        nc[ncni].utv[NC_IN_IDX], false) );
 	// Nib splines are locally convex and therefore have t value per slope
 	ns = nc[nci].on_nib->next;
-	no->nt = SplineSolveForUTanVec(ns, ut, 0.0);
+	no->nt = SplineSolveForUTanVec(ns, ut, 0.0, false);
 	if ( no->nt<0 ) {
 	    // At more extreme nib control point angles the solver may fail.
 	    // In such cases the tangent angle should be near one of the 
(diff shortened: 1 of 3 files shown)


fontforge/splinestroke.c
```diff
@@ -1303,7 +1303,7 @@ static bigreal SplineStrokeNextT(StrokeContext *c, Spline *s, bigreal cur_t,
 
     next_ut = SplineStrokeNextAngle(c, *cur_ut, is_ccw, &next_curved,
                                     reverse, nci_hint);
-    next_t = SplineSolveForUTanVec(s, next_ut, cur_t);
+    next_t = SplineSolveForUTanVec(s, next_ut, cur_t, false);
```
ctrlcctrlv: Forgive my ignorance, but it seems `picky` is never `true` - why PR this? You call it &laquo;optional&raquo;, but I don't see a way to enable it without editing this line (and the previous similar line)?

skef: I wrote `SplineSolveForUTanVec()` as a "general" facility that could be called in the future by other projects. I was nervous about adding this workaround for general use so I added a boolean to control whether it applies. 

When `picky` is passed as true the function works as it did before (`if (picky) return -1.0;`). As it happens I don't need that in the two places I call it -- I want the looser fallback if it's present.

skef: So, it's optional in the sense that future consumers of the function can avoid it.

ctrlcctrlv: OK, I understand. In that case I'll go ahead and merge this, you know best what's needed, and it's essentially a no-op in the current code.

skef: Well, as it happens the only two times this function is called have now changed, in that there is a section of `SplineSolveForUTanVec()` that looks for looser (i.e. less accurate) slope matches. 

But only after the tighter matches have failed, and the looser ones still have to be pretty close.

PR #4139, opened 2020-02-07
Prevent UBSan false positive

ctrlcctrlv:
Useful if compiled with `-DENABLE_SANITIZER=undefined`

### Type of change

- **Non-breaking change**


skef:
I'm not thrilled about the change of to the return value type without any corresponding changes to the types of the variables it is assigned to. Seems like they're (almost?) all still `int`.

PR #4141, opened 2020-02-07
Spiro crash fixes (2)

### LINKED ISSUE #1004: Crash while merging spiro points
Select spiro control points, Edit > Merge To Line, crash.

On Fedora 20

```
SPSelectPrevPoint (sp=0xdc1f50, state=state@entry=1) at charview.c:12934
12934       if( !sp->prev->from )
(gdb) bt
#0  SPSelectPrevPoint (sp=0xdc1f50, state=state@entry=1) at charview.c:12934
#1  0x00007ffff796e870 in _CVMergeToLine (elide=0, cv=0xa06660) at charview.c:8123
#2  CVMergeToLine (gw=<optimized out>, UNUSED_mi=<optimized out>, UNUSED_e=<optimized out>) at charview.c:8134
#3  0x00007ffff76621e2 in gmenu_mouse (m=<optimized out>, event=0x7fffffffbd70) at gmenu.c:1001
#4  0x00007ffff76626bb in gmenu_eh (w=0xdc21d0, ge=0x7fffffffbd70) at gmenu.c:1586
#5  0x00007ffff768aa01 in dispatchEvent (gdisp=gdisp@entry=0x8af5e0, event=event@entry=0x7fffffffbed0) at gxdraw.c:3502
#6  0x00007ffff768c3ca in GXDrawEventLoop (gd=0x8af5e0) at gxdraw.c:3601
#7  0x00007ffff7a8e3a2 in fontforge_main (argc=1, argv=<optimized out>) at startui.c:1265
#8  0x00007ffff1ab3d65 in __libc_start_main (main=0x400920 <main>, argc=1, argv=0x7fffffffe138, init=<optimized out>, fini=<optimized out>, 
    rtld_fini=<optimized out>, stack_end=0x7fffffffe128) at libc-start.c:285
#9  0x0000000000400951 in _start ()
```

### LINKED ISSUE #3569: Spiro crash when merging points

ctrlcctrlv:
I am extremely pleased to have the privilege of fixing these bugs.

A lot of us have tried. @JoesCat and @jtanx gave very worthwhile support in the related issues, but today I finally cracked them.

One commit closes #3569.
The other closes #1004.

### Type of change
- **Bug fix**

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -8673,10 +8673,12 @@ static void _CVMergeToLine(CharView *cv, int elide) {
     SplineCharMerge(cv->b.sc,&cv->b.layerheads[cv->b.drawmode]->splines,!elide);
 
     // Select the other side of the new curve
-    GList_Glib* gl = CVGetSelectedPoints( cv );
-    if( g_list_first(gl) )
-	SPSelectPrevPoint( (SplinePoint*)g_list_first(gl)->data, 1 );
-    g_list_free( gl );
+    if (!CVInSpiro(cv)) {
+        GList_Glib* gl = CVGetSelectedPoints( cv );
+        if( g_list_first(gl) )
+        SPSelectPrevPoint( (SplinePoint*)g_list_first(gl)->data, 1 );
+        g_list_free( gl );
+    }
 
     // And make the curve between the two active points a line
     _CVMenuMakeLine( (CharViewBase*) cv, 0, 0 );
--- a/fontforgeexe/cvaddpoints.c
+++ b/fontforgeexe/cvaddpoints.c
@@ -275,7 +275,6 @@ return( found!=NULL );
 
 // Return type: index into spiro_cp array, or -1 if not found.
 static int CVSpiroSelected(CharView *cv, SplineSet *ss) {
-    spiro_cp* selpt;
     int i, j = 0;
     bool found = false;
     for (i = 0; i < ss->spiro_cnt; i++) {
--- a/fontforgeexe/cvaddpoints.c
+++ b/fontforgeexe/cvaddpoints.c
@@ -273,6 +273,33 @@ return( found!=NULL );
     }
 }
 
+// Return type: index into spiro_cp array, or -1 if not found.
+static int CVSpiroSelected(CharView *cv, SplineSet *ss) {
+    spiro_cp* selpt;
+    int i, j = 0;
+    bool found = false;
+    for (i = 0; i < ss->spiro_cnt; i++) {
+        if (SPIRO_SELECTED(&ss->spiros[i])) {
+            j = i;
+            found = true;
+        }
+    }
+    if (!found) {
+        TRACE("CVSpiroSelected: We expected to find a selected spiro, but failed to.\n");
+        return -1;
+    }
+    TRACE("CVSpiroSelected: %d\n", j);
+    return j;
+}
+
+// If needed, reset the selected spiro.
+static void CVResetSelSpiro(CharView* cv, SplineSet *ss) {
+    if (CVInSpiro(cv)) {
+        int sel = CVSpiroSelected(cv, ss);
+        cv->p.spiro = cv->lastselcp = cv->active_cp = sel == -1 ? NULL : &ss->spiros[sel];
+    }
+}
+
 static void CVMergeSPLS(CharView *cv,SplineSet *ss, SplinePoint *base,SplinePoint *sp) {
     int order2 = cv->b.layerheads[cv->b.drawmode]->order2;
 
@@ -413,6 +440,8 @@ return;			/* We clicked on the active point, that's a no-op */
 
     cv->active_spl = ss;
     cv->active_cp = cp;
+    // We may have reallocated the spiros (above), so we need to reset the selected spiro.
+    CVResetSelSpiro(cv, ss);
     CVSetCharChanged(cv,true);
     CVInfoDraw(cv,cv->gw);
     SCUpdateAll(sc);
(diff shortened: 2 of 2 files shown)


jtanx:
While I'm sure this fixes some specific spiro crashes, it should be noted it still doesn't address the fundamental problem of how spiros are represented in FontForge - namely that it's stored as an array that gets reallocated whenever too many points are added - and where - in multiple places - points are just held into that array, which go invalid whenever that reallocation occurs.

You can still reproduce this on this branch quite easily by enabling ASan and just continually adding spiro points until ASan kicks in.

ctrlcctrlv:
As I see it the problem is not really the reallocation but the fact that stale pointers are being held onto. There's really no way to do away with the reallocation, is there? Not in a way that can support arbitrary numbers of Spiro points, you can get quite a few just by converting to Bezier and back a few times.

This releases stale pointers...I'm sure it needs to be done in other places as wewll

jtanx:
> There's really no way to do away with the reallocation, is there? 

There is, it's called use a linked list like is done for the spline sets.

jtanx:
(or otherwise see what I did in those commits on that issue)

ctrlcctrlv:
My recommendation would be: Merge this for now, revert if you plan to fundamentally change how Spiros are represented _everywhere_. Changing to linked lists would be a lot of work, but I'd happily welcome it if you want to do it ;-)

ctrlcctrlv:
Additional context from IRC so it doesn't get lost:

```
05:25:55 @copypaste jtan: do you really not like #4141? changing to linked lists everywhere might be more than i can do :)
05:26:33 @copypaste Spiro bugs remain pain points so i don't think incremental fixes are really that bad 
06:17:47 @jtan copypaste: yeah I'm not suggesting to rewrite it all at once, fixes like the one you posted are probably fine
06:19:47 @copypaste jtan: ok. yeah i mean it's been a major pain point for me for a year
06:19:51 @copypaste so im happy i found a somewhat concise fix
06:20:16 @copypaste i'd be happy with replacing spiros with linked lists everywhere but i don't know that i'm even capable of that
06:20:18*@copypaste even if i had the time
06:20:22 @copypaste which i don't :-)   
06:25:00*@jtan no one ever has enough time :)
```

skef:
I'm poking my head in here having only read the code in a very casual way, but given that the existing implementation works with an array, and the only identified, unaddressed problem is the realloc, couldn't we change the problem pointer into what -- in the old days, on the Mac -- was called a "handle"? That is:

1. Rather than hand out pointers to the array, hand out pointers to the element in the structure that points to the array.
2. Update the code that makes use of those pointers to dereference an additional time. 

That's an *annoying* change to implement, but a *simple* one, and if it's really only the realloc causing the problem it seems sufficient.

skef:
On second thought: If a natural solution is a linked list the pointers are presumably to individual elements of the array/spiro contour rather than the contour as a whole. 

If that's true, though, then a list approach would likely suffer the same problems as our current pointers to points: they don't handle the deletion case. The current non-modal point info dialog is a frequent source of crashes. 

When both moves and deletions are a problem one often uses some kind of two-way structure: The pointer is to some element in a list "next to" the object that itself has a pointer to the correct element. Those intermediate elements can then be updated as the list is. That layer of indirection means that you can use whatever "main" structure you want: list, array, etc. (With a linked list you only need to update the intermediate elements on deletion, with an array you have to update on realloc and any ordering changes.) 

(An alternative is to use a linked list and reference-count the elements, so that they can be "dead" for as long as there are pointers to them. You see that semantic in gdraw but not in the spline code, as far as I've seen.)

ctrlcctrlv:
In this specific case, since there's a merge happening, the pointers really should be NULLed, and that was always an error not to, regardless of data structure. Reference counting would just lead to a memory leak in this case. The fact it works at all sometimes on master is just coincidence and platform specific. That's why the crash didn't happen on Valgrind, literally where realloc puts the memory matters.

ctrlcctrlv:
I just pushed a nicer version of this, which continues to allow you to drag the point around after the merger.

ctrlcctrlv:
@JoesCat, do you mind taking a look at this?

frank-trampe:
That seems reasonable.

PR #4142, opened 2020-02-08
SVG-related fixes

skef:
1. Correct extend-cap and join limit parameters to match SVG spec
2. Enable stroke-miterlimit and additional joins in imported SVGs
3. Set default SVG miterlimit to 4.0 stroke-widths to match spec

There aren't really any standard miter-clip or arcs join tests currently -- they haven't yet been added to the wpt suite. However, I did find some miter tests that revealed I had been misinterpreting the miterlimit (or "Join Limit" in the Expand Stroke code). This corrects the relative versions of Join Limit and Extend Cap to be multiples of stroke with rather than radius. This has the benefit of being easier to describe and document. 

Beyond that these changes enable arcs and miter-clip joins in SVG imports and corrects some uses of `xmlStrcmp()`. `stroke-miterlimit` properties are also extracted from the style.  I also fixed the join length calculation (which is based on angle rather than actual length as the SVG spec requires).


frank-trampe:
So we had all those `xmlStrCmp` checks inverted?!

skef:
@frank-trampe Yes, they matched the other ones fixed earlier in https://github.com/fontforge/fontforge/pull/4029.

PR #4143, opened 2020-02-08
Make it possible to ask user for multiple choices (Python)

ctrlcctrlv:
Multiple of which can be selected.

This is needed for #703, which I'm implementing as a Python script.

This is the first of two new needed APIs. The second will add a new
function that works on `fontforge.glyph` objects, similar to `build`,
but will instead take a combining class type and place an anchor.

For easier testing, the script below may be used:

```python3
import fontforge

AnchorLocations = (
    ["Above", 0x100],
    ["Below", 0x200],
    ["Overstrike", 0x400],
    ["Left", 0x800],
    ["Right", 0x1000],
    ["Joins2", 0x2000],
    ["Center-left", 0x4000],
    ["Center-right", 0x8000],
    ["Centered outside", 0x10000],
    ["Outside", 0x20000],
    ["Left edge", 0x80000],
    ["Right edge", 0x40000],
    ["Touching", 0x100000]
)


def guess_anchors(data, font):
    al_strings = tuple([x[0] for x in AnchorLocations])
    print(al_strings)
    r=fontforge.askChoices("Anchor types", "What anchor type?", al_strings)
                         #tuple([True for i in range(len(al_strings))]), True)
    print(r)
    print(data, font)


pr = {}

fontforge.registerMenuItem(guess_anchors, None, pr, "Font", None,
                           "Guess Appropriate Anchors")
```

### Type of change

- **New feature**

### DIFF
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -838,7 +838,7 @@ fontforge.registerImportExport(importGlyph,exportGlyph,None,"foosball","foo","fo
     <TR>
       <TD><CODE>askChoices</CODE></TD>
       <TD><CODE>(title,question,answers,<BR>
-	[default,multiple])</CODE></TD>
+	[default=,multiple=])</CODE></TD>
       <TD>Similar to the above allows you to ask the user a multiple choice question.
 	It popups up a dialog posing the question with a scrollable list of choices
 	-- one for each answer.
@@ -857,6 +857,9 @@ fontforge.registerImportExport(importGlyph,exportGlyph,None,"foosball","foo","fo
 	<P>
 	The function returns the index in the answer array of the answer choosen
 	by the user. If the user cancels the dialog, a -1 is returned.
+	<P>
+	<CODE>default</CODE> and <CODE>multiple</CODE> may be passed by position if
+	desired.
 	<P>
 	Throws an exception if there is no user interface.</TD>
     </TR>
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -850,9 +850,10 @@ fontforge.registerImportExport(importGlyph,exportGlyph,None,"foosball","foo","fo
 	presses the [Return] key). If omitted the default answer will be the first.
 	<P>
 	The fifth argument means that multiple options can be selected. If true,
-	the fourth argument should be a tuple of boolean values. So, if there are
-	three options, it should look like <CODE>(True, False, True)</CODE>, which
-	would select the first and last option.
+	the fourth argument should be a tuple of boolean values or a single integer
+	index into the answer tuple. So, if there are three options, it should look
+	like <CODE>(True, False, True)</CODE>, which would select the first and
+	last option.
 	<P>
 	The function returns the index in the answer array of the answer choosen
 	by the user. If the user cancels the dialog, a -1 is returned.
--- a/doc/html/python.html
+++ b/doc/html/python.html
@@ -836,9 +836,9 @@ fontforge.registerImportExport(importGlyph,exportGlyph,None,"foosball","foo","fo
 	Throws an exception if there is no user interface.</TD>
     </TR>
     <TR>
-      <TD><CODE>askChoice</CODE>s</TD>
+      <TD><CODE>askChoices</CODE></TD>
       <TD><CODE>(title,question,answers,<BR>
-	[def])</CODE></TD>
+	[default,multiple])</CODE></TD>
       <TD>Similar to the above allows you to ask the user a multiple choice question.
 	It popups up a dialog posing the question with a scrollable list of choices
 	-- one for each answer.
@@ -848,6 +848,11 @@ fontforge.registerImportExport(importGlyph,exportGlyph,None,"foosball","foo","fo
 	the fourth and fifth arguments are option, the fourth is the index in the
 	answer array that will be the default answer (the one invoked if the user
 	presses the [Return] key). If omitted the default answer will be the first.
+	<P>
+	The fifth argument means that multiple options can be selected. If true,
+	the fourth argument should be a tuple of boolean values. So, if there are
+	three options, it should look like <CODE>(True, False, True)</CODE>, which
+	would select the first and last option.
 	<P>
 	The function returns the index in the answer array of the answer choosen
 	by the user. If the user cancels the dialog, a -1 is returned.
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -1734,15 +1734,17 @@ return( NULL );
 return( Py_BuildValue("i",ret));
 }
 
-static PyObject *PyFF_askChoices(PyObject *UNUSED(self), PyObject *args) {
+static const char *askChoices_keywords[] = { "title", "question", "answers", "default", "multiple", NULL };
+
+static PyObject *PyFF_askChoices(PyObject *UNUSED(self), PyObject *args, PyObject *kwargs) {
     char *title=NULL, *quest=NULL;
     char **answers; // receives answers written into `answero`
     PyObject* defo = NULL; // default answer index, or tuple of len cnt
     int def = 0; // default index, only used if not multianswer
     int cnt; // number of answers
-    PyObject *multipleo; // expected to be boolean
+    PyObject *multipleo = NULL; // expected to be boolean
     bool multiple = false;
-    PyObject *answero; // expected to be tuple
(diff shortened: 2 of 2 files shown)


skef:
I'm happy to review this and strongly approve of the change.

I'm going to start with an optional suggestion, which is that now that we have two optional arguments it would be good to throw some keywords onto them so that they can be specified either by position or by keyword. It's entirely backward compatible with the existing interface and not much more work -- the current `PyFFFont_autoWidth()` provides a good model of the semantic.

fontforge/python.c
```diff
@@ -1735,50 +1735,100 @@ return( Py_BuildValue("i",ret));
...
+        PyErr_Format(PyExc_EnvironmentError, "No user interface");
+        return( NULL );
     }
 
-    if ( !PyArg_ParseTuple(args,"esesO|i","UTF-8", &title, "UTF-8", &quest, &answero, &def) )
-return( NULL );
+    if ( !PyArg_ParseTuple(args,"ssO|OO", &title, &quest, &answero, &defo, &multipleo) )
+        return( NULL );
+
+    multiple = multipleo == Py_True;
```
skef: If we were python-3-only we could use `p` with a C int  here instead of `O` for the booleans. They didn't backport `p` to Python 2, unfortunately. 

However, what we care about is not whether these various booleans are identical to `Py_True`, we care if they are *true* in the Python sense. So replace `multipleo == Py_True` with `PyObject_IsTrue(multipleo)`.

fontforge/python.c
```diff
@@ -1735,50 +1735,100 @@ return( Py_BuildValue("i",ret));
...
+        }
+        int dcnt = PySequence_Size(defo);
+        if (dcnt != cnt) {
+            PyErr_Format(PyExc_ValueError, "Expected tuple/list of %d items, got %d items", cnt, dcnt );
+            PyMem_Free(title); PyMem_Free(quest); free(sel);
+            return( NULL );
+        }
+        for (int i = 0; i < cnt; i++) {
+            PyObject* temp = PyTuple_GetItem(defo, i);
+            sel[i] = temp == Py_True ? (char)1 : (char)0;
```
skef: `PyObject_IsTrue()` here as well.

fontforge/python.c
```diff
@@ -1735,50 +1735,100 @@ return( Py_BuildValue("i",ret));
...
     cnt = PySequence_Size(answero);
-    answers = calloc(cnt+1, sizeof(char *));
-    if ( def<0 || def>=cnt ) {
-	PyErr_Format(PyExc_ValueError, "Value out of bounds for 4th argument");
-	PyMem_Free(title);
-	PyMem_Free(quest);
-return( NULL );
+    char* sel = calloc(cnt, sizeof(char));
+
+    if (defo != NULL && defo != Py_None)
```
skef: Not a fan of bracketless conditionals that guard large blocks of code.

fontforge/python.c
```diff
@@ -1735,50 +1735,100 @@ return( Py_BuildValue("i",ret));
...
+    if (multiple) {
```
skef: The default specification seems overly restrictive to me, in the sense that if you can accept multiple answers but only want to specify one default, it seems like you should be able to do that with an integer. 

That is, why can't the logic be: "if default is a tuple then`multiple` should be true, otherwise it should be an integer"?

fontforge/python.c
```diff
@@ -1735,50 +1735,100 @@ return( Py_BuildValue("i",ret));
...
-    PyMem_Free(quest);
+    PyObject* reto;
+    if (multiple) {
+        char* buts2[2] = {_("OK"), _("Cancel")};
+        reto = PyTuple_New(cnt);
+        ret = ff_choose_multiple(title,(const char **) answers,sel,cnt,buts2,quest);
+        for (int i = 0; i < cnt; i++) {
+            char o = sel[i];
+            PyTuple_SetItem(reto, i, o == 1 ? Py_True : Py_False);
+            Py_INCREF(o == 1 ? Py_True : Py_False); // prevents crash on finalize
```
skef: Weird that this is needed but the docs bear it out. 

I think here I would do `PyObject o = (sel[i] == 1) ? PyTrue : Py_False;` and then just use `o` in the following two lines.

jtanx:
> I'm going to start with an optional suggestion, which is that now that we have two optional arguments it would be good to throw some keywords onto them so that they can be specified either by position or by keyword. It's entirely backward compatible with the existing interface and not much more work -- the current PyFFFont_autoWidth() provides a good model of the semantic.

100% agree with this, something that's been irking me is how we overload the signature of various functions, such that which *positional* arguments are valid depend on the context - something which you would almost never do if you were to write the interface in pure Python.

fontforge/python.c
jtanx: > If we were python-3-only we could use p with a C int here instead of O for the booleans. They didn't backport p to Python 2, unfortunately.

I think we're almost at the point where we should just drop Python 2 support entirely (not that the build system currently supports it anyway), which would allow us that.

skef: I do not personally object -- at all -- to dropping python 2 support soon, or even now. However, I have the sense that in the near term we'll still face significant pushback on that move.

frank-trampe: Yes. Let's wait a bit more before we burn that bridge. I find it encouraging, though, that there has been no complaint about crossing it.

fontforge/python.c
ctrlcctrlv: So, drop the `multiple` argument and infer it from the type of `default`? There are cases where you'd want default to be None but multiple selectable choices though.

skef: No, leave `multiple` and continue to reject passing a tuple for the defaults when `multiple` isn't true. But allow a caller to specify a single default with an integer when `multiple` is true if they feel like it. Just treat that as equivalent to passing the boolean tuple with one true value. 

This is tricky because that's definitely the wrong way to treat the return value -- when `multiple` is true you always want to return a tuple. Otherwise the caller would have to deal with annoying special cases when the user selects a single value. But input parameters are up to the caller and it's better not to force the author to build a tuple structure if they don't need to.

ctrlcctrlv: Understood, thanks :) will do it when I get home from church

fontforge/python.c
```diff
@@ -1734,51 +1734,105 @@ return( NULL );
...
+            def = (int)PyLong_AsLong(defo);
+            if ( def<0 || def>=cnt ) {
+                PyErr_Format(PyExc_ValueError, "Value out of bounds for 4th argument");
+                PyMem_Free(title); PyMem_Free(quest); PyMem_Free(sel);
+                return( NULL );
+            }
+            if (multiple) {
+                sel[def] = (char)1;
+            }
+        } else if ( PyTuple_Check(defo) ) {
```
skef: I think it's fine to also allow the bool-tuple method for specifying the value in the single choice case. But if you do that you should verify that there is only one true value if `multiple` is not true.

doc/html/python.html
```diff
@@ -836,9 +836,9 @@ <H3>
 	Throws an exception if there is no user interface.</TD>
     </TR>
     <TR>
-      <TD><CODE>askChoice</CODE>s</TD>
+      <TD><CODE>askChoices</CODE></TD>
       <TD><CODE>(title,question,answers,<BR>
-	[def])</CODE></TD>
+	[default,multiple])</CODE></TD>
```
skef: Need to document the keywords for the optional arguments somewhere here. (No need to do that for the mandatory ones.)

ctrlcctrlv: How would you do that? Put equals signs after them? Then it's not clear they're also possibly positional...

fontforge/python.c
ctrlcctrlv: https://github.com/fontforge/fontforge/pull/4143/commits/dd31daf0eac65bfdfa5df5210a46292a97aa0080 :)

doc/html/python.html
skef: You could follow the pattern similar to `addContextualSubtable()` and then note in the description that those parameters can also be passed by position if desired.

ctrlcctrlv: Sure, why not :)

https://github.com/fontforge/fontforge/pull/4143/commits/7d493b04a0905c76ec55c304aa2ff80015c3490c

PR #4146, opened 2020-02-09
Make open font dialog a toplevel window

### LINKED ISSUE #4145: Open Font dialog doesn't appear in taskbar
Only the splash screen window does (if enabled). This makes it difficult to get the window back on KDE, especially if the splash screen disappears before you open a font, and you tab away from the window. I have a fix for this, this is a tracking issue.

ctrlcctrlv:
If the splash screen window is worthy of this, surely the open font
dialog is. This closes #4145.

### Type of change

- **Bug fix**

### DIFF
--- a/fontforgeexe/fontview.c
+++ b/fontforgeexe/fontview.c
@@ -1016,19 +1016,6 @@ void MenuExit(GWindow UNUSED(base), struct gmenuitem *UNUSED(mi), GEvent *e) {
 	DelayEvent(_MenuExit,NULL);
 }
 
-char *GetPostScriptFontName(char *dir, bool mult, bool modal) {
-    unichar_t *ret;
-    char *u_dir;
-    char *temp;
-
-    u_dir = def2utf8_copy(dir);
-    ret = FVOpenFont(_("Open Font"), u_dir, mult, modal);
-    temp = u2def_copy(ret);
-
-    free(ret);
-return( temp );
-}
-
 void MergeKernInfo(SplineFont *sf,EncMap *map) {
 #ifndef __Mac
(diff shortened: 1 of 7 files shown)


skef:
Because this dialog is usually modal, shouldn't this flag depend on whether this is the initial dialog or some subsequent open-second-font scenario?

ctrlcctrlv:
I thought of that, but I really don't think it's necessary. The dialog does not really need to be modal since it opens a new font, it won't suddenly change an already open font.

ctrlcctrlv:
(Of course, in that case, there is also a definite argument for removing `wam_restrict` as well.)

fontforgeexe/fontview.c
```diff
@@ -1016,13 +1016,13 @@ void MenuExit(GWindow UNUSED(base), struct gmenuitem *UNUSED(mi), GEvent *e) {
 	DelayEvent(_MenuExit,NULL);
 }
 
-char *GetPostScriptFontName(char *dir, int mult) {
+char *GetPostScriptFontName(char *dir, bool mult, bool modal) {
```
skef: If you have an inkling to be truly heroic, you could:

1. Move `GetPostScriptFontName()` from this file to `openfontdlg.c`.
2. Remove the `FVOpenFont()` entry from `views.h`
3. Make `FVOpenFont()` static. 

This is how the code should be arranged given that `GetPostScriptFontName()` is basically a wrapper and the only place `FVOpenFont()` is called.

ctrlcctrlv: Thanks. I do rather like being heroic, (don't we all?), and if I don't do it it might not get done for another ten years, so I did it. Once CI passes I'll merge

PR #4150, opened 2020-02-10
Make PyFF_Glyph_get_font return a font if menu callback (Python)

### LINKED ISSUE #4149: Impossible to access font via glyph in Python menu callback
Title says it all. This is a tracking issue. I've already fixed this issue and a PR is on its way.

ctrlcctrlv:
To aid reviewers, this script shows the problem:

```python3
import fontforge

def issue4149(data, glyph):
    print(repr(glyph.font)) # Should not be `None` according to docs, yet is!

d = {}

fontforge.registerMenuItem(issue4149, None, d, "Glyph", None, "Issue №4149")
```

Open any glyph, and in that glyph's CharView, go to &laquo;Tools&rarr;Issue №4149&raquo;. Voil&agrave;.

This closes #4149 and is currently a blocker on #703.

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -299,9 +299,14 @@ PyObject* PyFF_FontForFV( FontViewBase *fv ) {
     if ( fv->python_fv_object==NULL ) {
         fv->python_fv_object = PyFF_FontType.tp_alloc(&PyFF_FontType,0);
         ((PyFF_Font *) (fv->python_fv_object))->fv = fv;
-        Py_INCREF( (PyObject *) (fv->python_fv_object) );	/* for the pointer in my fv */
+        Py_INCREF( (PyObject *) (fv->python_fv_object) );
     }
     return fv->python_fv_object;
+}
+PyObject* PyFF_FontForFV_I( FontViewBase *fv ) {
+    PyObject* pyfv = PyFF_FontForFV(fv);
+    Py_XINCREF(pyfv);
+    return pyfv;
 }
 static PyFF_Font* PyFF_FontForSF( SplineFont *sf ) {
     if ( sf==NULL )
@@ -1150,7 +1155,7 @@ static PyObject *PyFF_FontTuple(PyObject *UNUSED(self), PyObject *UNUSED(args)) 
     for ( fv=FontViewFirst(), cnt=0; fv!=NULL; fv=fv->next, ++cnt );
     tuple = PyTuple_New(cnt);
     for ( fv=FontViewFirst(), cnt=0; fv!=NULL; fv=fv->next, ++cnt )
-	PyTuple_SET_ITEM(tuple,cnt,PyFF_FontForFV(fv));
+	PyTuple_SET_ITEM(tuple,cnt,PyFF_FontForFV_I(fv));
 
 return( tuple );
 }
@@ -1160,7 +1165,7 @@ static PyObject *PyFF_ActiveFont(PyObject *UNUSED(self), PyObject *UNUSED(args))
     if ( fv_active_in_ui==NULL )
 Py_RETURN_NONE;
 
-return( PyFF_FontForFV( fv_active_in_ui ));
+return( PyFF_FontForFV_I( fv_active_in_ui ));
 }
 
 static PyObject *PyFF_ActiveGlyph(PyObject *UNUSED(self), PyObject *UNUSED(args)) {
@@ -1208,7 +1213,7 @@ return( NULL );
 return( NULL );
     }
     free(locfilename);
-return( PyFF_FontForFV( SFAdd( sf, openflags&of_hidewindow )));
+return( PyFF_FontForFV_I( SFAdd( sf, openflags&of_hidewindow )));
 }
 
 static PyObject *PyFF_FontsInFile(PyObject *UNUSED(self), PyObject *args) {
@@ -10047,10 +10052,9 @@ return( STRING_FROM_FORMAT( "<Selection for %s>", self->fv->sf->fontname ));
 }
 
 static PyObject *PyFFSelection_get_font(PyFF_Selection *self, void *UNUSED(closure)) {
-    PyObject* font = PyFF_FontForFV( self->fv );
+    PyObject* font = PyFF_FontForFV_I( self->fv );
     if ( font==NULL )
 Py_RETURN_NONE;
-    Py_INCREF(font);
     return font;
 }
 
@@ -16777,7 +16781,7 @@ return( NULL );
     if ( sf->fv==NULL )
 	EncMapFree(sf->map);
     newfv = SFAdd(InterpolateFont(fv->sf,sf,fraction, fv->map->enc ),false);
-return( PyFF_FontForFV(newfv));
+return( PyFF_FontForFV_I(newfv));
 }
 
 static PyObject *PyFFFont_CreateMappedChar(PyFF_Font *self, PyObject *args) {
@@ -18461,7 +18465,7 @@ return( (PyObject *) (all->python_data) );
 }
 
 static PyObject *PyFF_AWContext_getFont(PyFF_AWContext *self, void *UNUSED(closure)) {
-return( PyFF_FontForFV( self->base->fv ));
+    return PyFF_FontForFV( self->base->fv );
 }
 
 static PyObject *PyFF_AWContext_getEmSize(PyFF_AWContext *self, void *UNUSED(closure)) {
@@ -19698,7 +19702,7 @@ return;
     for ( pt=argtypes, i=0; *pt; ++pt, ++i ) {
 	PyObject *arg;
 	if ( *pt=='f' )
-	    arg = PyFF_FontForFV( va_arg(ap,FontViewBase *));
+	    arg = PyFF_FontForFV_I( va_arg(ap,FontViewBase *));
 	else if ( *pt=='g' )
 	    arg = PySC_From_SC_I( va_arg(ap,SplineChar *));
 	else if ( *pt=='s' )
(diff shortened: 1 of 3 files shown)


fontforgeexe/pythonui.c
```diff
@@ -144,6 +144,13 @@ return;
 static void cvpy_menuactivate(GWindow gw,struct gmenuitem *mi,GEvent *e) {
     CharView *cv = (CharView *) GDrawGetUserData(gw);
     PyObject *pysc = PySC_From_SC(cv->b.sc);
+    // The following two lines look like a noop but are very important. They
+    // set python_fv_object on the fv if it's not there, ensuring it's possible
+    // to do `glyph.font` on glyphs used as arguments in menus.  The pointer
+    // `fv` refers to is also accessible via cv->b.sc, as cv->b.sc->parent->fv,
+    // which is done (eventually) by PyFF_Glyph_get_font.
+    FontViewBase *fv = cv->b.fv;
+    PyObject *pyfv = PyFV_From_FV(fv);
```
skef: With "stronger" compiler warning flags this is likely to add an "assigned but not used" warning.

skef:
Here is one thing I don't yet understand about the code paths this change intervenes in. There are two interfaces, `PyFV_From_FV()` and `PyFF_FontForFV()`. Both take a `FontViewBase *` as an argument and return a (PyFF_Font *),  although with the former it is downcast to the more generic `PyObject *`. 

More relevantly, `PyFV_From_FV()` will go ahead and create an object if it's not there while `PyFF_FontForFV()` will shrug and hand you a NULL. The latter is why the current code is failing. But `PyFF_FontForFV()` obviously has the `fv`, and therefore everything it needs to populate `python_fv_object`. 

Looking at the code I don't yet understand the rationale for the split. I can *imagine* reasons one would want to do this -- mostly having to do with reference counting -- but the code doesn't seem to bear them out.

fontforgeexe/pythonui.c
ctrlcctrlv: I can remove the assignments, basically rewriting both lines as one line. Does that help?

ctrlcctrlv:
Perhaps the change should be made at `PySC_From_SC`? But that function is used in a lot of places, so I was hesitant.

skef:
Happily `python_fv_object` is only used in a small number of places, and is more or less always "handed out" via those two functions. `PyFF_Font_dealloc()` and `PyFFFont_close` clear the value, `PyFF_Font_new()` sets to `self`, `PyFF_FreeFV()` does some more subtle reference counting stuff, but that's it. The question is why there are two accessors. 

I suppose that it could be as simple as `PyFF_FontForFV()` being meant for intra-python use (it's static) while `PyFV_From_FV()` is for more general cases. But the only actual use of it in a different file is in `pythonui.c`, which isn't exactly exotic. 

The only reason I can see for having the two functions is if you want or need NULL to be returned in some cases when `fv` isn't NULL. If that isn't important we should just allocate in `PyFF_FontForFV()` and call it instead of `PyFV_From_FV()`. So that's the question: Why hand out NULL?

ctrlcctrlv:
Is that what really happens though? `PyFF_FontForFV` only checks for the parameter `fv` being NULL, while `PyFV_From_FV` again, checks `fv` for NULL, but if it's not, creates Python data. So, where do we get NULL when `(FontViewBase*)fv` is not itself NULL?

skef:
@ctrlcctrlv I'm not sure about all the cases, but one case pretty much has to be the one you're fixing in this PR. The side effect of your calling `PyFV_From_FV` is the setting of `python_fv_object`; it doesn't do anything else. So presumably without it `PyFF_FontForSC()` is calling `PyFF_FontForSF()` is calling `PyFF_FontForFV()` which is handing out a NULL `fv->python_fv_object`;

ctrlcctrlv:
Oh, I see what you mean...but that's a bug. We shouldn't ever _want_ this to happen. The bugs are pretty unambiguous that glyphs always have associated fonts.

I guess my concern is that `PyFF_FontForSF` and `PyFF_FontForSC` have similar semantics, and I won't be changing those. But then again, maybe that makes sense because `PyFF_FontForFV` is at the end of the chain. Then I could just drop the `PyFV_From_FV` function as no longer needed, using `PyFF_FontForFV` everywhere it used to be used.

skef:
Right, we don't want it to happen then. If we never want it to happen we can just collapse the two functions together and the need for your fix goes away because it's allocated "on demand". If we don't want to hand out references to closed fonts we could call `CheckIfFontClosed()`, although that probably isn't necessary.

So when *do* we want it to happen? Is the current code an accident or is there some reason for it? What is your suggested fix working around?

ctrlcctrlv:
The only reason I can think of for it is dummy glyphs. You're more familiar with the Python side of things than I am. Is it possible to create a dummy glyph from Python? We do it often in the C code but we never expose them.

My understanding of what I'm working around is that `PySC_From_SC` isn't setting the data on its parent FontView. We're not doing it the normal way through `PyFF_FontIndex`, which implements e.g. `font["A"]` to give us the glyph with unicode encoding `(char)'A'`, but rather calling `PySC_From_SC` in a context where the font hasn't already been initialized. I think this was just a simple oversight.

skef:
Python won't *make* a dummy glyph but I suppose one created somewhere else might "leak out" or interact. 

I know that dummy glyphs sometimes have a dummy `splinefont` parent. I think one having a view would be rarer, so the chain would almost always end with a NULL `fv`.

ctrlcctrlv:
OK, I'm going to push a version of this patch which collapses the two functions into one.

fontforge/python.c
```diff
@@ -1162,7 +1150,7 @@ static PyObject *PyFF_FontTuple(PyObject *UNUSED(self), PyObject *UNUSED(args))
     for ( fv=FontViewFirst(), cnt=0; fv!=NULL; fv=fv->next, ++cnt );
     tuple = PyTuple_New(cnt);
     for ( fv=FontViewFirst(), cnt=0; fv!=NULL; fv=fv->next, ++cnt )
-	PyTuple_SET_ITEM(tuple,cnt,PyFV_From_FV_I(fv));
+	PyTuple_SET_ITEM(tuple,cnt,PyFF_FontForFV(fv));
```
skef: Unfortunately `PyFV_From_FV_I()` is a convenience routine that also increments the reference count (presumably what the "I" is for), so either we still need that convenience or calls to it need to be replaced with an explicit increment (the former is probably easier). 

However, it can just call `PyFF_FontForFV` now, or we can just call it `PyFF_FontForFV_I` and put it next to `PyFF_FontForFV`, since you've already done the other relevant editing anyway.

ctrlcctrlv: Oh right, this is the same reason we needed to increment `Py_True`. Thanks, I'll do it.

ctrlcctrlv: Um, @skef, why can't `PyFF_FontForFV` just do the increment, and we decrement where it's not needed? Incrementing is the more common case.

ctrlcctrlv: (I pushed a version where `PyFF_FontForFV` always increments.)

skef: There are two separate issues here: the calling convention and the reference counts. 

I'm a little uncomfortable with this convention on a "that's not how this is typically done" basis, but I'm not sure I can make a principled argument. So I guess it's a preference. 

However, it is important to preserve the counts -- otherwise there will be (additional?) memory leaks. So if you change `PyFF_FontForFV()` to always increment, you have to change all the other callers to decrement. You can't (or shouldn't) just bump up the counts arbitrarily.

ctrlcctrlv: There are three places we shouldn't increment, so I've decremented there as appropriate. :)

fontforge/python.c
```diff
@@ -293,15 +293,20 @@ static PyFF_Layer *LayerFromLayer(Layer *,PyFF_Layer *);
...
+        ((PyFF_Font *) (fv->python_fv_object))->fv = fv;
+    }
+    Py_INCREF( (PyObject *) (fv->python_fv_object) );
     return fv->python_fv_object;
 }
 static PyFF_Font* PyFF_FontForSF( SplineFont *sf ) {
     if ( sf==NULL )
 	return NULL;
-    return PyFF_FontForFV( sf->fv );
+    return (PyFF_Font*)PyFF_FontForFV( sf->fv );
```
skef: @ctrlcctrlv You're now decrementing in the methods you've *changed*, but you haven't taken into account the other already present callers of `PyFF_FontForFV()` -- for example this one.

ctrlcctrlv: Oh yeah :) OK maybe we do need a `PyFF_FontForFV_I`

fontforge/python.c
```diff
@@ -293,15 +293,20 @@ static PyFF_Layer *LayerFromLayer(Layer *,PyFF_Layer *);
...
 /* ************************************************************************** */
-static PyFF_Font* PyFF_FontForFV( FontViewBase *fv ) {
+PyObject* PyFF_FontForFV( FontViewBase *fv ) {
     if ( fv==NULL )
 	return NULL;
+    if ( fv->python_fv_object==NULL ) {
+        fv->python_fv_object = PyFF_FontType.tp_alloc(&PyFF_FontType,0);
+        ((PyFF_Font *) (fv->python_fv_object))->fv = fv;
+    }
+    Py_INCREF( (PyObject *) (fv->python_fv_object) );
```
skef: Also, if you're going to take this approach you can't just move the increment, you'll have to increment *twice* on the creation path, once inside the NULL guard and once outside. Otherwise you'll be down a reference when the caller is expecting an increment. 

This is part of why the traditional approach may make more sense. Reference counting can get hairy.

PR #4152, opened 2020-02-11
Fix for #2421 Fails to import .svg

### LINKED ISSUE #2421: Fails to import .svg on latest Windows buld
this one: http://www.shapeoko.com/wiki/index.php/File:Tool_calipers_locking.svg

skef:
`SplineSetStroke()` strokes the full list of `SplineSet`s it is passed. This was true before the recent changes and remains true. 

The code being modified by this PR iterated through the SplineSets stroking each, making an initial copy for transformation with `SplinePointListCopy()` rather than `SplinePointListCopy1()`. As a result the output had one copy of the first stroked contour, two of the second, three of the third, etc. Add up enough SplineSets and Cairo is overwhelmed, hence the "hang" in #2421. 

The code wouldn't have been corrected by `SplinePointListCopy1()` because the loop was in the wrong place, so I just removed the inner loop as a cleaner solution. 

Closes #2421

### DIFF
--- a/fontforge/psread.c
+++ b/fontforge/psread.c
@@ -2890,7 +2890,7 @@ return( last );
 
 static SplinePointList *SplinesFromLayers(SplineChar *sc,int *flags, int tostroke) {
     int layer;
-    SplinePointList *head=NULL, *last, *new, *nlast, *temp, *each, *transed;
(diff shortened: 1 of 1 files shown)


skef:
Strictly speaking the hang at the time #2421 was filed was probably in the stroke code. If only the new code were slower the profiler might have been of some help ...

PR #4174, opened 2020-02-18
Prevent another UBSan false positive

ctrlcctrlv:
When parsing a TTF, this would appear:

../fontforge/mem.c:90:24: runtime error: left shift of 211 by 24 places cannot be represented in type 'int'

It's a false positive, not really a bug. I removed an unneeded function in the process of this, so it removes lines overall.

### Type of change

- **Non-breaking change**

### DIFF
--- a/fontforge/mem.c
+++ b/fontforge/mem.c
@@ -79,16 +79,21 @@ int get3byte(FILE *ttf) {
 }
 
 int32 getlong(FILE *ttf) {
-	int ch1 = getc(ttf);
-	int ch2 = getc(ttf);
-	int ch3 = getc(ttf);
-	int ch4 = getc(ttf);
+	unsigned int ch1 = getc(ttf);
+	unsigned int ch2 = getc(ttf);
+	unsigned int ch3 = getc(ttf);
+	unsigned int ch4 = getc(ttf);
 
 	if (ch4==EOF)
 		return EOF;
 
 	return (ch1<<24)|(ch2<<16)|(ch3<<8)|ch4;
 }
+
+uint32 getuint32(FILE *ttf) {
+    return (uint32)getlong(ttf);
+}
+
 
 real getfixed(FILE *ttf) {
 	int32 val = getlong(ttf);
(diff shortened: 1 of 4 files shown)


fontforge/parsettf.c
```diff
@@ -782,7 +782,7 @@ return;
 
     for ( i=0; i<info->numtables; ++i ) {
 	tabs[i].tag = getlong(ttf);
-	tabs[i].checksum = getlong(ttf);
+	tabs[i].checksum = getuint32(ttf);
```
ctrlcctrlv: This was already being silently converted to `uint32`, so that's why I don't also change the struct `tt_tables`.

fontforge/mem.c
```diff
@@ -79,17 +79,22 @@ int get3byte(FILE *ttf) {
 }
 
 int32 getlong(FILE *ttf) {
-	int ch1 = getc(ttf);
-	int ch2 = getc(ttf);
-	int ch3 = getc(ttf);
-	int ch4 = getc(ttf);
+	unsigned int ch1 = getc(ttf);
```
jtanx: Hold up, these changes look suspicious to me. This for sure would raise a signed-unsigned conversion warning if we actually bothered to turn up the warning levels.

ubsan in a sense is likely correct anyway, as we're not checking for EOFs on all the bytes.

ctrlcctrlv: So they should all be checked for EOF? That makes sense.

jtanx: I took a closer look at this and the EOF check at the end is fine, because if you don't get an EOF there, that implies that all the other reads succeeded (assuming a sane implementation of getc). 

However, both with and without these changes, there is still UB (or rather, implementation defined behaviour): https://stackoverflow.com/questions/13636397/unsigned-to-signed-conversion-in-c. It's just in this case, ubsan isn't complaining.

gcc does what the code (as written) expects it to, hence why nothing has fallen over.

ctrlcctrlv:
Voluntarily withdrawn.

PR #4179, opened 2020-02-21
Write out note as normal text element rather than dictionary

### LINKED ISSUE #3964: UFO import not handling note key correctly
The import UFO functionality should have the same behavior as the export regarding the mapping of fields. So the SFD field `Comment` is mapped to the UFO note `key` in the `dict` tag during the export, but not during import.

### FontForge version
Version: 20190801
OS: MacOS 10.14.6

### Expected behavior

If I import of UFO v2 file such as this one [here](https://github.com/eduardomourar/fiati/blob/763db1d11aa6fc0738b150fe6cf373785f43d2fd/sources/fiati-regular/fiati-regular.ufo/glyphs/uniE_C_43.glif#L34), I would expect the SFD file result to contain a `Comment` with the note value.

```
Comment: "fingeringFluteLHThumbBFlatClosed"
```

### Actual Behavior
But instead the note gets translated into a `PickledDataWithLists` for the glyph shown [here](https://github.com/eduardomourar/fiati/blob/763db1d11aa6fc0738b150fe6cf373785f43d2fd/sources/fiati-regular/fiati-regular.sfd#L1445).

```
PickledDataWithLists: "(dp0
Vnote
p1
VfingeringFluteLHThumbBFlatClosed
p2
s."
```

### Workaround

Modify SFD file after import by replacing the string from PickledDataWithLists to Comment.

skef:
The code to read a UFO 3 "note" element into the SplineChar comment field seems to be correct. The old code to write it out was using a key/value pair format rather than a simple text element. It appears that the glif `note` field is the latter. 

Closes #3964

### DIFF
--- a/fontforge/ufo.c
+++ b/fontforge/ufo.c
@@ -923,7 +923,8 @@ xmlNodePtr _GlifToXML(const SplineChar *sc, int layer, int version) {
 			    // "<anchor x=\"%g\" y=\"%g\" name=\"%s%s\"/>" ap->me.x ap->me.y ap->anchor->name
 			}
 	}
-    if (sc->comment) PListAddString(topglyphxml, "note", sc->comment);
+    if (sc->comment)
+	xmlNewChild(topglyphxml, NULL, BAD_CAST "note", BAD_CAST sc->comment);
     if ( sc->layers[layer].refs!=NULL || sc->layers[layer].splines!=NULL ) {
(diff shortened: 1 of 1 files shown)


skef:
Note that this will continue to discard the `note` when importing a glif into an individual character. This is probably the correct decision given the current code (one imports into a layer) but maybe there should be a "replace character" import option.

frank-trampe:
I agree that the status quo seems otherwise fine. We'll see whether anybody asks for more.

PR #4181, opened 2020-02-22
Fix for #3833 Incorrect actions of charview Tex/Math guess buttons

### LINKED ISSUE #3833: Incorrect actions of Guess-buttons
FontForge version 20190413 based on sources from 26-Jul-2019 git:279e69163e4da2be6240d1b35fdbb6438e792344
MacOSX 10.11.6

There seems to be a bug in the functions of the "Guess" buttons. They don't act on their own rows. => see screenshot

Go to Element > Glyph Info > TeX&Math
The 1st "Guess" button does nothing, it does not fill the first field "Height".
The 2nd "Guess" button fills the first line "Height".
The 3rd "Guess" button fills the second line "Depth".
The 4th "Guess" button fills the third line "Italic Correction".
There is no way to fill the last field "Top Accent Pos" with a button.

Glyph-Info-TeX

skef:
Closes #3833

### DIFF
--- a/fontforgeexe/charinfo.c
+++ b/fontforgeexe/charinfo.c
@@ -3539,7 +3539,7 @@ return( false );
 static int TeX_Default(GGadget *g, GEvent *e) {
     if ( e->type==et_controlevent && e->u.control.subtype == et_buttonactivate ) {
 	CharInfo *ci = GDrawGetUserData(GGadgetGetWindow(g));
-	int cid = GGadgetGetCid(g);
+	int cid = GGadgetGetCid(g), mcid = 0;
 	DBounds b;
 	int value;
 	SplineChar *basesc = NULL;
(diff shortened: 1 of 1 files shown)


fontforgeexe/charinfo.c
```diff
@@ -3573,31 +3573,39 @@ static int TeX_Default(GGadget *g, GEvent *e) {
...
 		value = rint(b.minx + (b.maxx-b.minx)/2);
 	    else
 		value = rint(b.minx + italic_off + (b.maxx - b.minx - italic_off)/2);
-	} else if ( (style&sf_italic) || sf->italicangle!=0 ) {
-	    value = rint((b.maxx-ci->sc->width) +
-			    (sf->ascent+sf->descent)/16.0);
-	} else
-	    value = 0;
-	sprintf( buf, "%d", value );
-	GGadgetSetTitle8(GWidgetGetControl(ci->gw,cid-5),buf);
```
jtanx: So it was effectively an off by one error here?

skef: Specifically -4, I think, but yes. 

I feel like one can be this clever if one gets it right (and it stays right), otherwise the explicit tokens go in.

PR #4183, opened 2020-02-22
Reorder cap and join enums to match PostScript

skef:
Turns out these enums were meant to match the `setlinejoin` and `setlinecap` PostScript commands. In both cases it is the first three values that are relevant. (Nothing else cares about the ordering that I know of.)


skef:
This is a small thing that I screwed up with the Expand Stroke code that I'm going to go ahead and merge in order to improve the #4191 debugging situation. I guess I can't claim it's completely low-risk to reorder enums given the need to, but I am putting them back in the order they started ...

PR #4189, opened 2020-02-24
More SEGV protection for extreme splineoverlap cases

### LINKED ISSUE #2832: "PS Interpretion" dialog may be causing Python script to hang

skef:
The last bits of #2832 . Note that you'll need to test with a release build or by commenting out the `NibIsValid` assertion. 

The test file is extreme in that it uses a stroke width of 0.067. Neither the Expand Stroke facility nor many other subsystems are designed to work well at such small fractions of an em-unit. So the goal in this case is not so much to produce good output as to avoid crashing. (I did try magnifying the file by 10x in Inkscape and reimporting and it looks OK. 30x would probably have worked better.)  

Closes #2832

### DIFF
--- a/fontforge/splineoverlap.c
+++ b/fontforge/splineoverlap.c
@@ -313,7 +313,7 @@ static int SSRmNullSplines(SplineSet *spl) {
 		/*  and itself where the spline covers no area (the two cps */
 		/*  point in the same direction) */
 		BpSame(&s->from->me,&s->to->me)) {
-	    if ( next==s )
+	    if ( next==s || next==NULL )
 return( true );
 	    if ( next->from->selected ) s->from->selected = true;
 	    s->from->next = next;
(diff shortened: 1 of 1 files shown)


skef:
@frank-trampe take a quick look at this one too if you get a chance. It's short.

frank-trampe:
Sorry for the delay. I never worked on this function, so I needed to read it carefully to make sure that I understood what it was doing.

PR #4204, opened 2020-03-03
Add glyph-level Import and Export parameters

### LINKED ISSUE #4153: Address accuracy in stroked SVG (and PS?) imports
The "1" numerals in the file referenced by #2421 don't import very well because they get "over-Simplified". That is due to the fact that the import process uses the default stroke accuracy value. 

There are two issues with that value. One is that a user might want it higher not considering transformations. The second is that SVGs have internal transformations and we don't correct for them. We should probably consider a preference value for the output accuracy in em-units and then track the scale up and down through the transformations to maintain output-relative accuracy.

### LINKED ISSUE #4198: Advice/Confirmation on Import/Export options
I'm in the process of adding some infrastructure to support import and export options. I'm opening this issue to summarize the preliminary design and, with luck, get some feedback. (I would wait for the PR but the release timing may call for earlier discussion.)

Note that this isn't about the specific options, just the infrastructure to support them. 

First, `fontforge/sd.h` will get these additions:

```
typedef struct importparams {
    int initialized;
    int shown_svg, shown_eps, shown_scale, show_always;

    /* omitted options */
} ImportParams;

typedef struct exportparams {
    int initialized;
    int shown_svg, show_always;

    /* omitted options */
} ExportParams;

extern void InitImportParams(ImportParams *ip);
extern ImportParams *ImportParamsState(void);
extern void InitExportParams(ExportParams *ep);
extern ExportParams *ExportParamsState(void);
```

The `Init` functions will set up the base-level defaults and the `State` functions will have `static` members that represent the in-dialog settings. 

`stroke_flags` in `uiinterface.h` will be replaced by
```
    void (*import_dlg)(ImportParams *ip);
    void (*export_dlg)(ExportParams *ep);
```
The non-UI versions of these will do nothing. The UI versions will display dialogs for setting the values.

The options dialogs will always be the same and have all the options for every import/export type. The chicken/egg problems are too annoying for format-specific dialogs and some options are multi-format anyway. 

Each dialog will be force-raised once for each new applicable filetype. (So, for example, if there is an SVG-specific setting and the dialog has not already been raised for that type, it will come up after hitting "OK" in the respective (Import or Export) file dialog.) The user can then choose whether to see the dialog on every import or not. The options dialog can also be raised with an "Options" button I'll add to the respective file dialogs. 

The native scripting interface will have a means of setting each option but will not interact with the `State` statics. Python will have a `system_options` keyword where string "set" uses the passed parameters to set the `State` values and string "use" takes the values from the `State`.

Finally, I'll add the `Param` pointer to all of the import functions and to the SVG output function (as there is only one SVG output option to start with). 

All of this is taking longer than I had hoped but with luck I'll have a draft tomorrow evening.

skef:
This is mostly discussed in #4198 -- refer to that issue for the details. The added options are described in the modifications to the API docs.

At present whether or not we scale on input seems to be based on the primary use case for the format. In the future it would probably be best to take a uniform approach, also scaling EPS and Plate files when `scale` is true and not when it is false. However, I'm not going to do that work right now.

One can easily imagine additional useful parameters. 

FYI I did this work now to aid round-tripping SVGs through FontForge, in effect converting stroked paths to filled paths. That gets messy without being able to turn scaling off on the input side. 

Closes #4198 
Closes #4153


skef:
@jtanx when I run CMake on my system I get no docs targets and neither the CMake nor the Sphinx Wiki pages says a word about building or dependencies. Installing `docutils` did not appear to change the situation. Is the expectation that one should read the main Sphinx/RST docs to learn how to generate HTML versions of modified documentation? Is `help()` now broken for CMake-driven local installations?

jtanx:
Ah, right, will update the cmake wiki page.

If you're rebased to master, there's an option called ENABLE_DOCS, which defaults to auto. It's looking for the `sphinx-build` command, which you can install via the sphinx package. Alternatively you can enable the `SPHINX_USE_VENV` option which will make cmake create a venv and install sphinx into that, which then makes it use that.

skef:
Thanks, installing `python-sphinx` on Arch Linux did add the target and build the docs automatically. 

The versions generated with the changes in this push (after a small fix) seem sufficiently consistent with the surrounding HTML.

fontforge/psread.c
```diff
@@ -2520,14 +2531,13 @@ static void _InterpretPS(IO *wrapper, EntityChar *ec, RetStack *rs) {
...
+		stack[sp].u.val = (miterlimit==JLIMIT_INHERITED?10.0:miterlimit);
 		stack[sp++].type = ps_num;
 	    }
 	  break;
 	  case pt_setmiterlimit:
-	    /* pop one item off stack */
-	    if ( sp>=1 )
-		--sp;
+	    if ( sp>=0 )
+		miterlimit = stack[--sp].u.val;
```
jtanx: This doesn't look right, if sp is 0 you're addressing -1 into the stack?

skef: Yeah that was just a typo. Fixed.

fontforge/python.c
```diff
@@ -8521,49 +8540,76 @@ return(NULL);
...
+    "simplify", "handle_clip", "handle_eraser", "scale", "accuracy",
+    "default_joinlimit", "usesystem", "asksystem", NULL };
+
+/* Legacy PostScript importing flags */
 static struct flaglist import_ps_flags[] = {
     { "toobigwarn", 0 },			/* Obsolete */
     { "removeoverlap", 0 },			/* Obsolete */
-    { "handle_eraser", sf_handle_eraser },
-    { "correctdir",  sf_correctdir },
+    { "handle_eraser", 1 },
```
jtanx: So these `psstrokeflags` aren't used any more? Can we remove them?

fontforge/python.c
```diff
@@ -8521,49 +8540,76 @@ return(NULL);
...
     PyObject *flags=NULL;
+    int psflags, use_system = false, ask_system = false;
+    bigreal jl_tmp = -1;
 
-    if ( !PyArg_ParseTuple(args,"es|O","UTF-8",&filename, &flags) )
-return( NULL );
+    InitImportParams(&ip);
+
+    if ( !PyArg_ParseTupleAndKeywords(args, keywds,
+                "es|$pppppddpp", glyph_import_keywords, "UTF-8", &filename,
```
jtanx: Can we avoid overloading this method? You more than likely wouldn't do this if you were writing this in pure python. How about:

```
es|O$pppppddpp
```

And check if flags is null or not?

fontforge/python.c
```diff
@@ -8620,9 +8686,21 @@ return( NULL );
 	format=2;
 
     if ( format!=-1 ) {
-	if ( !PyArg_ParseTuple(args,"O|ii",&foo,&pixels,&bits) ) {
-	    free(locfilename);
-return( NULL );
+	if ( obj1!=NULL ) {
```
jtanx: How is this equivalent to before? i.e. why did it parse out `bar` then never use it? A bug?

fontforge/scripting.c
```diff
@@ -2381,10 +2414,11 @@ static void bImport(Context *c) {
 	ok = FVImportMult(c->curfv,filename, back, bf_ttf);
     else if ( format==fv_pk )
 	ok = FVImportBDF(c->curfv,filename,true, back);
-    else if ( format==fv_image || format==fv_eps || format==fv_svg || format==fv_pdf )
-	ok = FVImportImages(c->curfv,filename,format,back,flags);
+    else if (    format==fv_image || format==fv_eps || format==fv_svg
```
jtanx: nit - not a fan of this style of putting the `||` on the next line and offsetting the first by that

fontforge/svg.c
```diff
@@ -1081,17 +1089,45 @@ int _ExportSVG(FILE *svg,SplineChar *sc,int layer) {
...
+	if (sc->parent!=NULL) {
+	    scc = SplineCharCopy(sc, sc->parent, NULL);
+	    FVTrans(scc->parent->fv,scc,trans,NULL,
+	            fvt_nopreserve|fvt_dontmovewidth|fvt_noupdate);
+	} else {
+	    // This is probably an isolated layer, so just transform the splines
+	    // there and back
+	    scc = sc;
+	    MatInverse(invtrans,trans);
+	    SplinePointListTransformExtended(scc->layers[layer].splines, trans,
```
jtanx: So we modify in place then reverse that? Are you sure this is absolutely idempotent, and not subject to fp error/instability?

fontforge/svg.c
```diff
@@ -3145,15 +3184,20 @@ return;
...
     int defh=0, defv=0;
     int has_font_face = false;
     xmlChar *name;
     SplineFont *sf;
     EncMap *map;
+    ImportParams *ip = ImportParamsState();
+    bigreal tmpjl = ip->default_joinlimit;
     int i;
 
+    if ( tmpjl==JLIMIT_INHERITED )
```
jtanx: I don't know how to feel about this approach, it seems brittle/easy to get wrong

fontforgeexe/cvexportdlg.c
```diff
@@ -55,6 +56,149 @@ struct sizebits {
...
+    return( true );
+}
+
+void _ExportParamsDlg(ExportParams *ep) {
+    GRect pos;
+    GWindow gw;
+    GWindowAttrs wattrs;
+    GGadgetCreateData gcd[5], boxes[4], *hvarray[5][4], *barray[4];
+    GTextInfo label[5];
+    int done = false;
```
jtanx: Why not make this `bool`

fontforgeexe/uiutil.c
```diff
@@ -692,6 +692,7 @@ struct ui_interface gdraw_ui_interface = {
 
     UI_TTFNameIds,
     UI_MSLangString,
-    (int (*)(void)) Ps_StrokeFlagsDlg
+    _ImportParamsDlg,
```
jtanx: Why are these underscored?

fontforgeexe/cvimportdlg.c
```diff
@@ -362,14 +470,38 @@ return( oldflags );
...
+    if ( err )
+	ip->accuracy_target = 0.25;
+    ip->show_always = GGadgetIsChecked(gcd[al_k].ret);
+
     GDrawDestroyWindow(gw);
-return( oldflags );
+}
+
+static void ShowImportOptions(ImportParams *ip, int shown, int type) {
+    if ( type==0 ) {
```
jtanx: Not a fan of these magic numbers

doc/sphinx/scripting/python/fontforge.rst
```diff
@@ -1207,11 +1207,28 @@ Layers may be compared to see if their contours are similar.
    Orients all contours so that external ones are clockwise and internal
    counter-clockwise. See also :meth:`contour.isClockwise()`.
 
-.. method:: layer.export(filename)
+.. method:: layer.export(filename[, KEYWORD])
```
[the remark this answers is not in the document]
skef: The "standard" syntax is mushes the non-positional argument issue.

fontforge/cvimages.c
```diff
@@ -44,11 +44,36 @@
...
 
-void SCAppendEntityLayers(SplineChar *sc, Entity *ent) {
+void InitImportParams(ImportParams *ip) {
+    assert( ip!=NULL );
+
+    memset(ip, 0, sizeof(ImportParams));
+
+    ip->initialized = true;
+
+    ip->correct_direction = no_windowing_ui ? true : false;
```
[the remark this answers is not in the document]
skef: I had thought this was an existing distinction I was preserving but I just went and looked again at the old `cvimportdlg.c` and it seems I was mistaken -- the initial default is `true` in both cases.

fontforge/python.c
skef: Apparently so

fontforge/python.c
skef: I don't think the interface should support simultaneous competing means of specifying the same flags and this is cleaner (and more comprehensive) than examining the keyword argument values to see if they've changed.

fontforge/python.c
skef: Before, `foo` and `bar` were just to make the initial `ParseTuple` work when used with both optional arguments to get the filename. Then that would be parsed and `ParseTuple` would be called again on `args` to extract the particular values depending on whether the format is raster or vector. 

Given that the layer can be specified as an int or a string and @ctrlcctrlv broke out the code to figure that out from a PyObject it seemed easier to just grab the two values as objects and examine them that way.

fontforge/scripting.c
skef: I find it makes the expressions much more readable.

fontforge/svg.c
skef: FontForge does this in other places even without `tpmask_dontTrimValues` (an option I added). It's seen in `effects.c`, for example. I guess I could sub-in a copy if that was preferred.

fontforge/svg.c
skef: I didn't want to have different miter/join limits for different formats, but still wanted it to be possible for each format to have its corresponding default when that was desired. This seemed like the most straightforward means of doing that.

fontforgeexe/cvexportdlg.c
skef: This code is copied from the import dialog case and I didn't want to change the `GetUserData` casts.

fontforgeexe/uiutil.c
skef: `ImportParamsDlg` and `ExportParamsDlg` are the defines that redirect through `ui_interface`. These are the UI-side implementations.

fontforgeexe/cvimportdlg.c
skef: I've since thought that a unified flag interface for both the state and the argument would be cleaner, but that can wait for a later straightening-up.

fontforge/svg.c
skef: I added the copy

fontforgeexe/cvimportdlg.c
skef: I did this anyway

fontforgeexe/cvimportdlg.c
```diff
@@ -456,48 +582,60 @@ return( true );
...
+	    else if ( format==fv_svg ) {
+		ShowImportOptions(ip, d->opts_shown, sp_svg);
+		ImportSVG(d->cv,temp,ip);
+	    } else if ( format==fv_glif )
+		ImportGlif(d->cv,temp,ip);
+	    else if ( format==fv_fig ) {
+		ImportFig(d->cv,temp,ip);
 #ifndef _NO_PYTHON
-	    else if ( format>=fv_pythonbase )
+	    } else if ( format>=fv_pythonbase )
```
jtanx: This bracket messes up compiling with python disabled.

fontforgeexe/cvexportdlg.c
```diff
@@ -191,12 +191,11 @@ void _ExportParamsDlg(ExportParams *ep) {
...
-    if ( type==2 ) {
-	if ( !shown && (!ep->shown_svg || ep->show_always) )
-	    _ExportParamsDlg(ep);
-	ep->shown_svg = true;
-    }
+static void ShowExportOptions(ExportParams *ep, int shown,
+                              enum shown_params type) {
+    if ( !shown && !(ep->shown_mask & type) )
+	_ExportParamsDlg(ep);
+    ep->shown_mask = (ep->shown_mask & ~type) | type;
```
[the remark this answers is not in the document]
skef: ok

fontforgeexe/cvimportdlg.c
skef: should be fixed.

fontforgeexe/cvimportdlg.c
```diff
@@ -488,20 +488,11 @@ void _ImportParamsDlg(ImportParams *ip) {
...
-    } else if ( type==2 ) {
-	if ( !shown && (!ip->shown_svg || ip->show_always) )
-	    _ImportParamsDlg(ip);
-	ip->shown_svg = true;
-    }
+static void ShowImportOptions(ImportParams *ip, int shown,
+                              enum shown_params type) {
+    if ( !shown && !(ip->shown_mask & type) )
+	_ImportParamsDlg(ip);
+    ip->shown_mask = (ip->shown_mask & ~type) | type;
```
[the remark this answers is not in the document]
skef: ok

jtanx:
Hmm, the 'show always' behaviour still seems a bit weird.

* New file
* Import
* Click Options -> set the show always to true, close
* Import SVG -> dialog does not show?

skef:
> Hmm, the 'show always' behaviour still seems a bit weird.

That's part of the design intent -- the params dialog doesn't come up again if it's already been up on "this" raise of the Import or Export dialog. Bringing it up again in that case would be annoying.

doc/sphinx/scripting/python/fontforge.rst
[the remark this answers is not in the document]
jtanx: Not entirely sure I follow that, but ok.

fontforge/scripting.c
jtanx: well it's definitely a unique coding style, usually you just have the || at the end of the line and align on the start of the bracket instead

fontforge/svg.c
jtanx: Yeah, just seems kind of dodgy to have to modify then unwind the change at the end of the call. I can imagine this would pretty easily break the next time someone comes along and returns before resetting the value.

Not necessarily for now, but something to think about - how about a separate field that defines the default value?  Then if the value is `inherited`, you use that, otherwise you use the normal value. Or maybe somewhat cleaner - you could also define a 'default type' field or something similar (i.e. indicate this is for SVG/etc.), then have the defaults defined in one place.

fontforgeexe/uiutil.c
jtanx: Yeah, I mean why isn't the naming scheme consistent with the rest of the other functions in this list?

fontforge/scripting.c
skef: This style of formatting communicates a higher level of commitment than that one because any logical connector at any level of nesting can wind up at the end of the line and be formatted that way. For example:

```
    if ( s==spline_to && (( t_fm<=t_to && !backward ) ||
         (t_fm>=t_to && backward))) {
```
vs
```
    if (    s==spline_to
         && (( t_fm<=t_to && !backward ) || (t_fm>=t_to && backward))) {
```
With the first you have to scan across counting parentheses to make sense of the logical structure. With the second I as the author have done the scanning and am communicating the structure to you the reader. 

Similarly with this from `utanvec.c`:
```
    if ( ys1d.a!=0 && (   Within16RoundingErrors(ys1d.a+ys1d.d,ys1d.d)
                       || Within16RoundingErrors(ys1d.a+yto,yto)))

```

fontforgeexe/uiutil.c
skef: I guess that's an idea but the other functions starting with `UI_` are defined in this file and the one I'm replacing that was not defined in this file didn't follow that convention.

doc/sphinx/scripting/python/fontforge.rst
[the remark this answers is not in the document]
skef: We have some interfaces with options that can be passed by position or by keyword and other interfaces where they can only be passed by keyword (and some interfaces with a mixture). Your string (if you add brackets) accurately communicates the first variety. I decided (when redoing the `stroke()` API docs) to follow something closer to the `flags` string tuple docs convention to make the distinction.

fontforge/scripting.c
jtanx: Typically in that case you'd just indent further on the second line to indicate a continuation of the subcondition.

skef: #830 is open. The project could arrive at a consensus style for new code; it hasn't done so. Until then I recommend against anyone presenting their own preferences as objectively better in code review contexts. I cannot believe you're unaware that there is wide disagreement about these questions.

PR #4207, opened 2020-03-05
Only adjust subdict fontmatrix for synthetic font dicts (fixes #4192)

### LINKED ISSUE #4192: Glyphs of certain CFF font not shown in overview
When opening the attached CFF font in fontforge its glyphs are so small that they are not visible in the overview.

From what I see it seems that FontMatrix from Top DICT ("0.001 0 0 0.001 0 0") is applied on FontMatrix "taken" from Private DICT ("0.001 0 0 0.001 0 0") in readcfftopdicts() inside of parsettf.c. The FontMatrix from Private DICT is actually a default value because it is not present there and I think that FontMatrix can not be present in Private DICT at all.

The resulting matrix is "0.000001 0 0 0.000001 0 0" which when applied makes the glyphs too small and hence invisible.

Fix could be to check whether the "fontmatrix_set" was set when reading Private DICT and apply the FontMatrix from Top DICT on it in that case and use FontMatrix of Top DICT otherwise (checking this right before multiplying the matrices in readcfftopdicts()).

I've discovered this issue when looking at https://bugzilla.redhat.com/show_bug.cgi?id=1803129.

- [x] Screenshot:
screenshot

- [x] The FontForge version and the operating system you're using:
Current FontForge master as of 25th of February 2020 (git hash: af998cc28fc69ccce56ffbaf2c62aa57d2790bef).
Fedora 31

- [x] The behavior you expect to see, and the actual behavior:
I expect to see the glyphs in the overview as in the screenshot-expected.png (done by manually modifying EM size to 1000 without scaling of outline in "General" tab of "Font Info"):
screenshot-expected

- [x] Steps to reproduce the behavior:
Open the attached CFF font in FontForge.

- [x] \(optional) Possible solution/fix/workaround:
Workaround is to modify EM size to 1000 in "General" tab of "Font Info" without scaling of outline.

[d00006-001.14.gz](https://github.com/fontforge/fontforge/files/4250672/d00006-001.14.gz)

skef:
See #4192 

Closes #4192

### DIFF
--- a/fontforge/parsettf.c
+++ b/fontforge/parsettf.c
@@ -3351,8 +3351,13 @@ return( NULL );
 	}
 	dicts[i] = readcfftopdict(ttf,fontnames!=NULL?fontnames[i]:NULL,
 		offsets[i+1]-offsets[i], info);
-	if ( parent_dict!=NULL && parent_dict->fontmatrix_set ) {
-	    MatMultiply(parent_dict->fontmatrix,dicts[i]->fontmatrix,dicts[i]->fontmatrix);
+	if (    parent_dict!=NULL
+	     && parent_dict->fontmatrix_set
+	     && dicts[i]->synthetic_base!=-1 ) {
+	    if ( dicts[i]->fontmatrix_set )
+		MatMultiply(parent_dict->fontmatrix,dicts[i]->fontmatrix,dicts[i]->fontmatrix);
+	    else
+		memcpy(dicts[i]->fontmatrix,parent_dict->fontmatrix,6*sizeof(real));
 	}
 	dicts[i]->cff_start = cff_start;
     }


frank-trampe:
I did not look at the Travis error, but I did restart it.

frank-trampe:
This is definitely a strictly positive change, but do you think that it would be better, in the case that the parent has a value and the current dict does not, simply to inherit the parent value rather than using the 0.001 defaults? Same question to @MarekKasik.

skef:
@frank-trampe Maybe -- I was wondering about that myself. 

Hunting down the spec, these seem to be the relevant entries: https://wwwimages2.adobe.com/content/dam/acom/en/devnet/font/pdfs/5176.CFF.pdf#page=27

The sub-dict doesn't seem to be private -- FontForge has a different function to read those. And FontForge doesn't support "synthetic fonts" (according to error messages in the code). So the relevant case seems to be a CID font with its own topdict. The spec doesn't seem to indicate that there should be any inheritance or relation for those, so maybe the correct thing to do is just remove the adjustment entirely.

frank-trampe:
Hmm. I wonder what the original case was that inspired this adjustment. @davelab6, can you think of anything?

skef:
... or make it conditional on `synthetic_base` somehow (`!=-1`?)

skef:
The way I read the spec it's definitely what you would want to do for a synthetic font (where the subdict fontmatrix is probably going to skew or something). But there's no indication of such a relation for CID fonts.

frank-trampe:
Okay. Let's follow that reading for now. I'll send a note to the user list to solicit feedback.

skef:
So then probably:

1. If not a synthetic font dict, do nothing (and therefore keep the default)
2. If a synthetic font dict and fontmatrix is present, multiply
3. If a synthetic font dict and fontmatrix is not present, inherit (treat the "default" as 1 0 0 1 0 0)

frank-trampe:
That sounds right.

skef:
@frank-trampe I think all this stuff is specific to CFF, and therefore Type1-as-TTF (?)

skef:
In any case I force-pushed an attempted implementation of that. It unsurprisingly does the seemingly right thing with the test font.

frank-trampe:
The changes look right to me. Let's wait a few days for mailing list feedback. We've been on a lucky streak of no objections for the past few months, and I hope that it keeps going.

davelab6:
you could have a look at the Adobe font development kit for open type source code now that it's published on a Libre license, there's a lot of magic in CFF. I'm sorry I can't be more specific I didn't really understand what the issue is...

skef:
Interesting: [It doesn't support synthetic fonts either](https://github.com/adobe-type-tools/afdko/blob/52e4f7f9fd0a496936286747b4cea712532edc47/c/spot/source/cffread.c#L1825).

This is how the matrix is used in the CID case: https://github.com/adobe-type-tools/afdko/blob/52e4f7f9fd0a496936286747b4cea712532edc47/c/spot/source/cffread.c#L2237 . It seems to be getting the dict from the glyph and using that value

But there is a multiply in `saveMatrix()`: https://github.com/adobe-type-tools/afdko/blob/52e4f7f9fd0a496936286747b4cea712532edc47/c/spot/source/cffread.c#L1518 

```
    h->font.unitsPerEm = (unsigned short)DBL2INT(1.0 / max);

    if (h->dict.fd == NULL)
        FM = h->dict.FontMatrix = h->dict.topDICTMatrix;
    else
        FM = h->dict.fd->FontMatrix = h->dict.fd->FDMatrix;

    for (i = 0; i < 6; i++)
        FM[i] = indexDbl(h, i) * h->font.unitsPerEm;

    if (h->dict.fd != NULL && h->dict.FontMatrix != NULL) {
        /* Multiply top DICT matrix and FD matrix */
        double *A = h->dict.FontMatrix;
        double B[6];

        memcpy(B, FM, sizeof(B));

        FM[0] = A[0] * B[0] + A[1] * B[2];
        FM[1] = A[0] * B[1] + A[1] * B[3];
        FM[2] = A[2] * B[0] + A[3] * B[2];
        FM[3] = A[2] * B[1] + A[3] * B[3];
        FM[4] = A[4] * B[0] + A[5] * B[2] + B[4];
        FM[5] = A[4] * B[1] + A[5] * B[3] + B[5];
    }
```
Even so:

1. The multiply comes after the `unitsPerEm` calculation
2. The only thing FontForge seems to use the matrix for at present is that calculation.

jtanx:
Where are we at with this? Good to merge?

skef:
I'm the OP so not the best judge, but this does solve the main problem and the further questions about the spec are esoteric and much less likely to be encountered in practice. I would say it's ready.

frank-trampe:
There were no objections on the mailing list, and it looks like the right change. I think that we ought to merge it.

PR #4208, opened 2020-03-05
Preserve ligature carets when opening charinfo dialog (fixes #4171)

### LINKED ISSUE #4171: Bug: null Lig. carets after change Glyph Color
Detail description (pdf with prinscreens) and example arabic font are in ZIP
[bug-null-lig-carets.zip](https://github.com/fontforge/fontforge/files/4214506/bug-null-lig-carets.zip)

skef:
The overall philosophy of CharInfo is to make a dialog-specific representation of (just about?) all the data and then copy it back out. But the dialog doesn't control or report the ligature caret positions -- that's done in the CharView. So the values were getting lost when the dialog is closed with "OK".

It's unexpected that this would have gone so long unreported, but not unprecedented. Maybe few users bother with ligature carets? 

Closes #4171

### DIFF
--- a/fontforgeexe/charinfo.c
+++ b/fontforgeexe/charinfo.c
@@ -837,7 +837,7 @@ static int CI_ProcessPosSubs(CharInfo *ci) {
     struct matrix_data *possub;
     char *buts[3];
     KernPair *kp, *kpprev, *kpnext;
-    PST *pst, *pstprev, *pstnext;
+    PST *pst, *pstprev, *pstnext, *lcpst=NULL;
 
     possub = GMatrixEditGet(GWidgetGetControl(ci->gw,CID_List+(pst_ligature-1)*100), &rows );
     cols = GMatrixEditGetColCnt(GWidgetGetControl(ci->gw,CID_List+(pst_ligature-1)*100) );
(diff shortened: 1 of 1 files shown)


fontforgeexe/charinfo.c
```diff
@@ -930,8 +930,31 @@ return( false );
     }
 
     /* If we get this far, then we didn't find any errors */
-    for ( pst = sc->possub; pst!=NULL; pst=pst->next )
-	pst->ticked = pst->type==pst_lcaret;
```
frank-trampe: What was this doing before?

fontforgeexe/charinfo.c
```diff
@@ -930,8 +930,31 @@ return( false );
...
+	;
+    if ( pst!=NULL ) {
+	if ( lcpst==NULL ) {
+	    lcpst = chunkalloc(sizeof(PST));
+	    lcpst->type = pst_lcaret;
+	    lcpst->next = sc->possub;
+	    sc->possub = lcpst;
+	}
+	if ( lcpst->u.lcaret.carets!=NULL )
+	    free(lcpst->u.lcaret.carets);
```
frank-trampe: Is there a case in which extant carets would change via the dialog?

fontforgeexe/charinfo.c
skef: Elements that aren't `ticked` will be modified or deleted by the code below. I assume this was there to save existing `pst_lcaret` entries but they never wound up being copied in the first place. Every `ticked` can be false now because the new code that copies the `lcaret`s sets `ticked` to true (and if there aren't any the entry can and should be deleted).

fontforgeexe/charinfo.c
skef: Yes; definitely a 1->0 count transition and probably any count transition (the dialog lets you choose how many lcarets there are, just not where they are).

PR #4214, opened 2020-03-07
Font Open flag improvements

### LINKED ISSUE #4212: fontforge moduel in python 2.7 wouldn't load some font files that actually work and get installed on windows
Fontforge from python wouldn't load some font files and throw the error.
`EnvironmentError: Open failed`
```
fontforge.open('font.ttf')
```
  '20170924' fontforge version 
Ubuntu 18.04 (wsl)

You can check the files here that wouldn't open here.

[nonloaded.zip](https://github.com/fontforge/fontforge/files/4300037/nonloaded.zip)

skef:
Document more Open flags
Add string tuple interface to python FontOpen API
Eliminate undocumented start-hidden flag

`of_hidewindow` was indeed undocumented, but more relevantly it doesn't currently work. Using the flag from `Execute Script...` in the UI results in an entry at the bottom of the Window menu that has no effect. 

Closes #4212

### DIFF
--- a/doc/sphinx/scripting/python/fontforge.rst
+++ b/doc/sphinx/scripting/python/fontforge.rst
@@ -444,10 +444,23 @@ Module functions
 
 .. function:: open(filename[, flags])
 
-   Opens a filename and returns the font it contains (if any). If the flags
-   argument is 4, then ff will load all glyphs in the 'glyf' table of a ttc file
-   (rather than just the glyphs used in the font picked). This will not load all
-   'glyf' tables though.
+   Opens a filename and returns the font it contains (if any). The optional ``flags`` argument can be string tuple or integer combination of the following flags:
+
+   .. object:: fstypepermitted (1)
+
+      The user has the appropriate license to examine the font no matter what the fstype setting is.
+
+   .. object:: allglyphsinttc (4)
+
+      Load all glyphs from the 'glyf' table of a ttc font (rather than only the glyphs used in the font picked).
+
+   .. object:: fontlint (8)
+
+      Report more error conditions
+
+   .. object:: alltables (32)
+
+      Retain all recognized font tables that do not have a native format
 
 .. function:: parseTTInstrs(string)
 
--- a/doc/sphinx/scripting/scripting-alpha.rst
+++ b/doc/sphinx/scripting/scripting-alpha.rst
@@ -1813,6 +1813,8 @@ the fourth argument you must specify the second and third arguments too.
      what the fstype setting is.
    * 4 => load all glyphs from the 'glyf' table of a ttc font (rather than only
      the glyphs used in the font picked).
+   * 8 => Report more error conditions
+   * 32 => Retain all recognized font tables that do not have a native format
 
 .. function:: Ord(string[,pos])
 
--- a/fontforge/autosave.c
+++ b/fontforge/autosave.c
@@ -111,7 +111,7 @@ return( false );
 	if ( (sf = SFRecoverFile(buffer,inquire,&inquire_state)) ) {
 	    any=true;
 	    if ( sf->fv==NULL )		/* Doesn't work, cli arguments not parsed yet */
-		FontViewCreate(sf,false);
+		FontViewCreate(sf);
 	    fprintf( stderr, " Done\n" );
 	}
     }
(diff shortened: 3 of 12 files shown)


skef:
Ignore this one for the upcoming release unless you happen to feel like reviewing it. I just felt like getting the issue on track for being closed.

doc/sphinx/scripting/python/fontforge.rst
```diff
@@ -444,10 +444,23 @@ Module functions
 
 .. function:: open(filename[, flags])
 
-   Opens a filename and returns the font it contains (if any). If the flags
-   argument is 4, then ff will load all glyphs in the 'glyf' table of a ttc file
-   (rather than just the glyphs used in the font picked). This will not load all
-   'glyf' tables though.
+   Opens a filename and returns the font it contains (if any). The optional ``flags`` argument can be string tuple or integer combination of the following flags:
```
jtanx: Pretty sure this is wrong - in that the integer value is not accepted, only the string representation is (and if I'm right I hope to keep it that way, otherwise it just seems bad to expose the underlying representation to the user as an option)

skef: I'm going to assume you're reviewing the docs before the code content here ...

Probably because it is one of the older methods the `open` interface *only* has the integer flags option in the current `master`. This PR adds the string tuple. I'm not happy about the integer option either but @frank-trampe strongly encourages backwards compatible interface changes when possible.

jtanx: Sorry, yep you're right.

jtanx:
> of_hidewindow...more relevantly it doesn't currently work.

I can think of a scenario where this could be used - open a font but don't display it in the UI, perform processing on it and close it from the script. 

Whether or not it's actually used like that (or used in the wild), I don't know...

skef:
@jtanx Yeah, I guess that would work, and I could see it being used with bulk-processing driven from the UI. In that case, though, "hide" isn't great terminology. I would expect "hide" to mean "start iconified" rather than "no FontView". And the fact that it appears in the window list is weird (although probably not important). 

Since there's no hurry let's just not put this in the upcoming release. Frank can ask on the mailing list and if that flag is used (or desired) we could figure out the right implementation at that point.

@frank-trampe what do you think?

frank-trampe:
Just to be clear, the code in question could not possibly be used right now and has never worked, right?

I would not have added this code without actually wiring it to work, but I also wouldn't remove it given that we might plausibly use it in the future. But, then again, the stakes are pretty low.

skef:
@frank-trampe No, it could conceivably be used in the way @jtanx suggests. My assertion that it doesn't work was based on an interpretation of how it was intended to work.

The flag has no effect when not run in a UI environment. When run in a UI environment the following happens (as best as I can tell):

1. The font is "visible" at the script level.
2. No FontView is created.
3. The font appears at the bottom of the `Window` menu but picking it has no effect.

Unless there's something way back in the old commits it does seem accurate that the flag was never documented.

frank-trampe:
Oh. In that case, I think that it would be better to keep and to document it. People may not know about it, but there are definitely cases in which they ought to be using it.

skef:
OK, I'll do that in the next hour or two and folks can see how they feel about it. I'll include a script sample so they can see how they feel. 

Again, this will just to provide the option of integrating. (I assume that with the smaller degree of change it might make more sense to merge now.)

skef:
Hmm. Do we care if, in the case where one of these fonts is opened and not subsequently closed, closing all the windows results in having no icons but the program still runs?

skef:
OK, here's what I'm going to do. I'm going to open another draft PR for the version with `hidewindow`, leave this PR as it is, and we discuss which version makes more sense -- presumably not for the up coming release.

frank-trampe:
Presumably, this would be a problem only if somebody starts the script from the scripting window in FontForge, right, since otherwise primary control should be in a separate terminal?

skef:
@frank-trampe yes, but the only use case for this flag is from that scripting window, right? Otherwise there's no window to suppress.

frank-trampe:
Oh, I guess that's right. That does seem mildly problematic, but I cannot think of an easy fix. We ought to make note of it somewhere.

skef:
@frank-trampe let's continue this over in #4219. If you can suggest some language I'll add it to the docs.

skef:
Given that the momentum points to keeping the `hidewindow` flag, I'm closing this.

PR #4219, opened 2020-03-12
Font Open flag improvements (keep hidewindow)

skef:
This is an alternate version of #4214 that keeps the `hidewindow` flag. As already noted:

1. When used in a standard script environment this has no effect
2. When used in the `Execute script` environment the flag adds an entry to the bottom of the Window menu for the font but does not create a window
3. Currently, if the font is not closed at the script level closing the other windows leaves the application running. 

I'm creating this draft PR so that folks can take a look and decide whether we want this and how it should work if we do.

### DIFF
--- a/doc/sphinx/scripting/python/fontforge.rst
+++ b/doc/sphinx/scripting/python/fontforge.rst
@@ -325,18 +325,18 @@ Module functions
 
    Returns n for FontForge internal table Unicode ``val=Ligature[n]``. If val
    does not exist in table, then return -1. Can execute with no current font.
-   
+
    .. note::
-      
+
       Count depends on chart version built into FontForge.
 
 .. function:: ucLigChartGetNxt(n)
 
    Returns FontForge internal table Unicode Ligature[n]. Return -1 if ``n<0``
    or ``n>=ucLigChartGetCnt()``. It can execute with no current font.
-   
+
    .. note::
-      
+
       Count depends on chart version built into FontForge.
 
 .. function:: ucOFracChartGetCnt()
@@ -345,7 +345,7 @@ Module functions
    described by www.unicode.org. It can execute with no current font.
 
    .. note::
-      
+
       Count depends on chart version built into FontForge.
 
 .. function:: ucOFracChartGetLoc(val)
@@ -354,7 +354,7 @@ Module functions
    val does not exist in table, then return -1. Can execute with no current font.
 
    .. note::
-      
+
       Count depends on chart version built into FontForge.
 
 .. function:: ucOFracChartGetNxt(n)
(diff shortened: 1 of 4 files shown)


skef:
The easiest way to test the hide option is with a native script like this in the `Execute Script...` dialog:

```
Open("LibertinusSerif-Regular.sfd", 16)
```

This python should also work:
```
import fontforge

fontforge.open("LibertinusSerif-Regular.sfd", ("hidewindow",))
```

(The trailing comma is, of course, to mark the 1-tuple.)

frank-trampe:
This all looks good to me. I cannot think of an appropriate warning that is sufficiently concise to go inline in the flags documentation. Maybe just a comment in `python.c`?

In the future, perhaps we could have an invisible windows list window that appears whenever (1) there is no font view window and (2) there are fonts open. It would have a buttons for opening a scripting window and for opening a font. Closing it would close FontForge.

skef:
@frank-trampe I decided to just "go verbose". I checked the HTML and the result is the expected short block paragraph.

frank-trampe:
Fine with me.

skef:
This version of the change is quite low-risk. Unless there is concern about advertising previously obscure flags, this can probably make the next release.

doc/sphinx/scripting/python/fontforge.rst
```diff
@@ -444,10 +444,27 @@ Module functions
...
+
+      Load all glyphs from the 'glyf' table of a ttc font (rather than only the glyphs used in the font picked).
+
+   .. object:: fontlint (8)
+
+      Report more error conditions.
+
+   .. object:: hidewindow (16)
+
+      Do not create a view window for this font even if the UI is active. Note: This option supports efficient bulk processing of fonts in scripts run through the UI but using it can be tricky. Open fonts will be listed at the bottom of the "Window" menu but choosing them will have no effect. If some fonts are not closed you may need to "force-quit" the application using your OS.
```
jtanx: For consistency, can we wrap this to ~80 chars/line?

skef: Should be done

doc/sphinx/scripting/scripting-alpha.rst
```diff
@@ -1813,6 +1813,9 @@ the fourth argument you must specify the second and third arguments too.
      what the fstype setting is.
    * 4 => load all glyphs from the 'glyf' table of a ttc font (rather than only
      the glyphs used in the font picked).
+   * 8 => Report more error conditions
+   * 16 => Do not create a view window for this font even if the UI is active. Note: This option supports efficient bulk processing of fonts in scripts run through the UI but using it can be tricky. Open fonts will be listed at the bottom of the "Window" menu but choosing them will have no effect. If some fonts are not closed you may need to "force-quit" the application using your OS.
```
jtanx: Ah, I guess we might as well wrap this too (and the other lines)

skef: Half the time I've tried to wrap the lines in these lists it has worked fine and the other half of the time it has failed and I don't know the difference. Could you please do the commit for this one? You have access to the branch via github.

jtanx: Sure.

For reference, the thing to keep in mind with lists is that the indentation has to line up like so:

```rst

* test
  note this is indented two spaces to line up with the line above

1. numbered
   this is indented by 3 spaces to line up with above

.. object:: stuff

   this is indented 3 spaces

   * stuff
     indented 5 spaces
```

PR #4220, opened 2020-03-14
Exclude fullpo from the default target and fix output directory creation

jtanx:
### Type of change

- **Bug fix**


jtanx:
The travis build for the PR builds fine, the other travis build is for coverity, which will always 'fail' (as I cancel all the builds except the 'full' variant, since you can only upload one version to coverity)

PR #4221, opened 2020-03-14
Fix a couple of coverity issues

jtanx:
### Type of change

- **Bug fix**


fontforge/python.c
```diff
@@ -9635,8 +9636,9 @@ static PyObject* PyFF_Glyph_BoundsAt(PyCFunction bounds_func, PyFF_Glyph *self,
             ss = ss->next;
             if (!PyTuple_Check(temp)) {
                 continue;
+            } else if (!PyArg_ParseTuple(temp, "dd", &tnmin, &tnmax)) {
```
skef: PyArg_ParseTuple sets an exception. If you want to just continue processing I think you need a `PyErr_Clear();` here.

jtanx: I made it return null as I guess it should never have an invalid tuple at that point

skef: That is also fine.

PR #4244, opened 2020-03-29
Avoid unaligned memory access in gethex function

floppym:
This resolves a runtime failure on platforms that do not support
unaligned loads/stores (like sparc).

Bug: https://bugs.gentoo.org/642756

### Type of change
- **Bug fix**

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -3475,6 +3475,7 @@ return( pt!=tokbuf?1:ch==EOF?-1: 0 );
 static int gethex(FILE *sfd, uint32 *val) {
     char tokbuf[100]; int ch;
     char *pt=tokbuf, *end = tokbuf+100-2;
+    uint32 u;
 
     while ( isspace(ch = nlgetc(sfd)));
     if ( ch=='#' )
@@ -3498,7 +3499,8 @@ static int gethex(FILE *sfd, uint32 *val) {
     }
     *pt='\0';
     ungetc(ch,sfd);
-    *val = strtoul(tokbuf,NULL,16);
+    u = strtoul(tokbuf,NULL,16);
+    memcpy(val, &u, sizeof(u));
 return( pt!=tokbuf?1:ch==EOF?-1: 0 );
 }
 


skef:
@floppym Can you provide more information about how you're compiling FontForge or what distribution you're getting it through? It would preferable to compile with `-xmemalign=8i` or its equivalent than to introduce a lot of these changes into the code.

floppym:
@skef See the linked bug report for Gentoo Linux. I'm really just forwarding the patch from there.

floppym:
Looking closer at the code, I would guess this line is the problem:

https://github.com/fontforge/fontforge/blob/20200314/fontforge/sfd.c#L8107

pfminfo.stylemap is a 16-bit int, but it gets cast to a 32-bit int pointer.

skef:
Hmm, so they're compiling with gcc which has no exact equivalent to `-xmemalign`. 

@jtanx is there any hope of handling this issue with compiler flags alone? Peppering the code with after-the-fact `memcpy`s isn't ideal.

floppym:
I think this code is just broken; it writes 4 bytes of data into a 2 byte field.

skef:
The new code casts an eight byte value into a four byte `uint32`. With the old code the compiler knows that `val` is a pointer to `uint32` so there's no reason it wouldn't do the same cast (that I'm aware of). The issue here remains one of alignment, I think.

floppym:
> The new code casts an eight byte value into a four byte uint32.

Yeah, but there are no pointers involved there, so the compiler handles the typecast just fine.

floppym:
In any case, I've asked the bug reporter to test #4245 to see if that is sufficient.

mattst88:
FWIW, dereferencing unaligned pointers in C is undefined behavior, regardless of whether the underlying CPU is capable of performing unaligned loads and stores.

skef:
I hadn't put together that the destination in this case was a 16 bit struct member, because of the location and nature of the "fix". The suggested code avoided the immediate fault but not the (presumable) data corruption.

jtanx:
If you're telling me it's because the input into this function is a 16 bit value, then this is still not the right fix, you're still writing 4 bytes into a 2 byte field. The proper fix is to use a uint32 on the calling end and then set that into the 2 byte field.

skef:
Closing in favor of #4245

PR #4245, opened 2020-03-29
Avoid integer size conflict when setting stylemap

floppym:
stylemap is a 16-bit integer, but gethex writes 32-bits.

Bug: https://bugs.gentoo.org/642756

### Type of change
- **Bug fix**

Possibly obsoletes #4244.


skef:
Ah, yes, this seems better. 

@jtanx, thoughts?

jtanx:
Yeah it's better.

DerDakon:
I can confirm that the testsuite now passes on sparc.

PR #4246, opened 2020-03-29
Stub-out MacServiceReadFDs() on non-Mac platforms

floppym:
This should fix a build failure on Linux with musl libc due to the
missing sys/select.h include.

MacServiceReadFDs seems to only be called in startui.c, also behind the
__Mac macro.

Bug: https://bugs.gentoo.org/706792

### Type of change
- **Bug fix**


anarchpenguin:
Tested, does correct the musl failure. Ack'd

PR #4250, opened 2020-04-01
Handle weird floating point problem with focus calculation

### LINKED ISSUE #4248: crash when importing glyph G as a svg-file
Hey, I hope you can help me.

- [x] Screenshot
fontforge1png
fontforge2png
FontForge Version 20200314
Windows 10 Pro / Version 1909

I want to import a svg-file for the glyph "G". I did it already with A-F with no problems. Then the problem with "G" ocurred. I tried some things about the file - for example made it one closed path instead of two (kind of) interlocking paths (like to c's you could imagine). One time I imported a PNG in the backround and started to retrace the glyph in Fontforge but at some point, it crashed again. So I redid the whole process several times, every time Fontforge crashed when I wanted to import. When I open a new project, I can import the glyph A, but still not G. I then went on in my old project and I could add glyph H, but G just won't work.
One time I got the message "This file is to complicated or has an error" when importing "G", but I can't recall when exactly that was. I just redid the whole process with the easier version of "G" to exclude the problem of being too complicated.
I prepare every letter exactly the same, what's it with the G?

thank you in advance!

skef:
Closes #4248

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -675,6 +675,7 @@ static int _IndexForUTanVec(NibCorner *nc, int n, BasePoint ut, int nci_hint) {
 	                         nc[N_NEXTI(n, nci)].utv[NC_IN_IDX],
 	                         false) )
 		break;
+    assert(nci < n);
     return nci;
 }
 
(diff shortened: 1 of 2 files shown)


skef:
I wonder if I will or could ever really be qualified to write floating point code ...

frank-trampe:
Only time shall tell whether it fixes the problem in all environments, but this change seems strictly positive to me.

skef:
I logged into travis via a different PR and restarted it for this one. For whatever reason the builds had succeeded but the state didn't get associated back.

PR #4255, opened 2020-04-03
Turn on UseNewIndicScripts by default

dscorbett:
The new Indic shapers have been around since at least 2006 and are well supported by now. They should be used by default.

### Type of change
- **Non-breaking change**

### DIFF
--- a/fontforge/tottfgpos.c
+++ b/fontforge/tottfgpos.c
@@ -44,7 +44,7 @@
 #include "utype.h"
 
 int coverageformatsallowed=3;
-int use_second_indic_scripts = false;
+int use_second_indic_scripts = true;
 
 #include "ttf.h"
 


frank-trampe:
As I'm sure you're aware, there are a lot of workflows that depend on getting identical results, and we try to avoid disrupting them. Perhaps we could instead just adjust the default "prefs" file?

dscorbett:
Doesn’t that come to the same thing? Upgrades of FontForge over existing installations would keep whatever the user’s current preference value is in ~/.config/fontforge/prefs, and fresh installations would get the new preference value.

frank-trampe:
My main concern is about headless script invocations.

ctrlcctrlv:
I would normally agree with Frank but there is precedent for changing an option if the default is obviously wrong in modern context, we did it for PreferSpacingAccents

frank-trampe:
@ctrlcctrlv, the due diligence involved in that was non-trivial. I would like to avoid it if there's an easier way to solve the immediate problem. If the old behavior is generally wrong in most contexts, we can start the process of changing the internal default, but that would necessarily involve a request for comment with context. To be honest, since I don't work with Indic scripts, even I don't entirely understand the functional difference between the two modes. Could somebody explain briefly?

ctrlcctrlv:
I'm happy to explain. 

You may have noticed some languages use OpenType script tags which weirdly contain the number 2. Eg. `dev2` for Devanagari instead of `deva`. You also may have noticed that the 2 versions are by far more common. 

It's because the original OpenType conception of how to handle these scripts was very flawed. Microsoft championed the fix, which is now standard and used by almost all modern Devanagari (and other Indic) fonts. 

@n8willis outlined the technical shortcomings which the new script tags solved here: https://github.com/n8willis/opentype-shaping-documents/blob/master/opentype-shaping-indic-general.md#the-old-indic-shaping-model

n8willis:
In my demonstrably non-Indic opinion, there's no reason to still be making new Indic1 fonts; Indic2 fixes lots of shortcomings in the original and mostly importantly HarfBuzz supports it (arguably better than anyone else) out of the box. So do the other shaping engines.

But I would rather have @pathumego or others who are actually in design & production of relevant fonts weigh in on the practical issues.

@frank-trampe You might could look into (say) whether or not there are any Indic1 fonts still in Debian and consider that a bellwether ... not sure where else you would encounter IndicN build scripts.

frank-trampe:
Agh. The first one I checked, Sahadeva, uses `deva`. I'm going to write a script to get numbers.

n8willis:
> Note the overlap.

FYI, that's totally fine. Lets the designer promote that the font binary could be used even on a legacy system with a legacy document.

Just to make sure we don't veer off course here, the issue we're talking about is not "removing `deva` support", it's "does changing the default user Preference break any batch scripts?" So we can look at those eight fonts in the `deva` but not the `dev2` list and maybe see if their build script employs it.

Although wait; the more I think about it, even that might be premature ... how is that preference even exposed in the scripting interface? Is it? 

The source SFD is already going to have the `deva` tag or it isn't. And I don't see a reference to it anywhere in the scripting docs.... Are you sure this scripting issue exists?

ctrlcctrlv:
@n8willis It is exposed via the scripting interface, but that is an extremely recent innovation. As in, literally it dates back to 20200314 release. I needed it for a script I was working on, see #4163.

Observe:

```python3
import fontforge as ff

for b in [True, False]:
  ff.setPrefs("UseNewIndicScripts", b)
  ff.postError('', ff.scriptFromUnicode(0x920))
```

This innovation is so recent that we can send a warning to the mailing list that UseNewIndicScripts is about to change by default and `ff.scriptFromUnicode` will start returning new script tags for Indic scripts as a result.

No fonts use this new function yet.

ctrlcctrlv:
@frank-trampe If we're going to merge this, it needs to be soonish. Before ff.scriptFromUnicode finds itself in circulation. We might already be too late. Let me know

frank-trampe:
Alright. I suppose that swift and decisive action here is much less likely to cause future problems than waiting. I'll send a note to the mailing list about the change.

PR #4258, opened 2020-04-04
cmake: adjust Python linkage

mistydemeo:
The current Python 3 linkage links the modules directly against Python itself. This means that these modules can only be used by the exact Python against which they were built. Python's distutils uses alternate linkage methods to ensure Python modules stay compatible: https://github.com/python/cpython/blob/9a77656/configure.ac#L2214-L2245

In CMake, switching fontforge's Python3::Python usage to PRIVATE and adding a Python3::Module link command to pyhook fixes this.

### Type of change
- **Non-breaking change**

cc @bo98, who pointed me at the fix


pyhook/CMakeLists.txt
```diff
@@ -14,6 +14,9 @@ endif()
 target_link_libraries(fontforge_pyhook PRIVATE fontforge)
 target_link_libraries(psMat_pyhook PRIVATE fontforge)
 
+target_link_libraries(fontforge_pyhook PRIVATE Python3::Module)
```
jtanx, line 17: It looks like this is only available in CMake 3.15 or later, so we'd need to update the backport accordingly

mistydemeo: Thanks for taking care of that!

mistydemeo:
Can confirm linkage looks good to me with the current state of this branch: the `fontforge` executable links directly against Python as appropriate, while the Python libraries don't and are importable by any compatible Python.

jtanx:
Now that I think about it, it seems a bit weird, the python modules are very thin shims that then link to libfontforge, which itself links to Python - would that cause issues?

jtanx:
Well I guess it's no worse off than before, so I'm inclined to merge unless there's any objections.

Bo98:
Apologies for the delay in responding here. I completely forgot about this.

> would that cause issues?

The answer is yes. As a test, I compiled this under 3.8.2 and then used it with under a 3.8.1 install located in a different location and I get a segmentation fault. To show it is because of `libfontforge`, I switched it to `Python::Module` and do not get a segmentation fault.

jtanx:
Is that Python::Module on the pyhook shims, or on libfontforge itself?

Bo98:
Both - but it was just as an experiment. Doing it in `libfontforge` is fine in the context of fontforgeexe and the pyhook shims since the symbols will be loaded via other means, but it would be problematic if there were third party applications linking with libfontforge that do not load Python, since it's still a strong link of symbols (evaluated on load) and not a weak import.

jtanx:
Ok great, it sounds like this is fine then, thanks for confirming.

PR #4271, opened 2020-04-09
Scripting: new options for glyph export

yurks:
Exporting SVG glyph contain some hard-coded stuff like fill property (sometimes it not needed) and viewBox adjustments (it changes glyph visual dimensions, which is bad thing for icons font)

Extended scripting function Export() with two extra flags only applicable for svg-export:
```
2 => Disable fill property for path
4 => Disable viewBox adjustment (useful for icons font)
```

### Type of change
- **New feature**

### DIFF
--- a/doc/sphinx/scripting/scripting-alpha.rst
+++ b/doc/sphinx/scripting/scripting-alpha.rst
@@ -871,6 +871,8 @@ the fourth argument you must specify the second and third arguments too.
    flags:
 
    * 1 => Flip the y-axis of exported SVGs with a transform element (instead of rewriting values)
+   * 2 => Disable fill property for path
+   * 4 => Disable viewBox adjustment (useful for icons font)
    * 256 => Use current Export dialog settings (other flags are ignored)
    * 512 => Present Export options dialog (if UI is enabled)
 
--- a/fontforge/scripting.c
+++ b/fontforge/scripting.c
@@ -2499,6 +2499,8 @@ static void bExport(Context *c) {
     if ( c->a.argc==4 ) {
 	flags = c->a.vals[3].u.ival;
 	ep.use_transform = !!(flags & 1);
+	ep.no_fillcolor = !!(flags & 2);
+	ep.no_padding = !!(flags & 4);
     }
     if ( flags&256 || flags&512 ) {
 	epp = ExportParamsState();
--- a/fontforge/sd.h
+++ b/fontforge/sd.h
@@ -198,7 +198,9 @@ typedef struct exportparams {
     int initialized;
     int shown_mask, show_always;
 
-    int use_transform;		// SVG
+    int use_transform;      // SVG
+    int no_fillcolor;       // SVG
+    int no_padding;		    // SVG
 } ExportParams;
 
 extern void InitImportParams(ImportParams *ip);
--- a/fontforge/svg.c
+++ b/fontforge/svg.c
@@ -1055,11 +1055,13 @@ return( ret );
 
 int _ExportSVG(FILE *svg,SplineChar *sc,int layer,ExportParams *ep) {
     char *end;
-    int em_size, xstart, xend, ascent;
+    int em_size, xstart, xend, ascent, padding;
     real trans[6] = { 1.0, 0.0, 0.0, 1.0, 0.0, 0.0 };
     DBounds b;
     SplineChar *scc;
     SplineSet *orig;
+
+    padding = ep->no_padding ? 0 : SVGMINLRPAD;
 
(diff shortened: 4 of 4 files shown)


skef:
A few problems in this initial version:

1. "Disable viewbox adjustment" doesn't accurately describe what this patch does. This solution certainly doesn't retain an arbitrary viewbox through a round-trip, which is what I would expect on reading that phrase.
2. As long as something like the padding parameter is being added it should probably be an integer value for the amount of padding rather than a flag.
3. More problematically this patch adds the options in the export structure and the native scripting language but doesn't add them to the python API or the GUI.

Beyond that I'm not sure how I feel about the no-fill option. It seems better/more general to offer the option of a single file-internal CSS style for the exported paths and provide the option for the user to alter the text put there.

yurks:
> 1. "Disable viewbox adjustment" doesn't accurately describe what this patch does. This solution certainly doesn't retain an arbitrary viewbox through a round-trip, which is what I would expect on reading that phrase.

Well, could be not clear at first-look. This is ability to bypass #3395 changes, as https://github.com/fontforge/fontforge/commit/f8636410041e6b3f964c11d2ca04d7dfa138b14e suddenly broke default export behavior without any chance to know why for me as user. My words "adjustment", "padding" -- is an effort to be a "human" :)

> 2. As long as something like the padding parameter is being added it should probably be an integer value for the amount of padding rather than a flag.

Probably. I propose flag because `#define SVGMINLRPAD 10` doesn't mean to be customizable (besides, why 10 and not 12 or 9). Actually, that SVGMINLRPAD-transformation looks like redundant as we talking about SVG -- any transformations available via CSS if someone need it. Export should mean "export" as it is.

> 3. More problematically this patch adds the options in the export structure and the native scripting language but doesn't add them to the python API or the GUI.

Agree, should be expanded.

> Beyond that I'm not sure how I feel about the no-fill option. It seems better/more general to offer the option of a single file-internal CSS style for the exported paths and provide the option for the user to alter the text put there.

We just do the export font glyph as it is. `fill="currentColor"` attribute is an extra care, like any custom colors or paddings -- this is not a scope of this action.

skef:
> Well, could be not clear at first-look. This is ability to bypass #3395 changes, as f863641 suddenly broke default export behavior without any chance to know why for me as user. My words "adjustment", "padding" -- is an effort to be a "human" :)

Even so, the description is then in relation to the previous behavior which was similarly arbitrary. If we wanted FontForge to preserve each ViewBox through a round-trip we could do that but it would be more work. "Preserving" past behavior is something different. 

> Probably. I propose flag because #define SVGMINLRPAD 10 doesn't mean to be customizable (besides, why 10 and not 12 or 9). Actually, that SVGMINLRPAD-transformation looks like redundant as we talking about SVG -- any transformations available via CSS if someone need it. Export should mean "export" as it is.

As *what* is? The left padding was added because the zero position of the glyph sometimes resulted in the left side being clipped. In SVG there is nothing magic about x or y at zero given that the viewbox can be positioned anywhere.  How are you defining "as it is?" 

You're suggesting that this issue has to do with SVG directly when it really has to do with a certain import/export convention. I have no idea how widespread that convention is or where it comes from. That's why I think it would make more sense to make the degree of padding configurable. We don't want the program to wind up with dozens of flags specific to this or that user's workflow. 

> We just do the export font glyph as it is. fill="currentColor" attribute is an extra care, like any custom colors or paddings -- this is not a scope of this action.

I guess I'm not necessarily against this approach, then. It does seem that a contour will preview as expected without the directive.

frank-trampe:
The padding we apply in exporting svg fonts is sort of arbitrary. Controlling its degree is not particularly useful and is a lot harder to implement than a flag. I favor a flag to disable the padding.

Is the fill color flag necessary, though? What is the use case? Could you just apply an XSLT to the output file rather than add it here?

We do need anything that gets merged to be exposed in the Python interface and in the graphical interface.

skef:
Note for clarification: I believe this path is just used on independent glyph export and not font export but I could be mistaken about that. 

If we go with a flag there's still the question of what the flag should do and why. I made the original change because some glyphs were getting clipped at x=0. 

It sounds like the desired option is something like: Set ViewBox to x=0 to advance horizontally, descender to ascender lines vertically, clipping be damned. If that makes sense for a large enough subset of workflows then I'm fine with it. I'm also fine with other heuristics as long as they can be accurately documented.

ctrlcctrlv:
Code has serious problems and no engagement from originator since April. Closing. Can be reopened if a new commit is made to [`yurks:master`](https://github.com/yurks/fontforge/tree/master).

PR #4272, opened 2020-04-09
Extra format for exported filename

yurks:
Ability to set filename for exporting glyph with char ligature if present.
```
``%N`` -- inserts the ligature name if present in glyph, or the glyph name if not (or the first 40 characters)
```

### Type of change
- **New feature**

### DIFF
--- a/fontforge/cvexport.c
+++ b/fontforge/cvexport.c
@@ -691,7 +691,7 @@ return( ret );
 static void MakeExportName(char *buffer, int blen,char *format_spec,
 	SplineChar *sc, EncMap *map) {
     char *end = buffer+blen-3;
-    char *pt, *bend;
+    char *pt, *bend, *ligature;
     char unicode[8];
     int ch;
 
@@ -705,7 +705,7 @@ static void MakeExportName(char *buffer, int blen,char *format_spec,
 	    if ( ch=='n' || ch=='N' ) {
             pt=sc->name;
             if (ch=='N') {
-                char *ligature = GetLigature(sc);
+                ligature = GetLigature(sc);
                 if (ligature != NULL) {
                     pt = ligature;
                 }
@@ -722,6 +722,9 @@ static void MakeExportName(char *buffer, int blen,char *format_spec,
 		for ( ; *pt!='\0' && buffer<bend; )
 		    *buffer++ = *pt++;
 #endif
+        if (ligature != NULL) {
+            free(ligature);
+        }
 	    } else if ( ch=='f' ) {
 		for ( pt=sc->parent->fontname; *pt!='\0' && buffer<bend; )
 		    *buffer++ = *pt++;
--- a/fontforge/cvexport.c
+++ b/fontforge/cvexport.c
@@ -733,10 +733,10 @@ static void MakeExportName(char *buffer, int blen,char *format_spec,
 		for ( pt=unicode; *pt!='\0' && buffer<bend; )
 		    *buffer++ = *pt++;
 	    } else if ( ch=='e' ) {
-        sprintf( unicode,"%d", (int) map->backmap[sc->orig_pos] );
-        for ( pt=unicode; *pt!='\0' && buffer<bend; )
-            *buffer++ = *pt++;
-        } else
+		sprintf( unicode,"%d", (int) map->backmap[sc->orig_pos] );
+		for ( pt=unicode; *pt!='\0' && buffer<bend; )
+		    *buffer++ = *pt++;
+	    } else
 		*buffer++ = ch;
 	}
     }
--- a/fontforge/cvexport.c
+++ b/fontforge/cvexport.c
@@ -702,17 +702,24 @@ static void MakeExportName(char *buffer, int blen,char *format_spec,
 	    ++format_spec;
 	    ch = *format_spec++;
 	    if ((bend = buffer+40)>end ) bend = end;
-	    if ( ch=='n' ) {
+	    if ( ch=='n' || ch=='N' ) {
+            pt=sc->name
+            if (ch=='N') {
+                char *ligature = GetLigature(sc);
+                if (ligature != NULL) {
+                    pt = ligature;
+                }
+            }
 #if defined( __CygWin ) || defined(__Mac)
 		/* Windows file systems are not case conscious */
 		/*  nor is the default mac filesystem */
-		for ( pt=sc->name; *pt!='\0' && buffer<bend; ) {
+		for ( ; *pt!='\0' && buffer<bend; ) {
 		    if ( isupper( *pt ))
 			*buffer++ = '$';
 		    *buffer++ = *pt++;
 		}
 #else
-		for ( pt=sc->name; *pt!='\0' && buffer<bend; )
+		for ( ; *pt!='\0' && buffer<bend; )
 		    *buffer++ = *pt++;
 #endif
 	    } else if ( ch=='f' ) {
@@ -726,10 +733,10 @@ static void MakeExportName(char *buffer, int blen,char *format_spec,
 		for ( pt=unicode; *pt!='\0' && buffer<bend; )
 		    *buffer++ = *pt++;
 	    } else if ( ch=='e' ) {
-		sprintf( unicode,"%d", (int) map->backmap[sc->orig_pos] );
-		for ( pt=unicode; *pt!='\0' && buffer<bend; )
-		    *buffer++ = *pt++;
-	    } else
+        sprintf( unicode,"%d", (int) map->backmap[sc->orig_pos] );
+        for ( pt=unicode; *pt!='\0' && buffer<bend; )
+            *buffer++ = *pt++;
+        } else
 		*buffer++ = ch;
 	}
     }
--- a/fontforge/cvexport.c
+++ b/fontforge/cvexport.c
@@ -703,7 +703,7 @@ static void MakeExportName(char *buffer, int blen,char *format_spec,
 	    ch = *format_spec++;
 	    if ((bend = buffer+40)>end ) bend = end;
 	    if ( ch=='n' || ch=='N' ) {
-            pt=sc->name
+            pt=sc->name;
             if (ch=='N') {
                 char *ligature = GetLigature(sc);
                 if (ligature != NULL) {
--- a/fontforge/svg.c
+++ b/fontforge/svg.c
@@ -770,18 +770,29 @@ return( SFGetChar(sc->parent,-1,pst->u.subs.variant));
 return( NULL );
 }
 
+#define LIGATURE_MAXLEN 150
+
 char *GetLigature(SplineChar *sc) {
     PST *best = HasLigature(sc);
     if ( best != NULL ) {
     	int32 univals[50];
-        char *buf = "";
-        int i, c;
+        char buf[LIGATURE_MAXLEN] = "";
+        char buf2[15];
+        int i, c, r, n = 0;
+
 		c = LigCnt(sc->parent,best,univals,sizeof(univals)/sizeof(univals[0]));
 		for ( i=0; i<c; ++i ) {
-		    if ( (univals[i]>='A' && univals[i]<='Z') ||  (univals[i]>='a' && univals[i]<='z') || (univals[i] >= '0' && univals[i] <= '9') || univals[i] == '_' || univals[i] == '-')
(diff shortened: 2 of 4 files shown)


fontforge/svg.c
```diff
@@ -766,60 +766,75 @@ SplineChar *SCHasSubs(SplineChar *sc, uint32 tag) {
...
+
     }
 return( NULL );
 }
 
+char *GetLigature(SplineChar *sc) {
+    PST *best = HasLigature(sc);
+    if ( best != NULL ) {
+    	int32 univals[50];
+        char *buf = "";
```
skef, line 778: An empty string isn't a buffer

fontforge/svg.c
```diff
@@ -766,60 +766,75 @@ SplineChar *SCHasSubs(SplineChar *sc, uint32 tag) {
...
+        int i, c;
+		c = LigCnt(sc->parent,best,univals,sizeof(univals)/sizeof(univals[0]));
+		for ( i=0; i<c; ++i ) {
+		    if ( (univals[i]>='A' && univals[i]<='Z') ||  (univals[i]>='a' && univals[i]<='z') || (univals[i] >= '0' && univals[i] <= '9') || univals[i] == '_' || univals[i] == '-')
+		    	asprintf(&buf,"%s%c", buf, univals[i]);
+		    else
+		    	asprintf(&buf, "%s&#x%x;", buf, (unsigned int) univals[i]);
+		}
+        return (  copy(buf)  );
```
skef, line 787: Seems like this won't be freed based on how the function is called.

skef: Some of our formatting functions use a static fixed-length array to get around this problem. One returns a pointer to the static value, which is retained until it gets written over by the next call. Because there's just the one buffer there's no need to free it (indeed, there's a need *not* to free it). 

It's not the greatest style but it might be the easiest solution for this case.

fontforge/cvexport.c
```diff
@@ -726,10 +733,10 @@ static void MakeExportName(char *buffer, int blen,char *format_spec,
 		for ( pt=unicode; *pt!='\0' && buffer<bend; )
 		    *buffer++ = *pt++;
 	    } else if ( ch=='e' ) {
-		sprintf( unicode,"%d", (int) map->backmap[sc->orig_pos] );
-		for ( pt=unicode; *pt!='\0' && buffer<bend; )
-		    *buffer++ = *pt++;
-	    } else
+        sprintf( unicode,"%d", (int) map->backmap[sc->orig_pos] );
```
skef, line 736: We try not to reformat blocks that aren't being changed otherwise so that `git blame` gives a more usefully direct answer.

fontforge/svg.c
```diff
@@ -766,60 +766,75 @@ SplineChar *SCHasSubs(SplineChar *sc, uint32 tag) {
...
-		unicode_alternates[encuni>>8]!=NULL &&
-		(alt = unicode_alternates[encuni>>8][encuni&0xff])!=NULL &&
-		alt[1]=='\0' )
-	    /* For arabic forms use the base representation in the 0600 block */
-	    fprintf( file, "unicode=\"&#x%x;\" ", alt[0]);
-	else if ( vs!=-1 )
-	    fprintf( file, "unicode=\"&#x%x;\" ", vs);
-	else
-	    fprintf( file, "unicode=\"&#x%x;\" ", encuni);
+		if ( encuni!=0x9 &&
```
skef, line 808: Same reformatting caveat here.

fontforge/svg.c
jtanx: I'm not a fan of that approach (of using a static fixed length buffer), but a lot of this code could be simplified by just using smprintf (and remembering to free the result afterwards)

frank-trampe: I also prefer dynamic allocation where practical. There are a lot of cases in which it is impossible to allocate a static buffer that is long enough for the top 0.1% but not ridiculously wasteful the rest of the time.

ctrlcctrlv:
I don't like this feature, and recommend against merging it. The right way to export to a filename based on glyph/font contents is the Python API. This also changes old behavior, doesn't it?&mdash;%N used to be literal and now is not. That could be confusing. Sorry, not a fan.

skef:
> %N used to be literal and now is not.

This is at least questionable -- when you have string substitution with percent symbols any currently unused letters are usually understood as being reserved.

fontforge/svg.c
yurks: There are a mix with tabs and spaces, so my IDE was confused i bit.
Agree, will be fixed

fontforge/svg.c
yurks: I would really appreciate if you could help to re-write this code in right way as c lang is not my main syntax.

yurks:
> I don't like this feature, and recommend against merging it. The right way to export to a filename based on glyph/font contents is the Python API. This also changes old behavior, doesn't it?—%N used to be literal and now is not. That could be confusing. Sorry, not a fan.

%n used for name for now, a lowercase. So, this not a change, but additional feature. We could choose %l instead, if it's ok

ctrlcctrlv:
I withdraw my concern.

fontforge/svg.c
yurks: refactored

fontforge/cvexport.c
```diff
@@ -702,17 +702,24 @@ static void MakeExportName(char *buffer, int blen,char *format_spec,
...
+		for ( ; *pt!='\0' && buffer<bend; ) {
 		    if ( isupper( *pt ))
 			*buffer++ = '$';
 		    *buffer++ = *pt++;
 		}
 #else
-		for ( pt=sc->name; *pt!='\0' && buffer<bend; )
+		for ( ; *pt!='\0' && buffer<bend; )
 		    *buffer++ = *pt++;
 #endif
```
skef, line 724: Freeing here is more of a pain because if there is no ligature you just write the glyph name, and you don't want to free *that* because there's still a pointer to it. So you could copy it also at the start but then you'd just have to free it right away if there is a ligature. So it's probably easiest to just do

```
if (pt!=sc->name)
    free(pt);
```
It's not safe to do `free(ligature);` here because if `ch!='N'` that won't be initialized/current. But you could do
```
if (ch=='N')
    free(ligature);
```
instead.

fontforge/svg.c
```diff
@@ -770,26 +770,50 @@ return( SFGetChar(sc->parent,-1,pst->u.subs.variant));
...
+		    	r = sprintf(buf2, "&#x%x;", (unsigned int) univals[i]);
+		    }
+		    n = LIGATURE_MAXLEN - strlen(buf) - 1;
+		    if (r > 0 && n >= r) {
+		        strncat(buf, buf2, n);
+		    } else {
+		        break;
+		    }
+		}
+        return (  copy(buf)  );
```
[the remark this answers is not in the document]
yurks: fixed, thank you.

fontforge/svg.c
```diff
@@ -770,26 +770,50 @@ return( SFGetChar(sc->parent,-1,pst->u.subs.variant));
...
+char *GetLigature(SplineChar *sc) {
+    PST *best = HasLigature(sc);
+    if ( best != NULL ) {
+    	int32 univals[50];
+        char buf[LIGATURE_MAXLEN] = "";
+        char buf2[15];
+        int i, c, r, n = 0;
+
+		c = LigCnt(sc->parent,best,univals,sizeof(univals)/sizeof(univals[0]));
+		for ( i=0; i<c; ++i ) {
```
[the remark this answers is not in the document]
yurks: Thank you for pointing to `str*` functions singularity, I optimize this part again.

(And yes, I know that sort of cultural issue where foreigners hears moralizing from locals about his verb form choosing. Sorry for my clueless with this code, now it looks like you have avenged for him :). Again, thank you for your time!

fontforge/cvexport.c
```diff
@@ -702,19 +702,29 @@ static void MakeExportName(char *buffer, int blen,char *format_spec,
...
+        if (ligature != NULL) {
```
skef, line 725: You just did what I said wasn't safe in my comment. It's not safe for two reasons:

1) If ch is not "N" before this line is reached (say it's "n"), then the ligature pointer will never be initialized. This could be fixed by initializing ligature to "NULL", but ... 
2) If ch has been "N" in the past and there was a ligature name, but it isn't currently "N", the `ligature` pointer will point to the previous entry, and this code will double-free it. 

Also note that freeing NULL is fine, the function knows to do nothing, so you never have to check for NULL prior to calling it. 

If you want to do this style of freeing, you would 

1. Set ligature to NULL at the top, and
2. Set ligature to NULL right after freeing it. 

That way it will only be non-NULL from a return from `GetLigature()`, and only for the lifetime of a single iteration.

fontforge/svg.c
```diff
@@ -770,26 +770,55 @@ return( SFGetChar(sc->parent,-1,pst->u.subs.variant));
...
+		    remain = LIGATURE_MAXLEN - stored;
+		    if ( (univals[i]>='A' && univals[i]<='Z') || (univals[i]>='a' && univals[i]<='z') || (univals[i] >= '0' && univals[i] <= '9') || univals[i] == '_' || univals[i] == '-') {
+		        r = snprintf( buf + stored, remain, "%c", univals[i] );
+		    } else {
+                r = snprintf( buf + stored, remain, "&#x%x;", (unsigned int) univals[i] );
+		    }
+
+            if ( r < 0 || r >= remain) {
+               if ((r - remain) > 0) {
+       		       snprintf( buf + stored, remain, "");
```
skef, line 794: This is a little strange. If the goal here to ensure zero-termination as if the last `snprintf()` hadn't been run it would be clearer to just do that directly, as in `*(buf + stored) = 0;` .

fontforge/svg.c
```diff
@@ -770,26 +770,55 @@ return( SFGetChar(sc->parent,-1,pst->u.subs.variant));
 return( NULL );
 }
 
+#define LIGATURE_MAXLEN 150
+
+char *GetLigature(SplineChar *sc) {
+    PST *best = HasLigature(sc);
+    if ( best != NULL ) {
+    	int32 univals[50];
```
jtanx, line 778: The indentation in this function is all over the place

fontforge/svg.c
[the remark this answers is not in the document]
jtanx: I'd be preferential to replace all of this with using a string builder like [this one](https://developer.gnome.org/glib/stable/glib-Strings.html#g-string-append-printf). I'm pretty sure we've used it in at least one other place too.

fontforge/svg.c
```diff
@@ -770,26 +770,55 @@ return( SFGetChar(sc->parent,-1,pst->u.subs.variant));
...
+        char buf[LIGATURE_MAXLEN] = "";
+        int i, c, r, stored, remain;
+        stored = 0;
+		c = LigCnt(sc->parent,best,univals,sizeof(univals)/sizeof(univals[0]));
+
+		for ( i=0; i<c; ++i ) {
+		    remain = LIGATURE_MAXLEN - stored;
+		    if ( (univals[i]>='A' && univals[i]<='Z') || (univals[i]>='a' && univals[i]<='z') || (univals[i] >= '0' && univals[i] <= '9') || univals[i] == '_' || univals[i] == '-') {
```
jtanx, line 786: `isalnum(univals[i]) || univals[i]=='_' || univals[i]=='-'`

fontforge/svg.h
```diff
@@ -19,5 +19,6 @@ extern SplineFont *SFReadSVGMem(char *data, int flags);
 extern SplineSet *SplinePointListInterpretSVG(char *filename, char *memory, int memlen, int em_size, int ascent, int is_stroked, ImportParams *eip);
 extern void SFLSetOrder(SplineFont *sf, int layerdest, int order2);
 extern void SFSetOrder(SplineFont *sf, int order2);
+extern char *GetLigature(SplineChar *sc);
```
jtanx, line 22: It's kind of weird having this function in the svg source/header. The name might be a bit misleading too, something more like `SCGetLigatureDescription`?

ctrlcctrlv:
Code has very serious problems and no engagement from originator since April. Closing—can be reopened if a new commit is made to [`yurks:ligature-format`](https://github.com/yurks/fontforge/tree/ligature-format).

PR #4289, opened 2020-04-20
When a glyph name matches a block name, go to the glyph

dscorbett:
The “Goto” dialog says “Enter the name of a glyph in the font”, but entering the name of a Unicode range works too. If a glyph and a range share a name, it goes to the range. I think this is surprising behavior, considering the dialog only mentions glyphs, not ranges. I noticed this when trying to go to the glyph `Vai` in Noto Sans Bhaiksuki. (Vai is, coincidentally, a Unicode block.)

### Type of change
- **Breaking change**

### DIFF
--- a/fontforgeexe/gotodlg.c
+++ b/fontforgeexe/gotodlg.c
@@ -91,7 +91,10 @@ static int Goto_OK(GGadget *g, GEvent *e) {
 	gw = GGadgetGetWindow(g);
 	d = GDrawGetUserData(gw);
 	ret = GGadgetGetTitle8(GWidgetGetControl(gw,CID_Name));
-	if ( d->ranges!=NULL ) {
+	d->ret = NameToEncoding(d->sf,d->map,ret);
+	if ( d->ret<0 || (d->ret>=d->map->enccount && d->sf->cidmaster==NULL ))
+	    d->ret = -1;
+	if ( d->ret==-1 && d->ranges!=NULL ) {
 	    for ( i=0; d->ranges[i].text!=NULL; ++i ) {
 		if ( strcmp(ret,(char *) d->ranges[i].text)==0 ) {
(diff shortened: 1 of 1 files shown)


fontforgeexe/gotodlg.c
```diff
@@ -91,7 +91,10 @@ static int Goto_OK(GGadget *g, GEvent *e) {
 	gw = GGadgetGetWindow(g);
 	d = GDrawGetUserData(gw);
 	ret = GGadgetGetTitle8(GWidgetGetControl(gw,CID_Name));
-	if ( d->ranges!=NULL ) {
+	d->ret = NameToEncoding(d->sf,d->map,ret);
+	if ( d->ret<0 || (d->ret>=d->map->enccount && d->sf->cidmaster==NULL ))
```
skef, line 95: Non-rhetorical question: what happens here if `d->ret>=d->map->enccount` but `d->sf->cidmaster!=NULL`?

dscorbett: `d->ret>=d->map->enccount` would normally mean that `d->ret` is not a valid encoding slot, but if `d->sf->cidmaster!=NULL` the implication is that it is valid if you take the CID map into account. So, in that case, the function ends up returning the current valid `d->ret`, without popping up the error dialog.

skef: Would `d->ret>=d->map->enccount` be an unexpected result when `d->sf->cidmaster==NULL` that the code is protecting against just in case? Otherwise, can there be the analogous problem when `d->sf->cidmaster!=NULL` (it's a CID font but the value is still bogus)?

dscorbett: I don’t know. I just swapped the two blocks, so if it was bogus before, it is still bogus.

PR #4284, opened 2020-04-18
CID fixes

jtanx:
* Fix use after free on the call to SFFlatten (it frees the sf
  that the cidmaster belongs to). Matches behaviour to what
  FVMenuFlatten does
* Ensure that the FontViewReformatAll call actually runs if
  executing the script from the GUI, prevents another use after
  free on stale FV show/filled BDFFonts
* Fix GID allocation on cid keyed fonts in SFAddGlyphAndEncode

<details>
<summary>Asan error for second case</summary>

```
=================================================================
==7233==ERROR: AddressSanitizer: heap-use-after-free on address 0x61a0000924f0 at pc 0x7f2213311a27 bp 0x7fff662193f0 sp 0x7fff662193e0
READ of size 4 at 0x61a0000924f0 thread T0
    #0 0x7f2213311a26 in BDFPieceMeal ../fontforge/splinefill.c:1677
    #1 0x557fe410ac1e in FVDrawGlyph ../fontforgeexe/fontview.c:262
    #2 0x557fe4114da0 in FVExpose ../fontforgeexe/fontview.c:6014
    #3 0x557fe4125f47 in v_e_h ../fontforgeexe/fontview.c:6733
    #4 0x557fe435a90d in _GWidget_Container_eh ../gdraw/gcontainer.c:281
    #5 0x557fe4383c0e in _GGDKDraw_CallEHChecked ../gdraw/ggdkdraw.c:346
    #6 0x557fe4385c44 in _GGDKDraw_DispatchEvent ../gdraw/ggdkdraw.c:1249
    #7 0x7f2211e2f764  (/usr/lib/x86_64-linux-gnu/libgdk-3.so.0+0x37764)
    #8 0x7f2211e3f7f4  (/usr/lib/x86_64-linux-gnu/libgdk-3.so.0+0x477f4)
    #9 0x7f2211e3f897  (/usr/lib/x86_64-linux-gnu/libgdk-3.so.0+0x47897)
    #10 0x7f2211e409f5  (/usr/lib/x86_64-linux-gnu/libgdk-3.so.0+0x489f5)
    #11 0x7f2211e40bef  (/usr/lib/x86_64-linux-gnu/libgdk-3.so.0+0x48bef)
    #12 0x7f221143d10c in g_closure_invoke (/usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0+0x1010c)
[...]

PR #4293, opened 2020-04-23
Add image/x-inkscape-svg to accepted mime types + cleanup

### LINKED ISSUE #4291: Inkscape copy&paste always a bitmap
= SYSTEM =
FontForge 2020-03-14
Windows 10, x64, up-to-day
Inkscape 0.92.4

= ISSUE =
If you copy & paste the char "o", FontForge renders it as a bitmap image always.

If you try to import it, FontForge will crash always.

Note 1: the output doesn't change if you try to import it as two paths or as a single one.
Note 2: the output doesn't change if you try to import a "plain SVG" version of the file.
Note 3: copy & paste seems broken as everything that is pasted from Inkscape is rendered as a bitmap image.

= EXPECTED BEHAVIOR =
To be able to copy the shapes as intended.

= HOW TO REPRODUCE =
Open SVG in Inkscape. Copy the shape. Open the character in FontForge and paste it: it will be a bitmap.

Open the character in a new font in FontForge and try to import the SVG. It will crash.

Thanks in advance,
   Franz

[o.zip](https://github.com/fontforge/fontforge/files/4516267/o.zip)
(Seriously... add support for SVG files...)

jtanx:
So I took a further look and it seems like the problem is Inkscape is sending `image/x-inkscape-svg`, followed by image formats like png/bmp. So make FF accept that mime type. Not sure if Inkscape changed the mime type sent or something.

Fixes #4291

### Type of change
- **Non-breaking change**

### DIFF
--- a/fontforge/cvundoes.c
+++ b/fontforge/cvundoes.c
@@ -2107,43 +2107,52 @@ static void PasteNonExistantRefCheck(SplineChar *sc,Undoes *paster,RefChar *ref,
     }
 }
 
+static const char* SCGetPasteableClipboardType(int* type) {
+    static const char* mimes[] = {
+        "image/svg+xml",
+        "image/svg-xml",
+        "image/svg",
+        "image/x-inkscape-svg",
+        "image/eps",
+        "image/ps",
+#ifndef _NO_LIBPNG
+        "image/png",
+#endif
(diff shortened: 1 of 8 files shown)


jtanx:
#4284 has the travis fixes by the way.

PR #4297, opened 2020-04-26
Fix size specifiers in the python binding

### LINKED ISSUE #4295: Test failures on ppc64
On Gentoo Linux, we get a few test failures on ppc64 with fontforge-20200314. Please advise on any additional information you need to diagnose the issue.
```
The following tests FAILED:
	54 - test0004_py (Failed)
	55 - test0004_pyhook (Failed)
	80 - test1009_py (Failed)
	81 - test1009_pyhook (Failed)
```
[LastTest.log](https://github.com/fontforge/fontforge/files/4530100/LastTest.log)

Gentoo bug: https://bugs.gentoo.org/719058

CC: @asarubbo

jtanx:
This should fix #4295. I believe the problem came from ppc64 having different sized int/longs vs on amd64. Read the commit messages for more information.

### Type of change
- **Bug fix**

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -760,7 +760,7 @@ static PyObject *PyFF_UnicodeAnnotationFromLib(PyObject *UNUSED(self), PyObject 
     char *temp;
     long val;
 
-    if ( !PyArg_ParseTuple(args,"|i",&val) )
+    if ( !PyArg_ParseTuple(args,"|l",&val) )
 	return( NULL );
 
     if ( (temp=unicode_annot(val))==NULL ) {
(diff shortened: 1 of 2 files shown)


floppym:
> I believe the problem came from ppc64 having different sized int/longs vs on amd64. 

I think sizeof(long) == 8 on both systems, but using the wrong format string has different effects. Assuming sizeof(int) == 4, using "i' in the format string would set the first 4 bytes in the resulting value.

On little-endian, the low order bytes get set.
On big-endian, the high order bytes get set.

That would explain why the code seems to work on little-endian ppc64.

floppym:
The CI failure looks unrelated.
```
Warning, treated as error:
/home/travis/build/fontforge/fontforge/doc/sphinx/glossary.rst:23:term alphabet not found in case sensitive match.made a reference to Alphabet instead.
ninja: build stopped: subcommand failed.
```

gyakovlev:
I confirm this fixes the tests on ppc64be
and still passes the test on ppc64le

build+test log from big-endian machine attached

[fontforge-build.log](https://github.com/fontforge/fontforge/files/4540422/fontforge-build.log)

jtanx:
Thanks for confirming, and yeah our own CI is busted until #4284 is merged.

PR #4160, opened 2020-02-13
If converting UTF-8⇔UTF-8, use existing functions

ctrlcctrlv:
The versions at these points were non-standard conforming, and encoded characters as surrogate pairs if possible, even though surrogate pairs are explicitly forbidden in UTF-8.

https://unicodebook.readthedocs.io/issues.html#non-strict-utf-8-decoder-overlong-byte-sequences-and-surrogates

(It took me more than three hours to come up with this solution, as I had to check basically all the code that could possibly corrupt this, and I wasted the first hour thinking that the filenames _themselves_ had the problem.)

This closes #4159, which contains a test case. But in short, the easiest way to test this now that I know the problem is to just make a file named an emoji and observe the crash.

### Type of change

- **Bug fix**

PR #4301, opened 2020-04-27
ustring.c: u2utf8_copyn: truncate at the first invalid utf8 char instead of returning null

### LINKED ISSUE #4294: Crash when getting surrogate glyph info
Opening the “Glyph Info” window for a surrogate crashes FontForge. I am using version 20200314 on macOS 10.14.6. To reproduce:

* Select a surrogate in the main grid, e.g. U+D800.
* Open the “Element” menu.
* Click “Glyph Info...”.

[Here is the crash report.](https://github.com/fontforge/fontforge/files/4529202/fontforge_2020-04-24-094345.txt)

jtanx:
Fixes #4294. e.g. from that case, a single surrogate is not valid utf-8

### Type of change

- **Bug fix**

### DIFF
--- a/Unicode/ustring.c
+++ b/Unicode/ustring.c
@@ -655,18 +655,19 @@ char *u2utf8_copy(const unichar_t *ubuf) {
 
 char *u2utf8_copyn(const unichar_t *ubuf,int len) {
 /* Make a utf8 string copy of unichar string ubuf[0..len] */
-    char *utf8buf, *pt;
+    char *utf8buf, *pt, *pt2;
 
     if ( ubuf==NULL || len<=0 || (utf8buf=pt=(char *)malloc(len*6+1))==NULL )
 	return( NULL );
 
-    while ( (pt=utf8_idpb(pt,*ubuf++,0)) && --len );
-    if ( pt ) {
-	*pt = '\0';
-	return( utf8buf );
-    }
-    free( utf8buf );
-    return( NULL );
+    while ( (pt2=utf8_idpb(pt,*ubuf++,0)) && --len )
+	pt = pt2;
+
+    if ( pt2 )
+	pt = pt2;
+
+    *pt = '\0';
+    return( utf8buf );
 }
 
 int32 utf8_ildb(const char **_text) {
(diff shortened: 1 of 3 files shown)


ctrlcctrlv:
Did I cause this regression? Change looks fine to me.

Unicode/ustring.c
```diff
@@ -655,18 +655,19 @@ char *u2utf8_copy(const unichar_t *ubuf) {
...
-	return( utf8buf );
-    }
-    free( utf8buf );
-    return( NULL );
+    while ( (pt2=utf8_idpb(pt,*ubuf++,0)) && --len )
+	pt = pt2;
+
+    if ( pt2 )
+	pt = pt2;
+
```
ctrlcctrlv, line 668: Now that I'm on my PC I can confirm that I did not cause this. I thought I may have in #4160, but that didn't touch this function.

I think a developer `TRACE` warning should be added here on truncation. It seems like the kind of thing to make debugging difficult when we're expecting a string of one length and getting a length of another. We may even want to consider changing the function signature to int *len so lengths can be verified as needed.

inc/basics.h
```diff
@@ -67,9 +67,9 @@ typedef uint32 unichar_t;
 /* A macro to print a string for debug purposes
  */
 #ifndef NDEBUG
-#define TRACE(...) printf(__VA_ARGS__)
+#define TRACE(...) fprintf(stderr, __VA_ARGS__)
 #else
-#define TRACE(...)
+#define TRACE(...) while(0)
```
ctrlcctrlv, line 72: I don't understand how this even compiles (I assumed you'd want `do {} while (0)`), but it does, so this looks fine to me now :-)

skef:
```
TRACE("u2utf8_copyn: truncated on invalid char 0x%x\n", ubuf[-1]);
```
(for example) is replaced by:
```
while(0);
```
which one could read as:
```
while (false)
    ;
```

ctrlcctrlv: Oh, I see, okay, the semicolon makes it work. I thought a statement was needed, but I just tested it, and `if(true);` also is a no-op. Neat!

skef: Various different conventions turn up for this but they're often partly intended to "eat" the trailing semicolon on the TRACE or ASSERT or whatever, either because leaving it would be syntactically illegal or (more commonly) to avoid lint or compiler warnings.

PR #4313, opened 2020-05-02
If point numbering shown, update as splines change

### LINKED ISSUE #4312: When using the "Number Points" feature, changes to splines aren't seen without pressing it again
View&rarr;Number Points&rarr;PostScript&reg;

Adding points results in multiple points labeled "0". Deleting points results in wrong point numbering.

This is a tracking issue, I already fixed this locally.

ctrlcctrlv:
Close #4312

Definitely helps when making two glyphs compatible (in the variable fonts sense).

### Type of change

- **Bug fix**

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -5243,6 +5243,8 @@ static void _CV_CharChangedUpdate(CharView *cv,int changed) {
 	/* If we changed the grid then any character needs to know it */
 	FVRedrawAllCharViewsSF(cv->b.sc->parent);
     }
+    if ( cv->showpointnumbers || cv->show_ft_results )
+	SCNumberPoints(cv->b.sc, cvlayer);
     cv->recentchange = false;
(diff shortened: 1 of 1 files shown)


jtanx:
Hmm, so if I understand correctly calling SCNumberPoints iterates the whole spline set to renumber them. Sounds a bit expensive to be doing this on every CV update.

ctrlcctrlv:
@jtanx It won't do the iteration before you let go of the mouse, or before a change is made. It is a bit expensive, yes, but there's no reason to enable the point numbering and then want to ever see incorrect point numbers, and especially not two points numbered "0" or "??" as happens now.

I think George originally added this feature to help with MM fonts, and that is its main purpose today too. As variable fonts are becoming more and more common, along with the production of the same, what you do to make two glyphs compatible is you open two fonts, two chars, enable the point numbering in each.

jtanx:
Yeah ok, I can't think of a better catch-all apart from inserting this at every possible place a point could be added/removed so this is probably fine.

PR #4309, opened 2020-05-01
Add an option to suppress the FFTM table

dscorbett:
Closes #2389.

### Type of change
- **New feature**

PR #4317, opened 2020-05-03
Fix UI segmentation fault introduced in #4309

ctrlcctrlv:
Per https://github.com/fontforge/fontforge/pull/4309#issuecomment-623050441

### Type of change

- **Bug fix**

### DIFF
--- a/fontforgeexe/savefontdlg.c
+++ b/fontforgeexe/savefontdlg.c
@@ -847,8 +847,8 @@ static void SaveOptionsDlg(struct gfc_data *d,int which,int iscid) {
     gcd[k].gd.cid = CID_TTF_TeXTable;
     gcd[k++].creator = GCheckBoxCreate;
     hvarray2[37] = &gcd[k-1]; hvarray2[38] = GCD_ColSpan; hvarray2[39] = NULL;
-    hvarray2[41] = GCD_Glue; hvarray2[42] = GCD_Glue; hvarray2[43] = NULL;
-    hvarray2[44] = NULL;
+    hvarray2[40] = GCD_Glue; hvarray2[41] = GCD_Glue; hvarray2[42] = NULL;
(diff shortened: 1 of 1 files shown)


jtanx:
Hmm, the segfault is fixed but it looks like the UI doesn't reflect what the setting actually is. e.g. uncheck the FFTM table option, generate. Open the generate dialog again and go to options -> FFTM Table option is checked. But when generating, the FFTM table is omitted. You have to un-check and recheck the option to get the FFTM table generated again.

skef:
I think lines 471-2:
```
    GGadgetSetChecked(GWidgetGetControl(gw,CID_TTF_FFTMTable),
            (flags&~ttf_flag_noFFTMtable) && which!=0);
```

Should be something like:
```
    GGadgetSetChecked(GWidgetGetControl(gw,CID_TTF_FFTMTable),
            !(flags&ttf_flag_noFFTMtable) && which!=0);
```
but I don't understand how this line isn't overridden by line 510:
```
    GGadgetSetEnabled(GWidgetGetControl(gw,CID_TTF_FFTMTable),which!=0);
```
(Or, for that matter, what's going on with lines 505 and 508.) 

Actually, maybe I do understand a little. given that `ps_flag_nohints` follows the same pattern (check off maps to flag on) perhaps line 510 should be replaced by:
```
    if ( which!=0 && (flags&ttf_flag_noFFTMtable) )
        GGadgetSetChecked(GWidgetGetControl(gw,CID_TTF_FFTMTable), false);
```
To turn the selection off in that case. But I'm still confused about the relation between the earlier and later calls.

ctrlcctrlv:
Let's just revert. Clearly the changeset is bad. I tested it with Python, and that works, but you're all right that the UI is messed up. Closing, will open a revert PR and merge that.

PR #4320, opened 2020-05-04
Add an option to suppress the FFTM table

### LINKED ISSUE #2389: Generate: Don't include FFTM

dscorbett:
This redoes #4309 and closes #2389.

~It seems like the only problem with #4309 was in the GUI code, so let’s restore the scripting code and I can figure out the GUI crash later. (I did test the options dialog window before opening the first PR, so it may be a while before I can even reproduce the crash.)~

### Type of change
- **New feature**


ctrlcctrlv:
Sorry, I'm not inclined to accept this without you exposing the option in the GUI. The GUI shouldn't be treated as second class

jtanx:
The crash was fixed with #4317, it was just the behaviour around it being checked/unchecked not reflecting what the true state was

skef:
@dscorbett If you add the GUI code from the last PR back into his one I'll attempt to get it working. I have medium-good grasp on what the problems were.

skef:
I'll take a look at this this evening and if there are remaining changes to be made I'll try to do so.

skef:
(My inability to distinguish `GGadgetSetChecked` from `GGadgetSetEnabled` led to confusion in my comments on the last PR.)

I've both reviewed this code as is and tested it and it seems fine in appearance and behavior now. I think the previous remaining issue was the now-correct flag check. @jtanx can you verify?

jtanx:
Yeah seems fine now

Issue #4125, opened 2020-02-02
Rethink nonextcp/noprevcp for cubic splines


skef:
#2601 is an example of a more pervasive problem: FontForge doesn't systematically address the problem of a missing cubic spline control point. There are many cases where FF will let you just eliminate one CP and then produce poor output downstream.

A cubic spline with two missing (or, equivalently, on-curve-point-colocated) control points is a line, and is fine. It should be output in point form without both points. A cubic spline with one missing control point is mostly a problem -- ideally there would be none of these. 

One can of course just check to see if one is outputting a cubic spline and output a co-located control point, and just editing our various "output" routines would be one approach. However, these splines have undefined slopes at the co-located end, which could screw up some calculations in external tools. It would be better, either always or optionally, to output a control point 1) very slightly away from the on-curve point at 2) the slope calculated at a small t-jump away (small such as 1e-10). That would retain the shape while creating a fully defined spline.

However, there are many contexts *within* FontForge where one would want that correction. So one could imagine doing this instead on setting nonextcp/noprevcp and every subsequent change to the on-curve points or the remaining control point. The place for that is probably in SplineRefigure(); I suppose the question with that is whether and when folks consider it kosher to modify SplinePoint values via that function. Given that SplineRefigure (often through SplineMake) is supposed to be called on every geometry change it would be a tidy place to implement the change. 

I suppose the problem with "always" instead of "optionally" doing that is that some people will want rounded point values and therefore  won't be happy with the "tiny-slope" fix. But we can check nonextcp and noprevcp on behalf of those folks and just output the colocated point, given some Pref.

ctrlcctrlv:
@skef Question: I thought a bit about the rounding problem. If someone asks for rounded control points, and a control point is almost colocated, I suppose the problem is that the point will become colocated, with no "almost".

So, in the rounding routine, why can't we correct for this? If rounding will cause a control point that is not already colocated to become colocated, increase the tiny-slope until it's above where rounding would happen.

That way we can always write out colocated CPs.

skef:
Since I wrote this I've browsed the cubic spline code in more tools and it seems as though most just special-case the co-located point problem. (Which, to be explicit, is that there is no mathematically defined tangent at the endpoint of the spline.) 

My idea of positioning the point to reproduce the "implicit tangent" may be nifty but it's probably impractical. 

[...]

PR #4321, opened 2020-05-07
Optionally only interpolate compatible glyphs/layers

ctrlcctrlv:
"Compatible" here meaning, "would work in a variable font". So, same number of contours/points (on/off curve), and same contour directions.

* Adds `only_compatible` flag to font.interpolateFonts, and `strict` to layer.interpolateNewLayer, glyph.createInterpolatedGlyph.
* Exposes "Only compatible glyphs" option in the GUI (Element→Interpolate Fonts).

This expands on my work in #4313 and is part of a greater effort to make FontForge better at working with variable fonts.

### Type of change

- **New feature**


fontforge/splineutil.c
```diff
@@ -7953,3 +7953,19 @@ bigreal DistanceBetweenPoints( BasePoint *p1, BasePoint *p2 )
...
+
+int SplineSetNumberPoints(SplineSet *ss, int order2) {
+    int pnum=0;
+    if ( order2 ) {
+        return SSTtfNumberPoints(ss);
+    } else {
+        for ( SplinePoint *sp = ss->first; ; ) {
+            pnum++;
+            if ( sp->next==NULL ) break;
+            if ( !sp->nonextcp || !sp->next->to->noprevcp ) pnum += 2;
```
skef, line 7965: I don't understand this -- if you have one cubic contour with eight on-curve points and fourteen off-curve points, and another cubic contour with 20 on-curve points and two off-curve points, are they somehow interpolation-compatible? I would have thought interpolation mostly cares about on-curve points and can fudge transitioning lines to curves by considering missing off-curve points to be co-located.

fontforge/splineutil.c
```diff
@@ -7953,3 +7953,19 @@ bigreal DistanceBetweenPoints( BasePoint *p1, BasePoint *p2 )
     t = sqrt( t );
     return t;
 }
+
+int SplineSetNumberPoints(SplineSet *ss, int order2) {
+    int pnum=0;
+    if ( order2 ) {
+        return SSTtfNumberPoints(ss);
```
skef, line 7960: Seems like this clause is a better place for `SplinePointListCopy1()` given that we don't need it for the cubic case. Might even be better to make a `SSTtfNumberPoints1` with its "innards" and that takes a boolean to control whether it modifies its argument. Then `SSTtfNumberPoints()` can just call that for each SS with `false` and you can call it here with `true` (or the other way around -- I don't have any intuition for the boolean polarity).

fontforgeexe/fvfontsdlg.c
```diff
@@ -315,8 +318,10 @@ return( true );
 	}
 	if ( fv==NULL )
 	    InterAskFilename(d->fv,last_amount/100.0);
-	else
-	    FontViewCreate(InterpolateFont(d->fv->b.sf,fv->b.sf,last_amount/100.0,d->fv->b.map->enc),false);
+	else {
```
skef, line 321: I'll check the GUI stuff closer next time around.

fontforge/splineutil.c
ctrlcctrlv: Yes, they are compatible. The distinction does not matter to the interpolator, it only considers points, not whether they are on or off curve.

skef: Do the results make sense? I take it the idea with this PR is to add what amounts to an "only do this if it makes sense" flag. We might want to differentiate between cases the algorithm can operate on and cases with a chance of having a good result.

ctrlcctrlv: Sure, that makes sense to me. I'll modify it to count off- and on-curve points separately, as we are calling this "strict" mode in the API so it makes sense that that would be a strict interpretation.

skef: I'm transitioning from reviewing to spit-balling here. 

Even assuming a relatively boring interpolator that is not trying to help the user *much* I think having the number of on-curve points match is what should matter. If a segment with control points is being matched up with a line there are some pretty simple heuristics the code can use for the line, for example (considering cubics):

1. Add co-linear control points 1/3 and 2/3 of the way along the line, then interpolate. 
2. (Getting a bit more sophisticated) Look at the distance of each control point on the curved segment compared with the distance between the on-curve points and add co-linear control points to the line at corresponding ratios, then interpolate. 

Now, whether FontForge's interpolation code can do this is another matter.

ctrlcctrlv: It can't. The standard also rejects such heuristics. Given that our (my) ultimate goal with adding this is to help people produce CFF2 variable fonts, I disagree with your suggestion that I should ignore the number of off curve points. Improving FontForge's interpolator does nothing to improve the interpolator actually in use by e.g. FreeType.

skef: That's fine but I think it means that you can't take this approach at all. Knowing the number of off-curve points is pretty meaningless if you don't know how they are distributed among the on-curve points (isn't it?). Unless you want to do something exotic you'll probably need to iterate through the contour pairs to make sure they match up.

Incidentally, what does the standard say about co-located control points? The current code in the PR seems to be mapping the single point case onto one in which the other control point is co-located with the on-curve point. That's probably fine, and FontForge probably shouldn't be marking those `no?cp` anyway (see #4125) but the spec might have something to say about it if it prohibits lines from getting the same treatment.

(An example of "exotic" would be forming a bit string where 0 means a line and 1 means a curve and returning its checksum.)

skef: If you'd rather chat about this in some other setting sometime I'm happy to do that too. The other review issues are pretty ordinary while this one is close to the heart of the feature.

ctrlcctrlv: Still got time? I'm in IRC

ctrlcctrlv: By the way, for future reference, just joining IRC (#fontforge) and saying "copypaste" will alert me if I'm at PC :-)

My client is more relaxed for other channels where people usually say my name on accident (e.g. "copypaste this into a terminal") but I watch #fontforge very closely

fontforge/fvfonts.c
```diff
@@ -1647,6 +1697,11 @@ return( NULL );
 	new->layers[i].order2 = base->layers[i].order2;
     }
     for ( i=0; i<base->glyphcnt; ++i ) if ( base->glyphs[i]!=NULL ) {
+	if (!InterpolationSanity(base->glyphs[i]->layers[ly_fore].splines, other->glyphs[i]->layers[ly_fore].splines,
+	                         base->glyphs[i]->layers[ly_fore].order2)) {
+		LogError(_("Glyph %s is incompatible."), base->glyphs[i]->name);
```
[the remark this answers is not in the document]
ctrlcctrlv: To your first review item, that's done in the latest revision.

To the second, I disagree. You'll already get a bunch of warnings in master&mdash;https://github.com/fontforge/fontforge/blob/ce383f93c8e35e7755f41c8dd8a1f61c7d30d426/fontforge/fvfonts.c#L1367

I removed this as my warnings are much more exact, so now you get about the same number of warnings as before. If the user isn't interested in why their glyphs look all weird, well, I don't understand how they wouldn't be :-)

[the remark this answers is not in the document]
skef: The previous code didn't have your new option, so any glyphs that weren't compatible were errors in (close to) the same sense that trying to interpolate incompatible layers is an error. So it's proper that the old code warned on every incompatible glyph.

Your new option allows only compatible glyphs to be interpolated, which makes those that are skipped not errors in the same sense.

ctrlcctrlv:
I have no idea what I did to break the AppVeyor build and no ability to restart it. Any idea @jtanx ? Tests pass for me on Linux (Xubuntu VM).

I guess I might have to figure out how to build FontForge on Win32 if you don't know.

fontforge/fvfonts.c
```diff
@@ -1267,6 +1267,86 @@ return;
...
+            bsp = bsp->next->to;
+            osp = osp->next->to;
+            if ( bsp==bss->first && osp==oss->first ) break;
+            else if ( bsp==bss->first || osp==oss->first ) {
+                ret = false; break;
+            }
+        }
+    }
+
+    if (prefix[0] != '\0') free(prefix);
```
skef, line 1346: I think it's better in general for the allocating level to do the free -- it knows how the memory was allocated and the lifecycle of the string.

jtanx:
Can you see if you can rerun it now? Wouldn't be surprised if some package we install got updated and some of the deps are too far behind (we usually don't run `pacman -Syu`)

fontforge/fvfonts.c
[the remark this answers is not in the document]
ctrlcctrlv: @skef I still think it's helpful to know _why_ FontForge didn't include a certain glyph if you were expecting it to. If not, I don't really see who the warnings are harming, they can be easily ignored. (Despite FontForge's strange function name, `LogError` really calls `(ui_interface->logwarning)`, but I'm sure you know that. `IError` is for actual errors.)

ctrlcctrlv:
I can re-run it now, and did so. Let's see if that helps :-)

fontforge/fvfonts.c
```diff
@@ -1267,6 +1267,86 @@ return;
...
+        int bicw = SplinePointListIsClockwise(bss), oicw = SplinePointListIsClockwise(oss);
+
+        if (bicw == -1 || oicw == -1)
+            TRACE(_("%sInterpolationSanity: Assuming self-intersecting contour %d to be sane"), prefix, contour_count);
+        else if (bicw != oicw) {
+            LogError(_("%sIncompatible contour directionality (contour %d)"), prefix, contour_count);
+            ret = false;
+        }
+
+        int bpnum = SplineSetNumberPoints(bss, order2);
```
skef, line 1313: You don't need this anymore, the "after segment" error on (new) line 1326 reports the same info.

You therefore don't need the `order2` parameter (as long as you're not checking quadratic/cubic compatibility at this level).

ctrlcctrlv:
Nope, failed again with same message.

fontforge/fvfonts.c
```diff
@@ -1612,7 +1690,7 @@ static void InterpFixupRefChars(SplineFont *sf) {
 }
 
 SplineFont *InterpolateFont(SplineFont *base, SplineFont *other, real amount,
```
skef, line 1692: See my note about `PyFFFont_CreateInterpolatedGlyph()` below, which also applies here.

jtanx:
Let me try building locally. Fwiw it's pretty easy to build nowadays, install [this](https://www.msys2.org/), pacman install the dependencies and build as you would on linux.

fontforge/python.c
```diff
@@ -17105,16 +17119,16 @@ return( PySC_From_SC_I( sc ));
 static PyObject *PyFFFont_CreateInterpolatedGlyph(PyFF_Font *self, PyObject *args) {
     FontViewBase *fv;
     SplineFont *sf;
-    PyObject *from, *to;
+    PyObject *from, *to, *stricto=NULL;
```
skef, line 17122: The layer code seems complete to me. This function calls `SplineCharInterpolate()` which calls `LayerInterpolate()` which calls `SplineSetsInterpolate()` *and* `InterpRefs()`. A complete compatibility check also needs to verify that the references are interpolation-compatible.

ctrlcctrlv:
Thanks. It's 4:20AM here so that'll be all from me for the next six to eight hours, but if you don't find the problem I'll figure it out tomorrow. 👍 

I already have MSYS2 installed, so am allegedly halfway there! 👀

fontforge/splineutil.c
```diff
@@ -7953,3 +7953,21 @@ bigreal DistanceBetweenPoints( BasePoint *p1, BasePoint *p2 )
     t = sqrt( t );
     return t;
 }
+
```
skef, line 7956: Don't need this anymore.

tests/test1015.py
```diff
@@ -0,0 +1,23 @@
+﻿#Needs: fonts/InterpolateCompatibleCheck.sfd
+#Needs: fonts/InterpolateCompatibleCheck2.sfd
+
+import fontforge
+import sys
+
+f1 = fontforge.open(sys.argv[1])
+f2 = fontforge.open(sys.argv[2])
+
+f1["A"].layers["Fore"].interpolateNewLayer(f2["A"].layers["Fore"], 0.5, True)
```
skef, line 10: Tests could be more thorough in matching error conditions but that's not a requirement.

fontforge/fvfonts.c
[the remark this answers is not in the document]
skef: That's why I suggested an opt-in.

jtanx:
I think I know what the issue is, gcc was recently updated and it's reporting the dependencies to ninja incorrectly. Needs an upstream msys2 fix.

fontforge/fvfonts.c
ctrlcctrlv: Sorry, not sure I understand. This function is the allocating level. (Yes I said I was going to sleep, having trouble with that.)

fontforge/splineutil.c
ctrlcctrlv: Yeah, you're right. My excuse for keeping it in is pretty thin, it's "it's kind of nice to know how many points off we are if we're working on making a glyph compatible."

If you're convinced by that, I'll remove this tomorrow

fontforge/fvfonts.c
skef: Oh, I see, you're assigning above. I'm not sure I'm thrilled about that in terms of generality but I guess it's fine. Easy enough to refactor later.

fontforge/splineutil.c
skef: If you really want to keep it you can but you'll have to correct the Cubic path (it should only count on-curve points) and verify that `SSTtfNumberPoints()` really does what it needs to. Given that the function is unnecessary the work also seems unnecessary.

ctrlcctrlv: I guess an alternative is, in InterpolationSanity, if we hit the `(bnonextpt != ononextpt)` branch, just `int oextras = 0; while (osp != NULL) { extras++; osp = osp->next->to; }` if `bnonextcp`, then similar if `ononextcp`, then we can report in the warning something like &laquo;(base/other has %d extra unexpected points)&raquo;.

That's what I'll end up doing tomorrow. If I get some sleep :P

skef: Or just have a quick internal two-line loop to increment looking for `first` or `NULL`

ctrlcctrlv:
@skef Sorry for the early push&mdash;I'm too tired to think about some of the issues right at the moment, but I wanted to give you a chance to review the fixes I did make to make sure I'm on the right track, if you're up for that before I push the other half of the changeset, which will tackle the references, and do much more tests with quadratic fonts. :-)

fontforge/fvfonts.c
```diff
@@ -1304,40 +1307,42 @@ bool InterpolationSanity(SplineSet* base, SplineSet *other, int order2, char* pr
...
             if ( bcurve != ocurve ) {
                 char* wrongpttype = bcurve ? _("base") : _("other");
                 LogError(_("%sSegment %d in the %s spline is a curve when it should be a line"), prefix, segment_count, wrongpttype);
-                ret = false; break;
+                ret = false;
             }
-            if ( bnonextpt ) break;
-            bsp = bsp->next->to;
-            osp = osp->next->to;
+
```
skef, line 1326: The code below seems strangely organized to me. I would do something like:

```
           // This assumes you're verifying both are closed or open up front, which is worth doing to give it its own error message. 
            bool bnonextpt = ( bsp->next==NULL || bsp->next->to==bss->first );
            bool ononextpt = ( osp->next==NULL || osp->next->to==bss->first );

            if ( bnonextpt && ononextpt ) 
                  break;
            if ( bnonextpt != ononextpt ) {
                   char *problemspline = bnonextpt ? _("base") : _("other");
                   /* extra point count and error here */
                   ret = false;
                   break;
             }
             bsp = bsp->next->to;
             osp = osp->next->to;
```
My major substantial argument for this is that lines 1329 and 1330 selectively advance the individual pointers before the remaining checks and that's just really confusing.

skef:
I took a quick look at the reference stuff; I hadn't thought about this much before. The "reference philosophy" is to attempt to confirm that the referenced characters are the same and then interpolate the transformation matrices. 

Does that make sense with the new flag? It's complicated. In some cases you'll get a false positive: if the referenced glyphs don't interpolate then the derived glyphs shouldn't be considered to. You could address that with a recursive call to InterpolationSanity. In other cases you'll get a false negative: If the underlying glyphs are different (suppose one accent is taken from a spacing character and the other is from a composing character) InterpRefs will return false even though the contours might be interpolable. 

One way towards a more complete solution would be to add another option, called something like "allow_reference_breaking". If that's false you would just verify that the references are to the same glyph and that glyph is interpolable. If it's true and they aren't to the same glyph you would check the interpolability of the referred-to glyphs and do the interpolation directly (gettng the ordering right between the contour interpolation and the transformation matrix, if that matters -- haven't thought it through). 

Anyway, this needs a bit more thought about the scope of the fix, so it can be accurately described in the docs and/or extended in the implementation.

ctrlcctrlv:
I don't know...references will have to break in at least some cases if the glyphs they refer to are not interpolatable, because they won't appear in the file at all.

skef:
Hmm -- thinking about this a bit more it's silly to worry about the possibility of interpolating "poorly matched" fonts -- they'll almost never have compatible contours. So why attempt to deal with mismatched characters? 

Therefore I think what's needed here is to add the recursive call to InterpolationSanity somewhere (maybe in InterpRefs when "check" is enabled?) to verify that the underlying characters interpolate. That will tell you if that character will "go through" when interpolating. 

The other approach to that problem would be based on ordering -- always process the characters without references first and then check to see whether the referenced character exists in the new font. 

(All this also raises tricky questions about what to do with the glyph-level method.)

skef:
> I don't know...references will have to break in at least some cases if the glyphs they refer to are not interpolatable, because they won't appear in the file at all.

Well, this depends on whether you want to consider those derived glyphs interpolable or not. It's easiest to just say "if the referenced glyph doesn't interpolate then the derived glyph isn't 'sane'."

ctrlcctrlv:
Thanks for the prompt review as always. I think I'll be taking the ordering approach but I'm out of energy for tonight. :-)

ctrlcctrlv:
Oh, and also, for the glyph-level method, I'd say: "interpolate the refs as normal if the new glyph will be in the same font as glyphs A and B; otherwise if glyphs A and B are from different fonts or if the new glyph will be, flatten first."

fontforge/fvfonts.c
```diff
@@ -1280,35 +1283,138 @@ static RefChar *InterpRefs(RefChar *base, RefChar *other, real amount, SplineCha
...
+    }
+
+    SplinePoint* sp;
+    int contour_count = 1;
+    for ( bss = base, oss = other; bss!=NULL; bss=bss->next, oss=oss->next, contour_count++ ) {
+        if (!ret) break;
+        int bicw = SplinePointListIsClockwise(bss), oicw = SplinePointListIsClockwise(oss);
+
+        if (bicw == -1 || oicw == -1)
+            TRACE(_("%sInterpolationSanity: Assuming contour %d of undefined direction to be sane\n"), prefix, contour_count);
```
jtanx, line 1365: I wouldn't mark TRACE strings as translatable fwiw, especially as they're only for debugging

fontforge/fvfonts.c
```diff
@@ -1280,35 +1283,138 @@ static RefChar *InterpRefs(RefChar *base, RefChar *other, real amount, SplineCha
...
-	} else
-	    LogError( _("In character %s, could not find reference to %s\n"),
-		    sc->name, base->sc->name );
+	    if ( !check_only ) {
+		cur = RefCharCreate();
+		free(cur->layers);
+		*cur = *base;
+		cur->orig_pos = cur->sc->orig_pos;
+		for ( i=0; i<6; ++i )
+			cur->transform[i] = base->transform[i] + amount*(other->transform[i]-base->transform[i]);
```
skef, line 1298: This comes under a "this isn't your code and may not be your problem" disclaimer. 

It seems to me like the intent of this loop structure is to iterate through each element of "base" looking for a matching "other" temporarily called `test`. At the bottom of the loop we advance `other` if it's equal to `test` because if it happens to be the first in the list it's "checked" now and we don't have to look at it again (and we're GW so any performance gain automatically trumps code complexity).

If that's right, fine, but what I don't get is why we're then using `other` in the transform interpolation itself. Why isn't it `test` in (new) line 1298? That's what we just marked `checked`. 

Maybe it only makes sense to match up the `RefChar` entries 1:1 in order when interpolating, and therefore to grab the first off the list, but in that case why bother with `test` at all? 

So I'm 85% confident that that `other` should be `test`, with the other 15% being lack of confidence I've grasped this code at all.

fontforge/fvfonts.c
```diff
@@ -1280,35 +1283,138 @@ static RefChar *InterpRefs(RefChar *base, RefChar *other, real amount, SplineCha
...
+                         prefix, problemspline, extras, segment_count);
+                ret = false; break;
+            }
+
+            bsp = bsp->next->to;
+            osp = osp->next->to;
+        }
+    }
+
+    if (prefix[0] != '\0') free(prefix);
```
skef, line 1414: This works but is a little opaque. Seems better just to use the same logic as at allocation time (i.e. `glyphname!=NULL`)

fontforge/fvfonts.c
```diff
@@ -1280,35 +1283,138 @@ static RefChar *InterpRefs(RefChar *base, RefChar *other, real amount, SplineCha
...
+
+    if (bmcc != omcc) {
+        LogError(_("%sIncompatible contour counts: %d vs %d."), prefix, bmcc, omcc);
+        ret = false;
+    }
+
+    SplinePoint* sp;
+    int contour_count = 1;
+    for ( bss = base, oss = other; bss!=NULL; bss=bss->next, oss=oss->next, contour_count++ ) {
+        if (!ret) break;
```
skef, line 1361: I may be in a minority but I prefer `ret && bss!=NULL` as the loop test rather than adding additional flow control inside the loop (given that the `break` test comes first anyway).

fontforge/fvfonts.c
```diff
@@ -1522,13 +1628,13 @@ static void LayerInterpolate(Layer *to,Layer *base,Layer *other,real amount,Spli
...
     to->splines = SplineSetsInterpolate(base->splines,other->splines,amount,sc);
-    to->refs = InterpRefs(base->refs,other->refs,amount,sc);
+    to->refs = InterpRefs(base->refs,other->refs,amount,sc->name,false);
     if ( base->images!=NULL || other->images!=NULL )
 	LogError(_("I can't even imagine how to attempt to interpolate images in layer %d of %s\n"), lc, sc->name );
 }
 
 SplineChar *SplineCharInterpolate(SplineChar *base, SplineChar *other,
-	real amount, SplineFont *newfont) {
+	real amount, SplineFont *newfont, bool only_compatible) {
```
skef, line 1637: Does it matter that we're not doing the compatibility check in the multilayer case? I know you did some work with multilayer support before.

fontforgeexe/fvfontsdlg.c
```diff
@@ -315,8 +318,10 @@ return( true );
 	}
 	if ( fv==NULL )
 	    InterAskFilename(d->fv,last_amount/100.0);
```
skef, line 320: Seems to me that we should pass `only_compatible` into `InterAskFilename()` and make the same change there as just below. (Or given that it's only called here just merge the `InterAskFilename()` code into this function.)

fontforge/fvfonts.c
skef: (I personally have no issue with skipping multilayer support for the flag and just adding an "XXX todo" comment.)

jtanx:
Before I start looking too closely at this, what's the state of this/do we wish to merge this?

ctrlcctrlv:
I've marked this PR a draft. The reference issue is very hairy and I no longer have interest in figuring it out as I use other tools to deal with this now.

PR #4329, opened 2020-05-09
ggdkdraw.c: re-set transient status for restricted windows when un-hiding

### LINKED ISSUE #4328: Generate BDF stops working after one run
- [ ] Screenshot

## The FontForge version and the operating system you're using
FF 2020-03-14

## Steps to reproduce the behavior 
1. Open TTF file
1. Element > Bitmap Strikes Available... > (enter pixel size of font)
1. File > Generate Fonts... > Generate > OK

(works)

### Now repeat this step:
1. File > Generate Fonts... > change pixel size of output > Generate > OK

(error noise, no files are produced)

## Possible solution/fix/workaround
1. close and relaunch FF
1. redo the generate

jtanx:
The transient state was only being set on window creation. But sometimes windows
are held around and only hidden/shown. While the transient state was being
cleared on hide events, it was not reinstated when the window was shown again.

Fixes #4328

### Type of change

- **Bug fix**

### DIFF
--- a/gdraw/ggdkdraw.c
+++ b/gdraw/ggdkdraw.c
@@ -1481,6 +1481,9 @@ static void GGDKDrawSetVisible(GWindow w, int show) {
         _GGDKDraw_OnFakedConfigure(gw);
 #endif
         gdk_window_show(gw->w);
+        if (gw->restrict_input_to_me && gw->transient_owner == NULL && gw->display->mru_windows->length > 0) {
+            GGDKDrawSetTransientFor((GWindow)gw, (GWindow) - 1);
+        }
     } else {
         GGDKDrawSetTransientFor((GWindow)gw, NULL);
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
For testing purposes, what issue does this solve?

frank-trampe:
This sounds reasonable. I think that it makes sense to leave further testing to the master branch builds.

gingerbeardman:
@jtanx what is the best way to get a build that includes this change?

I tried to build the current source on macOS, but it failed part way through the ninja phase with some code errors.

Thanks for the fix!

jtanx:
@gingerbeardman you should be able to find a dmg here: http://dl.bintray.com/fontforge/fontforge/

Scroll to the bottom for the latest.

gingerbeardman:
Thanks so much!

PR #4332, opened 2020-05-13
Add LGTM

jtanx:
Given how bad the current situation is with warnings (basically all ignored), I'm hoping to add this, just to give visibility to some more obvious issues introduced in code changed from PRs.

### Type of change
- **New feature**
- **Non-breaking change**


jtanx:
(LGTM build is actually fine, it's marked as failing because the master build isn't working and won't work until this is merged)

PR #4333, opened 2020-05-14
Python Unicode/UTF8 cleanup

### LINKED ISSUE #4331: Crash when setting a glyph comment containing a surrogate
The following script crashes FontForge 20200314. I am using macOS 10.14.6.
```python3
import fontforge
fontforge.font().createChar(0x30).comment = '\ud800'
```
[Here is a crash report.](https://github.com/fontforge/fontforge/files/4618043/fontforge_2020-05-12-153621.txt)

ctrlcctrlv:
Closes #4331, and certainly other unreported bugs.

### Type of change

- **Bug fix**
- **Non-breaking change**

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -97,8 +97,6 @@
 extern int old_sfnt_flags;
 extern int prefRevisionsToRetain;
 
-PyMODINIT_FUNC fontforge_python_init(const char* modulename);
-
 /* ========== MODULE DEFINITIONS ========== */
 /* The following types are used to define Python Modules.  For every
  * module, you must create a 'module_definition' structure describing
@@ -145,7 +143,7 @@ typedef struct {
      */
     struct {
 	PyObject *module;
-	PyMODINIT_FUNC (*modinit_func)(void);
+	PyObject* (*modinit_func)(void);
 	PyModuleDef pymod_def;
     } runtime;
 } module_definition;
@@ -19179,8 +19177,6 @@ static int AddPythonTypesToModule( PyObject *module, python_type_info* typelist 
 PyMODINIT_FUNC CreatePyModule_##name(void) {\
     return CreatePyModule(&module_def_##name);\
 }
-#define ff_fixmod(name) \
-    module_def_##name.runtime.modinit_func = CreatePyModule_##name
 
 /* ===== MODULE REGISTRY -- LIST OF ALL PYTHON MODULES ===== */
 static module_definition * all_modules[] = {
@@ -19193,14 +19189,9 @@ ff_crmod(fontforge)
 ff_crmod(psMat)
 ff_crmod(ff_internals)
 
-static void fixup_module_definitions(void) {
-    ff_fixmod(fontforge);
-    ff_fixmod(psMat);
-    ff_fixmod(ff_internals);
-}
-/* ===== END MODULE REGISTRY ===== */
-
 #define NUM_MODULES (sizeof(all_modules) / sizeof(module_definition *))
+
+/* ===== END MODULE REGISTRY ===== */
 
 
 static const char *spiro_names[] = { "spiroG4", "spiroG2", "spiroCorner",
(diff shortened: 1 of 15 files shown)


fontforge/python.c
```diff
@@ -221,42 +222,6 @@ static wchar_t *copy_to_wide_string(const char *s) {
 #endif /* NEED_WIDE_CHAR */
```
ctrlcctrlv, line 224: This entire function was pointless because `PyUnicode_AsUTF8String` sets a more verbose exception if one exists.

jtanx:
The methods you deleted do in fact serve a purpose, at least on Python 2. On Python 2, `PYGETSTR` only works with non-unicode/ascii encoded strings only. You can check this out by comparing


```py
glyph.glyphname = u'Á' # Uses PYGETSTR, this will throw an exception
glyph.comment = u'Á' # Uses the unicode methods, so this will work
```

So the question comes down to, do we want to continue supporting Python2 in code. As I've said before, I say no.

But in which case, we can do much better and delete the PYGETSTR macro altogether, and use `PyUnicode_AsUTF8` directly, which is much nicer to use because you don't need to free anything (so no more ENDPYGETSTR equivalents).

ctrlcctrlv:
Right @jtanx, I do remember that we discussed making 20200314 the last release that could be built against Python 2. 

This is as good a time as any to break it. I'll leave this comment a few days, and if no one resists, push a deeper change which will remove all Python 2 ifdefs

ctrlcctrlv:
OK, I've allowed more than enough time to pass. I'll hack it out in the next hour.

frank-trampe:
I would have preferred not to break this, but it probably doesn't make sense to put positive work into restoring what's broken here.

ctrlcctrlv:
it doesn't

frank-trampe:
I mean for Python 2. Hence why it doesn't make sense to fix it.

ctrlcctrlv:
@jtanx Let me know what you think :-)

jtanx:
Sorry about the delay, I'll try to take a closer look over the weekend. I'm tempted to take this opportunity to go a bit further and get rid of PYGETSTR too, the behaviour is non-obvious and it's a bit weird now that ENDPYGETSTR is deleted by PYGETSTR isn't. iirc there was also at least one point where it was returning (in the error case) without doing proper cleanup.

I think there's also some opportunities to simplify some other usages to `PyUnicode_AsUTF8`.

I'd be happy to do the changes if you're ok with that (or if you want to do them then that's fine too).

ctrlcctrlv:
@jtanx Very happy to let you do the changes! I invited you to collaborate (in Github's phrasing) on ctrlcctrlv/fontforge. As I understand it, once you accept, you can push directly atop `issue4331`.

fontforge/python.c
```diff
@@ -1717,50 +1608,42 @@ static PyObject *PyFF_ask(PyObject *UNUSED(self), PyObject *args) {
...
-	    PyMem_Free(title);
-	    PyMem_Free(quest);
-	    FreeStringArray( i, answers );
-return( NULL );
-	}
-	answers[i] = copy(PyBytes_AsString(utf8_name));
-	Py_DECREF(utf8_name);
+        PyObject *item = PySequence_GetItem(answero, i);
+        answers[i] = PyUnicode_AsUTF8(item);
+        Py_XDECREF(item);
```
jtanx, line 1629: I haven't checked in depth whether or not it's ok (decrementing the reference while holding the string rep), but my reading of the docs for PyUnicode_AsUTF8 (combined with assuming the sequence is a list/tuple) suggests that it should be, as long as that reference to the object in the list/tuple still exists

jtanx:
Pushed a bunch of changes, but haven't tested in depth. Ran into a few cases where strings and references weren't being handled correctly along the way. 

Most notably, FF leaks references all over the place when using PySequence_GetItem and not releasing the reference after. I've fixed quite a few, but there's a heap more all over the place. Can probably leave that for another time though.

ctrlcctrlv:
I've been using the branch for the past months, time to take the plunge.

Issue #2796, opened 2016-08-29
adding classes in contextual chaining lookups incorrectly modifies all previous rules


RolfNoyer:
Create a contextual chaining substitution lookup subtable with the options “By Classes” and “Simple” (Dialog); close the lookup; reopen and try to add a new back class or ahead class to the list. As soon as the new class is created and populated with at least one glyph, suddenly all the substitution rules are changed for no reason. In particular it seems that the context after the first pipe in the first rule is added to all rules in the list. 

Workaround: go through every substitution rule in the list, deleting the incorrectly added context. Make sure you know what your rules were before the extra garbage is added so you can restore them.  

Using FontForge 20160402 on Mac Laptop running OS 10.11.6

PR #4338, opened 2020-05-17
Fix contextual chaining lookups rules modification (#2796)

o-sdn-o:
### Type of change

- **Bug fix**

### Fix contextual chaining lookups rules modification (#2796)
Rule section markers are not zeroed in the next iterations.

### Specify untracked files that Git should ignore when using Visual Studio 2019
Add exceptions in .gitignore for Visual Studio 2019.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 build/
-out/
 
 .vscode/
 .vs/
--- a/fontforgeexe/contextchain.c
+++ b/fontforgeexe/contextchain.c
@@ -1748,7 +1748,8 @@ static void RenameClass(struct contextchaindlg *ccd,char *old,char *new,int sect
 	}
     } else {
 	for ( i=0; i<rows; ++i ) {
-	    char* end_back = NULL, * end_match = NULL;
+	    char *end_back = NULL;
+	    char *end_match = NULL;
 	    char *oldrule = classrules[cols*i+0].u.md_str;
 	    char *newrule;
 	    for ( pt=last_name=oldrule; *pt; ) {
--- a/fontforgeexe/contextchain.c
+++ b/fontforgeexe/contextchain.c
@@ -1748,10 +1748,10 @@ static void RenameClass(struct contextchaindlg *ccd,char *old,char *new,int sect
 	}
     } else {
 	for ( i=0; i<rows; ++i ) {
-	    char* end_back = NULL, * end_match = NULL;
-	    char *oldrule = classrules[cols*i+0].u.md_str;
-	    char *newrule;
-	    for ( pt=last_name=oldrule; *pt; ) {
+		char* end_back = NULL, * end_match = NULL;
+		char *oldrule = classrules[cols*i+0].u.md_str;
+		char *newrule;
+		for ( pt=last_name=oldrule; *pt; ) {
 		while ( isspace(*pt)) ++pt;
 		if ( *pt=='|' ) {
 		    if ( end_back == NULL )
--- a/fontforgeexe/contextchain.c
+++ b/fontforgeexe/contextchain.c
@@ -1748,7 +1748,7 @@ static void RenameClass(struct contextchaindlg *ccd,char *old,char *new,int sect
 	}
     } else {
 	for ( i=0; i<rows; ++i ) {
-        char* end_back = NULL, * end_match = NULL;
+		char* end_back = NULL, * end_match = NULL;
 	    char *oldrule = classrules[cols*i+0].u.md_str;
 	    char *newrule;
 	    for ( pt=last_name=oldrule; *pt; ) {
--- a/fontforgeexe/contextchain.c
+++ b/fontforgeexe/contextchain.c
@@ -1748,7 +1748,7 @@ static void RenameClass(struct contextchaindlg *ccd,char *old,char *new,int sect
 	}
     } else {
 	for ( i=0; i<rows; ++i ) {
-		char* end_back = NULL, * end_match = NULL;
+        char* end_back = NULL, * end_match = NULL;
 	    char *oldrule = classrules[cols*i+0].u.md_str;
 	    char *newrule;
 	    for ( pt=last_name=oldrule; *pt; ) {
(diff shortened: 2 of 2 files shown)


ctrlcctrlv:
Why can't you tell VS to use the `build` folder?

fontforgeexe/contextchain.c
```diff
@@ -1748,6 +1748,7 @@ static void RenameClass(struct contextchaindlg *ccd,char *old,char *new,int sect
 	}
     } else {
 	for ( i=0; i<rows; ++i ) {
+		char* end_back = NULL, * end_match = NULL;
 	    char *oldrule = classrules[cols*i+0].u.md_str;
```
ctrlcctrlv, line 1752: Spacing needs to be fixed

jtanx:
Are you actually building this in VS, or just using VS as an IDE?

o-sdn-o:
> Are you actually building this in VS, or just using VS as an IDE?

Yes, it builds and run with the debugger in Visual Studio 2019 in Windows 10, it works fine.

I use Windows 10 + WSL2 + Xming:
- turn on WSL feature in your Win10
- install Xming on your Windows 10, run it
- add two environment variables in your Windows 10:
  - DISPLAY=localhost:0.0
  - WSLENV=DISPLAY/u
- run WSL CLI and install all needed fontforge's dependencies in WSL2 (same as linux)
- run Visual Studio and clone the project (https://github.com/monotty/fontforge.git) to local folder
  - manage configurations
    - Remove selected
	- x64-Debug
    - Add two new configurations
	- WSL-GCC-Debug
	- WSL-GCC-Release
   - Save all files (via menu)

In the CMAKE targets view, select 'fontforgeexe' as Startup Project (for Debug/Release)

  - Run Debug/Release

ctrlcctrlv:
Why is Xming needed? If you're going to be contributing more than just this, you really should be building the GDK release, because that's what 99% of FontForge users using current builds are using.

ctrlcctrlv:
Actually, maybe I understand. WSL provides no Xorg. And if you build using it, you're not using our MSYS2 build system, so you need some display server even if building the GDK version. Presumably Wayland would also work if there was something like Xming for it.

OK. That makes sense. Sorry to double post.

skef:
@o-sdn-o You're getting a few off-topic questions because we thought the project was further away from potential VS compilation than it apparently is.

If you're able we'd be curious about whether you could get it working with the GTK port available in vcpkg referenced [here](https://www.gtk.org/docs/installations/windows/#using-gtk-from-vcpkg-packages). (And ideally with the vcpkg port of python 3 as well.)

fontforgeexe/contextchain.c
o-sdn-o: I changed the spaces to tabs, but something is wrong and I don’t know what is wrong.

skef: @o-sdn-o For historical reasons much of FontForge is intended with a mixture of spaces and tabs. I suggest copying the whitespace from a line with the same indentation (such as the one below right below) and pasting it.

jtanx:
So WSL is effectively Linux, this isn't a native MSVC/Windows build by any means.

o-sdn-o:
> Why can't you tell VS to use the `build` folder?

Of course, it is possible to change the path to the "build" folder, but just the default value for the `"Build root"` parameter in the `WSL-GCC-Release/Debug` configuration in Visual Studio is set to `${projectDir}\out\build\${name}"`

skef:
@jtanx Given that vcpkg appears to be a source-based package manager that has python 3 and GTK and doesn't seem to be WSL-dependent, wouldn't that put things pretty close?

jtanx:
This is going completely off topic now, but what you're looking for is to effectively compile FontForge and all of its dependencies against the same MSVCRT as used by Python, with one of those paths being to compile FontForge with MSVC itself. 

As it stands, even if you had all your dependencies built to satisfy that condition (probably doable, except with a lot of effort), FontForge itself cannot be compiled with MSVC directly because of some GCCisms and/or *nix oriented code that is handled by mingw but not MSVC.

You can also just compile FontForge with GCC but link to the same runtime as the official Python builds, but again you're stuck with trying to get all your dependencies to do the same, which means building all of them yourself.

fontforgeexe/contextchain.c
o-sdn-o: ok, I will do so, and apparently something is wrong with PR checks

Details:
```
continuous-integration/appveyor/p - AppVeyor` build failed
```
```
loading package files...
error: could not open file /var/cache/pacman/pkg/msys2-runtime-3.1.4-1-x86_64.pkg.tar.zst: Child process exited with status 127
error: could not open file /var/cache/pacman/pkg/msys2-runtime-devel-3.1.4-1-x86_64.pkg.tar.zst: Child process exited with status 127
error: failed to commit transaction (cannot open package file)
Errors occurred, no packages were upgraded.
Command exited with code 1
```

skef:
@jtanx do we know what's going on with appveyor? Is there a persistent problem or is this just intermittent crankiness?

fontforgeexe/contextchain.c
ctrlcctrlv: If you're going to be contributing a lot, in future you should enable visible whitespace in your editor. That's what I do.

image

We might want to recommend this in CONTRIBUTING.md.

o-sdn-o: Ok.

o-sdn-o: Now everything looks consistent
image

jtanx:
Probably msys2/msys2-packages#1960

.gitignore
```diff
@@ -1,6 +1,8 @@
 build/
+out/
```
ctrlcctrlv, line 2: Because @jtanx asked for my opinion:

I continue to disagree with the change to the Git ignore. There is no good reason for us to have two build folders. Configure your IDE. We shouldn't be expected to bend over backwards to support the defaults of every IDE. If there's a configuration file we can add to tell VS to use our build folder, that's acceptable too.

o-sdn-o: Ok.

PR #4343, opened 2020-05-20
Fix a crash when pressing a key that doesn't exist in the current encoding

jtanx:
Noticed this while working on cid fonts.

Steps to repro:
* Create a new font
* Set encoding to custom, remove unused slots and add 256 (or any amount > 255)
* Fill all slots with something so they're not empty
* Press the spacebar or any key on your keyboard
* `i` is now >256 which exceeds the backmap size (backmax) --> garbage data or crash

Alternatively, open source han sans, and press the spacebar in the fontview

### Type of change

- **Bug fix**


ctrlcctrlv:
I definitely appreciate all the CID fixes lately, Jeremy! 🙌

PR #4370, opened 2020-06-06
Fix SSRemoveTiny crash

### LINKED ISSUE #4364: SplineSetRemoveOverlap crash, calling SSRemoveTiny with a NULL SplineSet (test case inside). Triggered via «Change Weight...» operation.
In the process of making a **bold** font (why does Github make it **semibold** when that's barely any different from Regular?), I encountered an issue called Glyph of glitch. When the glyph is attempted to be emboldened, it causes glitches in all of FontForge. Use the Element\\Style\\Change Weight... operator (within the interactive grid of glyphs) and set to embolden by 51 units. Not only this will crash FontForge, but also create a glitched recovery file. Using the operator within the glyph editor will work, but will lead to glitches later, like attempting to clear the glyph from the interactive grid of glyphs will crash.

[Glyph of glitch.zip](https://github.com/fontforge/fontforge/files/4739538/Glyph.of.glitch.zip)

ctrlcctrlv:
It wasn't properly checking base for NULL, which led to `sp` pointing at
junk.

This closes #4364.

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/splineoverlap.c
+++ b/fontforge/splineoverlap.c
@@ -3490,7 +3490,7 @@ static SplineSet *SSRemoveTiny(SplineSet *base) {
 
     while ( base!=NULL ) {
 	ssnext = base->next;
-	for ( sp=base->first; ; ) {
+	for ( sp=base->first; !( base == NULL || sp==base->first ); ) {
 	    if ( sp->next==NULL )
 	break;
 	    nsp = sp->next->to;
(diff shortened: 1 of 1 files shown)


skef:
@ctrlcctrlv When I test the code in the issue I get a `malloc_consolidate(): invalid chunk size` in a different place. Let's hold off on this for a bit.

fontforge/splineoverlap.c
```diff
@@ -3490,7 +3490,7 @@ static SplineSet *SSRemoveTiny(SplineSet *base) {
 
     while ( base!=NULL ) {
 	ssnext = base->next;
-	for ( sp=base->first; ; ) {
+	for ( sp=base->first; !( base == NULL || sp==base->first ); ) {
```
skef, line 3493: With a for loop the guard is checked immediately after the initialization. At that point sp==base->first so the guard will be false. Therefore this addition just disables the loop.

ctrlcctrlv: Thank you - I've closed this as now that you're looking into it I'm sure you'll come up with a better solution. :-)

PR #4138, opened 2020-02-07
Avoid deleted contours in SSRemoveTiny.

frank-trampe:
Avoid working with a splinepoint or a spline post-deletion in SSRemoveTiny.

This closes #4095 and #4117 .

@skef, please review, and @arturrataj please test.

PR #4371, opened 2020-06-06
Clear sp in SSRemoveTiny so the guards after the loop don't execute.

### LINKED ISSUE #4364: SplineSetRemoveOverlap crash, calling SSRemoveTiny with a NULL SplineSet (test case inside). Triggered via «Change Weight...» operation.

skef:
Similar/related to #4138 

Closes #4364

### DIFF
--- a/fontforge/splineoverlap.c
+++ b/fontforge/splineoverlap.c
@@ -3510,6 +3510,7 @@ static SplineSet *SSRemoveTiny(SplineSet *base) {
 			    prev->next = ssnext;
 			base = NULL;
 			nsp = NULL;
+			sp = NULL;
 			break;
 		    } else {
 			// We want to remove the spline following sp.


ctrlcctrlv:
Indeed, I was right, the solution you found was much better than mine :-D

I'll test this but I see no reason it shouldn't work.

PR #4407, opened 2020-07-21
Adjustments for floating point corner cases (Expand stroke and SVG import)

### LINKED ISSUE #4406: FontForge crashes when importing SVG
### Reporting a bug/issue:

- [x] Screenshot
- [ ] (optional) Possible solution/fix/workaround

### Short description:

The program crashes when I'm trying to import several SVGs to a glyph.

### Screenshot (to comply with the guidelines, otherwise irrelevant):

FontForge Character View

### Confirmed to crash on:

- Character `U+F064`
- [These SVG files](https://github.com/fontforge/fontforge/files/4947644/fontforge-test-import-svgs.zip)
- Manjaro Linux 20.0.3 (Lysia)
- FontForge 20200314

I think the most relevant factors are SVGs themselves, because the _share.svg_ file manages to crash program when imported to every glyph I tested. The character code might also be somewhat relevant because, at one time, _robot.svg_ crashed the program when imported at `U+F544`, but not when at `U+F553`; although, I was unable to reproduce it since then and now it crashes every time, regardless of the character.

### Technical details:

- **OS:** Manjaro Linux
  - **Version:** 20.0.3 (Lysia)
  - **Architecture:** x86-64
  - **Kernel:** Linux kernel 5.7.0
    - **Release:** 5.7.0-3-MANJARO
  - **Desktop environment:** KDE Plasma 5.18.5
- **FontForge:**
  - **Version:** 20200314
  - **Build:** 2020-03-18 19:16 UTC-ML-D-GDK3
  - **OS package:** _extra/fontforge 20200314-2_ (Manjaro)
    - **Package file:** [_fontforge-20200314-2-x86_64.pkg.tar.zst_](https://manjaro.moson.org/stable/extra/x86_64/fontforge-20200314-2-x86_64.pkg.tar.zst)
    - **SHA1 checksum:** `16e5aa2a93983ddadc793307f638217b55bb7b1d`

### Steps to reproduce:

1. Create a new font file.
2. In the Character View window, select the character with code `U+F064`.
3. Press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>I</kbd>.
4. In _Format_, select _SVG_.
5. In _Options_, choose the following parameters (and click _OK_):
   FontForge Import Parameters
6. Select one of the files I provided.
7. Observe a crash.

**Note:** If unable to reproduce, try the second file.

**Note:** I was unable to reproduce the same behaviour on a Windows 10 PC with the official latest Windows build from the website; everything seems to work just fine. So it may be specific to just Linux, Manjaro, or even this specific build.

### GDB output:

```
% gdb fontforge
GNU gdb (GDB) 9.2
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from fontforge...
(No debugging symbols found in fontforge)
(gdb) run
Starting program: /usr/bin/fontforge
Copyright (c) 2000-2020. See AUTHORS for Contributors.
Version: 20200314
Based on sources from 2020-03-18 19:16 UTC-ML-D-GDK3.

0x00007ffff7b3e92e in ?? () from /usr/lib/libfontforge.so.4
(gdb) cont
Continuing.

Program terminated with signal SIGSEGV, Segmentation fault.
The program no longer exists.
```

skef:
Adjust UTMARGIN for case just outside unit vector accuracy
Protect against bad calculations of circular SVG args that
slip through the SVG 1.1 standard test

closes #4406

### DIFF
--- a/fontforge/svg.c
+++ b/fontforge/svg.c
@@ -1301,14 +1301,18 @@ static void SVGTraceArc(SplineSet *cur,BasePoint *current,
 	y1p =-ang.y*(current->x-x)/2 + ang.x*(current->y-y)/2;
 	/* Correct for bad radii */
 	lambda = x1p*x1p/(rx*rx) + y1p*y1p/(ry*ry);
-	if ( lambda>1 ) {
+	if ( lambda>=1 ) {
 	   lambda = sqrt(lambda);
 	   rx *= lambda;
 	   ry *= lambda;
 	   cxp = cyp = 0;
 	} else {
 	    factor = rx*rx*ry*ry - rx*rx*y1p*y1p - ry*ry*x1p*x1p;
-	    factor = sqrt(factor/(rx*rx*y1p*y1p+ry*ry*x1p*x1p));
+	    factor = factor/(rx*rx*y1p*y1p+ry*ry*x1p*x1p);
+	    if (factor < 0) // Ideally the lambda check will prevent this, but ...
+		factor = 0;
+	    else
+		factor = sqrt(factor);
 	    // This part of the center calculation deals with large_arc
 	    // leaving sweep as a parameter to SSAppendArc
 	    if ( large_arc==sweep )
--- a/fontforge/utanvec.c
+++ b/fontforge/utanvec.c
@@ -34,10 +34,12 @@
 #include <math.h>
 #include <assert.h>
 
-#define UTMARGIN (1e-8)
-#define UTSOFTMARGIN (1e-5)
+#define UTMARGIN (1e-7)     // Arrived at through testing
+#define UTSOFTMARGIN (1e-5) // Fallback margin for some cases
 #define BPNEAR(bp1, bp2) BPWithin(bp1, bp2, UTMARGIN)
-#define UTRETRY (UTMARGIN/10.0)
+#define UTRETRY (1e-9)      // Amount of "t" to walk back from the 
+                            // ends of degenerate splines to find
+                            // a slope.
 
 BasePoint MakeUTanVec(bigreal x, bigreal y) {
     BasePoint ret = { 0.0, 0.0 };


fontforge/svg.c
```diff
@@ -1301,14 +1301,18 @@ static void SVGTraceArc(SplineSet *cur,BasePoint *current,
 	y1p =-ang.y*(current->x-x)/2 + ang.x*(current->y-y)/2;
 	/* Correct for bad radii */
 	lambda = x1p*x1p/(rx*rx) + y1p*y1p/(ry*ry);
-	if ( lambda>1 ) {
+	if ( lambda>=1 ) {
 	   lambda = sqrt(lambda);
 	   rx *= lambda;
 	   ry *= lambda;
 	   cxp = cyp = 0;
```
frank-trampe, line 1308: Is this line the only one that is operative in the `==` case?

frank-trampe: Or is the issue that the `else` code causes problems in that case?

fontforge/utanvec.c
```diff
@@ -34,7 +34,7 @@
 #include <math.h>
 #include <assert.h>
 
-#define UTMARGIN (1e-8)
+#define UTMARGIN (5e-7)
```
frank-trampe, line 37: Did you compute this somehow or is it just the result of testing against old and new targets?

skef: Actually this is just wrong -- so good catch.  What I intended was either 1e-7 or 5e-8. 

This value went through a lot of testing with many sources but one of the #4406 cases had a difference just slightly larger than 1e-8. (Scrolling back up the value was
`{x = 1, y = 1.1978936413822867e-08}`, which needed to match `{1,0}`.) So I first switched it to 1e-7 and then intended to move it back down to 5e-8 and futzed the edit. 

Which would you recommend, or is there some more principled way of determining? I suppose I could trace back through the execution patch to figure out where in particular the values are coming from ...

fontforge/svg.c
skef: With the "robot.svg" file:

1. gdb prints `lambda` as exactly 1
2. `factor` before the square root is negative, resulting in NANs.

There's no problem I can see with changing the inequality, so I did that. However, I don't trust that as a general solution (and maybe not even in this case -- gdb print might be rounding), so if the `factor` calculation goes negative I just set it to zero.

fontforge/utanvec.c
ctrlcctrlv: Whatever is decided on, certainly there should be a comment explaining why all "random-seeming" constants were chosen. :-)

skef: I guess. Half the constants I've run into in Skia are either uncommented or say "I chose this at random". I can put "arrived at through testing" on each of them but I'm not sure what good it will do.

Anyway absent other feedback I'm going to go with 1e-7 for this one.

ctrlcctrlv:
It passes my usual tests for this sort of thing, so yes. :-)

PR #4423, opened 2020-08-08
Fix checkbox double toggle due to et_char and et_charup events

### LINKED ISSUE #4422: Question: how can I set a shortcut key to toggle "Use My Metrics" option in "Reference info" menu
Hi
I'm using the latest fontforge version on windows 10 and need to check a huge number of glyphs(which are combined of at least 2 components) and set the "Use My Metrics" option(for one of the components) in the Reference Info menu. I can't find it in the hotkey shortcuts file. so what is the solution?
(I need to work with GUI to see and check the components)
thank you.

fontforge_2020-08-08_20-51-43

skef:
@jtanx this seems to fix #4422 under GTK and works fine under X11 (I didn't check whether the problem reproduces with the X11 backend). Does it look OK to you?

Closes #4422

### DIFF
--- a/gdraw/gradio.c
+++ b/gdraw/gradio.c
@@ -454,7 +454,7 @@ return(false);
 	    event->u.chr.keysym == GK_BackTab || event->u.chr.keysym == GK_Escape )
 return( false );
 
-    if (event->u.chr.chars[0]==' ' ) {
+    if (event->type == et_char && event->u.chr.chars[0]==' ' ) {
 	GRadioChanged(gr);
 	_ggadget_redraw(g);
 return( true );


jtanx:
Yeah this looks fine. Interesting though that it doesn't toggle immediately when you press the accelerator but instead just focuses on it.

Copying from `gbutton_focus`, this would fix that too.

```diff
diff --git a/gdraw/gradio.c b/gdraw/gradio.c
index 258fb06a6..61b53ce64 100644
--- a/gdraw/gradio.c
+++ b/gdraw/gradio.c
@@ -468,7 +468,8 @@ static int gradio_focus(GGadget *g, GEvent *event) {
     if ( !g->takes_input || (g->state!=gs_enabled && g->state!=gs_active ))
 return(false);

-    if ( event->u.focus.mnemonic_focus==mf_shortcut ) {
+    if ( event->u.focus.mnemonic_focus==mf_shortcut ||
+           event->u.focus.mnemonic_focus==mf_mnemonic ) {
        GRadioChanged(gr);
     }
 return( true );

```

skef:
I can see a consistency argument for it not changing the value but I don't know what the normal convention is.

PR #4431, opened 2020-08-18
Fix parsing of reverse-chaining substitutions.

clsn:
Adding reverse-chaining lookups formerly resulted in a segfault: a NULL pointer was being passed to a function that tried to count how many glyphs were in the substitution list.  Reverse chaining lists are unlike other tables, and they were not being parsed properly (most crucially, the substitutions weren't ending up in the same structure as the match-list).  They now parse correctly, at least in my tests.

### Type of change
- **Bug fix**

### DIFF
--- a/fontforge/lookups.c
+++ b/fontforge/lookups.c
@@ -4701,6 +4701,8 @@ static void GrowBufferAddClass(GrowBuf *gb,int class_n,char **classnames, int cl
 int GlyphNameCnt(const char *pt) {
     int cnt = 0;
 
+    /* Avoid a segfault if something is done wrong. */
+    if (!pt) { return 0; }
     while ( *pt ) {
 	while ( isspace( *pt )) ++pt;
 	if ( *pt=='\0' )
@@ -4910,6 +4912,8 @@ return( smprintf( _("Unterminated coverage table, starting at: %.20s..."), start
 		if ( cnt==0 )
 return( smprintf( _("Replacements must follow the coverage table to which they apply: %s"), start-4 ));
 		ch = *end; *end = '\0';
+		/* Must back up one! */
+		cnt--;
 		parsed[cnt].replacements = copy(start);
 		*end = ch;
 		rcnt = GlyphNameCnt(parsed[cnt].replacements);
@@ -4929,8 +4933,9 @@ return( smprintf( _("Replacements must follow the coverage table to which they a
 		    parsed[cnt].replacements = newr;
 		} else
 return( smprintf( _("There must be as many replacement glyphs as there are match glyphs: %s => %s"),
-		    parsed[cnt].entity, parsed[cnt].replacements));
-    continue;
+		  parsed[cnt].entity, parsed[cnt].replacements));
+		// And advance back to where we were
+		cnt++;
 	    }
 	} else if ( *start!='@' && *start!='<' && !(*start=='=' && start[1]=='>') ) {
 	    /* Just a normal glyph or class name. (If we expect a coverage table we'll treat it as a table with one glyph) */
@@ -4943,8 +4948,9 @@ return( smprintf( _("There must be as many replacement glyphs as there are match
 return( smprintf( _("No replacement lists may be specified in this contextual lookup, use a nested lookup instead, starting at: %.20s..."), lpt ));
 	    if ( do_replacements )
 return( smprintf( _("Only one replacement list may be specified in a reverse contextual chaining lookup, starting at: %.20s..."), lpt ));
-	    do_replacements = true;
+	    do_replacements = 1;
 	    lpt += 2;
+	    continue;
 	} else {
 	    /* A lookup invocation */
 	    if ( fpst->format==pst_reversecoverage )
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
Where they only broken when added through the GUI, or through feature files, or both?

clsn:
Well, I know they were broken in the GUI, and they were also broken when I tried to use the Python library to set them.  The latter was the bug I specifically fixed; the parsing from Python string to table was wrong.  Now that I'm trying with the GUI, yes, still dumping core (at least with the specific case I'm using), in a different function now, when trying to write out the table.  So the current fix IS, I think, valid for what it does, but doesn't fix the larger problem.  I'm looking into the issue with the GUI, etc.  I'll submit a separate request for that.  I don't know if fixing it will actually make reverse-chaining lookups WORK (as in render in FF); with luck they'll at least write out correctly so that OpenType font renderers will be able to use them.

frank-trampe:
@clsn, could you post the stack trace? This seems like a good time to determine the appropriateness of the upstream function passing a null pointer.

ctrlcctrlv:
Yes and of course a test case would help us as well. :-)

Thanks for contributing

clsn:
Oh, I absolutely know what the problem is, and I think I've fixed it in the GUI as well.  The replacements in reverse-chaining coverage were never getting saved into their structure, so when you tried to output them, you're facing a NULL.  Now, as to whether or not FF deals with the tables correctly when it comes to generating or displaying them, that's another matter.
I'll commit and push as soon as I write a test (for the python bit)

clsn:
Pushed, with test.  I think things are still being done wrong by something else wrt reverse chaining tables, because they don't seem to work (however, my own understanding of them is very shaky, so I might not know what working looks like.)
Meanwhile, I hesitate to file an issue over this, but the freshly-compiled fontforge apparently crashes my windows-manager (Enlightenment) whenever it tries to create a new window for editing a table.  Quite disturbing.

ctrlcctrlv:
FontForge's MetricsView doesn't necessarily support every feature in FontForge.

For example, `rand`; Cursive Attachment; etc.

If you attempt to use HarfBuzz to output the font generated, does it look OK?

https://docs.microsoft.com/en-us/typography/opentype/spec/gsub#lookuptype-8-reverse-chaining-contextual-single-substitution-subtable

clsn:
Yeah, I did try the font so generated in a program that I know can handle Iosevka's thousands-grouping (which uses reverse-chaining, among other things).  To the extent that I understand what should happen, it doesn't seem to work, no.  But at least it isn't crashing; presumably there's something wrong with how it's being output into the ttf file or whatever. Maybe I can find some other font tool I can compare with.  I wish FF would stop crashing my window-manager...

skef:
@clsn You might try recompiling against the X11 backend instead of GDK (or vice versa if you're using the X11 backend currently) to see if that works around the WM crashing problem.

clsn:
@skef hey, thanks for the advice!  That worked!  I wouldn't have thought of that.

clsn:
OK, when I generate a ttf or otf font from the sfd and then open that up with fontforge, the reverse chaining table and subtable are there, but the subtable is empty.  So presumably it's not being written to the output correctly.  At least it's correctly being stored in the sfd though!

ctrlcctrlv:
You might want to install `ttx` (`fonttools`) and dump the font and compare it.

clsn:
@ctrlcctrlv Thanks!  I had ttx, didn't know how to use it.  So, the ttx dump shows that the ttf/otf file written FF has a spot for the table and all, but just says `<ReverseChainSingleSubst index="0" Format="3"></ReverseChainSingleSubst>` where it should be.  And the format is wrong, too: it should be `Format="1"`.  I can fill in the XML in the ttx based on what I've seen in another font (Iosevka) and re-convert with ttx and it works, yay.  Still a problem for FF, though.

I think my current patch should be merged in (and humbly beg that it be), as it prevents some crashes that have been happening for a long time and *does* let you put reverse-chaining tables into your font.  Yes, they are only readable by FF at the moment, but at least they are being saved correctly there.  And then when some future patch (which I'll probably be working on) comes out which fixes the output problem, you'll be ready just to load up and generate your font files.

fontforge/lookups.c
```diff
@@ -4701,6 +4701,8 @@ static void GrowBufferAddClass(GrowBuf *gb,int class_n,char **classnames, int cl
 int GlyphNameCnt(const char *pt) {
     int cnt = 0;
 
+    /* Avoid a segfault if something is done wrong. */
+    if (!pt) { return 0; }
```
ctrlcctrlv, line 4705: It would be better to avoid calling this function with a null pointer than adding this check.

fontforge/lookups.c
```diff
@@ -4943,8 +4948,9 @@ return( smprintf( _("There must be as many replacement glyphs as there are match
 return( smprintf( _("No replacement lists may be specified in this contextual lookup, use a nested lookup instead, starting at: %.20s..."), lpt ));
 	    if ( do_replacements )
 return( smprintf( _("Only one replacement list may be specified in a reverse contextual chaining lookup, starting at: %.20s..."), lpt ));
-	    do_replacements = true;
+	    do_replacements = 1;
```
ctrlcctrlv, line 4951: It might make sense to add more boolean flags than to use this as an integer.

clsn:
> Overall, I do really appreciate the work you've done here, but don't see the rush to merge. If we know font output is broken, we should fix both in one PR, rather than merge a half-fix. No?

On one hand, I actually do think that half a fix is better than none, and this half-fix does really make something useful and usable that wasn't.  On the other hand... yeah, should probably get it doing what it ought to do.  I've been working on that, trying to fix the dumpg___ContextChainCoverage function.  As before, I'm partway there: got the coverage tables outputting okay, but not the substitutions.  I'll get it.  It's a matter of keeping track of the offsets I think.

> It would be better to avoid calling this function with a null pointer than adding this check.
Fair enough.  We aren't calling it with a null pointer anymore anyway.  I like to have safeguards to keep things from dying a screaming horrible death when the unexpected happens, but it's a matter of personal preference: maybe a horrible screaming death is a good way of telling you something is very very wrong. (well, in some applications.  Sometimes you really really need it not to die a horrible screaming death.)

> It might make sense to add more boolean flags than to use this as an integer.
It already was being used as an integer, and indeed I would have changed that to `1` instead of `true` even if I hadn't used it in other ways.  It was already getting set to 1 here (though it was called `true`) and checked for being `1` (not boolean!) up on line 4907 (in the original code) and set to `2` in line 4909.  I just found that 0-1-2 wasn't quite enough and I needed a 3.

Let me see what I can do about the ttf output.  One thing, though: for some reason, what I'm doing is helping stuff when I generate a font in the GUI, but doing `font.generate()` in python doesn't seem to call the same functions, even when I remembered the flags for it.  Have to see what I'm missing there.

clsn:
All right!  Got it to output properly, looks good with ttx and looks good when I test the font out in a suitable application!  I'm amazed at how tiny the changes wind up being by the time I've tried all kinds of things.  Also removed that null-pointer check.

Issue #4378, opened 2020-06-11
Trouble importing Octicons v10 Heart Fill Icon


FredHappyface:
* Provide a general summary of the issue in the **Title** above.  
* Before you open an issue, please check if a similar issue already exists or has been closed before.

### Important
Mark with [x] to select. Leave as [ ] to unselect.

### When reporting a bug/issue:
- [x] Screenshot
- [x] The FontForge version and the operating system you're using
- [x] The behavior you expect to see, and the actual behavior
- [x] Steps to reproduce the behavior 
- [ ] \(optional) Possible solution/fix/workaround

### When you open an issue for a change/improvement/feature request:
- [ ] A description of the problem you're trying to solve, including _why_ you think this is a problem
- [ ] If the feature changes current behavior, reasons why your solution is better
- [ ] \(optional) Possible solution/fix/workaround

### Version
FontForge 20200314 as available https://ci.appveyor.com/project/fontforge/fontforge/build/artifacts
Windows 10 1909

### Description and screenshots 

When importing the heart-fill.svg from Octicons v10, FontForge 20200314 crashes. Using https://ci.appveyor.com/project/fontforge/fontforge/build/artifacts prevents the crash but the icon looks rather unusual

![image](https://user-images.githubusercontent.com/41634689/84390146-153bde80-abef-11ea-8a25-46fe85096dce.png)

### Sample SVG

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16"><path fill-rule="evenodd" d="M7.655 14.916L8 14.25l.345.666a.752.752 0 01-.69 0zm0 0L8 14.25l.345.666.002-.001.006-.003.018-.01a7.643 7.643 0 00.31-.17 22.08 22.08 0 003.433-2.414C13.956 10.731 16 8.35 16 5.5 16 2.836 13.914 1 11.75 1 10.203 1 8.847 1.802 8 3.02 7.153 1.802 5.797 1 4.25 1 2.086 1 0 2.836 0 5.5c0 2.85 2.045 5.231 3.885 6.818a22.075 22.075 0 003.744 2.584l.018.01.006.003h.002z"/></svg>
```

### FontForge render 

![image](https://user-images.githubusercontent.com/41634689/84391048-54b6fa80-abf0-11ea-8980-bd239d49ec71.png)

### SVG render

Using https://github.com/lishu/vscode-svg2 the svg file renders as such: 

[...]

PR #4441, opened 2020-09-08
Add font.compact() to Python API

ctrlcctrlv:
Adds a function to the Python API that has the identical effect as &laquo;Encoding&rarr;Compact&raquo;.

### Type of change

- **Small new feature**

### DIFF
--- a/doc/sphinx/scripting/python/fontforge.rst
+++ b/doc/sphinx/scripting/python/fontforge.rst
@@ -3180,7 +3180,8 @@ This type may not be pickled.
 .. attribute:: font.encoding
 
    The name of the current encoding. Setting it will change the encoding used
-   for indexing
+   for indexing. To compact the encoding, first set it to your desired encoding
+   (e.g. ``UnicodeBMP``), then set it to ``compacted``.
 
 .. attribute:: font.familyname
 


khaledhosny:
`font.encoding = "Compact"` used to do the same thing.

skef:
A little bird pointed us to the existing `SFReencode` interface for this, which is to assign the string "`compacted`" to the encoding.

I don't have any strong opinions about whether to add this interface. (I do wish the python API wasn't so fixated on properties and assignments but it doesn't seem detrimental in this case.)

ctrlcctrlv:
Yeah the reason I'm adding this is because I made an icon font recently as mentioned at https://github.com/fontforge/fontforge/issues/4378#issuecomment-688711482 ... due to how most icon fonts are expected to work, the icons are always going to be way off screen when opening SFD. I suppose a Custom encoding is usable, but trying to set `font.encoding = Custom` didn't go so well without a way to compact out the empty blocks, I just got the flood of red question marks.

I'll take a look at `SFReencode` when I get a chance if you think it's better for adding this function.  :-)

ctrlcctrlv:
Oh and here's my use case:

[icons.ttf](https://github.com/mfeq/Qglif/blob/f9052673e60d445015cc7c092e3e29004b8506fa/resources/fonts/icons.ttf)

https://github.com/mfeq/Qglif/blob/f9052673e60d445015cc7c092e3e29004b8506fa/resources/icons/gen.py#L1-L39

skef:
@ctrlcctrlv I think the suggestion is to try (with the existing code) a) opening the font in python, b) do `fnt.encoding = 'UnicodeFull'`, and then c) do `fnt.encoding = 'compacted'` . 

I just tried it and it seems to work.

ctrlcctrlv:
@skef Force pushed, this is now a documentation-only change. Thanks.

skef (approved):
Looks good!

PR #4442, opened 2020-09-08
Fix outputting reverse chaining tables into feature files

clsn:
Fontforge refuses to output reverse chaining tables into feature files, mostly because it *claims* it does not know how to, when in fact it actually (mostly) does and it's mainly the check that is preventing it.  Not *only* the check, though, since it would output "ignore" all the time, but that is fixed as well.

### Type of change

- **Bug fix**
- **New feature**


ctrlcctrlv:
Do you have an example SFD?

clsn:
Here's a version of Pointfree I did when I was experimenting with making digits automatically put thousands-grouping commas into the glyphs.  Exporting the feature file works, yay.
[PointThousand.zip](https://github.com/fontforge/fontforge/files/5337382/PointThousand.zip)

frank-trampe:
We'll see how this works in the wild.

PR #4447, opened 2020-09-13
Suppresswarn

clsn:
### Type of change

- **Bug fix**

Suppress the "Invalid Unicode alert" message when scrolling past the Unicode surrogate characters in the BMP.  We know they're like that, they're supposed to be.


ctrlcctrlv:
Sorry it took me so long to comment. Thinking about it more, does it even make sense to display these characters in the FontView at all if they aren't in the font already? They should be treated like `-1`'s, IMO.

clsn:
> Sorry it took me so long to comment. Thinking about it more, does it even make sense to display these characters in the FontView at all if they aren't in the font already? They should be treated like `-1`'s, IMO.

That's a different call to make.  These are Unicode surrogate characters, so they should NEVER have any glyphs, I believe.  They're "there" and so they take up space that people might be expecting to have to scroll past and could be surprised that they don't have to, though I'll grant that's kind of reaching, as far as possible user-affecting changes go.  Me, I was just annoyed by the "Invalid Unicode" errors that always showed up when I was scrolling past them, and so that's what I suppressed.

Oh, I think I misunderstood: I thought you meant maybe that whole section should simply be excised from the font view entirely and not even take up space.  Which is a more drastic thing... and yet _still_ isn't entirely a bad idea.  Treating them like -1s, in the sense of showing the block for them but not trying to print a character in that block?  Yes, definitely, since they should never have glyphs and no other font should ever be defining glyphs for them either.

frank-trampe:
This is a positive change. We can discuss further improvements in the future.

Issue #4456, opened 2020-10-03
Make quadratic off curve points individually selectable


yg8ijvjvjv:
As a TrueType font designer, off-curve points should be selectable individually. Select, move, remove, add, etc. individually.

A button for adding off curve points just as there is for adding an on curve point.
![image](https://user-images.githubusercontent.com/68180132/94991787-94319a00-0585-11eb-9707-92ba4cd6208a.png)

The adding off curve point being performed on a straight line.
![image](https://user-images.githubusercontent.com/68180132/94991856-36ea1880-0586-11eb-8f24-b9ad7b2af64d.png)

The adding off curve point being performed on a quadratic curve. The interpolated point is not actually a defined point so it is no longer displayed (an alternative is to display it but make it no longer selectable, similarly to missing extrema view).
![image](https://user-images.githubusercontent.com/68180132/94991931-b1b33380-0586-11eb-9e25-9a97197509ad.png)

And removing points would work similarly. Removing an off curve point between two on curve points makes a straight line (not an open contour, because open contours don't exist in TrueType). Removing an on curve point between two off curve points would make it interpolated.

Moving: The on curve points being moved no longer affects any off curve points. Moving no longer changes point order, and points cannot automatically become interpolated or not (instead, this is now done by adding or removing the on curve point).

Rounding: Each defined point rounds individually, not relative to the on curve point.

skef:
Adding: Shift-Alt (typically) lets you drag a control point out of an off-curve point that lacks at least one. 

Removing: The two cases you describe seem to match "Make Line" (in the Point and right-click menus) and "Center Between Control Points" (in the Point menu).

I'm not sure I understand the other two issues.

yg8ijvjvjv:
> 
> 
> Adding: Shift-Alt (typically) lets you drag a control point out of an off-curve point that lacks at least one.
> 

I meant the direct add operation. Like how selecting the add point button and clicking on the outline would make an on curve point, selecting the add off curve point would do the same for an off curve point. When adding an off curve point when at least one of the points next to it is off curve, it would make an interpolated point (but not visible).

> Removing: The two cases you describe seem to match "Make Line" (in the Point and right-click menus) and "Center Between Control Points" (in the Point menu).

I meant as in the actual remove operation (select the point and Delete key, etc.) would do this, and the interpolated point would not even be visible.

> 
> I'm not sure I understand the other two issues.

[...]

PR #4457, opened 2020-10-03
Get the fundamental concept of new TrueType mode

yg8ijvjvjv:
We're still long way to implementing new TrueType mode but I got the basics. I got the point engine and the saving prompt.

### #4456
- **Bug fix**: Using the new TrueType mode fixes many bugs such as rounding introducing missing extrema or the Non Linear Transform causing glitching in the outline.
- **New feature**: It is now possible (in theory) to add on curve and off curve points separately, but it is still necessary to implement a graphical interface over that.
- **Breaking change**: It is no longer possible to save sfd files in the new TrueType mode.
- **Non-breaking change**: As a result, moving points changes so that each point, whether on curve or off curve (interpolated points don't count), can be selected and moved individually.

### DIFF
--- a/fontforge/bezctx_ff.c
+++ b/fontforge/bezctx_ff.c
@@ -103,24 +103,24 @@ bezctx_ff_quadto(bezctx *z, double xm, double ym, double x3, double y3, int poin
 		nancheck(bc);
 		xm = ym = x3 = y3 = 0;
 	    }
-	    if ( (sp=pointon?SplinePointCreate(x3,y3):SplinePointCreate(xm,ym))!=NULL ) {
-	  	  if(pointon){
+	    if ( (sp=pointoncurve?SplinePointCreate(x3,y3):SplinePointCreate(xm,ym))!=NULL ) {
+	  	  if(pointoncurve){
 		bc->ss->last->nextcp.x = x3;
 		bc->ss->last->nextcp.y = y3;
-		bc->ss->last->nonextcp = pointon; //Now determines whether it is on-curve (1) or off-curve (0)
+		bc->ss->last->nonextcp = pointoncurve; //Now determines whether it is on-curve (1) or off-curve (0)
 		sp->prevcp.x = x3; 
 		sp->prevcp.y = x3;
-		sp->noprevcp = pointon; //Now determines whether it is on-curve (1) or off-curve (0)
+		sp->noprevcp = pointoncurve; //Now determines whether it is on-curve (1) or off-curve (0)
 		if ( SplineMake3(bc->ss->last,sp)!=NULL )
 		    bc->ss->last = sp;
 		  }
 		  else{
 		bc->ss->last->nextcp.x = xm;
 		bc->ss->last->nextcp.y = ym;
-		bc->ss->last->nonextcp = pointon; //Now determines whether it is on-curve (1) or off-curve (0)
+		bc->ss->last->nonextcp = pointoncurve; //Now determines whether it is on-curve (1) or off-curve (0)
 		sp->prevcp.x = xm;
 		sp->prevcp.y = ym;
-		sp->noprevcp = pointon; //Now determines whether it is on-curve (1) or off-curve (0)
+		sp->noprevcp = pointoncurve; //Now determines whether it is on-curve (1) or off-curve (0)
 		if ( SplineMake3(bc->ss->last,sp)!=NULL )
 		    bc->ss->last = sp;
 		  }
--- a/fontforge/bezctx_ff.c
+++ b/fontforge/bezctx_ff.c
@@ -92,8 +92,9 @@ static void bezctx_ff_lineto(bezctx *z, double x, double y) {
 /* This could create a quadratic spline, except FontForge only is prepared */
 /* to deal with cubics, so convert the quadratic into the equivalent cubic */
 /* If NewTrueTypeModeEnabled, FontForge is prepared for TrueType. */
+int pointoncurve;
 static void
-bezctx_ff_quadto(bezctx *z, double xm, double ym, double x3, double y3, int pointoncurve) {
+bezctx_ff_quadto(bezctx *z, double xm, double ym, double x3, double y3) {
     if(NewTrueTypeModeEnabled){
     
 	    bezctx_ff *bc = (bezctx_ff *)z;
--- a/fontforge/bezctx_ff.c
+++ b/fontforge/bezctx_ff.c
@@ -92,7 +92,6 @@ static void bezctx_ff_lineto(bezctx *z, double x, double y) {
 /* This could create a quadratic spline, except FontForge only is prepared */
 /* to deal with cubics, so convert the quadratic into the equivalent cubic */
 /* If NewTrueTypeModeEnabled, FontForge is prepared for TrueType. */
-int pointoncurve;
 static void
 bezctx_ff_quadto(bezctx *z, double xm, double ym, double x3, double y3) {
     if(NewTrueTypeModeEnabled){
--- a/fontforge/bezctx_ff.c
+++ b/fontforge/bezctx_ff.c
@@ -91,8 +91,42 @@ static void bezctx_ff_lineto(bezctx *z, double x, double y) {
 
 /* This could create a quadratic spline, except FontForge only is prepared */
 /* to deal with cubics, so convert the quadratic into the equivalent cubic */
+/* If NewTrueTypeModeEnabled, FontForge is prepared for TrueType. */
 static void
-bezctx_ff_quadto(bezctx *z, double xm, double ym, double x3, double y3) {
+bezctx_ff_quadto(bezctx *z, double xm, double ym, double x3, double y3, int pointoncurve) {
+    if(NewTrueTypeModeEnabled){
+    
+	    bezctx_ff *bc = (bezctx_ff *)z;
+	    SplinePoint *sp;
+
+	    if ( !isfinite(xm) || !isfinite(ym) || !isfinite(x3) || !isfinite(y3)) {
+		nancheck(bc);
+		xm = ym = x3 = y3 = 0;
+	    }
+	    if ( (sp=pointon?SplinePointCreate(x3,y3):SplinePointCreate(xm,ym))!=NULL ) {
+	  	  if(pointon){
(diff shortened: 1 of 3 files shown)


frank-trampe:
I'll let @skef take a first run at the technical issues since you've discussed them extensively with him. But we definitely cannot merge something that is not compatible with SFD. All significant SplineFont data must persist in SFD. We can bump the SFD version if necessary, but the current version tolerates unknown single-line key/value pairs, and we might be able to make that work here.

yg8ijvjvjv:
> 
> 
> I'll let @skef take a first run at the technical issues since you've discussed them extensively with him. But we definitely cannot merge something that is not compatible with SFD. All significant SplineFont data must persist in SFD. We can bump the SFD version if necessary, but the current version tolerates unknown single-line key/value pairs, and we might be able to make that work here.

But it won't crash (you know, like crash), what happens is that only when the new TrueType mode is enabled saving is replaced with a prompt. SFD still persists for the PostScript mode and the legacy quadratic mode, and for the new TrueType mode it is recommended to save in ttf.

yg8ijvjvjv:
A prompt when a user tries to save or FontForge tries to autosave a new TrueType font, with four options: Save as ttf, Convert to legacy quadratic, Cancel, Never Save.

yg8ijvjvjv:
Completing the implementation of "Save as ttf" remains an exercise to the reader!

yg8ijvjvjv:
Hope it becomes inspirational to FontForge developers, it might seem overly complicated to effectively rebuild most of the code and there might be a lot of glitching in the way of doing this, however in the long run it will be more than compensated by the lesser trouble with functions glitching (due to every point being defined separately) and TrueType font designers. In future, sfd might be deprecated for quadratic fonts and they will be saved in the ttf format.

Other potential differences in the new TrueType format:
* Em size no longer split into Ascent and Descent
* Is Offset for metrics no longer available, because the TrueType design is guided around the font height boundaries, not the other way around which would be optically incorrect.
* Underline Position uses the TrueType convention rather than the PostScript convention

skef:
@frank-trampe Since I don't agree with some of the approach in #4456, and since this seems to be more of an aspirational PR that doesn't yet have the intended implementation, I'm not sure I'm the right person to review. 

Does the project accept shell PRs at present? I would have thought it required new features to be done prior to merging.

ctrlcctrlv:
> Completing the implementation of "Save as ttf" remains an exercise to the reader!

> Hope it becomes inspirational to FontForge developers

This isn't appropriate. PR's are not submitted to inspire others to do work. Sometimes others are willing to help but this shouldn't be expected or required. Of course, to get your PR submitted, you need to be prepared to do everything required. We can't merge anything incomplete that requires the "reader" (user?) to do "exercises" to make it complete.

frank-trampe:
Certainly. We can't merge work-in-progress pull requests, and I was not proposing that we do so. It seems to me that @yg8ijvjvjv is sufficiently interested to complete the work with the right support. If the core logic makes sense (and I have not looked at that at all), we need to sketch out what this needs to be viable. One thing of which I am certain is that the path mode must be set at font level and stored in SFD.

@yg8ijvjvjv, are you sure that this code tests correctly even for your narrow use case? I don't see what the point of using `pointoncurve` uninitialized is.

skef:
> One thing of which I am certain is that the path mode must be set at font level and stored in SFD.

I agree (https://github.com/fontforge/fontforge/issues/4456#issuecomment-703113712).

The broad idea of a different interaction mode seems reasonable. I'm generally disinclined toward new pallette tools unless the use case requires "dragging" but I've already gone over that stuff in the bug.

yg8ijvjvjv:
> 
> 
> Certainly. We can't merge work-in-progress pull requests, and I was not proposing that we do so. It seems to me that @yg8ijvjvjv is sufficiently interested to complete the work with the right support. If the core logic makes sense (and I have not looked at that at all), we need to sketch out what this needs to be viable. One thing of which I am certain is that the path mode must be set at font level and stored in SFD.
> 

I don't think sfd will be supported for the new TrueType mode, instead the user gets a nice prompt.

> @yg8ijvjvjv, are you sure that this code tests correctly even for your narrow use case? I don't see what the point of using `pointoncurve` uninitialized is.

I haven't got it to compile with (LGTM, AppVeyor, Travis CI) but I need a proper TrueType interface (each point moves individually, no open contours, etc.) and I do allow edits by maintainers too!

ctrlcctrlv:
> I don't think sfd will be supported for the new TrueType mode

SFD is FontForge's native format. Generating a TTF then editing that TTF and regenerating is an anti-pattern.

yg8ijvjvjv:
> 
> 
> > I don't think sfd will be supported for the new TrueType mode
> 
> SFD is FontForge's native format. Generating a TTF then editing that TTF and regenerating is an anti-pattern.

And making the new TrueType format would require remaking most of FontForge's natives all over again. Of course the current FontForge's native format cannot handle the new TrueType mode because it does not have the new natives necessary! The new TrueType mode is literally a TrueType mode so it would make sense for ttf.

ctrlcctrlv:
There's certainly no code here that implements a new file format. What is it that you're asking me to review? Do you have a standard for this new file format? Do you want to make TTF (SFNT) FontForge's native format?

yg8ijvjvjv:
> 
> 
> There's certainly no code here that implements a new file format. What is it that you're asking me to review? Do you have a standard for this new file format? Do you want to make TTF (SFNT) FontForge's native format?

FontForge would have a new mode

• Cubic mode — uses existing FontForge's natives in sfd
• Quadratic mode — uses existing FontForge's natives in sfd
• Mixed mode — uses existing FontForge's natives in sfd
• New TrueType mode — uses new natives so cannot be in sfd. There will probably not be a native format at all, meaning every autosave gets replaced with a prompt. If the user chooses Save as ttf in the prompt, FontForge can restore autosave from the ttf. If the user chooses Convert to legacy quadratic, the font will process conversion to sfd quadratic mode and then autosave proceeds as existing autosave. If the user chooses Cancel the autosave does not get saved. If the user chooses Never Save, the autosave does not ever get saved until the user restarts FontForge.

ctrlcctrlv:
Who is going to do the work to implement that?

frank-trampe:
@yg8ijvjvjv, if the new mode can be displayed from the SplineFont, SplineChar, and SplineSet data types, it can be stored in SFD. If the only state you add consists of the `NewTrueTypeModeEnabled` variable and the uninitialized `pointoncurve` variable (which I still don't understand), then one would simply put those somewhere in the SplineFont and round-trip them in SFD.

To clarify previous comments, we will not merge a change that adds font state that is not stored in SFD or that impairs the storage of existing state in SFD.

yg8ijvjvjv:
> 
> 
> @yg8ijvjvjv, if the new mode can be displayed from the SplineFont, SplineChar, and SplineSet data types, it can be stored in SFD. If the only state you add consists of the `NewTrueTypeModeEnabled` variable and the uninitialized `pointoncurve` variable (which I still don't understand), then one would simply put those somewhere in the SplineFont and round-trip them in SFD.
> 
> To clarify previous comments, we will not merge a change that adds font state that is not stored in SFD or that impairs the storage of existing state in SFD.

pointoncurve is used in the font editor, not the font itself. It is to determine which tool (off curve point or on curve point) is being used.
image

ctrlcctrlv:
Also I'm finding this PR to be disruptive as it's causing many email notifications every time a minor push is made. This is clearly nowhere near merge-ready. It should be closed and reopened when closer to ready. This is an issue masquerading as a PR, IMHO.

yg8ijvjvjv:
> 
> 
> Also I'm finding this PR to be disruptive as it's causing many email notifications every time a minor push is made. This is clearly nowhere near merge-ready. It should be closed and reopened when closer to ready. This is an issue masquerading as a PR, IMHO.

The new TrueType mode is still a long way to be done and most of functions need to be rewritten to handle it.

ctrlcctrlv:
OK, so let's close this for now. A _pull request_ is a _request that code be merged_. They should be submitted when the author thinks the code is done or close to it. You're welcome to continue working on your fork, and to discuss things in appropriate issues, and then reopen a new PR when you're much closer to done.

yg8ijvjvjv:
When this will be done, most of the FontForge files will have changed with major modifications to the source code as necessary to handle new TrueType mode.

ctrlcctrlv:
I certainly understand that. 

Large changes are allowed just like small changes are allowed. It might take longer to review but if you come with a large change somebody will review it. If you're touching many areas we might need multiple people to help. Every open pull request more than a month old has gotten reviewed and the reason it's stuck is not because it's waiting for a review. Most of the time the problem is that people have found another way to do the same thing and don't have the time to fix the review issues

yg8ijvjvjv:
Here is a proof of concept of what such an editor interface would look like. https://ufile.io/3ex8tqan
Demonstration: https://www.youtube.com/watch?v=HmL-iCX_0jM (note how each point selects and moves independently and interpolated points are not points)

yg8ijvjvjv:
Not implemented in the proof of concept:
FontForge - This is not FontForge code and cannot be merged.
Insertion and deletion - deletion would be straightforward in TrueType point order, insertion on spline regarding the side of off-curve point will likely follow which side of the spline midpoint it is.
Rectangle selection - should logically map to the points.
Undo
Redo
Snapping (except for detecting a nearby point clicked on) - should snap to any point including off-curve excluding interpolated
Incrementally building up the outline - as TrueType has no open outlines, it might be an interface of repeatedly inserting points after the currently selected point, whereas a new contour is made if less or more than one point is selected.
Shift-moving (in 45° increments) - should be relative to the original point position, not a nearby on-curve point like FontForge does with off-curve points.

yg8ijvjvjv:
The Point Info in FontForge quadratic mode looks like this:

○ Normal ○ Interpolated
- [ ] Never Interpolate
Base:   ⃞   ⃞
Prev CP: ☐ Default
Offset   ⃞   ⃞
Dist   ⃞   ⃞
Curvature: 
Next CP: ☐ Default
Offset   ⃞   ⃞
Dist   ⃞   ⃞
Curvature: 

Type: (curve, HV curve, corner, tangent)

However, the Point Info in TrueType would look like this:

position: x:   ⃞ y:   ⃞
type: (off-curve, on-curve)

PR #4463, opened 2020-10-07
gmenu.c: prevent the menu key from creating a menu if one is already present

### LINKED ISSUE #4462: issue about right-click button on windows OS
Hi
in the latest fontforge version on windows(and maybe in other versions), right-click button(on the keyboard) working continuously(stacking right-click menu bars as long as keep pushing on the right-click keyboard key).

### LINKED ISSUE #4662: Why does the non-standard LANGUAGE environment variable exist? (FontForge locale is strange)
This works fine for me with qBittorrent:

```bash
export L=ja_JP.UTF-8; echo $L && LC_ALL=$L LANG=$L qbittorrent
```
image

But not FontForge:

```bash
export L=ja_JP.UTF-8; echo $L && LC_ALL=$L LANG=$L fontforge -new
```

image

Even supplying `LANGUAGE` doesn't work…:

```bash
export L=ja_JP.UTF-8; echo $L && LC_ALL=$L LANG=$L LANGUAGE=$L fontforge -new
```

(still launches in English)

But strangely:

```bash
export L=ja_JP.UTF-8; echo $L && LANGUAGE=$L fontforge -new
```

image

I confirmed that all of this is true for `/usr/bin/fontforge` (Arch's `fontforge` package) as well as `/opt/bin/fontforge` (my compiled `fontforge` w/debug symbols which I actually use in production).

So, what's going on here?

* Why does `LANGUAGE` exist at all?
* Why are `LC_ALL` and `LANG` ignored?

jtanx:
Fixes #4462

### Type of change

- **Bug fix**

### DIFF
--- a/gdraw/gmenu.c
+++ b/gdraw/gmenu.c
@@ -2026,7 +2026,7 @@ int GMenuBarCheckKey(GWindow top, GGadget *g, GEvent *event) {
     }
 
 //    TRACE("menubarcheckkey(e2)\n");
-    if ( event->u.chr.keysym==GK_Menu )
+    if ( event->u.chr.keysym==GK_Menu && most_recent_popup_menu==NULL )
 	GMenuCreatePopupMenu(event->w,event, mb->mi);
 
     return( false );


frank-trampe:
`GMenuCreatePopupMenuWithName`, which ultimately sets `most_recent_popup_menu`, is called in `fontview.c` and `cvpalettes.c` and maybe other places. Do we need to guard those calls too?

jtanx:
Those should be fine since they're only called in response to right clicks, and the mouse is grabbed while a menu is active. But there's no harm in adding another check there so I've added it.

jtanx:
Actually change of mind, I'll just keep the original fix.

PR #4486, opened 2020-11-07
Add a 20th anniversary splash screen.

frank-trampe:
Add a temporary splash screen, modified from the 2019 artwork by @ctrlcctrlv in #3754, commemorating the 20th anniversary of FontForge.

The artwork files follow. The files going into the source tree are scaled 50% to match existing conditions.

FF20_1
[FF20_1.xcf.zip](https://github.com/fontforge/fontforge/files/5505269/FF20_1.xcf.zip)


frank-trampe:
If you're building and not installing, you need to copy fontforgeexe/pixmaps/tango to build/share/fontforge/pixmaps for testing.

fontforgeexe/startui.c
```diff
@@ -203,19 +203,28 @@ static void InsCharHook(GDisplay *gd,unichar_t ch) {
...
 static GWindow splashw;
 static GTimer *autosave_timer, *splasht;
 static GFont *splash_font, *splash_italic, *splash_mono;
 static int as,fh, linecnt;
 static unichar_t msg[546];
 static unichar_t *lines[32], *is, *ie;
 
+static void SplashImageInit() {
+    if (splashimagep == NULL)
+        splashimagep = GGadgetImageCache("splash2020.png");
```
skef, line 217: My one comment is that if you were to name this "spash_special.png" (or something) instead, future releases could have a dedicated splash-screen with no code changes. Not sure if that's worth it or not.

skef (approved):
I didn't test this but the changes look fine to me.

fontforgeexe/startui.c
frank-trampe: I thought about doing it that way. The major deciding factor for me was that it is easier to track text changes than to track binary file replacements. As it is, the regular splash screen is named splash2019.png, for example, so switching back is as easy as a one-line patch of two characters.

PR #4493, opened 2020-11-11
Better heuristics for correction of direction in Expand Stroke

### LINKED ISSUE #4492: Changed behavior in version 20200314 when stroking a self-intersecting contour
The following script generates a font with a glyph of a hollow square in version 20190801. In versions 20200314 and 20201107, the glyph is filled in. I am using macOS 10.14.6.
```python
import fontforge

font = fontforge.font()
glyph = font.createChar(0x30)
pen = glyph.glyphPen()
pen.moveTo((0, 0))
pen.lineTo((0, 100))
pen.lineTo((100, 100))
pen.lineTo((100, 0))
pen.lineTo((10, 0))
pen.endPath()
glyph.stroke('circular', 30, 'round')
font.generate('bug.otf')
```
This happens when the contour intersects itself or almost intersects itself. The width of the stroke is relevant: with narrower strokes, the glyph stays hollow. What is the new intended way to stroke a hollow shape?

skef:
Closes #4492 

I'm unhappy with the heuristics I have available at this stage of the processing but this is an improvement. 

This issue was caused by my forgetting that `RemoveOverlap` can return multiple contours in any order, and I was futzing with the direction of the first one. I shouldn't touch its output as it knows more about direction than I do, hence the ordering change for the counter-clockwise case.

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -2288,6 +2288,10 @@ static SplineSet *OffsetSplineSet(SplineSet *ss, StrokeContext *c) {
 		    left = SplineSetRemoveOverlap(NULL,left,over_remove);
 		// Open paths don't always produce clockwise output
 		is_ccw_mid = SplinePointListIsClockwise(left);
+		if ( is_ccw_mid==0 && left->next==NULL ) {
+		    SplineSetReverse(left);
+		    is_ccw_mid=1;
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
Doesn't break any of my fonts and is tiny. Seems like better behavior.

PR #4498, opened 2020-11-11
Add mnemonic underscores to Python registerMenuItem

ctrlcctrlv:
This allows Python scripts to add "mnemonics" (alt-key sequences).

Note underscored letters in example:

image

```python
fontforge.registerMenuItem(autoNameGlyphPoints, None, None, "Glyph", None, "_Higher-Order Interpolation", "Attempt to automatically _name points")
```

In `master`, this would not work, underscores would appear in the text and <kbd>alt-T-h-n</kbd> wouldn't work.

### Type of change

- **New feature**

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -1427,7 +1427,7 @@ void python_call_onClosingFunctions()
 /* ************************ User Interface routines ************************* */
 /* ************************************************************************** */
 
-static PyObject *PyFF_registerMenuItemStub(PyObject *UNUSED(self), PyObject *UNUSED(args)) {
+static PyObject *PyFF_registerMenuItemStub(PyObject *UNUSED(self), PyObject *UNUSED(args), PyObject *UNUSED(kwargs)) {
 //    printf("PyFF_registerMenuItemStub()\n");
     /* This is a stub which will be replaced when we've got a UI */
 Py_RETURN_NONE;
@@ -18787,7 +18787,7 @@ PyMethodDef module_fontforge_methods[] = {
     /* Access to the User Interface ... if any */
     { "hasUserInterface", PyFF_hasUserInterface, METH_NOARGS, "Returns whether this fontforge session has a user interface (True if it has opened windows) or is just running a script (False)"},
     { "registerImportExport", PyFF_registerImportExport, METH_VARARGS, "Adds an import/export spline conversion module"},
-    { "registerMenuItem", PyFF_registerMenuItemStub, METH_VARARGS, "Adds a menu item (which runs a python script) to the font or glyph (or both) windows -- in the Tools menu"},
+    { "registerMenuItem", (PyCFunction)PyFF_registerMenuItemStub, METH_VARARGS | METH_KEYWORDS, "Adds a menu item (which runs a python script) to the font or glyph (or both) windows -- in the Tools menu"},
     { "getConvexNib", PyFF_getConvexNib, METH_VARARGS, "Sets the specified 'Convex' to the layer/contour argument."},
     { "setConvexNib", PyFF_setConvexNib, METH_VARARGS, "Returns the specified 'Convex' nib as a layer."},
     { "logWarning", PyFF_logError, METH_VARARGS, "Adds a non-fatal message to the Warnings window"},
@@ -18875,12 +18875,12 @@ static module_definition module_def_fontforge = {
 /* ************************************************************************** */
 /* ************************* initializer routines *************************** */
 /* ************************************************************************** */
-void FfPy_Replace_MenuItemStub(PyObject *(*func)(PyObject *,PyObject *)) {
+void FfPy_Replace_MenuItemStub(PyObject *(*func)(PyObject *,PyObject *,PyObject *)) {
     int i;
     PyMethodDef *methods = module_fontforge_methods;
     for ( i=0; methods[i].ml_name!=NULL; ++i )
 	if ( strcmp(methods[i].ml_name,"registerMenuItem")==0 ) {
-	    methods[i].ml_meth = func;
+	    methods[i].ml_meth = (PyCFunction)func;
 return;
 	}
 }
--- a/fontforgeexe/pythonui.c
+++ b/fontforgeexe/pythonui.c
@@ -231,7 +231,9 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv, unichar_t*
     }
 
     for ( i=5; i<cnt; ++i ) {
-	unichar_t *submenuu = utf82u_copy(PyUnicode_AsUTF8(PyTuple_GetItem(args, i)));
+	// 0xFFFF means mnemonic unset
+	unichar_t t_mnemonic = mnemonic[i-5] == 0xFFFF ? 0 : mnemonic[i-5];
+	unichar_t *submenuu = utf82u_mncopy(PyUnicode_AsUTF8(PyTuple_GetItem(args, i)), &t_mnemonic);
 	for (unichar_t* p = submenuu; *p != '\0'; p++) {
 	    if (*p == '_') *p = 0x200B; // zero-width space
 	}
@@ -253,8 +255,7 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv, unichar_t*
 	if ( mmn[j].ti.text==NULL ) {
 	    mmn[j].ti.text = submenuu;
 	    mmn[j].ti.fg = mmn[j].ti.bg = COLOR_DEFAULT;
-	    // 0xFFFF means mnemonic unset
-	    mmn[j].ti.mnemonic = mnemonic[i-5] == 0xFFFF ? 0 : mnemonic[i-5];
+	    mmn[j].ti.mnemonic = t_mnemonic;
 	    // If in first or middle submenu
 	    if ( i!=cnt-1 ) {
 		mmn[j].mid = -1;
--- a/fontforgeexe/pythonui.c
+++ b/fontforgeexe/pythonui.c
@@ -213,7 +213,7 @@ return( fvpy_menu_cnt++ );
     }
 }
 
-static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv) {
+static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv, unichar_t* mnemonic) {
     int i, j, cnt;
     PyObject *func, *check, *data;
     char *shortcut_str = NULL;
@@ -232,6 +232,9 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv) {
 
     for ( i=5; i<cnt; ++i ) {
 	unichar_t *submenuu = utf82u_copy(PyUnicode_AsUTF8(PyTuple_GetItem(args, i)));
+	for (unichar_t* p = submenuu; *p != '\0'; p++) {
+	    if (*p == '_') *p = 0x200B; // zero-width space
+	}
 
 	j = 0;
 	if ( *mn != NULL ) {
@@ -250,10 +253,14 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv) {
 	if ( mmn[j].ti.text==NULL ) {
 	    mmn[j].ti.text = submenuu;
 	    mmn[j].ti.fg = mmn[j].ti.bg = COLOR_DEFAULT;
+	    // 0xFFFF means mnemonic unset
+	    mmn[j].ti.mnemonic = mnemonic[i-5] == 0xFFFF ? 0 : mnemonic[i-5];
+	    // If in first or middle submenu
 	    if ( i!=cnt-1 ) {
 		mmn[j].mid = -1;
 		mmn[j].moveto = is_cv ? cvpy_tllistcheck : fvpy_tllistcheck;
 		mn = &mmn[j].sub;
+	    // If at end of tree
 	    } else {
 		mmn[j].shortcut = shortcut_str;
 		mmn[j].invoke = is_cv ? cvpy_menuactivate : fvpy_menuactivate;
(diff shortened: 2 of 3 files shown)


fontforge/python.c
```diff
@@ -18787,7 +18787,7 @@ PyMethodDef module_fontforge_methods[] = {
     /* Access to the User Interface ... if any */
     { "hasUserInterface", PyFF_hasUserInterface, METH_NOARGS, "Returns whether this fontforge session has a user interface (True if it has opened windows) or is just running a script (False)"},
     { "registerImportExport", PyFF_registerImportExport, METH_VARARGS, "Adds an import/export spline conversion module"},
-    { "registerMenuItem", PyFF_registerMenuItemStub, METH_VARARGS, "Adds a menu item (which runs a python script) to the font or glyph (or both) windows -- in the Tools menu"},
+    { "registerMenuItem", (PyCFunction)PyFF_registerMenuItemStub, METH_VARARGS | METH_KEYWORDS, "Adds a menu item (which runs a python script) to the font or glyph (or both) windows -- in the Tools menu"},
```
ctrlcctrlv, line 18790: I did this because going forwards I want to be able to pass kwargs into this function. Passing kwargs in master is currently an error. This is a relaxation of that so in a future release we can consider expanding this function further.

skef:
How is a mnemonic different from a shortcut. The shortcut system currently doesn't work but there is an interface for it. Or is this about alt-codes?

ctrlcctrlv:
>Or is this about alt-codes?

Yeah.

skef:
Should we consider putting those by convention before the pipe in the shortcut string? That's sort of the rough intention as things stand.

ctrlcctrlv:
Can you show me an example? The shortcut string didn't seem to do anything in my tests.

ctrlcctrlv:
I did it this way because it's the convention in the C GDraw code, by the way~

skef:
Ah, well.

The shortcut string hasn't done anything since the hotkey file changes (assuming it worked before). I have it recorded in some other bug to change it to serve as a "suggestion", which will be activated as long as the code isn't already in use. 

However, you can see from [the docs](https://fontforge.org/docs/scripting/python/fontforge.html#fontforge.registerMenuItem) that the previous convention was "Name|Ctrl-H" and so forth. I'm basically suggesting an interface where "_A|" would be an alt code only, "|Ctrl-H" would be a shortcut only, and "_A|Ctrl-H" would be both. 

That convention is broadly consistent with what used to be true and doesn't require any interface changes.

fontforgeexe/pythonui.c
```diff
@@ -232,6 +232,9 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv) {
 
     for ( i=5; i<cnt; ++i ) {
 	unichar_t *submenuu = utf82u_copy(PyUnicode_AsUTF8(PyTuple_GetItem(args, i)));
+	for (unichar_t* p = submenuu; *p != '\0'; p++) {
+	    if (*p == '_') *p = 0x200B; // zero-width space
```
jtanx, line 236: care to explain this one a bit more?

fontforgeexe/pythonui.c
```diff
@@ -250,10 +253,13 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv) {
 	if ( mmn[j].ti.text==NULL ) {
 	    mmn[j].ti.text = submenuu;
 	    mmn[j].ti.fg = mmn[j].ti.bg = COLOR_DEFAULT;
+	    mmn[j].ti.mnemonic = mnemonic[i-5];
```
jtanx, line 256: i-5?

fontforgeexe/pythonui.c
```diff
@@ -275,13 +281,30 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv) {
...
 /* (function,check_enabled,data,(char/font),shortcut_str,{sub-menu,}menu-name) */
-static PyObject *PyFF_registerMenuItem(PyObject *self, PyObject *args) {
+static PyObject *PyFF_registerMenuItem(PyObject *self, PyObject *args, PyObject *kwargs) {
     int i, cnt;
     int flags;
+    unichar_t* mnemonic;
 
     if ( !no_windowing_ui ) {
 	cnt = PyTuple_Size(args);
+	mnemonic = calloc(cnt, sizeof(unichar_t));
```
jtanx, line 307: leaks

fontforgeexe/pythonui.c
```diff
@@ -300,20 +323,26 @@ return( NULL );
...
 return( NULL );
 	}
+	PyObject* temp_u;
+	int idx;
 	if ( PyTuple_GetItem(args,4)!=Py_None ) {
-	    if (PyUnicode_AsUTF8(PyTuple_GetItem(args, 4)) == NULL) {
+	    if ((temp_u = PyTuple_GetItem(args, 4)) == NULL) {
 		return NULL;
+	    } else {
+		idx = append_mnemonic(temp_u, mnemonic);
```
jtanx, line 332: what is idx used for

fontforgeexe/pythonui.c
```diff
@@ -275,13 +281,30 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv) {
...
+    int idx = 0;
+    for (char* p = PyUnicode_AsUTF8(in); *p != '\0'; p++) {
+	if (*p == '_' && *(p+1) != '\0') {
+	    temp[0] = *(p+1);
+	    u_strcat(mnemonic, temp);
+	    return idx;
+	}
+	idx++;
+    }
+    u_strcat(mnemonic, temp);
```
jtanx, line 295: surely this would do nothing?

fontforgeexe/pythonui.c
ctrlcctrlv: So that a double underscore isn't shown to the user, we replace it. Otherwise user would see something like «_<ins>H</ins>igher-Order Interpolation».

fontforgeexe/pythonui.c
ctrlcctrlv: `i` comes from Python argument index, which will at this point always be ≥ 5, which is why in the parent loop `i` = 5, because `cnt` == number of Python arguments.

fontforgeexe/pythonui.c
ctrlcctrlv: Oh, you're right. I'll have to use the guaranteed non-character, U+FFFF, instead of 0. (I attended the last UTC meeting and contribute to Unicode when I can so I'm sure that'll never become a character, they are exceedingly cautious about backwards compatibility.)

ctrlcctrlv:
On further reflection, a problem with @skef's observation of possible former `shortcutstr` behavior is that `registerMenuItem` _also_ defines submenus, so the convention of `_A|` cannot actually replicate all the behavior I implemented, because then how would you add the mnemonic for the parent menu, you could only add it for the bottom-most item, right?

ctrlcctrlv:
FYI @jtanx 
image
```python3
fontforge.registerMenuItem(subdivide, None, None, "Glyph", None, "_Higher-Order Interpolation", "_Subdivide selected spline")
fontforge.registerMenuItem(subdivide, None, None, "Glyph", None, "_Higher-Order Interpolation", "hi @jtanx", "hi _@jtanx", "_Subdivide selected spline")
```

skef:
> On further reflection, a problem with @skef's observation of possible former `shortcutstr` behavior is that `registerMenuItem` _also_ defines submenus, so the convention of `_A|` cannot actually replicate all the behavior I implemented, because then how would you add the mnemonic for the parent menu, you could only add it for the bottom-most item, right?

Well, I suppose that's true, but on the other hand different calls to `registerMenuItem()` could register different shortcuts to the higher menus and we'd have to sort that shit out. (I guess you could do first-wins or last-wins. We could also just make them different menus because the strings are different but that would be terrible.)

If we do this we need, at minimum, a way of escaping an underscore so that it can appear in the menu, because even if we don't care about that someone might. Is there already a means for that?

ctrlcctrlv:
I don't care about that and don't think it's important but I do understand your concern. 

I could make it so that 2 underscores are interpreted as one with no mnemonic. Is that acceptable?

Making it so that registerMenuItem is meant to be invoked multiple times will break more existing scripts than this patch—or otherwise create a confusing API where you only need multiple invocations when you want menu mnemonics.

skef:
Let me think about the other questions a bit.

> Making it so that registerMenuItem is meant to be invoked multiple times will break more existing scripts than this patch—or otherwise create a confusing API where you only need multiple invocations when you want menu mnemonics.

My point was about sub-menus. In almost all cases where you have a submenu the reason is to put multiple things there. So submenus are almost always registered multiple times in different calls to registerMenuItem for the "leaf entries"

ctrlcctrlv:
It's true that submenus almost always involve multiple invocations, but in every case, no current script would register the submenu first before its children (to my knowledge).

ctrlcctrlv:
Pinging @jtanx~

fontforgeexe/pythonui.c
jtanx: I think I'm missing something here. Why do the compiled-in menu items show fine without having to do this?

jtanx: This looks relevant:

https://github.com/fontforge/fontforge/blob/4af47a671c0c32609e6e4b3395fd26b4639317bd/gdraw/gtextinfo.c#L276

ctrlcctrlv: Ah-ha @jtanx :)

So the real bug here is:

https://github.com/fontforge/fontforge/blob/4af47a671c0c32609e6e4b3395fd26b4639317bd/fontforgeexe/pythonui.c#L234

[Fixed. Now we're using `utf82u_mncopy` as we always should've been.](https://github.com/fontforge/fontforge/pull/4498/commits/142a37cc6c90a45e6a6a689eebd0c49bd6fc0de2#diff-75edc152c5687a95f25ee0a7ceaa2a2c5e3f92c0d5b8d60c7c01111004658c64L234)

jtanx: So now you're using it, why do you need to do this zero-width space thing and why do you still try to find the mnemonics yourself with `append_mnemonic`?

ctrlcctrlv: @jtanx Well we have to populate `t_mnemonic` somehow. The `mncopy` version must know which character is the mnemonic. Note how there are two fields at play here, `text` and `mnemonic`. Because everywhere else they're hardcoded, we just set e.g. 
```c
gg[i].text = _("_Go");
gg[i].mnemonic = 'G'; // this line matters!
```

Consider this patch:

```patch
diff --git a/fontforgeexe/charinfo.c b/fontforgeexe/charinfo.c
index 44b6d7544..fd36435e4 100644
--- a/fontforgeexe/charinfo.c
+++ b/fontforgeexe/charinfo.c
@@ -4382,19 +4382,19 @@ return;
        memset(&ubox,0,sizeof(ubox));
        memset(&ulabel,0,sizeof(ulabel));
 
-       ulabel[0].text = (unichar_t *) _("Gl_yph Name:");
+       ulabel[0].text = (unichar_t *) _("Glyp_h Name:");
        ulabel[0].text_is_1byte = true;
        ulabel[0].text_in_resource = true;
        ugcd[0].gd.label = &ulabel[0];
        ugcd[0].gd.pos.x = 5; ugcd[0].gd.pos.y = 5+4; 
        ugcd[0].gd.flags = gg_enabled|gg_visible;
-       ugcd[0].gd.mnemonic = 'N';
+       ugcd[0].gd.mnemonic = 'y';
        ugcd[0].creator = GLabelCreate;
        uhvarray[0] = &ugcd[0];
 
        ugcd[1].gd.pos.x = 85; ugcd[1].gd.pos.y = 5;
        ugcd[1].gd.flags = gg_enabled|gg_visible;
-       ugcd[1].gd.mnemonic = 'N';
+       ugcd[1].gd.mnemonic = 'y';
        ugcd[1].gd.cid = CID_UName;
        ugcd[1].creator = GListFieldCreate;
        ugcd[1].data = (void *) (-2);
```

If you compile and open the «Glyph Info» dialog (<kbd>Ctrl + I</kbd>), you'll see <ins>h</ins> marked as the mnemonic, but note that `y` does indeed work!

So, `utf82u_mncopy` only does a special copy into the `text` field _for the purposes of display of the mnemonic string_. But `mnemonic` still must get set somehow—`utf82u_mncopy` can't do it, it doesn't get a reference to that.

jtanx: I don't buy this. By inspection, `utf82u_mncopy` does not read in `mn` (the second parameter) - it *stores* the mnemonic it stripped out. 

In your example, I would not be surprised if there are flags that control whether or not to use the mnemonic specified or something along those lines.

Even without that part, I still don't understand why you have this zero-width-space logic after having run through `utf82u_mncopy`.

I just ran this branch locally with this patch:

```c
diff --git a/fontforgeexe/pythonui.c b/fontforgeexe/pythonui.c
index 3d36e0089..ed8dd6fe8 100644
--- a/fontforgeexe/pythonui.c
+++ b/fontforgeexe/pythonui.c
@@ -232,11 +232,8 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv, unichar_t*

     for ( i=5; i<cnt; ++i ) {
        // 0xFFFF means mnemonic unset
-       unichar_t t_mnemonic = mnemonic[i-5] == 0xFFFF ? 0 : mnemonic[i-5];
+       unichar_t t_mnemonic = 0;// = mnemonic[i-5] == 0xFFFF ? 0 : mnemonic[i-5];
        unichar_t *submenuu = utf82u_mncopy(PyUnicode_AsUTF8(PyTuple_GetItem(args, i)), &t_mnemonic);
-       for (unichar_t* p = submenuu; *p != '\0'; p++) {
-           if (*p == '_') *p = 0x200B; // zero-width space
-       }

        j = 0;
        if ( *mn != NULL ) {
```

and it seemed to work fine for me - which suggests to me that `append_mnemonic` and this zero-width-space business is not needed.

skef:
After thinking about this more, from the perspective of contributed scripts the whole problem of Tools menu accelerators (as opposed to sub-menus) is a mess.

I've been favoring adding hotkey support as a "suggestion", where the hotkey is enabled if it's not already in use. The same convention could work for alt-accelerators with a given menu: the entry gets it as long as it's free within the menu, otherwise its skipped.

That leaves the need (or desire) for a user to configure their own. One natural solution to this is extending the syntax of the `hotkeys` file to allow underscores to configure alt-accelerators. That leaves two main questions:

1. Should we add this support for all menus or just Tools menu entries, and
2. How hard will it be to implement? 

(I'm assuming that only a "present" underscore would have an effect, so if there isn't one in a hotkey file the originally chosen accelerator is retained.)

jtanx:
Not entirely sure I follow what makes this a mess.

Given #4682, I think the entirety of this PR can be reduced to:

```diff
diff --git a/fontforgeexe/pythonui.c b/fontforgeexe/pythonui.c
index 7b49bf3..ea98798 100644
--- a/fontforgeexe/pythonui.c
+++ b/fontforgeexe/pythonui.c
@@ -231,7 +231,8 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv) {
     }

     for ( i=5; i<cnt; ++i ) {
-       unichar_t *submenuu = utf82u_copy(PyUnicode_AsUTF8(PyTuple_GetItem(args, i)));
+       unichar_t mnemonic = 0;
+       unichar_t *submenuu = utf82u_mncopy(PyUnicode_AsUTF8(PyTuple_GetItem(args, i)), &mnemonic);

        j = 0;
        if ( *mn != NULL ) {
@@ -249,6 +250,7 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv) {
        mmn = *mn;
        if ( mmn[j].ti.text==NULL ) {
            mmn[j].ti.text = submenuu;
+           mmn[j].ti.mnemonic = mnemonic;
            mmn[j].ti.fg = mmn[j].ti.bg = COLOR_DEFAULT;
            if ( i!=cnt-1 ) {
                mmn[j].mid = -1;
@@ -264,6 +266,7 @@ static void InsertSubMenus(PyObject *args,GMenuItem2 **mn, int is_cv) {
                mn = &mmn[j].sub;
            else {
                char *temp = u2utf8_copy(submenuu);
+               mmn[j].ti.mnemonic = mnemonic;
                mmn[j].shortcut = shortcut_str;
                mmn[j].invoke = is_cv ? cvpy_menuactivate : fvpy_menuactivate;
                mmn[j].mid = MenuDataAdd(func,check,data,is_cv);
```

If we're happy with that I can update this PR accordingly.

jtanx:
Actually looking at this again, does #4682 supersede this completely?

skef:
I believe this is entirely superseded by #4682 .

PR #4501, opened 2020-11-15
Fix double free crash when flattening CID fonts

ahyangyi:
### Type of change
- **Bug fix**
This PR fixes a crash that happens when one flattens a CID font with anchors and immediately saves it to an SFD file.

I traced the problem with Address Sanitizer and apparently it was because the anchor in the master font is freed but still referenced in the flattened fonts.

Obviously this fix introduces a memory leak of a few bytes, but I doubt it's of any concern when we still have crash bugs to deal with.

I'm bothered to create an issue for an obvious one-liner fix.

Way to reproduce (more reliably when also running with Address Sanitizer, since double freeing does not guarantee a crash on all systems):
1. open NotoSansCJK
2. flatten
3. save to an sfd file

### DIFF
--- a/fontforge/encoding.c
+++ b/fontforge/encoding.c
@@ -1529,6 +1529,7 @@ return(NULL);
 	fvs->sf = new;
 	FVSetTitle(fvs);
     }
+    cidmaster->anchor = NULL;
     FontViewReformatAll(new);
     SplineFontFree(cidmaster);
 return( new );


ahyangyi:
Note: after applying this fix, I no longer get crashes but the anchor points in the flattened fonts are still wrong/missing.

(still, not crashing is an improvement, so I'm not willing to devote further time before this gets merged first)

ctrlcctrlv:
Thank you! I'm going to leave this open as I'm not sure if others agree on merging a fix we know still has issues. Now that we know the problem thanks to your research maybe we can fix the anchor points. Can you provide a sample font?

jtanx:
> Obviously this fix introduces a memory leak of a few bytes, but I doubt it's of any concern when we still have crash bugs to deal with.

Disagree with this, this change obscures the true bug while introducing another. It sounds more like `AnchorPointsFree` is being called twice and/or it being called without the field being nulled b the caller.

jtanx:
From the trace I think it's pretty clear the anchor classes aren't being set properly on the flattened `SplineFont` and that neither are the anchor points on the `SplineChars` being updated (so they still point to the now freed `AnchorClass`es)

ahyangyi:
> AnchorClassesFree is being called twice
AnchorClassesFree was only called once, and it was on the CID font.

>  it being called without the field being nulled b the caller.
That's exactly what this one-liner patch does.

Of course, this patch is indeed "incorrect" in that the flattened font still has no anchors, and glyphs in the flattened font have invalid references. (which results in tons of warnings upon opening the dumped font)

A hypothetical "more correct" fix would also create (or move) the anchors to the flattened font, and update the anchor references in the glyphs.

I think I agree with you on this. I just wonder whether we have a developer who is willing to create the more correct fix. And if not, whether the temporary hack does more good or harm.

If you feel the change "obscures the real problem", what about adding a comment about the nature of the nulling?

ctrlcctrlv:
(It's somewhat questionable to me at least whether or not crashing is preferable to generating a font we know to be wrong, that is to say, a font without anchors when we know it should have anchors.)

ahyangyi:
In my use case, not crashing is definitely preferred -- since I am basically pasting the CJK characters from NotoSansCJK to my own Latin font, and the anchors are mostly not related to the CJK characters.

I understand that the priority is different for everyone though.

ahyangyi:
Anyway, I would also argue that while missing anchors might be incorrect, crashing is *never* correct.

jtanx:
Maybe, but two wrongs don't make a right. If it completely dropped the anchors that's one thing, but this is *knowingly* using the anchors off the now disposed of instance.

ahyangyi:
That's because I don't know how to correctly drop all anchors once and for all.

Anyways, I am encountering another crashing bug related to CID fonts, so I'll work on that for now and perhaps that one would be connected to this one. I think that alone is a good reason we wait some time and do not rush to merge this.

Since the only proponent of this PR (that is, me) now falls back, I guess we now have a consensus... :P

ahyangyi:
P. S. If there is an easy way to simply clear all anchors from the glyphs, please tell me.

I won't use that in merge requests, but they might help my private branches, and they might help me understand how to implement the proper thing.

ctrlcctrlv:
I would often prefer a program crash than silently produce defective output~

ahyangyi:
While I definitely prefer (slightly) defective output to crashing, I notice that it's now a choice between always crashing and "sometimes crashing and otherwise defective output". And even I can't say I prefer the latter.

To be precise, when I call from Python, it crashes; and when I manually reproduce a similar sequence of operations, it does not crash. And it's somewhat hard to produce precise stack traces when I call from Python. I leave all those debug information, frame points etc, but I still only managed to obtain very mutilated stack traces so far, making it very hard to track the root cause down.

All I can say for now is that this problem is deeper than it looks. :(

ctrlcctrlv:
I'm closing this as it's been almost a month, and this change has two of us against it because it doesn't address the deeper problem.

If users want to add kludges to FontForge's source code to just get _some_ output, that's their prerogative, but we shouldn't be merging patches we know are going to lead to defective output.

I'm sorry we disagree on the way to develop the source code, but silent errors are always worse in my opinion. While anchors might be rare in CID fonts, and aren't oft-used in CJK languages, that doesn't excuse us.

A new PR which solves the problem is welcome.

ahyangyi:
Fair, I also wanted to close this.

The real kludge is to remove the call to `SplineFontFree(cidmaster)` instead. That's a real fix to crash at the cost of memory leak. I'm switching to that in my local branch.

And of course, the real fix is to get the anchors working again.

This PR is neither of the above, and is even uglier than removing the call to `SplineFontFree`.

We can agree to disagree on software development approaches, but we can agree that this particular PR isn't worth it...

PR #4506, opened 2020-11-19
[gdraw] Run filter to determine if directory should be treated as file

### LINKED ISSUE #4495: Open dialog: Double-clicking UFO acts differently than clicking once and hitting «OK»
Double-clicking opens the UFO directory, very unlikely to be user's desired action.

Clicking once and pressing OK opens UFO in FontForge as expected.

ctrlcctrlv:
Previously filters were only run on files for the purpose of deciding
whether or not to display them. I realized that it also makes sense to
run them on directory names so that FontForge correctly opens a UFO file
on double click.

This closes #4495.

### Type of change

- **Bug fix**

### DIFF
--- a/gdraw/gfilechooser.c
+++ b/gdraw/gfilechooser.c
@@ -727,7 +727,6 @@ return( true );
 
 /* Handle events from the file list list */
 static int GFileChooserFListSelected(GGadget *gl,GEvent *e) {
-    GFileChooser *gfc;
     int i;
     int32 listlen; int len, cnt, dirpos, apos;
     unichar_t *dir, *newdir;
@@ -765,8 +764,8 @@ return( true );
     }
     if ( apos==-1 )
 return(true);
-    gfc = (GFileChooser *) GGadgetGetUserData(gl);
     ti = GGadgetGetListItem(gl,apos);
+    GFileChooser *gfc = (GFileChooser *) GGadgetGetUserData(gl);
     if ( e->u.control.subtype==et_listselected && cnt==1 ) {
 	/* Nope, quite doesn't work. Goal is to remember first filename. But */
 	/*  if user types into the list box we'll (probably) get several diff*/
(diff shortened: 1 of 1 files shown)


skef:
I seem to be the party pooper tonight.

Looking at the filter list it seems to include (in addition to a lot of dot extensions) "TeX Bitmap Fonts `*{pk,gf}`". Does this mean that users can no longer descend into directories with names that end in those strings?

Seems better to have a notation somewhere indicating what formats can be directories (presumably starting with just UFO) and use that.

ctrlcctrlv:
I don't mind adding such a string, but do you mind if I add it to the default GFileChooser constructor?

GDraw was imagined to be useful for other applications so a lot of it is needlessly generic. This genericism led to my patch in line with it. 

If I am not allowed to add it to the default constructor we would need to set this string every time we construct the class. That seems pointless to me given GDraw's unpopularity outside FontForge.

frank-trampe:
I've long noted this phenomenon but thought that it was intentional and at worst benign. In fact, I feel like I've seen it in other software. Does clicking the open button bother you, or is this intended to make life easier for newcomers?

ctrlcctrlv:
I work with a lot of UFO's and it's annoying behavior yes. The UFO directory should never contain another font to open, FontForge treating it as such is anyways wrong.

ctrlcctrlv:
Pinging for comment, @skef?

> I don't mind adding such a string, but do you mind if I add it to the default GFileChooser constructor?

skef:
@ctrlcctrlv I'm not sure exactly what you mean by that proposal. 

Looking through the code, I would do something like:

1. Add a (better named?) `unichar_t *dirwildcard` to `ggadgetP.h:gfilechooser`
2. Add a `void GFileChooserSetDirFilterText(GGadget *g,const unichar_t *dirwildcard)` to `gfilechooser.c` adapted from `GFileChooserSetFilterText`
3. Use `dirwildcard` in your new check **if it's not null**. Don't do the check otherwise. 
4. Call `GFileChooserSetDirFilterText` in `FVOpenFont()` passing in a new filter we maintain that currently just matches `*.ufo`

That way the dialog can handle this issue in other contexts in the future but the amount of change needed is pretty small.

PR #4508, opened 2020-11-19
Switch back erroneously flipped layer palette mnemonics

### LINKED ISSUE #4504: Keyboard shortcut of Select Fore Layer / Back Layer
- [x] Screenshot
圖片
Version: FontForge-2020-03-14-Windows, Fontforge 20th Anniversary Edition for Windows.

The previous version (20100501-ML) used Alt+B for keyboard shortcut for going to Back Layer, and Alt+O for going to Fore Layer. These two versions mis-assigned these keyboard shortcuts. Please reverse them back.

ctrlcctrlv:
Close #4504

### Type of change
- **Bug fix**

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -8060,9 +8060,6 @@ return;
     } else if ( event->u.chr.keysym=='\\' && (event->u.chr.state&ksm_control) ) {
 	/* European keyboards need a funky modifier to get \ */
 	CVDoTransform(cv,cvt_none);
-    } else if ( (event->u.chr.keysym=='F' || event->u.chr.keysym=='B') &&
-            !(event->u.chr.state&(ksm_control|ksm_meta)) ) {
-        CVLSelectLayer(cv, event->u.chr.keysym=='F' ? 1 : 0);
     } else if ( (event->u.chr.state&ksm_control) && (event->u.chr.keysym=='-' || event->u.chr.keysym==0xffad/*XK_KP_Subtract*/) ){
 	    _CVMenuScale(cv, MID_ZoomOut);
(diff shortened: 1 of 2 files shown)


skef:
Original PR was #1103

skef:
I have to say that I don't much like either the "old" or the "new" picture here.

The new picture, which this PR reverses, is bad because the mnemonics are hard-coded and could conflict with assigned hotkeys.

The old picture is bad primarily because the text of the layer widget doesn't bother adding the underscores, meaning the user needs to know magic or dig deep into the documentation (?) to find them. It's also bad because it doesn't offer a hotkey interface.

In any case I'm confused about #4504 because it seems like the "new" code is stripping out `ksm_meta`, which should be Alt- (among other potential keys), so why do the two solutions conflict? 

The ideal solution here would be to add a "CharView.Layer" section to the hotkey system where "CharView.Layer.Fore" identifies the foreground, "CharView.Layer.Blah" identifies the Blah layer, and so forth. Then we could add Shift-F and Shift-B to the `default` file and these strokes would be retained but not interfere with the Alt assignments. (With that option it seems fine to let the underscores slide.) However, that would be a bit of work. 

I'm not going to approve because I think there should be some discussion of these issues before this gets merged.

ctrlcctrlv:
That's fine, no harm done. Often I open PRs to start discussion, as you know

frank-trampe:
I'm sort of inclined to bear those ills we have pending a complete fix.

skef:
@frank-trampe If we're bearing ills it's not clear to me that the later changes that this PR reverses don't win out. How is someone supposed to know about Alt-o?

PR #4491, opened 2020-11-10
Expand Stroke join fixes

skef:
Properly select contours to join to in SplineSetJoin
Add option to SplineSetJoin (in underscore version) to choose whether to close individual contours
Express join intention in Expand Stroke code.

closes #4452 
closes #4315

Issue #4494, opened 2020-11-11
Crash when pressing Tab or Right in glyph info dialog


blobject:
Version:
- FontForge 20201107 on void linux

Bug description:
- Upon opening the "Glyph Info" dialog and pressing the "Tab" key or the "Right" (directional right) key, FontForge crashes with a segmentation fault.
- Saw the same crashing behavior on several different `.sfd` and `.ttf` files, so it seems to be a general GUI issue rather than tied to particular glyphs or fonts.

Reproduce:
1. Start FontForge
1. Open up an `.sfd` or a `ttf` file
1. Select any glyph, then open the "Glyph Info" dialog (either by pressing `ctrl-i` or clicking through the menu, `Element` --> `Glyph Info...`)
1. Press the "Tab" key or the "Right" key

Further remarks:
- Perhaps similar to issue 3676, but unlike that issue, crashing takes place when pressing a key, not when the dialog opens.
- Once the dialog has opened, if the mouse clicks a gui element like a text entry line, then pressing "Tab" or "Right" does not cause a crash.
- The behavior is not seen (ie. does not crash) with the "Font Info" dialog.

skef:
@jtanx it appears that `GMatrixEdit_StartSubGadgets` is using `r` in offset calculations even when its -1, which seems like a reason for suspicion. (This function also defines a different `GRect r` in a subblock, which is presumably fine at the code level if not the style level.)

jtanx:
Yeah seems like it. More than one way to fix it I guess, but this seems to work


```diff
diff --git a/gdraw/gmatrixedit.c b/gdraw/gmatrixedit.c
index d445e4c56..9d3e0ce6e 100644
--- a/gdraw/gmatrixedit.c
+++ b/gdraw/gmatrixedit.c
@@ -1392,7 +1392,7 @@ static void GMatrixEdit_StartSubGadgets(GMatrixEdit *gme,int r, int c,GEvent *ev
        gme->wasnew = true;
     }

-    if ( c==gme->cols || r>=gme->rows || gme->col_data[c].disabled )
+    if ( c<0 || r<0 || c==gme->cols || r>=gme->rows || gme->col_data[c].disabled )
 return;
     oldr = gme->active_row;
     gme->active_col = c; gme->active_row = r;

```

skef:
This was fixed by #4509 .

PR #4509, opened 2020-11-21
Fix for #4494 Crash when pressing Tab or Right in glyph info dialog

skef:
Closes #4491 

This code was suggested by @jtanx and I've tested it to confirm it resolves the crash. It doesn't leave the dialog in a perfect accelerator state (tab works to advance the left menu position if you fiddle with it but not to start with). However, we have a lot of that sort of thing across the code base and I don't think its worth having individual open issues for them.


jtanx (approved):
👍

frank-trampe:
The Travis Macintosh build is stalling, but it seems unrelated to this change.

PR #4512, opened 2020-11-22
Switch to GitHub Actions for CI

### LINKED ISSUE #4476: Switch CI to Github actions
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

If we switch to GH actions, can probably can do all three (win/mac/linux) builds at once, plus have build artifacts on GH actions (from time to time I have to manually clear out bintray as there's a 10gb limit, with GH actions this instead becomes a 90 day period from last build).

### LINKED ISSUE #4487: Bundled Python 3.7.9 is out of date
The AppImage associated with [release 20201107 in GitHub](https://github.com/fontforge/fontforge/releases/tag/20201107) uses Python 3.7.9. It would be convenient if the bundled Python were kept up to date. Compare the macOS disk image, which bundles Python 3.9.0.

jtanx:
Probably going to be a couple more changes required after this as some things can only be tested on master.

This does all three targets (linux, mac, windows) via github actions. The Mac build is now done with macOS Catalina (10.15) which bumps the minimum version from High Sierra (10.13), but High Sierra was already dropped from Homebrew (hence recent failures/long build times) so such is life.

Appveyor build kept for now but will make it not mandatory. GitHub actions also allows for storing artifacts on builds, but it's not quite so nice, since it always zips the result. I still also push to bintray for now because why not... If the artifact retrieval mechanism on Github actions was nicer I'd be more inclined to drop it.

Coverity build probably doesn't work but I'll fix that later if needed.

### Type of change

- **Bug fix**
- **New feature**
- **Non-breaking change**

Closes #4476
Closes #4487


jtanx:
If no objections I'll merge tomorrow.

frank-trampe:
Fine with me.

Issue #3950, opened 2019-09-23
Remove Overlap cases differing from Preview


skef:
The stroke/offset code I've been developing occasionally generates output that produces unexpected results when fed through Remove Overlap. By "unexpected" I primarily mean unexpected *by me*, but for the purposes of this issue "unexpected" can more objectively mean output with an appearance different in Preview mode. 

I add all these caveats because I've run into these cases as I've been working on other problems and have not thought much about this. At a glance it doesn't look like Remove Overlap handles these correctly (especially given the open contours), but I'm not confident. Perhaps both RO and Preview are wrong.

I'm filing this now in the hope that someone else might take a look while I'm finishing the other parts of this new system. (Anyway, because the offset project is becoming a bit more of a reality its problems are also becoming more of a reality.)

The "theme" with these geometries, to the extent there is a theme, is contour segments that exactly double-back on themselves. Here are four representative samples, or more likely two representative samples that happen to differ slightly in other details: 

[roprobs.sfd.txt](https://github.com/fontforge/fontforge/files/3640691/roprobs.sfd.txt)

PR #4521, opened 2020-12-04
Track and fix unpaired deduplication in overlap removal.

frank-trampe:
I cannot find the original issue. I have attached the sample file demonstrating the problem. The file includes segments that are replicated in odd quantities greater than one. The old cancellation logic ignored parity and just removed all but one, which produced bad results in those cases.

[remove.overlap.sfd.gz](https://github.com/fontforge/fontforge/files/5639816/remove.overlap.sfd.gz)

### DIFF
--- a/fontforge/splineoverlap.c
+++ b/fontforge/splineoverlap.c
@@ -2297,7 +2297,7 @@ static Intersection *FindIntersections(Monotonic **ms, enum overlap_type ot) {
 	continue;		/* Can't intersect since they don't have overlapping bounding boxes */
 	    // ValidateMonotonic(m1); ValidateMonotonic(m2);
 	    if ( CoincidentIntersect(m1,m2,pts,t1s,t2s) ) {
-		// If the splines are nearly coincident , we add up to 4 preintersections with the close flag.
+		// If the splines are nearly coincident, we add up to 4 preintersections with the close flag.
 		for ( i=0; i<4 && t1s[i]!=-1; ++i ) {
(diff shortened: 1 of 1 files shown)


skef:
Based on the nature of the fix my first guess would be #3950

frank-trampe:
It might solve that one too, but there was definitely a guy named Yoshi who posted the sample I attached.

frank-trampe:
This solves problems with all but the first glyph in the sample from #3950. The first one presents a rather unusual case of non-idempotent behavior by the overlap removal code. It removes some overlap on its first execution (with no errors) and the remainder of the overlap on the second execution (with a few errors reported).

frank-trampe:
@skef, the actual code change is only one line, and the logic is pretty straightforward.

skef:
@frank-trampe I did look and do plan to review this but I want to do a little bit of research on the variable added to the line. Promise I'll do it in the next day or two.

skef (approved):
OK, I can't claim to understand this in detail but I take it that isunneeded on at least one of the splines is being cleared later and this prevents it from being incorrectly set again. Approved.

frank-trampe:
It's much simpler than that. The problem was entirely in that code block. If there were three monotonics between A and B, the old code would cancel out monotonics 1 and 2 together and then cancel monotonics 1 and 3 together, which results in zero survivors instead of one. The new logic ensures that each monotonic cancels only one other monotonic.

Issue #3085, opened 2017-06-09
enhance scripting interface to manipulate the forward/backward encoding vector.


HinTak:
It is a bit daft, really, even the documentation says that, if more than one code point encodes to the same glyph, querying the glyph for encoding with `GlyphInfo("Encoding")` (or the python interface) would randomly return either of them. This cause problems when re-encoding, as all except one of them is lost.

This is the 2nd problem with contributing to https://github.com/fontforge/fontforge/issues/3080 .
 
fontforge needs to provide direct scripting access to the forward and backward encoding vector.

You need that to manipulating the Adobe Source CJK / Google Noto CJK fonts properly. They have about 400 glyphs having multiple encoding. I can work around  https://github.com/fontforge/fontforge/issues/3080 with an external script (either parsing `ftdump -V` with perl or using `freetype-py` with python) but I'd prefer not to.

HinTak:
to workaround https://github.com/fontforge/fontforge/issues/3080 , I'll just use freetype-py , assuming eventually fontforge's own scripting interface will provide equivalent functionality.

HinTak:
It would be nice if `MultipleEncodingsToReferences()`  does the right thing - it currently does not.

HinTak:
My freetype-py script to fix fontforge's encoding problem is up at

https://github.com/HinTak/freetype-py/blob/fontval-diag/examples/subfonts-script-generate.py

The 20 lines in freetype-py should be possible with fontforge's scripting interface.

jtanx:
@HinTak thanks for reporting, but please refrain from cross-posting the same message across multiple issues. If they are all linked, it should be sufficient to link back to a single issue (this one?) and continue the discussion there.

HinTak:
Well, there seems to be a number of people and bug reports saying that fontforge cannot process Source Sans properly. Most of them are a bit unclear about what they think is wrong. In my case, I found out what's wrong, and offered my workaround. It works for me. It is really up to these other people to decide whether it fixes their problem too, or offers some insights to where their problems is.

I am okay if these other people want to continue here, but it is their decision. For my part, all I can offer is - this is what I think the problem is, this is my answer. Have a look and see if it works for you too, or can be adapted to work in your case. I don't decide that they should continue here, they do.

mskala:
> Well, there seems to be a number of people and bug reports

All the more reason not to file the same report ten times with cut and paste.

HinTak:
[...]

PR #4522, opened 2020-12-04
[Python] Add function: font.isWorthOutputting()

HinTak:
This is mostly the same as 
https://github.com/fontforge/fontforge/pull/3109 , updated for bitrot to work for current master, to address some of the issues raised in #3107


tests/test1016.py
```diff
@@ -0,0 +1,54 @@
...
+try:
+    font[256]
+except TypeError:
+    # expect this!
+    pass
+else:
+    raise ValueError("Unexpected Error")
+
+font.selection.select(256)
+if ( font.isWorthOutputting() != False ):
```
ctrlcctrlv, line 27: Rewrite as:
```python3
if font.isWorthOutputting():
    ...
```

tests/test1016.py
```diff
@@ -0,0 +1,54 @@
...
+    raise ValueError("Wrong return from fontforge.fonts after selecting unused slot")
+
+font.selection.all()
+if ( font.isWorthOutputting() != True ):
```
ctrlcctrlv, line 31: And this as:
```python3
if !font.isWorthOutputting():
    ...
```

, etc. for all of these.

tests/test1016.py
```diff
@@ -0,0 +1,54 @@
+#Needs: fonts/feta20.pfb
+
+# feta is a font for music symbols, so reencode
+# to ucs4 means no encoding slots are used; but
+# you can still select by the symbol names.
```
ctrlcctrlv, line 5: The font itself being used for these tests doesn't matter, right? So let's just say that these are tests for `isWorthOutputting` and we're using a music symbol font. This information isn't really necessary to understand the tests.

tests/CMakeLists.txt
```diff
@@ -118,6 +118,7 @@ if(ENABLE_PYTHON_SCRIPTING_RESULT)
   add_py_test(test1013.py "NotoSerifTibetan-Regular.ttf" "Read glyph names from unusual post table")
   add_py_test(test1014.py "PfEd layers round tripping")
   add_py_test(test1015.py "Caliban.sfd" "Reverse chaining tables")
+  add_py_test(test1016.py "feta20.pfb" "selection-based isWorthOutputting method")
```
ctrlcctrlv, line 121: I would write as `Tests for font.isWorthOutputting`

doc/sphinx/scripting/python/fontforge.rst
```diff
@@ -4910,6 +4910,13 @@ See the :class:`selection` type for how to alter the selection.
    Leaves only areas in the intersection of contours in all selected glyphs.
    See also :meth:`font.removeOverlap()`.
 
+.. method:: font.isWorthOutputting()
+
+   Returns whether the current selection is worth outputting into a font file.
+   Basically a glyph is worth outputting if it contains any contours, or
+   references or has had its width set.
+   See also :meth:`glyph.isWorthOutputting()`.
```
ctrlcctrlv, line 4918: This is good, but what happens if the font is just opened and nothing is selected? It seems wrong that `font.isWorthOutputting()` would consider 0 glyphs instead of all the ones in the font by default, and that's not done for any other function.

`font.isWorthOutputting` should only consider the selection if one is set. No one will ever want to know if 0 glyphs are worth outputting.

ctrlcctrlv (changes requested):

ctrlcctrlv:
Overall I don't oppose this @HinTak, although another thing I'm wondering about now is why you didn't also add `glyph.isWorthOutputting`?

khaledhosny:
There is already [`glyph.isWorthOutputting()`](https://fontforge.org/docs/scripting/python/fontforge.html#fontforge.glyph.isWorthOutputting) and it has been there since forever, not sure why a font-level one is needed at all.

ctrlcctrlv:
Thanks @khaledhosny — I'm not sure why I didn't see it 14 hours ago. I did <kbd>Ctrl+F</kbd> through the documentation. Perhaps I accidentally had «Whole Words» checked, I inadvertently do that often. :man_facepalming:

tests/test1016.py
HinTak: Done

tests/test1016.py
HinTak: Done

doc/sphinx/scripting/python/fontforge.rst
HinTak: It has been a while but I think this is the case - `select.none()` causes `isWorthOutputting()` to return false (it is part of the test case).

tests/CMakeLists.txt
HinTak: There is already a glyph-based method as @khaledhosny noted.

ctrlcctrlv: …so? Unrelated.

doc/sphinx/scripting/python/fontforge.rst
ctrlcctrlv: Right, that's the issue.

HinTak: The behavior you requested is already implemented, and as tested/verified around line 34 of the test case on `select.none()`.

tests/CMakeLists.txt
HinTak: "Selection-based" vs the existing "glyph-based". I think the current wording is appropriate.

tests/test1016.py
HinTak: A symbol font means you can't select by unicode value. The information is relevant to understand what the test does and does not do (unless you are asking to rewrite the test with a text font).

ctrlcctrlv:
@HinTak keeps marking my review comments as resolved despite the fact that they aren't.

So I am not going to keep commenting using that system. 

Generally we mark resolved only if we make the change. We allow the commenter to mark resolved if we reply without making a change. This isn't a rule, just etiquette. 

I agree with Khaled. This function isn't even really needed. My original comments were made before I knew the glyph level function exists due to browser search issue. One could just do:

```python3
any([g.isWorthOutputting() for g in font.glyphs()])
```

HinTak:
From my point of view, "resolved" meant "addressed". Sometimes, "addressed" means "I don't see any problem with that; no change required". Is that too difficult to understand? The pull is an update for a bit-rotten pull from 3 years ago - to be honest, I don't remember all the details.

Bit-rotten means no-longer merge cleanly; partly it is move to cmake, partly it is that test case is numbered sequentially. When I wrote the test case, it was test case 10; an unrelated test case 10 got merged, so it could not be. So it is now test case 16.

tests/CMakeLists.txt
frank-trampe: I guess that's okay. But the capitalization is definitely inconsistent with that of the other test descriptions.

yg8ijvjvjv:
Font is not worth outputting if it is a TrueType font and has no .notdef or more glyphs than 16-bit can allocate.

yg8ijvjvjv:
Font is also not worth outputting if it is a TrueType font and has mixed references with outlines or out of bounds reference transformation or if FontForge has not enough RAM to output.

HinTak:
@yg8ijvjvjv please read #3107 and #3085 before you make any further comments. Thanks.

ctrlcctrlv:
Any idea why your test case fails on Linux, etc? https://github.com/fontforge/fontforge/pull/4522/checks?check_run_id=1645240966

frank-trampe:
I was going to try to restart the checks, but I can't figure out how to do it.

jtanx:
I think the failure reason is pretty clear, this is not how you write conditionals in Python:

```py
if ( !font.isWorthOutputting() ):
```

HinTak:
> I think the failure reason is pretty clear, this is not how you write conditionals in Python:
> 
> ```python
> if ( !font.isWorthOutputting() ):
> ```

That change was made in a rush, as a response to @ctrlcctrlv's request. Yes, it should be "not ...", I guess.

frank-trampe:
@HinTak, I'm trying to get pull requests merged or closed. Do you still need this functionality? What is the use case?

HinTak:
@jtanx what do you close this for?

jtanx:
No answer to the question above for over two months + questionable usefulness considering existing functionality.

HinTak:
The original pull https://github.com/fontforge/fontforge/pull/3109 was 4 years ago. So it is very slow going, and quite intentionally - @jtanx please leave it open.

HinTak:
@jtanx don't you have better things to do than messing with github on a Sunday?

jtanx:
Wow, no need to take things so personally.

Can I please just get an answer to the question posed?

HinTak:
@jtanx as I said, don't you have better things to do than messing with github on a Sunday? I do.

jtanx:
Indeed, I have better things to do than reviewing 4 year old PRs with no reasonable explanation provided, so I guess we'll keep this closed then.

HinTak:
@jtanx please leave the issue open.

HinTak:
@jtanx nobody asked you to work on Sunday so please don't. Leave this open.

skef:
@HinTak Beyond the occasional modest grant everyone who works on FontForge is a volunteer. I understand how this particular action annoyed you but the particular times at which people choose to volunteer are not your business.

Just answer the question, or not, as you can on your own schedule and the same thing can happen as if it were being answered today. A few days or a week of delay won't be held for or against the viability of the PR.

Issue #4529, opened 2020-12-13
OpenType `name` table: "Mac" platform encoded strings no longer required


ctrlcctrlv:
@kenlunde, a subject matter expert who is an endless font of knowledge¹ on the OpenType standard, PostScript, and Adobe's CID font stuff, noticed when going through my TT2020 font's `name` table that I had non-Unicode strings for the "Macintosh" platform ID (1, 0, 0). This is a legacy format and OS X does not require it; we can understand therefore "Macintosh" there quite literally as referring to the era when people actually called Macs Macintoshes :wink:

As people care quite a bit about keeping fonts as small as possible for bandwidth purposes (which is why @dscorbett removed FFTM in #4320), we ought to also stop writing these unnecessary Mac strings.

Obviously I can just hack TT2020's `name` table after the fact, (I actually just edited the binaries to make my latest release,) but Lunde has been recommending this since at least 2015 ([_What’s in a ‘name’ (Table)?_](https://blogs.adobe.com/CCJKType/2015/11/whats-in-a-name-table.html), 20151107).

So this is a tracking issue for us to stop writing these strings. Does anyone oppose? It's not so difficult, we just delete `static void AddMacName` from `tottf.c`. We've already even got a flag for it: `ttf_flag_nomacnames`.

If some, such as @frank-trampe, wish to be more conservative, I suppose we can just default `ttf_flag_nomacnames` to on. But these strings really serve no purpose but to bloat the `name` table unnecessarily, and me thinking they were necessary² led me to write some really ugly code recently…:

https://github.com/ctrlcctrlv/TT2020/blob/master/build/version_0.2_patch.py#L16-L22

<sub>
¹ heh<br>
² Reasonable assumption given they're default even with “Apple” unselected in «Generate…», no?
</sub>

khaledhosny:
There is an option to not write mac names, you can propose making it the default though it would be a behavior change so be ready to defend it.

PR #4531, opened 2020-12-13
Stop adding Reserved Font Names (RFNs) to OFL

### LINKED ISSUE #4434: Reserving a font name is probably not a sane default when adding OFL license
http://web.archive.org/web/20160822030019/https://lwn.net/Articles/552178/

I consider Reserved Font Names (RFNs) harmful; Google Fonts (@davelab6) does not allow them, nor should they.

Reserving a font name primarily serves to detract from the work I put into my fonts, as fork authors feel forced to rename them, rather than acknowledge my work.

Furthermore, many RFNs are in practice legally null and void; for example, if I tried to declare an RFN on my KJV1611 font, that would be preposterous. Second, let's say a user opens a font under another FOSS license compatible with the SIL OFL, and decides to relicense their font as SIL OFL; to declare an RFN means that they are telling original author they can't use their own font name! (Obviously legally null and void.)

So, I think the GUI button "Add OFL" should not add an RFN by default. Perhaps a dialog should pop up asking, which explains a little bit about benefits and drawbacks of RFNs.

Comments?

Edit: Of course, we should not change people's saved fonts. Only if they click "Add OFL" should it change. By adding a dialog, we can inform old users that the effect of the button has changed.

ctrlcctrlv:
FontForge has no way to figure out whether it's making a legally possible request when adding an RFN, and a minority of OFL license users even use RFN's in the first place. 

**Important to note: This will not change anyone's fonts, only new fonts!** It's only triggered when "Add OFL" is pressed, as in, it's a convenience feature where we try to do the sane thing for most font authors who want to free their font.

Close #4434

### DIFF
--- a/fontforge/ofl.c
+++ b/fontforge/ofl.c
@@ -34,9 +34,6 @@
 
 static char *ofl_eng[] = {
 /* sil.org license by Authors Nicolas Spalinger & Victor Gaultney, 2007feb26 */
-    "Copyright (c) %1$d, %2$s (<URL|email>),",	/* <Date> <Copyright Holder> */
-    "with Reserved Font Name %s.",		/* <Font Family Name> */
-    "",
     "This Font Software is licensed under the SIL Open Font License, Version 1.1.",
     "This license is copied below, and is also available with a FAQ at:",
     "http://scripts.sil.org/OFL",
--- a/fontforgeexe/fontinfo.c
+++ b/fontforgeexe/fontinfo.c
@@ -3266,10 +3266,7 @@ static int GFI_AddOFL(GGadget *g, GEvent *e) {
 	int rows;
 	struct matrix_data *tns, *newtns;
 	int i,j,k,l,m, extras, len;
-	char *all, *pt, **data;
-	char buffer[1024], *bpt;
-	const char *author = GetAuthor();
-	char *reservedname, *fallback;
+	char *all, *pt, **data, *bpt;
 	time_t now;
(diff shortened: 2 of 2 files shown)


fontforge/ofl.c
```diff
@@ -34,9 +34,6 @@
 
 static char *ofl_eng[] = {
 /* sil.org license by Authors Nicolas Spalinger & Victor Gaultney, 2007feb26 */
-    "Copyright (c) %1$d, %2$s (<URL|email>),",	/* <Date> <Copyright Holder> */
-    "with Reserved Font Name %s.",		/* <Font Family Name> */
-    "",
     "This Font Software is licensed under the SIL Open Font License, Version 1.1.",
```
ctrlcctrlv, line 37: I fixed this while I was here. This was another issue noted by @kenlunde (_cf._ #4529). We're already writing the dedicated OTF `name` table string for the copyright string, so this is needless duplication which bloats files.

frank-trampe:
I'm not qualified to provide a second set of eyes here. @davelab6, please advise.

ctrlcctrlv:
@frank-trampe I based my changes on Dave's comment here: https://github.com/fontforge/fontforge/issues/4434#issuecomment-687647591. Unless I missed something I followed them to the letter.

frank-trampe:
Ah, okay. Let's merge this.

frank-trampe (approved):

PR #4533, opened 2020-12-13
Make «Glyph Info…→Unicode→Glyph Name» mnemonic `y`

### LINKED ISSUE #4532: «Element→Glyph Info…→Unicode→Glyph Name» mnemonic doesn't work
It's mismatched.

If you browse to «Element→Glyph Info…→Unicode», you'll see «Glyph Name» has got the wrong character highlighted, <ins>y</ins>. But the code says that its mnemonic is actually <ins>N</ins>, which in any event can't be, as that collides with «<ins>N</ins>ext >». 

End result being, a mnemonic broken in two ways. Very simple patch incoming. Discovered while thinking about https://github.com/fontforge/fontforge/pull/4498#discussion_r541925438

ctrlcctrlv:
`N` is already taken by `Next >`.

Also, `y` is what we're telling users works, when it doesn't.

Close #4532

### Type of change

- **Bug fix**

### DIFF
--- a/fontforgeexe/charinfo.c
+++ b/fontforgeexe/charinfo.c
@@ -4388,13 +4388,13 @@ return;
 	ugcd[0].gd.label = &ulabel[0];
 	ugcd[0].gd.pos.x = 5; ugcd[0].gd.pos.y = 5+4; 
 	ugcd[0].gd.flags = gg_enabled|gg_visible;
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
This is such a small change, and is so obviously correct, that I'm just going to merge it given we're short staffed

Issue #4524, opened 2020-12-08
Endless corrupted file, crash on opening


yg8ijvjvjv:
Permacrash of file. Pasted bitmap references. Crash of FontForge

This file download is limited so be sure to test it out before it expires.
https://ufile.io/1q9cmx14

yg8ijvjvjv:
Opening progress quickly fills up to half the progress which is followed by a crash.

jtanx:
```
==============================================================
==5294==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61d000053d08                                                                                         at pc 0x7f040795b2bc bp 0x7ffe77c7cd00 sp 0x7ffe77c7ccf0
READ of size 8 at 0x61d000053d08 thread T0
    #0 0x7f040795b2bb in SFDFixupBitmapRefs ../fontforge/sfd.c:6179
    #1 0x7f040795c14c in SFDGetBitmapFont ../fontforge/sfd.c:6282
    #2 0x7f04079780f1 in SFD_GetFont ../fontforge/sfd.c:9005
    #3 0x7f0407978e1e in SFD_Read ../fontforge/sfd.c:9088
    #4 0x7f0407979385 in _SFDRead ../fontforge/sfd.c:9121
    #5 0x7f04079e984a in _ReadSplineFont ../fontforge/splinefont.c:1179
    #6 0x7f04079eb17a in ReadSplineFont ../fontforge/splinefont.c:1322
    #7 0x7f04079eb62b in LoadSplineFont ../fontforge/splinefont.c:1380
    #8 0x7f04074f5f9b in ViewPostScriptFont ../fontforge/fontviewbase.c:1348
    #9 0x5557f18630e2 in fontforge_main ../fontforgeexe/startui.c:1401
    #10 0x5557f11cd3f9 in main ../fontforgeexe/main.c:33
    #11 0x7f0405249bf6 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2                                                                                        1bf6)
    #12 0x5557f11cd2f9 in _start (/home/jeremy/ffstaging/build/bin/fontforge+0x1                                                                                        5c2f9)

Address 0x61d000053d08 is a wild pointer.
SUMMARY: AddressSanitizer: heap-buffer-overflow ../fontforge/sfd.c:6179 in SFDFi                                                                                        xupBitmapRefs
Shadow bytes around the buggy address:
[...]

PR #4535, opened 2020-12-14
sfd.c: perform better validation of some inputs

jtanx:
Fixes crash noted in #4524.

### Type of change

- **Bug fix**


skef (approved):
lgtm

yg8ijvjvjv:
Does this also fix the glitching of cross-font references when copying bitmap references?

ctrlcctrlv:
No, just the crash.

yg8ijvjvjv:
But note that crashes are usually caused by glitchy state. Because the glitchy state of cross-font references is not fixes, there are still many crashes left. For instance, closing the font the references are pointing to causes a crash when FontForge tries to read the references by dereferencing their memory address and the only way to fix it is to fix bitmap reference clipboard.

PR #4549, opened 2020-12-29
striptcc: checksumming with filecheck() requires read+write access.

NSGod:
The `stripttc` program takes a TrueType Collection font (.ttc) and splits it into separate .ttf files. Part of that process includes setting the proper `checkSumAdjustment` value in the `'head'` table, which involves checksumming the entire font. As such, the .ttf file should be opened with both read and write access to be able to complete the checksumming process.

### Type of change

- **Bug fix** not sure if this applies as there's no existing bug filed regarding this issue.
- **Non-breaking change**


ctrlcctrlv:
Thanks for your contribution, I agree with your analysis.

PR #4553, opened 2021-01-02
Fix CVCurEnc

ivo-s:
### Type of change

- **Bug fix**
### Bug description
Open the first or the last glyph in a font. Move to other glyphs (Next/Prev). The other Next/Prev option in the menu will remain disabled, even though you are not on the first/last glyph anymore.
### Description
Upon glyph opening, FontForge treats it as a single-glyph view. When switching to other glyphs, square brackets are added to the word list, e.g. `[a]`. From now on, the `_CVMenuChangeChar()` function treats the view as a multi-glyph view and the `cv->enc` variable is not updated (probably because `CharView` now, in theory, displays multiple glyphs, and a single index would not be a valid descriptor). At the same time, some code relies on `CVCurEnc()` to get the current glyph index, which normally uses the `cv->enc` variable. This also happens when checking the View menu for first/last glyphs in `cv_vwlistcheck()`. I commented out this part, forcing `CVCurEnc()` to always look up the index of the current `SplineChar`.

### DIFF
--- a/fontforgeexe/charview.c
+++ b/fontforgeexe/charview.c
@@ -3469,9 +3469,6 @@ return( icon );
 
 static int CVCurEnc(CharView *cv)
 {
-    if ( cv->map_of_enc == ((FontView *) (cv->b.fv))->b.map && cv->enc!=-1 )
-        return( cv->enc );
-
     return( ((FontView *) (cv->b.fv))->b.map->backmap[cv->b.sc->orig_pos] );
 }
 


fontforgeexe/charview.c
```diff
@@ -3469,8 +3469,13 @@ return( icon );
...
 {
-    if ( cv->map_of_enc == ((FontView *) (cv->b.fv))->b.map && cv->enc!=-1 )
-        return( cv->enc );
+/* This is because cv->enc is not updated when FontForge thinks multiple
+ * glyphs are open, which is almost always. See the condition in
+ * _CVMenuChangeChar (testing GGadgetGetTitle8), which triggers whenever
+ * squared brackets are displayed in the word list ("[x]").
+ */
+//    if ( cv->map_of_enc == ((FontView *) (cv->b.fv))->b.map && cv->enc!=-1 )
+//        return( cv->enc );
```
ctrlcctrlv, line 3478: I'm not sure why we need comments for code you _disabled_ as if it would in some other context still be valid to re-enable…certainly the commit/PR notes would suffice for this?

ctrlcctrlv (changes requested):

fontforgeexe/charview.c
ivo-s: I guess I was paranoid, sorry.

ivo-s:
Hey @ctrlcctrlv I just checked back and noticed that you highlighted all of the comments, so now it should be all right.

ctrlcctrlv:
Thanks for the poke. This looks good to me now

PR #4561, opened 2021-01-05
Update the language system tag list to OT 1.8.4

dscorbett:
I updated the list of language systems to match [the list](https://docs.microsoft.com/en-us/typography/opentype/spec/languagetags) in OpenType 1.8.4.

### Type of change
- **Non-breaking change**


skef (approved):
You seem to have omitted a few of the parentheticals compared with the Microsoft-hosted list but that seems to be a matter of judgment. Approved.

PR #4567, opened 2021-01-10
ApproximateSplineFromPointsSlopesAccurate

### LINKED ISSUE #4552: Make Merge/splinefit/ApproximateSplineFromPoints* more accurate
This is a improvement request. When using "Merge" in FontForge, the resulting merged curve is quite inaccurate in many cases.
The following image shall illustrate this inaccurateness:
merge0
If several segments are to be merged, FontForge's algorithm (right) is even worse in comparison with my algorithm (left): (only the upper right segments have been merged)
merge2

I think splinefit.c and the ApproximateSplineFromPoint* methods are responsible for "Merge". 

The main problem of FontForge's algorithm IMHO is the assumption that the point at time t of the approximation contour is near the point at time t of the original contour. But because FontForge's contour are not parameterized by arc length, this assumption is bad. 

If you want you may use my algorithm: I have varied the handle lengths until the error of the approximation contour has become minimal. The error is calculated as follows:
- calculate few (9) points inbetween the start and the end point of the original contour
- calculate more (99) points inbetween the start and the end point of the approximation contour
- for each of the original 9 points find the smallest squared distance to the 99 points; the sum of the squared distances is the error

I have made a quick Python plugin to test the algorithm. It seems to work okay (only the User Interface function is bad).

Example font and Python plugin in [mergebezier.zip](https://github.com/fontforge/fontforge/files/5760261/mergebezier.zip)
(I have updated the plugin because I had implemented a small part which made the error calculation faster but worse in some situations. Now it really should follow the description that is already given here.)

linusromer:
Added an additional function ApproximateSplineFromPointsSlopesAccurate() in splinefit.c which has a boolean at the end. When the boolean is false, it behaves as before but if the boolean is true, it supplies a more accurate (but slower) approximation. The "Merge" function which lives inside splineutil2.c now uses the more accurate version. "Expand stroke" and "Simplify" do still use the original function.

Closes #4552


fontforge/splinefit.c
```diff
@@ -731,248 +731,315 @@ return( SplineMake3(from,to));
...
+						from->nextcp = nextcp;
+						to->prevcp = prevcp;
+					}
+				}
+			}
+		}
+return( SplineMake3(from,to));	
+	}
+	else { /* original and fast function */
+		pt_pf_x = to->me.x - from->me.x;
```
skef, line 797: In the past some project members have supported not changing the indentation of large blocks in cases like this so as not to change `git blame` ownership (among other reasons). I'm agnostic about shorter blocks but I think for something this large it would be better not to change the lines. 

The lines around the block can be left as-is (as far as I am concerned).

fontforge/splinefit.c
```diff
@@ -731,248 +731,315 @@ return( SplineMake3(from,to));
 	from->nextcp = from->me; to->prevcp = to->me;
 return( SplineMake3(from,to));
     }
+    /* This is the generic case, where a generic part is approximated by a cubic */
+    /* bezier spline. */
+	if ( is_accurate ) { /* More accurate but slower function by Linus Romer */
```
skef, line 736: I think I *basically* understand this, and would characterize it roughly as "brute force based on limits already calculated for the other algorithm".

One thing I don't understand is that I would expect the "fine-tuning" run to search in a narrower range of measures than the "narrowing" run. However, I'm having a little trouble understanding the motivation for what seems to amount to `ftlen*fromunit.x*((31*fromhandle)/30.0)/60.0` in run 0. Is the idea more or less that this is close enough to `ftlen*fromunit.x*fromhandle/60` that it's not worth two separate statements or is there something more subtle going on?

linusromer: Well, "brute force based on limits already calculated for the other algorithm" is fine for me.

I do not understand how you got `ftlen*fromunit.x*((31*fromhandle)/30.0)/60.0`. If `run==0` the part should read as `ftlen*fromunit.x*( fromhandle )/60.0`.

While I see that this part is not well readable, it is much more compact than a more readable version. I will try to explain it for fromhandle:
run 0:
fromhandle runs from 1 to 60, the handle is divided in parts from 1/60,2/60,...,60/0. The best is chosen, say 45/60.
run 1:
fromhandle runs from -30 to 30, the handle is divided in parts from 44/60, (45-29/30)/60, ..., 45/60, ... , (45+29/30)/60, 46/60

But I agree, that fromhandle in run 1 should only take values from -29 to 29.

skef: I suspect I was just confused in that I was getting my ternary operator precedence wrong.

There's a school that advocates only adding parentheses when they're needed and a different school according to which it's good to add redundant parentheses for cases less frequently encountered. I suppose you're using spacing to echo the precedence but the FontForge source base does unusual things with spacing so that may do much in context.

skef:
If the big block gets restored to its old indentation I will approve this. I'll leave other stylistic choices to the discretion of the submitter and other reviewers (if any).

Someone might quibble with adding `Accurate` to the function name as, well, not quite accurate: the resulting function is either accurate or not depending on the boolean. The usual FontForge convention of a leading underscore for the enhanced function isn't awesome either, but it's easy and (presumably?) sufficient.

ctrlcctrlv:
@skef, I actually think in this case the spacing might not need to be reverted. 

Clearly this function was entirely changed by the commit, and it is unlikely that any bugs that arise from this would not be blamed on this commit. 

Furthermore, if it is found that the bugs go older than that, I don't think it's that big a deal to go back one commit because of the already mentioned fact that this commit changes entirely the behavior of the function. 

I don't know if that's enough to sway you, but I would be happy to merge this now if it is. It is an obviously useful feature to have a better merge.

ctrlcctrlv:
I would also say that because the entire function was reviewed by @linusromer and significantly changed, that it would be worse to have some of it spaced one way and some of it spaced another. If it was not such a large overhaul to the function I would have a different opinion.

skef:
@ctrlcctrlv I don't agree that the function is "entirely changed" by the commit. There's an added section that corresponds to the new boolean. The only change to the rest of the function (as far as I can tell from verifying a fraction of them) is the indentation. That indentation change obscures that the rest of the function was syntactically unmodified by the commit.

ctrlcctrlv:
Yeah but how could adding extra indentation be avoided? There's a new else block so some kind of new indentation is needed isn't it? And indentation that sticks the way we usually do things would still create changed lines wouldn't it?

skef:
Although it would be difficult to pick out specific PRs at this point, I take myself to be following a review convention I and other contributors have been informed of in similar cases. In general modifying very large blocks of existing code only for indentation has been discouraged. 

Maybe @frank-trampe can weigh in on whether this is still relevant. 

> Yeah but how could adding extra indentation be avoided? There's a new else block so some kind of new indentation is needed isn't it? 

It's C -- you avoid the extra indentation by not adding it. The project certainly doesn't use a style checker. The `/* original and fast function */` comment would make it quite clear what's going on.

ctrlcctrlv:
I don't know Skef. I don't think that we have ever required someone to not add indentation that they otherwise should have.

I agree, @frank-trampe or @jtanx should weigh in. 

I think not adding the indentation would harm the readability of the function and that that matters more than the blame.

skef:
@linusromer GW's indentation style (still used in most of the files) is idiosyncratic and a pain to reproduce. I followed through on your intention in a different way and added a commit to your repository. Make sure you pull it down before editing any further (if you do). 

(If this is surprising, there's a little checkbox at submit time that gives project reviewers write access to the source branch of a PR and it's checked by default. Just a handy github thing.)

Since the latest commit is mine I'll let you sign off on it before approving but the current code looks good to me. (The windows build failure looks unrelated but I'll follow up on that.)

skef:
@jtanx Looks like the windows build is unhappy -- the last few commits on this have failed to build for windows in a way that doesn't seem tied to the source changes.

frank-trampe:
Sorry for my silence on the indentation issue. @skef was on point, and I missed the subsequent ping. It seems to be resolved.

skef (approved):
This looks good to me. Unless someone swoops in I'll merge in a day or two.

skef:
After thinking about this I could hear @jtanx asking "Are there really enough existing calls that we need to add a new signature?" It turns out that there are only 5 so the answer is "no". I added a commit to go back to a single function and change those existing calls so I'll sit on this another day before merging.

Issue #4558, opened 2021-01-04
FontForge makes 'out of bounds' glyph indexes (cmap table generation problem when targeting TTF when ~50k glyphs are in the font)


yg8ijvjvjv:
https://ufile.io/i860xgf6

In Windows 7, the resulting fonts do install but when looking at the character set some of the characters render glitchy incorrect glyphs. Vista has this glitching as well.

In Windows 10, the fonts fail to get recognized, with only the Upper working.

In FontForge, Glyph index out of bounds. Was 55292, must be less than 22528.
  In attempt to associate a glyph with encoding d7fc in segment 0
  with platform=3, specific=1 (in 'cmap')
Glyph index out of bounds. Was 55293, must be less than 22528.
  In attempt to associate a glyph with encoding d7fd in segment 0
  with platform=3, specific=1 (in 'cmap')
Glyph index out of bounds. Was 55294, must be less than 22528.
  In attempt to associate a glyph with encoding d7fe in segment 0
  with platform=3, specific=1 (in 'cmap')
Glyph index out of bounds. Was 55295, must be less than 22528.
  In attempt to associate a glyph with encoding d7ff in segment 0
  with platform=3, specific=1 (in 'cmap')

Why does FontForge save make 'out of bounds' glyph indexes in this font?

yg8ijvjvjv:
Someone suggested things like performing Autotrace again to fix the digitization or contacting Unicode to attempt to resolve the Unicode glitching.

yg8ijvjvjv:
This is important, how else am I supposed to export the font?

yg8ijvjvjv:
When is this being fixed? I need it to work properly

yg8ijvjvjv:
Because I need a usable font and when I try to export the usable font the usable font glitches in a usable renderer

frank-trampe:
@yg8ijvjvjv, finding the specific problem here seems likely to take a bit of work. If you need this fixed quickly, I would suggest that you do the footwork yourself and submit a pull request.

ctrlcctrlv:
I took some pity on you Piotr.

So, there's a problem writing the `cmap` table. This seems to be a legitimate bug. I'm not promising to fix it, but here's more information.

At the time of writing, which you for some reason omitted in your message, we get the dreaded `Attempt to output %d into a 16-bit field. It will be truncated and the file may not be useful.` error.

```
(gdb) break GDrawIError
[...]

PR #4585, opened 2021-01-19
(Generally) improved cmap "format 4" heuristics

skef:
Discussion towards the end of #4558. 

The new heuristics don't always save memory, which means there's a chance it will fail to output a table when the old code could. However the worst relative case is the sequence "mapped unmapped unmapped mapped" with four or more unmapped slots on either side, and you would need a whole lot of these to use up the available space (assuming that's possible). 

On the other hand this does very well on the `XZ_Kissinger_13.0.051.sfd` test font from the issue, encoding the whole cmap with just 18 delta-mapped segments (including the terminating segment). 

You'd think I can write code to output arcane tables like this just because I'm "of that age" but I had to stare at the spec for a while. The whole modulo 65536 delta thing is nifty if a bit opaque to the likes of me.


skef:
I'm opening this as a draft because obviously it needs more testing and it will be a while before I can construct a test font with the needed corner cases. There are probably some bugs. 

This way @yg8ijvjvjv might use it in the meantime, which is only fitting in light of his unwavering and selfless support of FontForge and its community.

yg8ijvjvjv:
"this does very well on the XZ_Kissinger_13.0.051.sfd"

So this works well with Kissinger, but does it also work well in Kissinger JP? When FontForge updates I will release an updated version of Kissinger and when I happen to be banned from FontForge, I will modify FontForge to remove the ban screen because it is free software.

skef:
@yg8ijvjvjv you are a strange dude. If you're banned it will be from these forums, not the software. How would that even work with a normal open source license? 

All that aside, are you in a position to compile FontForge for testing? It's polite to be willing to test PRs that correspond to issues one has filed. If not I'll try to dig up the info on the build artifacts from PRs.

skef:
Actually, never mind. I should build the test font for this before anyone bothers with other testing.

yg8ijvjvjv:
"@yg8ijvjvjv you are a strange dude. If you're banned it will be from these forums, not the software. How would that even work with a normal open source license?"

Detect the user identity (for instance by browser logins) and if it matches the banlist immediately jump the code execution to the ban screen. In case of FontForge, while most users won't understand the code, it is possible to look into the history of the commits and selectively revert the ones that add the banlist system. The LGTM, etc. systems will compile automatically when it is added as a pull request.

skef:
Alright, I guess that's how one would do it ...

That's never going to happen in any mainstream open source project.

skef:
I've added a test and removed the Draft status on this. Please review and let me know if there are other cases I should add to the test.

"Bite-sized" my foot.

skef:
I've stared that that LGTM warning multiple times and I can't figure it out. `first_delta` is set to -1 in the "unmapped" branch of the code. I don't see anything that forces it back to a positive value before that guard.

I know that 99% of the time it's you that's wrong rather than the tool but so far I haven't been able to see it.

dscorbett:
If the condition on line 4596 is under consideration, then the condition on line 4585 must have been false, so `first_delta==-1` must have been false, so checking `first_delta!=-1` is redundant.

skef:
Oh, hmm, so it's just a narrow complaint about that part of the guard.  In that case I'm inclined to leave it as is -- the point is to protect the sanity second comparison. I'm happy to let the compiler optimize it out if the other guards are still arranged such that it isn't needed.

frank-trampe:
I don't see any obvious problems, and the LGTM warning seems practically spurious. I say we merge it if there are no objections.

frank-trampe (approved):

skef:
This one scares me a little, although not as much as it would had I not constructed the test font. 

One thing we could do is run a bunch of ttfs and/or otfs through an open/generate cycle and verify that the cmap is the same via ttx (or that if and when they aren't the same it's for other reasons besides the encoding.)

ctrlcctrlv:
"CMAPEncTest" doesn't have any copyright issues, does it? We've since learned that "Kissinger" is copyright infringement.

skef:
It does, probably, although they should be easy to resolve. The glyphs are from Libertinus Mono which carries a SIL 1.1 license. Technically that requires shipping the license text but I don't think even the Libertinus project follows that to the letter. 

How about I add a "glyphs taken from" message to the copyright string and add the License and License URL fields to the TTF names?

ctrlcctrlv:
That should be fine. I was mostly concerned that you borrowed Piotr's font. (I didn't try to open it.)

skef:
I pushed that data into the font. 

As for the font, the potential CMAP table problems involve individual glyphs being mismatched, so I wanted a really obvious sign of the correspondence between glyph and encoding slot. So I wrote a little script to depict hex values using references:

ffCMAP

Not the kind of thing anyone is likely to get up in arms about but good to give proper credit and copyright info.

fontforge/tottf.c
```diff
@@ -4499,34 +4499,42 @@ return( NULL );
...
+#define SEGBREAKEVEN 4
 static FILE *NeedsUCS2Table(SplineFont *sf,int *ucs2len,EncMap *map,int issymbol) {
     /* We always want a format 4 2byte unicode encoding map */
     /* But if it's symbol, only include encodings 0xff20 - 0xffff */
-    uint32 *avail = malloc(65536*sizeof(uint32));
-    int i,j,l;
-    int segcnt, cnt=0, delta, rpos;
-    struct cmapseg { uint16 start, end; uint16 delta; uint16 rangeoff; } *cmapseg;
-    uint16 *ranges;
+    int32 *avail = malloc(65536*sizeof(int32));
```
jtanx, line 4507: Can you explain the sign change?

skef: The previous code stored glyph IDs in `avail` and then used other logic later to go from glyph ID to `ttf_glyph`. The new code stores `ttf_glyph` values in `avail` to start with. Being able to check against -1 is handy but it's mostly to match the type of `ttf_glyph`.

frank-trampe:
Can we merge this?

skef:
> Can we merge this?

Here's what I did: 

I have one of those "massive collections of fonts!" with the "A" directory unpacked (it's too big to leave all of it uncompressed). I use these as a test sample sometimes. So I did a `find` to put all the files ending in `.ttf` or `.otf` into `/tmp/aa`. Then I ran this:

```
#!/usr/bin/env python

import fontforge
from fontTools.ttLib import TTFont

flns = open("/tmp/aa", 'r')
good = open("good.txt", "w")
bad = open("bad.txt", "w")

for l in flns:
    name = l.strip()
    ext = name[-4:None]
    print(ext)
    try:
        f = fontforge.open(name)
    except:
        continue
    name2 = "/tmp/bbb" + ext
    f.generate(name2)
    f.close
    blah1 = TTFont(name)
    blah2 = TTFont(name2)
    cm1 = blah1.get('cmap')
    cm2 = blah2.get('cmap')
    for t1 in cm1.tables:
        if t1.format != 4:
            continue
        t2 = cm2.getcmap(t1.platformID, t1.platEncID)
        if t2 is None or t2.format != 4:
            continue
        c1 = t1.cmap
        c2 = t2.cmap
        # c.pop(97, None)
        if (c1!=c2):
            print(name, file=bad)
            df = { k : (v, c2.get(k)) for (k, v) in c1.items() if v != c2.get(k) }
            df2 = { k : (v, None) for (k, v) in c2.items() if c2.get(k) is None }
            print({ **df, **df2 }, file=bad)
        else:
            print(name, file=good)
    blah1.close()
    blah2.close()

flns.close()
good.close()
bad.close()
``` 

So basically this goes through file pairs looking for matching format 4 cmap tables, where one file is a TrueType or OpenType font from the collection and the other is the (roughly) equivalent file generated with FontForge. Any where the tables don't match are output to `bad.txt` with a representation of what didn't match. 

(Oh, key caveat: the code I did this with also had #4586 merged, so that multiply-encoded glyphs would be recorded properly (?))

So what didn't match? Interestingly, none of the 193 `.otf` files had any mismatch whatsoever. 

There were 969 "comparable" (both files had matching format 4 cmaps) .ttf files of which 175 had differences *of some kind*.  The most common pattern of difference was akin to:

```
{176: ('degree#1', 'degree'), 8211: ('endash#1', 'endash'), 8226: ('bullet#1', 'bullet')}
```

I think this is just FontForge stripping the numerical codes off the names (FF-generated glyph names are second). That's bolstered by a smaller number of entries like these:

```
{160: ('.notdef#2', 'uni00A0'), 166: ('.notdef#10', 'glyph171'), 173: ('.notdef#3', 'uni00AD'), 178: ('.notdef#16', 'uni00B2'), 179: ('.notdef#18', 'uni00B3'), 185: ('.notdef#6', 'uni00B9'), 188: ('.notdef#8', 'glyph160'), 189: ('.notdef#9', 'glyph163'), 190: ('.notdef#19', 'glyph209'), 208: ('.notdef#20', 'glyph210'), 215: ('.notdef#15', 'multiply'), 221: ('.notdef#17', 'glyph203'), 222: ('.notdef#22', 'glyph225'), 240: ('.notdef#14', 'glyph188'), 253: ('.notdef#12', 'glyph174'), 254: ('.notdef#11', 'glyph172'), 305: ('onequarter#1', 'onequarter'), 321: ('.notdef#4', 'glyph142'), 322: ('.notdef#5', 'glyph148'), 352: ('.notdef#13', 'glyph180'), 353: ('.notdef#21', 'glyph222'), 732: ('onequarter', 'glyph130'), 8722: ('.notdef#7', 'glyph153')}
```

The FontForge names for the glyphs match the encoding slots whereas the glyphs in the original file are being given "non-names" for whatever reason. 

Then there are some like this:
```
{916: ('Delta#1', 'Delta'), 8710: ('Delta', 'glyph168')}
```

But again this looks like FontForge sorting out name conflicts for itself ("Delta#1" was first and "claimed" the name "Delta", when "Delta" was reached it was assigned "glyph168" to avoid a conflict). 

And that's basically it. I found one case of an extra entry in a `platEncID=3` table but that seems to be a "difference of opinion" about what belongs there. 

So based on this survey I would say the PR is not likely to cause problem, which is not to say that I'm sure it won't cause any.

Issue #3080, opened 2017-06-07
Reencode() silently mis-encodes type1 fonts with Unused/detached glyphs


HinTak:
* Provide a general summary of the issue in the **Title** above.  
* Before you open an issue, please check if a similar issue already exists or has been closed before.

### Important
Mark with [x] to select. Leave as [ ] to unselect.

### When reporting a bug/issue:
- [ ] Screenshot
- [x] The FontForge version and the operating system you're using

Based on source from git with hash: b8e5ff8f24955f4d7d59ac73c903cc088b21bdb6, fedora linux, x86_64

- [x] The behavior you expect to see, and the actual behavior

For example, for subfont 44, this is what ttf2tfm gives:

```
/a1sh44Encoding [
/index0x1AD1 /index0x1AD2 /index0x1AD3 /index0x1AD4
/index0x1AD5 /index0x1AD6 /index0x1AD7 /index0x1AD8
/index0x1ADA /index0x1ADB /index0x1ADC /index0x1ADD
/index0x1ADE /index0x1ADF /index0x1AE0 /index0x1AE1
/index0x1AE2 /index0x1AE3 /index0x1AE4 /index0x1AE5
/index0x1AE6 /index0x1AE7 /index0x1AE8 /index0x1AE9
/index0x1AEB /index0x1AEC /index0x1AED /index0x1AEE
/index0x1AF0 /index0x1AF1 /index0x1AF2 /index0x1AF3
...
```
This is what fontforge does (it is really glyph id rather than uniXXXX) - whenever fontforge seems to be randomly uses /uni44*, it encodes the wrong glyph. The glyph ids should be uni1AD1, etc. So half of subfont 44 is wrong, in my case.

```
/a1sh44-Enc [
/uni4400 /uni1AD2 /uni4402 /uni1AD4
/uni4404 /uni4405 /uni1AD7 /uni4407
/uni4408 /uni1ADB /uni1ADC /uni440B
/uni1ADE /uni1ADF /uni440E /uni1AE1
/uni1AE2 /uni4411 /uni4412 /uni1AE5
/uni4414 /uni1AE7 /uni1AE8 /uni4417
/uni4418 /uni1AEC /uni441A /uni441B
/uni1AF0 /uni441D /uni441E /uni1AF3
...
```

- [x ] Steps to reproduce the behavior 

```
fontforge -script cjk/utils/subfonts/subfonts.pe SourceHanSerifTC_EL-M/SourceHanSerifTC
[...]

Issue #1831, opened 2014-10-21
Outdated or wrong CIDMaps?


ahyangyi:
I noticed that some glyphs are missing when I use Fontforge to open Noto Sans CJK SC. The fonts can be downloaded at https://www.google.com/get/noto/cjk.html .

An example would be the character '量', which, according to all systems and applications, are present in Noto Sans CJK SC. Also it's a commonly used character in all variants of CJK languages. But If you use fontforge to import the fonts you won't be able to find this character.

I am still using the old CIDMaps from the old fontforge website. Does anyone know of a newer version, or is there any potential fix to this problem?

HinTak:
I think it is a variation of https://github.com/fontforge/fontforge/issues/3085 and https://github.com/fontforge/fontforge/issues/3080 . Noto Sans CJK fonts have about 400 glyphs encoding to two code points. About a dozen to 3 code points, actually.

HinTak:
It would be nice if  `MultipleEncodingsToReferences()` does the right thing - it currently does not.

HinTak:
What's happening is that the Source CJK fonts / Noto CJK fonts have about 400 glyphs having more than 2 coding points; actually about a dozen have 3. All but one is silently dropped by fontforge. So you get about 400 coding points missing when re-encoding.

The worst part of fontforge behaviour is that it uses the last one it sees as authoritative - that's often the CJK Compat variant range, rather than the lower CJK Unified region. So the lower and more often-used CJK Unified code range ended up having about 400 glyphs missing.

HinTak:
fontforge cannot cope with glyphs mapping to multiple encoding points. In summary.

HinTak:
The cidmap is not wrong - it is just that fontforge cannot currently cope. The author(s) of fontforge seems not to be able to comprehend the need for one glyph to multiple coding points properly. Source CJK/Noto CJK have about 400 of those, and it is missing from the lower CJK Unified range, as fontforge prefers to use the upper Compat variant as authoritative encoding for such glyphs.

HinTak:
My freetype-pay script to fix fontforge's encoding problem is up at

https://github.com/HinTak/freetype-py/blob/fontval-diag/examples/subfonts-script-generate.py

yg8ijvjvjv:
It's because the fonts are OpenType OTF, but FontForge is misinterpreting them as CID files.

ahyangyi:
[...]

Issue #1534, opened 2014-07-16
Noto Sans CJK or Source Han Sans cannot be opened properly


moyogo:
Open NotoSansCJK-Regular.otf (https://code.google.com/p/noto/) or SourceHanSans-Regular.otf (http://sourceforge.net/projects/source-han-sans.adobe/files/ in SourceHanSansOTF-1.000.zip).
Only U+FF21..FF5A are displayed with glyphs.

HinTak:
Argh, is this still the case? am going through the issue list to see the mis-re-encoding ( https://github.com/fontforge/fontforge/issues/3080 ) has been seen by somebody else. This looks like it.

HinTak:
It would be nice if  `MultipleEncodingsToReferences()`  does the right thing - it currently does not.

What's happening is that the Source CJK fonts / Noto CJK fonts have about 400 glyphs having more than 2 coding points; actually about a dozen have 3. All but one is silently dropped by fontforge. So you get about 400 coding points missing when re-encoding.

The worst part of fontforge behaviour is that it uses the last one it sees as authoritative - that's often the CJK Compat variant range, rather than the lower CJK Unified region. So the lower and more often-used CJK Unified code range ended up having about 400 glyphs missing.

HinTak:
My freetype-py script to fix fontforge's encoding problem is up at

https://github.com/HinTak/freetype-py/blob/fontval-diag/examples/subfonts-script-generate.py

ipsmile:
Followed the link and installed freetype-py, it still didn't fix the problem. I still see blanks for all glyphs after opening the font file in FontForge. 

Here is the installation outputs which was successful. Should the problem be fixed after freetype-py installed? 

ip@LINUXMINT182 ~ $ git clone https://github.com/rougier/freetype-py.git
Cloning into 'freetype-py'...
remote: Counting objects: 998, done.
remote: Total 998 (delta 0), reused 0 (delta 0), pack-reused 998
Receiving objects: 100% (998/998), 1.12 MiB | 791.00 KiB/s, done.
Resolving deltas: 100% (620/620), done.
Checking connectivity... done.
ip@LINUXMINT182 ~ $ cd freetype-py
ip@LINUXMINT182 ~/freetype-py $ python setup.py install
running install
running bdist_egg
running egg_info
creating freetype_py.egg-info
writing freetype_py.egg-info/PKG-INFO
writing top-level names to freetype_py.egg-info/top_level.txt
[...]

PR #4586, opened 2021-01-21
Record unicode cmap encodings when one glyph is in multiple slots

skef:
The problem addressed by this PR is referenced in multiple bugs, including #1534, #1831, #3085 and possibly #3080. However, the scope of the problem and the complete fix is as yet unclear.

### DIFF
--- a/fontforge/parsettf.c
+++ b/fontforge/parsettf.c
@@ -4537,6 +4537,23 @@ static int PickCMap(struct cmap_encs *cmap_encs,int enccnt,int def) {
 return( ret );
 }
 
+static void addttfencoding(SplineChar *sc, int unc) {
+    struct altuni *alt;
+    if ( unc==-1 )
+	return;
+    // This doesn't check for duplication but it easily could
+    if (sc->unicodeenc!=-1) {
+	sc->unicodeenc = unc;
+    } else {
+	alt = chunkalloc(sizeof(struct altuni));
+	alt->unienc = unc;
+	alt->vs = -1;
+	alt->fid = 0;
+	alt->next = sc->altuni;
+	sc->altuni = alt;
+    }
+}
+
 /* 'cmap' table: readttfcmap */
 static void readttfencodings(FILE *ttf,struct ttfinfo *info, int justinuse) {
     int i,j, def, unicode_cmap, unicode4_cmap, dcnt, dcmap_cnt, dc;
@@ -4738,7 +4755,7 @@ return;
 			if ( map!=NULL )
 			    map->map[i] = table[i];
 			if ( dounicode && trans!=NULL )
-			    info->chars[table[i]]->unicodeenc = trans[i];
+			    addttfencoding(info->chars[table[i]], trans[i]);
 		    }
 		} else if ( table[i]<info->glyph_cnt && info->chars[table[i]]!=NULL )
 		    info->inuse[table[i]] = 1;
@@ -4801,8 +4818,8 @@ return;
 				}
 			    } else {
 				if ( uenc!=-1 && dounicode ) used[uenc] = true;
-				if ( dounicode && info->chars[(uint16_t) (j+delta[i])]->unicodeenc==-1 )
-				    info->chars[(uint16_t) (j+delta[i])]->unicodeenc = uenc;
+				if ( dounicode )
+				    addttfencoding(info->chars[(uint16) (j+delta[i])], uenc);
 			        if ( map!=NULL && lenc<map->enccount )
 				    map->map[lenc] = (uint16_t) (j+delta[i]);
 			    }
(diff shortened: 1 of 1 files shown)


skef:
@HinTak

I know there is a complex and perhaps unpleasant history with this family of issues but I'm hoping you can help me out with taking a look at a potential fix. Keep in mind that I know *next to nothing* about CID mapped fonts so you might have to slowly walk me through steps like verifying problems and/or fixes in the UI. 

Since you've made at least one PR I assume you can compile FontForge. The code for this PR is in this branch: https://github.com/skef/fontforge/tree/bg1831 , which you can look at by cloning `https://github.com/skef/fontforge` and doing a `git checkout bg1831` "inside" the repository. 

Based on preliminary testing (of `Open`/`Generate` pairs) this seems to resolve the noted problem in #1831 and I suspect it gets most or all of the way with #1534. But again I'm very green with this CID stuff.

skef:
@HinTak I should have supplied more context about this fix.

The "altuni" functionality and interface already exists. The main UI is in the "Alternate Unicode Encodings" section of the "Unicode" (first) tab of the "Glyph Info..." dialog. The python interface is described [here](https://fontforge.org/docs/scripting/python/fontforge.html#fontforge.glyph.altuni). (I don't know if there is a native FontForge scripting interface -- at first glance there doesn't seem to be.)

All this PR does is add the code to make use of it on ttf/otf import. I don't know why it wasn't doing so before except that looking in the repository history the code was very early and perhaps GW never got around to addressing the case.

I'm pretty sure there's already some "deduplication" code present, much like there is for   single-encoding cases. You'll have to play with it a bit. Regardless, most of what you've mentioned is already in the code base.

skef:
There's more info about the UI [here](https://fontforge.org/docs/ui/dialogs/charinfo.html). I may not currently be setting all the variables correctly. (It looks like 0 would have been a better choice than -1 for Variation Selector? Or maybe not given that this is a CID font being loaded? It's a little hard to say because it's a bit of one and a bit of the other.).

(Also, to make this a complete response altuni data are already stored in the sfd file.)

HinTak:
Okay... A few years ago when I looked into this I saw the altuni struct and couldn't figure out what it is supposed to do. Maybe it is just unfinished code.

skef:
@HinTak These CID/altuni problems are unlikely to ever be fixed without test cases, or at least detailed guidance, from someone with more familiarity with the (practical) issues and problems. I've already run into at least one gap in the altuni implementation ( #4598 ) but I don't have a good picture for what more, if anything, is missing. 

You seem to be the main interested party so far. If we work together I think it's likely we could make some progress. Otherwise this is likely to sit in the PR queue and eventually get deleted when @ctrlcctrlv does a cleanup pass.

HinTak:
It is relatively simple to test the correctness of the change: the Adobe source cjk fonts are known to be exhibit this kind of behavior, and you should be able to get at the multiple mappings via the new code and compare what my freetype-py script does. I just don't have either the time nor the inclination to build fontforge from source at the moment.

frank-trampe:
@skef, this pull request seems to fix real problems, and I see no significant risk of regression from it. Is there any reason we couldn't merge it now?

skef:
Yeah, I guess the likelihood of this causing problems is low. The risk is not tracking what else is needed but that's always a risk. 

It would be much better if we had a full idea of how the altuni system should work with test cases.

HinTak:
It should be possible to rewrite part of, and compare with the other freetype-py dependent script I wrote without freetype-py .

skef:
I looked at this a bit more. Fixing the altuni stuff more completely probably requires an overhaul of the Encoding code. At the very least it would require going through that stuff with a fine tooth comb. So by the "improves things and doesn't cause obvious harm" standard this could probably go in.

HinTak:
As I replied earlier some months ago, it should be possible to re-write my freetype-py dependent script with the new functionality and remove the freetype-py dependency, and also run the two side by side to verify that the new functionality works correctly (if you provide a python binding to the new functionality).

skef:
@HinTak The current issues have less to do with bottom-level glyph state than with having the GUI update correctly when manipulated via scripting.

fontforge/parsettf.c
```diff
@@ -4558,6 +4558,30 @@ static int PickCMap(struct cmap_encs *cmap_encs,int enccnt,int def) {
...
 }
 
+/* This doesn't worry about putting the most appropriate encoding
+   in sc->unicodeenc because (in theory) NameConsistancyCheck()
+   will swap it in. */
+static void addttfencoding(SplineChar *sc, int unc) {
+    struct altuni *alt;
+    if ( unc==-1 ) {
+	return;
+    } else if (sc->unicodeenc==-1 || sc->unicodeenc==unc) {
```
jtanx, line 4568: Seems like you could move `sc->unicodeenc==unc` to the if statement above

jtanx (approved):

fontforge/parsettf.c
skef: done

jtanx:
Change looks fine but needs to be rebased to fix conflicts with master

skef:
Rebased

Issue #4595, opened 2021-01-29
TTF Version string has trailing spaces


deepestblue:
- FontForge version: 20201107, MacOS 11.1

- Generate TTF file. Run it through fontbakery.

Expected:
- Generated TTF file to pass FontBakery Universal tests.
Actual:
- Fails FontBakeryCheck:com.google.fonts/check/name/trailing_spaces, with an output like so:

```
Name table records must not have trailing spaces.
Check ID: <FontBakeryCheck:com.google.fonts/check/name/trailing_spaces>
🔥 MyFont.ttf
🔥 FAIL
Name table record with key = (1, 0, 0, 5) has trailing spaces that must be removed: 'Version 0.9 '
🔥 FAIL
Name table record with key = (3, 1, 1033, 5) has trailing spaces that must be removed: 'Version 0.9 '
```

PR #4596, opened 2021-01-29
Remove trailing space from Version string in TTF output

deepestblue:
- **Bug fix** for #4595
- Trivial change to not generate trailing spaces in TTF Version strings. [According to Fontbakery](https://font-bakery.readthedocs.io/en/latest/fontbakery/profiles/universal.html#com.google.fonts/check/name/trailing_spaces), trailing spaces are verboten in TTF name records, and this change removes them. Trailing spaces were only being generated for a user-specified string and not for the default `Version 1.0` string, so this change also brings in consistency.

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -3792,9 +3792,9 @@ void DefaultTTFEnglishNames(struct ttflangname *dummy, SplineFont *sf) {
 	dummy->names[ttf_fullname] = utf8_verify_copy(sf->fullname);
     if ( dummy->names[ttf_version]==NULL || *dummy->names[ttf_version]=='\0' ) {
 	if ( sf->subfontcnt!=0 )
-	    sprintf( buffer, "Version %f ", (double)sf->cidversion );
+	    sprintf( buffer, "Version %f", (double)sf->cidversion );
 	else if ( sf->version!=NULL )
-	    sprintf(buffer,"Version %.20s ", sf->version);
+	    sprintf(buffer,"Version %.20s", sf->version);
 	else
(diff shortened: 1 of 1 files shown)


skef (approved):
Looks fine to me. Anyone else have an issue with this?

frank-trampe:
It seems odd to me that the extra space would occur twice, particularly since the erratic spacing around `buffer` indicates that the lines were written separately rather than copied one from another. The only explanation that comes to mind is that the intent was to copy more text to the string later in those cases but that said code never materialized or was removed. I have searched for other code that appends to the TrueType version string and have found nothing relevant.

This change may alter the result of some font generation workflows. I will send a note out when we merge this to let people know.

ctrlcctrlv (approved):
Font Bakery isn't a standards authority, it can't declare anything «_verboten_». It's a linter, and it makes suggestions on best practice.

There's nothing disallowed in terms of having that space there. The Version record is allowed to have any characters following the number, and any characters terminate the minor version number. It's common to write even to the point of absurdity e.g.:

```plain
Version 1.0;fontmake 2.3.0;FontForge 20201107;ttx 4.18.2;2d848da95feea343bb1b9e1541af18d5bb7f2b3b
```

I've also seen:

```plain
Version 1.0 2020-01-31
```

However, this change is also fine, don't get me wrong. It looks better to have no space and it saves a byte. :shrug: It's likely the space was there because George was considering writing some other version info there, which he'd be allowed to do. Although these days the standard reads:

> A character such as “;” is helpful to separate different pieces of version information. 

So we'd probably want to follow this _de facto_ standard instead. Font Bakery also likely enforces it :wink:

PR #4597, opened 2021-01-29
Avoid freeing values[] set with findnumbers()

### LINKED ISSUE #4592: Importing PFA file results in segfault
I am currently dealing with parsing PDF files that use fonts that have no `ToUnicode` map associated with them, so in a first step, I wanted to extract the respective font and analyze it using FontForge. For that, I first executed `pdftops input.pdf` and from the resulting PostScript file, I extracted the part between the lines `%%BeginResource` and `%%EndResource` into a new file.

With that process, I created [this](https://gist.github.com/sigalor/125d60346977def3f17a90a87204f275) PostScript font (PFA) file. Unfortunately, when trying to import it into FontForge, the application segfaults.

- FontForge version: 20201107
- operating system: Ubuntu 20.04

skef:
Closes #4592

### DIFF
--- a/fontforge/parsepfa.c
+++ b/fontforge/parsepfa.c
@@ -2679,7 +2679,8 @@ void PSCharsFree(struct pschars *chrs) {
 return;
     for ( i=0; i<chrs->next; ++i ) {
 	if ( chrs->keys!=NULL ) free(chrs->keys[i]);
-	free(chrs->values[i]);
+	if ( chrs->lens!=NULL && chrs->lens[i]!=0 )
+	    free(chrs->values[i]);
     }
     free(chrs->lens);
     free(chrs->keys);


fontforge/parsepfa.c
```diff
@@ -2679,7 +2679,8 @@ void PSCharsFree(struct pschars *chrs) {
 return;
     for ( i=0; i<chrs->next; ++i ) {
 	if ( chrs->keys!=NULL ) free(chrs->keys[i]);
-	free(chrs->values[i]);
+	if ( chrs->lens!=NULL && chrs->lens[i]!=0 )
```
jtanx, line 2682: Care to explain this further? How does the value of `lens` dictate whether or not `values` should be freed?

skef: Normally `values[i]` is either NULL (and therefore OK to free) or points to a chunk of memory of length `lens[i]`, as in the allocation in `findstring[]`. 

However, there is a special case in `findnumbers()` where `values[i]` is set with a `(void *)` cast directly to the output of `strtol()`. In this case `lens[i]` is helpfully set to `0`. 

The crash in #4592 is caused by attempting to free `0x...01` or `0x...02`.  (I don't remember which at this point.) So to me this looks like a special use case not getting a matching special deallocation case, although I did not go through every use of `lens` and `values` to determine exactly what's going on with the full subsystem.

skef: Oh, and I guess it's worth clarifying that line 1151 is the allocation of `values`, not `values[i]` and this PR changes the pattern of deallocation of `values[i]` only.

jtanx (approved):

PR #4598, opened 2021-01-30
Don't change to custom encoding on setting altuni

skef:
... when the encoding is already unicode compatible.

`PyFF_Glyph_set_altuni()` currently ends with:

```
    for ( fvs=self->sc->parent->fv; fvs!=NULL; fvs=fvs->nextsame ) {
        fvs->map->enc = &custom;
        FVSetTitle(fvs);
    }
```

This is always executed no matter what the existing encoding. As far as I can see:

1. There's no reason to ever do this when the current encoding is UnicodeFull
2. There's no reason to do this when the current encoding is UnicodeBMP and none of the altuni values for the glyph are outside of the BMP range. 

Anyway this is super-annoying when working with a script because (for example) `createMappedChar()` stops working right after you add an altuni unless you set the encoding back to UnicodeBMP (for example) after every addition.

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -6728,7 +6728,7 @@ return( ret );
 static int PyFF_Glyph_set_altuni(PyFF_Glyph *self,PyObject *value, void *UNUSED(closure)) {
     int cnt, i;
     struct altuni *head, *last=NULL, *cur;
-    int uni, vs, fid;
+    int uni, vs, fid, bmp_compat=true;
     PyObject *obj;
     FontViewBase *fvs;
 
@@ -6754,6 +6754,8 @@ return( -1 );
 	    else
 		last->next = cur;
 	    last = cur;
+	    if (uni > 0xFFFF)
+		bmp_compat = false;
 	}
     }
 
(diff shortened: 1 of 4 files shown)


skef:
At the time of filing this is a partial fix, because while it does eliminate the problem it doesn't cause the character in question to be displayed at the altuni-corresponding place(s) in the FontView. You can see this if you save the file and reopen it. 

I played around with various combinations of code to make that work and didn't have much luck. @frank-trampe any ideas?

fontforge/python.c
```diff
@@ -6754,15 +6754,19 @@ return( -1 );
...
     AltUniFree(self->sc->altuni);
     self->sc->altuni = head;
 
     for ( fvs=self->sc->parent->fv; fvs!=NULL; fvs=fvs->nextsame ) {
-	fvs->map->enc = &custom;
-	FVSetTitle(fvs);
+	if ( !IsUnicodeEncoding(fvs->map->enc, bmp_compat) ) {
+	    fvs->map->enc = &custom;
+	    FVSetTitle(fvs);
+	} // XXX else rebuild map?
```
jtanx, line 6769: Is this this?

https://github.com/fontforge/fontforge/blob/9dff1f785b8fe1f0b7b7ea249705232a03ce4d2b/fontforge/encoding.c#L2073

PR #4610, opened 2021-02-03
Better handling of closed self-intersecting paths in Expand Stroke

skef:
The overlap removal heuristics for closed contours were only appropriate for source contours that don't self-intersect. The output contours in such cases are kind of bizarre and processing them individually will have strange results. (For this reason it also makes little sense to use `removeinternal`/`removeexternal` with such contours, but caveat excogitator.)

This removes most of the special handling when using `removeoverlap=='layer'` and improves it when using `removeoverlap=='contour`` (by processing the left and right sides of an additional source contour together).

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -2319,12 +2319,8 @@ static SplineSet *OffsetSplineSet(SplineSet *ss, StrokeContext *c) {
 	                         &closed, true);
 	    if ( !closed )
 		LogError( _("Warning: Left contour did not close\n") );
-	    else if ( c->rmov==srmov_contour ) {
-		left = SplineSetRemoveOverlap(NULL,left,over_remove);
-		SplineSetsCorrect(left, &trash);
-	    }
+
 	    cur = left;
-	    left = NULL;
 	}
 	if ( right!=NULL ) {
 	    CalcNibOffset(c, ut_ini, true, &no, -1);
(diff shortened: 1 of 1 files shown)


skef:
@dimitre Try this for (some of) your remaining problems. Analogous archive is here: 
[ses.zip](https://github.com/fontforge/fontforge/files/5921439/ses.zip)

dimitre:
totally solved the self intersecting even for bizarre complex shapes. 
Screen Shot 2021-02-03 at 18 33 54
Cheers!

skef:
Good! Are there any remaining problems with your sources?

dimitre:
@skef regarding expand stroke there is only one but it is super non standard, it used to work. 
If you think it is not useful I agree, if you want to look I can make a test case.
Screen Shot 2021-02-03 at 18 55 51

I still trying to debug a SEGFAULT that happens from time to time (in recent versions), not sure if it has something related to expand stroke or not.
And a spiro issue which I'll try to make the minimal case, most probable unrelated to the expand stroke.

skef:
We could put that zero problem in an issue (with whatever specific precision is needed to reproduce it) but whether it winds up on the "someday maybe" list depends on what the problem is. If it's extremely sensitive to setup-conditions then it might not be worth it. 

Obviously building with `-DCMAKE_BUILD_TYPE=Debug` will make the SEGV easier to identify once you hit it again.

@ctrlcctrlv No hurry on this since there's nothing further I need to PR, but if and when you do any testing try it with both `removeoverlap="layer"` (the default) and `removeoverlap="contour"` (or the dialog equivalents).

frank-trampe:
The code path is sufficiently tangled that I'm inclined to take a test-based review approach. If there are no objections, let's merge this.

frank-trampe (approved):

ctrlcctrlv:
No objection. That's the best way.

skef:
Just for the sake of completeness, let me explain the code a bit in a way that may make it easier to review:

We've just finished stroking a closed path. This could be an "0" or it could be a glyph contour we're turning into an "outline" style or making smaller or bigger (with internal only or external only) or it could be something more exotic like a stroked-font figure eight. 

The old code assumed everything was like an 0 or a glyph outline: a closed non-self-intersecting clockwise contour. Then (prior to processing) the "left" contour is the clockwise exterior and the "right" contour is the clockwise interior. So then you can just run Remove Overlap on the left (in case there were cusps). For the right you can reverse it, draw a big clockwise box around it, run Remove Overlap on that combination, and take the box away. 

If the source contour is self-intersecting or not clockwise these assumptions break down. In all cases but one it's therefore better to just run Remove Overlap on the *combination* of the two contours to get a proper result. You would think this causes a chicken-and-egg problem because you don't know the "clockwiseness" of the contours to begin with because they self-intersect, but it appears that the Remove Overlap algorithm is indifferent to the direction of the contours passed to it. However, it does *preserve* those directions, so when it's finished you need to get the them properly arranged. This is why I introduce SplineSetsCorrect() on that path. 

The special case is when the user chooses Internal Only or External Only. Then I just to back to the original assumption of a clockwise non-intersecting input contour because there's really nothing else to do with the tools I have. Anyone stroking a figure eight with one of those options probably doesn't have a solid handle on what they're doing anyway.

PR #4614, opened 2021-02-04
Protect against ceil(float) to int index overflow in splinefill.c

skef:
also: Fix ELOrder() leak in _SplinePointsListIsClockwise() (from Feb 2020)

Closes 4613

### DIFF
--- a/fontforge/splinefill.c
+++ b/fontforge/splinefill.c
@@ -276,7 +276,10 @@ return;
     }
     if ( es->interesting ) {
 	/* Mark the other end of the spline as interesting */
-	es->interesting[(int) ceil(e->mmax)]=1;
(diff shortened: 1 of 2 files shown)


frank-trampe:
@skef, what's the specific floating-point madness going on?

skef:
This is one of those cases where the arithmetic logic looks sound but because of floating point mushiness `cnt` can wind up as (arbitrary example) 99 and `mmax` on a given run can wind up as 98.0000000000034, which `ceil()` pushes to 99 which exceeds the allocated memory in the array. I could go back and reproduce to get the specific numbers but conceptually that's how it went. 

It's possible I missed something about how `mmax` is calculated so that the logic isn't sound but it looked that way to me.

frank-trampe (approved):

PR #4619, opened 2021-02-06
Allow 0-len mult subs in UI, feature files (GSUB)

### LINKED ISSUE #4618: [fea] Deleting glyphs should be possible
It is apparently a long-standing "secret" that it is actually possible to delete glyphs in OpenType feature code. This was brought to my attention by fonttools/fonttools#2170, which builds on work @simoncozens did in adobe-type-tools/afdko#1251 (cf. adobe-type-tools/afdko#1234). Both fonttools and AFDKO now support this type of multsub.

As Simon notes, this feature was brought to @behdadesfahbod's attention by [Firefox bug №1128658](https://bugzilla.mozilla.org/show_bug.cgi?id=1128658).

This is explicitly against the specification, but that is going to change. (Cf. MicrosoftDocs/typography-issues#673) As can be seen there, many others are accommodating this.

FontForge supports it—I tried to open and rewrite fonts with 0-length mult subs and it works fine. 

However, it doesn't support adding such mult subs via the UI or via feature files. I will fix this in a PR.

ctrlcctrlv:
This allows glyph deletion in OpenType Layout.

Cf. MicrosoftDocs/typography-issues#673, fonttools/fonttools#2170, adobe-type-tools/afdko#1251

Close #4618

FontForge already supported this implicitly: reading an OpenType font which uses this mechanism works, as does writing it back. Adding one in the UI was broken, as were feature files in both directions. This PR fixes that.

### Type of change
- **Non-breaking change**

### DIFF
--- a/fontforge/featurefile.c
+++ b/fontforge/featurefile.c
@@ -1405,7 +1405,11 @@ return;					/* No support for apple "lookups" */
 			    fprintf( out, "    sub " );
 			    dump_glyphname(out,sc);
 			    fprintf( out, " by " );
-			    dump_glyphnamelist(out,sf,pst->u.subs.variant );
+			    if (*(pst->u.subs.variant) != '\0') {
+				dump_glyphnamelist(out,sf,pst->u.subs.variant );
+			    } else {
+				fprintf( out, "NULL" ); // really possible! see GH: №4618
+			    }
 			    fprintf( out,";\n" );
 			  break;
 			  case gsub_alternate:
@@ -4399,7 +4403,7 @@ static struct feat_item *fea_process_sub_multiple(struct parseState *tok,
 	len += strlen(g->name_or_class);
 	mult[len++] = ' ';
     }
-    mult[len-1] = '\0';
+    mult[len>1?len-1:0] = '\0';
     item = chunkalloc(sizeof(struct feat_item));
     item->type = ft_pst;
     item->next = sofar;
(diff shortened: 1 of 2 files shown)


fontforge/featurefile.c
```diff
@@ -4732,14 +4736,20 @@ static void fea_ParseSubstitute(struct parseState *tok) {
...
 		++tok->err_count;
-	    } else if ( rpl->has_marks ) {
+	    } else if ( !next_is_null && rpl->has_marks ) {
 		LogError(_("No marked glyphs allowed in replacement on line %d of %s"), tok->line[tok->inc_depth], tok->filename[tok->inc_depth] );
 		++tok->err_count;
 	    } else {
-		if ( cnt==1 && rpl->next==NULL ) {
+		if (next_is_null) {
+		    tok->sofar = fea_process_sub_multiple(tok,glyphs,NULL,tok->sofar);
+		    fea_ParseTok(tok);
```
skef, line 4751: Can you help me out with a quick explanation of why we need to consume an extra token here but not in the other cases?

ctrlcctrlv: The string "NULL" is handled specially by the FEA parser, it is in its own token class, `tk_NULL`. Therefore, it causes all sorts of issues if it comes up in the next run, and this tk_NULL will never be properly consumed any other way. In the normal cases, where the glyph name in question actually resolves to a glyph, the swallowing of the token is was already done during the function `fea_ParseMarkedGlyphs`.

skef: OK --

This pattern makes me a bit nervous given that `fea_UnParseTok()` only sets `backedup` and seems to be there to match up with potential further calls to `fea_ParseTok()`. Before we were calling `fea_process_sub_single()`, `fea_process_sub_multiple()`, etc. with the saved state of the *previous* token. Now we're parsing again, setting `backedup` to true, and then calling those functions.

Is the idea that if parsing a NULL only modifies the `type` and those functions never look at the type then everything works out? That would be OK if the next token *is* a "NULL" but what if it's something else? Do we have an exhaustive list of what it could be? And what if the file has a syntax error that makes the token different -- won't we wind up with weird processing here rather than a syntax error message?

Maybe I just need more guidance about what's going on.

ctrlcctrlv: We already read in what the replacement is going to be into `rpl`. The only valid thing it can be at this point is either `NULL` (in which case we consume the NULL) or something already read into `rpl`. We still get logical errors if they exist in the input feature file.

Setting `backedup` is enough because it causes the next `fea_ParseTok` to fall through.

skef: OK, I've looked through the code in a bit more depth. I guess `tok` is only passed into the `_sub_` functions for error reporting purposes. This isn't great given [feature file whitespace handling](http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#2b-white-space), which will lead to incorrect error messages in some cases.

However, assuming that's acceptable I would prefer to centralize the logic this way:

```
	    fea_ParseTok(tok);
	    bool next_is_null = tok->type == tk_NULL;
	    if ( !next_is_null )
	        fea_UnParseTok(tok);
```
preferably with a comment about what's going on.

ctrlcctrlv: Force pushed, thanks.

skef: Oh, sorry, the idea of the added guard is to eliminate the need for `fea_ParseTok(tok);` on the next_is_null branch below. I left that part implicit.

ctrlcctrlv: As much my fault as yours :-) Really tired.

Force pushed again.

skef:
This looks fine to me relative to the caveats about error reporting. I'm going to put off approving for a day or two in case @frank-trampe or anyone else has a contrary opinion.

skef (approved):

PR #4626, opened 2021-02-10
fsys: resolve symlinks on Unix platforms

liushuyu:
This will make `fontforge` resolve the symbolic links on the Unix platforms when loading its assets.

This resolves the issue where if `fontforge` is invoked through symlink, it will not be able to load the resources.

### Type of change
- **Non-breaking change**

---
This change is [Reviewable](https://reviewable.io/reviews/fontforge/fontforge/4626)

### DIFF
--- a/gutils/fsys.c
+++ b/gutils/fsys.c
@@ -533,6 +533,7 @@ char *_GFile_find_program_dir(char *prog) {
 	free(tmppath);
     }
 #else
+    char* buf = NULL;
     if ( (pt = strrchr(prog,'/'))!=NULL )
 	program_dir = copyn(prog,pt-prog);
     else if ( (path = getenv("PATH"))!=NULL ) {
@@ -552,6 +553,10 @@ char *_GFile_find_program_dir(char *prog) {
 		program_dir = copy(path);
 	}
     }
+	if ( (buf = realpath(program_dir, NULL)) ) {
+	   free(program_dir);
+	   program_dir = buf;
+	}
 #endif
 
     if ( program_dir==NULL )


gutils/fsys.c
```diff
@@ -552,6 +553,10 @@ char *_GFile_find_program_dir(char *prog) {
 		program_dir = copy(path);
 	}
     }
+	if ( (buf = realpath(program_dir, NULL)) ) {
+		if ( program_dir!=NULL ) free(program_dir);
```
skef, line 557: You don't need a NULL check when calling `free()`.

liushuyu: Addressed.

skef (approved):
This seems fine to me. Anyone else have concerns?

jtanx:
Can you explain a full example? imo if you want to resolve symlinks this should be done on the path to the fontforge binary itself instead of the folder containing fontforge (like there is a potential difference resolving `/a/b/c` vs `/a/b/c/fontforge`).

liushuyu:
> Can you explain a full example? imo if you want to resolve symlinks this should be done on the path to the fontforge binary itself instead of the folder containing fontforge (like there is a potential difference resolving `/a/b/c` vs `/a/b/c/fontforge`).

This could happen in many Linux distro where `/bin/` is a symlink to `/usr/bin`. The fontforge will look for resource files in `/share` in this case, which is incorrect. This fix will hopefully resolve such issue.

Regarding your concern with the path resolution, I will change my patch to resolve the path to the binary.

jtanx:
Which distro is this? Sounds more like a problem that if you symlink /bin to /usr/bin, then so too should /share to /usr/share

gutils/fsys.c
```diff
@@ -552,6 +553,10 @@ char *_GFile_find_program_dir(char *prog) {
 		program_dir = copy(path);
 	}
     }
+	if ( (buf = realpath(program_dir, NULL)) ) {
+	   free(program_dir);
+	   program_dir = buf;
+	}
```
ctrlcctrlv, line 559: The whitespace is misaligned.

ctrlcctrlv (changes requested):

frank-trampe:
@liushuyu, just pinging regarding what @jtanx said.

PR #4631, opened 2021-02-14
Fix UFO features.fea relative include path implementation

### LINKED ISSUE #4629: Including files in UFO features.fea files: Our implementation is wrong
The [specification](https://unifiedfontobject.org/versions/ufo3/features.fea/) is clear:

> Any `include()` statements must be relative to the UFO path, not to the `features.fea` file itself.

Meaning, the path the UFO _is in_ is the relative path used in these files. When you have a UFO named `a.ufo` and it has a `features.fea` containing:

```fea
include(../mark.fea);
```

This means:

* include `a.ufo/../../mark.fea`

Not:

* include `a.ufo/../mark.fea`

That's the meaning used by `fontmake`. Our implementation is wrong and should change, but it's going to break everyone's UFO's.

What to do, what to do?

ctrlcctrlv:
We're including the wrong path when the features.fea include path is
relative. It's supposed to be relative to the directory the UFO is in,
not relative to features.fea.

I fixed this by following @skef's advice of checking the correct
directory first, and if it fails, then trying the deprecated directory,
and emitting a warning.

Close #4629

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/featurefile.c
+++ b/fontforge/featurefile.c
@@ -2276,6 +2276,7 @@ struct parseState {
     unsigned int in_vkrn: 1;
     unsigned int backedup: 1;
     unsigned int skipping: 1;
+    unsigned int in_ufo: 1;
     SplineFont *sf;
     struct scriptlanglist *def_langsyses;
     struct glyphclasses *classes; // TODO: This eventually needs to merge with the SplineFont group storage. For now, it needs to copy from it at first invocation.
@@ -2391,7 +2392,7 @@ static void fea_ParseTok(struct parseState *tok);
 
 static void fea_handle_include(struct parseState *tok) {
     FILE *in;
-    char namebuf[1025], *pt, *filename;
+    char namebuf[1025], *pt, *filename, *filename_deprecated;
     int ch;
 
     fea_ParseTok(tok);
@@ -2440,21 +2441,49 @@ return;
 return;
     }
 
-    if ( *namebuf=='/' ||
-	    ( pt = strrchr(tok->filename[tok->inc_depth],'/') )==NULL )
-	filename=copy(namebuf);
-    else {
-	*pt = '\0';
-	filename = GFileAppendFile(tok->filename[tok->inc_depth],namebuf,false);
-	*pt = '/';
+
+    pt = strrchr(tok->filename[tok->inc_depth],'/');
+    if ( pt != NULL ) *pt = '\0';
+    filename_deprecated = GFileAppendFile(tok->filename[tok->inc_depth],namebuf,false);
+    if ( pt != NULL ) *pt = '/';
+    if ( tok->in_ufo && *namebuf!='/' ) {
+        // we know this is a UFO's features.fea file
+        filename = GFileDirName(tok->filename[tok->inc_depth]);
+        char* filename_above = GFileAppendFile(filename, "..", false);
+        char* filename_fea = GFileAppendFile(filename_above, namebuf, false);
+        TRACE("fea_handle_include: FEA: %s, deprecated FEA: %s\n", filename_fea, filename_deprecated);
+        free(filename_above);
+        free(filename);
+        filename = filename_fea;
+    } else {
+        if ( *namebuf=='/' || pt == NULL ) {
+            filename = copy(namebuf);
+        } else {
+            filename = filename_deprecated;
+        }
     }
+
     in = fopen(filename,"r");
     if ( in==NULL ) {
-	LogError(_("Could not open include file (%s) on line %d of %s"),
-		filename, tok->line[tok->inc_depth], tok->filename[tok->inc_depth] );
-	++tok->err_count;
-	free(filename);
+        if ( filename != filename_deprecated ) {
+            free(filename);
+            filename = filename_deprecated;
+            in = fopen(filename,"r");
+        }
+        if ( in == NULL ) {
+            LogError(_("Could not open include file (%s) on line %d of %s"),
+                     filename, tok->line[tok->inc_depth], tok->filename[tok->inc_depth]);
+            ++tok->err_count;
+            free(filename);
 return;
+        } else {
+            LogError(_("Warning: FontForge's implementation of \
+include statements in UFO features.fea files was wrong. Please \
+correct your path `%s` as per the specification for guaranteed \
+future compatibility. (See GitHub issue №4629 for more info.)"), namebuf);
+        }
+    } else if ( filename != filename_deprecated ) {
+        free(filename_deprecated);
     }
 
     ++tok->inc_depth;
@@ -7309,6 +7338,10 @@ static void CopySplineFontGroupsForFeatureFile(SplineFont *sf, struct parseState
   }
 }
 
+static bool fea_isInUFO(char *filename) {
+    return g_regex_match_simple("^.*ufo[23]?[/\\\\]features.fea$", filename, 0, 0);
+}
+
 void SFApplyFeatureFile(SplineFont *sf,FILE *file,char *filename) {
     struct parseState tok;
     struct glyphclasses *gc, *gcnext;
@@ -7316,10 +7349,12 @@ void SFApplyFeatureFile(SplineFont *sf,FILE *file,char *filename) {
     struct namedvalue *nvr, *nvrnext;
 
     memset(&tok,0,sizeof(tok));
+
     tok.line[0] = 1;
     tok.filename[0] = filename;
     tok.inlist[0] = file;
     tok.base = 10;
+    tok.in_ufo = fea_isInUFO(filename);
     if ( sf->cidmaster ) sf = sf->cidmaster;
     tok.sf = sf;
     CopySplineFontGroupsForFeatureFile(sf, &tok);


fontforge/featurefile.c
```diff
@@ -7304,17 +7329,29 @@ static void CopySplineFontGroupsForFeatureFile(SplineFont *sf, struct parseState
   }
 }
 
+static bool fea_isInUFO(char *filename) {
```
skef, line 7332: I understand what this is doing but don't have a strong sense how "safe" it is.

fontforge/featurefile.c
```diff
@@ -2435,21 +2438,43 @@ return;
...
-	filename = GFileAppendFile(tok->filename[tok->inc_depth],namebuf,false);
-	*pt = '/';
+
+    pt = strrchr(tok->filename[tok->inc_depth],'/');
+    if ( pt != NULL ) *pt = '\0';
+    filename_deprecated = GFileAppendFile(tok->filename[tok->inc_depth],namebuf,false);
+    if ( pt != NULL ) *pt = '/';
+    if ( tok->in_ufo ) {
+        // we know this is a features.fea file
+        filename = copy(tok->filename[tok->inc_depth]);
```
skef, line 2448: I know the spec page you linked to says  "Any `include()` statements must be relative to the UFO path, *not* to the features.fea file itself." but are you sure they didn't mean "Any *relative* `include()` statements ..."? That is, are you sure they're ruling out absolute includes? I'm not sure why that would be.

skef:
@ctrlcctrlv should we pick this back up again?

ctrlcctrlv:
Yes, we should. Apologies for long delay.

fontforge/featurefile.c
ctrlcctrlv: I'm afraid I don't quite understand the question. Can you say what you think my patch does wrong in the case of absolute includes?

ctrlcctrlv: I'm afraid I don't quite understand the question. Can you say what you think my patch does wrong in the case of absolute includes?

fontforge/featurefile.c
ctrlcctrlv: From what I can tell, it's completely safe. If someone tries to import a fea file that's part of a foreign UFO, for example, they'd still want the imports to be resolved as they would be if a fea file is part of a UFO.

fontforge/featurefile.c
skef: The old code path had a special case for `*namebuf=='/'` where it just copied the whole filename (thereby handling absolute paths). I don't see the equivalent in the new code path.

ctrlcctrlv: I don't really see any evidence absolute includes are supported in UFO. What about Windows, where it's C:, D:, etc? The old code wouldn't have worked with that. I think UFO's strong use of "any" means that even leading slashes denote relative files.

skef: The spec has been updated to include language about UFO resolution and it doesn't rule them out: http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#3-including-files . And I'm not sure why one *would* rule them out. Most paths can be relative or absolute.

ctrlcctrlv: Do I also need to support Windows notation, as `C:/WINDOWS/example.ufo/features.fea`? I _assume_ that this can happen, but that we're saying even on Windows to use <kbd>/</kbd>, as elsewhere? I'm just not sure absolute includes make any sense in a UFO file. UFO is supposed to be self-contained, movable around the system. The reason the FEA's live in the dir above is MM fonts, made of multiple UFO's under a project root. Why would you ever even have a UFO with an absolute include?

skef: I believe that somehow, perhaps via posix interfaces, the slash/backslash problem with pathnames is taken care of, so the windows notation issue would be related to switching drive letters. Allowing that would be virtuous but probably not mandatory.

I don't read the spec the way you do and I take the possibility of an absolute path as a potential work-around, perhaps a temporary one, that should still be available. If you want this to go in with this functionality change you should probably be asking @frank-trampe or someone else on the core team.

ctrlcctrlv: It's true that on modern Windows you can mount a drive letter under any other drive letter. If we're saying that's the solution here, then I can deal with that and go ahead and update the patch to consider `/` to be absolute, even in a UFO, as before. I don't know that that's exactly to-spec, but I can see why you might think it should work, and it's probable other software supports it. The UFO spec has some ambiguities, this is one, so I can resolve it this way, that's fine, just leaving the resolution of drive letters up to the user to fix, either through mounting or what have you.

fontforge/featurefile.c
```diff
@@ -27,6 +27,8 @@
  * ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <libxml/xmlregexp.h>
```
jtanx, line 30: This should come after fontforge-config.h

fontforge/featurefile.c
```diff
@@ -2435,21 +2438,43 @@ return;
...
-	*pt = '/';
+
+    pt = strrchr(tok->filename[tok->inc_depth],'/');
+    if ( pt != NULL ) *pt = '\0';
+    filename_deprecated = GFileAppendFile(tok->filename[tok->inc_depth],namebuf,false);
+    if ( pt != NULL ) *pt = '/';
+    if ( tok->in_ufo ) {
+        // we know this is a features.fea file
+        filename = copy(tok->filename[tok->inc_depth]);
+        char* filename_above = GFileAppendFile(GFileDirName(filename), "..", false);
```
jtanx, line 2449: This leaks the result from GFileDirName

fontforge/featurefile.c
jtanx, line 2448: Why do you need to copy this at all?

fontforge/featurefile.c
```diff
@@ -7304,17 +7329,29 @@ static void CopySplineFontGroupsForFeatureFile(SplineFont *sf, struct parseState
   }
 }
 
+static bool fea_isInUFO(char *filename) {
+    // Note: Per standard, ^ and $ are implicitly placed at tail and end, so
+    // these aren't defined and can't be used.
+    static xmlChar *regex_pattern = ".*ufo[23]?[/\\\\]features.fea";
```
jtanx, line 7335: mark const

You could also probably just replace the whole thing with something like

```
return g_regex_match_simple("^.*ufo[23]?[/\\\\]features.fea$", filename, 0, 0);
```

fontforge/featurefile.c
```diff
@@ -2435,21 +2438,43 @@ return;
...
     }
     in = fopen(filename,"r");
     if ( in==NULL ) {
-	LogError(_("Could not open include file (%s) on line %d of %s"),
-		filename, tok->line[tok->inc_depth], tok->filename[tok->inc_depth] );
-	++tok->err_count;
-	free(filename);
+        in_deprecated = fopen(filename_deprecated, "r");
+        if ( in_deprecated != NULL ) {
+            LogError(_("Warning: Prior to February 2021, FontForge's implementation of \
```
jtanx, line 2465: I would drop the date from this

fontforge/featurefile.c
iterumllc: I think both these problems would be solved by just doing `filename = GFileDirName(tok->filename[tok->inc_depth]);` on the first line and `GFileAppendFile(filename, "..", false);` on the second.

skef: Whoops -- alter ego there.

dimasnirwan:
Hello,

I have downloaded it several times, but the file is broken.
The file downloaded, size is not the same as listed.

ctrlcctrlv:
Sorry? I don't know what that has to do with this PR.

fontforge/featurefile.c
jtanx: Now that I think about this again, can we not just compare that this filename is contained by `sf->filename` (possibly applying `GFileGetAbsoluteName`)?

ctrlcctrlv: I don't think so, because of  it's valid to have non-`features.fea` `.fea` files in a UFO. You'd at least need to check that the file is named `features.fea`.

jtanx: Right, but I guess if you checked that + against the filename that's a stronger check than some random path that may or may not come from the same ufo

ctrlcctrlv: Wait, what? I don't understand what's wrong with the way I'm doing it. Who cares if the file is contained by `sf->filename`? That could e.g. be the case with `sfdir`, the directory version of SFD. I think also those weird "Apple font suitcases" we saw discussed elsewhere could also apply here. Just checking that the path ends with `UFO[23]?` is much simpler than assuring we're "contained" by `sf->filename`, but then also making sure that it's not a false positive (we're not in a font suitcase or in an SFD or any other way I'm not thinking of which could trigger true).

Also, I consider the fact that we're not making sure that `features.fea` comes from the same UFO a feature and not a bug. If the `features.fea` is part of any UFO, than obviously `include` resolution should happen exactly as in the UFO spec. Imagine a variable font with 16 masters…if you include a `features.fea` inside a foreign UFO, which way do you want `include` to resolve? Obviously the same as it would in that UFO.

So yeah, not convinced.

jtanx: Ok, fair.

jtanx (approved):

PR #4637, opened 2021-02-18
Fix bug caused by possibly negative unienc.

ahyangyi:
### Type of change
- **Bug fix**
This oneliner fixes potential crash when `unienc` is a negative value other than -1. I haven't traced why such a negative value is even possible, but at least when it happens, it does not crash the whole thing.


ahyangyi:
Temporarily closing this PR, since it looks like needing further investigations.

PR #4639, opened 2021-02-19
Stop UniFromName from returning random-looking numbers

ahyangyi:
### Type of change
- **Bug fix**

UniFromName will return random-looking numbers for glyphs without encoding slot and named like "u10B3110B19". This oneliner fixes the problem.

This bug is the underlying problem of what I was trying to fix with #4637.

### DIFF
--- a/fontforge/namelist.c
+++ b/fontforge/namelist.c
@@ -142,7 +142,7 @@ int UniFromName(const char *name,enum uni_interp interp,Encoding *encname) {
 	if ( *end )
 	    i = -1;
 	_recognizePUA = true;
-    } else if ( name[0]=='u' && strlen(name)>=5 ) {
+    } else if ( name[0]=='u' && (strlen(name)==5 || strlen(name)==6)) {
 	i = strtol(name+1,&end,16);
 	if ( *end )
 	    i = -1;
--- a/fontforge/namelist.c
+++ b/fontforge/namelist.c
@@ -171,6 +171,7 @@ int UniFromName(const char *name,enum uni_interp interp,Encoding *encname) {
     }
     if ( !_recognizePUA && i>=0xe000 && i<=0xf8ff )
 	i = -1;
+assert(i >= -1);
 return( i );
 }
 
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
Excellent. Am I to understand that this is a type of ligature notation, u10B3110B19? I suppose that it denotes U+10B31 U+10B19?

ahyangyi:
Yep, and the particular glyph name is from NotoSansAvestan.

skef:
I think it's OK to leave the code that was going to be fixed by #4637 but it would be better to add `assert( unienc >= 0 );`. (Leaving it means not hiding problems that cause it to have a bad value.)

ahyangyi:
@skef I added the assertion according to discussion. Note that it's `i>=-1` instead of `i>=0` since UniFromName uses `-1` to denote cases where one cannot derive a unicode slot from the name.

skef:
I was thinking of the check near the `index` assignment, where we've already checked it isn't -1, but this is probably better.

ctrlcctrlv:
Are you going to fix the test failure, @ahyangyi ?

```
../fontforge/namelist.c:174:1: error: implicit declaration of function ‘assert’; did you mean ‘g_assert’? [-Werror=implicit-function-declaration]
 assert( i >= -1 );
 ^~~~~~
 g_assert
```

ctrlcctrlv:
@frank-trampe Before you merge this, know elsewhere that we tend to include `assert.h`

ctrlcctrlv:
(The C compiler links it by default on some platforms, which caused it to work for Yi Yang but not us)

fontforge/namelist.c
```diff
@@ -171,7 +171,7 @@ int UniFromName(const char *name,enum uni_interp interp,Encoding *encname) {
     }
     if ( !_recognizePUA && i>=0xe000 && i<=0xf8ff )
 	i = -1;
-assert( i >= -1 );
+g_assert( i >= -1 );
```
jtanx, line 174: don't do this, `#import <assert.h>` instead

ahyangyi:
@ctrlcctrlv actually, I attempted to fix it via the Github online editor because I was not at my home computer. Hence the version without the proper #include statement would probably break everyone's build, include the one of myself.

Since you seem to have reached a consensus, you are welcome to go ahead and correct it for me and merge this.

Alternatively, I'll fix it when I have time and in front of my home computer.

ahyangyi:
@ctrlcctrlv done.

ahyangyi:
Friendly ping.

PR #4646, opened 2021-02-25
Add selection status to reference get/set routines

### LINKED ISSUE #4303: Expose reference selection in python API
The current python API can adjust glyph contours (by point) and also references (by transformation matrix). However, in order to build interactive scripts that register menu items it helps to have some sort of UI. For contours the UI is typically based on point selection, which is how a lot of FontForge built-ins work. References can also be selected in FontForge but their selection status is not exposed in the API. We should do that. 

My inclination would be to add a third boolean element to the reference tuple to indicate if it is selected. Of course, this is likely to break some existing scripts. We can make the assignment code tolerant to omitting the value to reduce that breakage somewhat.

skef:
Closes #4303 

I first wrote this up on the model of `glyph.anchorPointsWithSel` by copying the `PyFF_RefArray` object and its related functions. But then after realizing that the `selected` element could and should be optional adding it didn't seem likely to cause problems. Consumers of the output of `glyph.references` and `glyph.layerrefs` will see an extra element at the end of the tuple, and that could break particularly sensitive scripts, but for the most part this won't create new issues.

@frank-trampe If you want me to do the more conservative thing instead I can. Still, this seems like a candidate for mailing a warning to the email list and calling it a day.

### DIFF
--- a/doc/sphinx/scripting/python/fontforge.rst
+++ b/doc/sphinx/scripting/python/fontforge.rst
@@ -1517,10 +1517,12 @@ This type may not be pickled.
    Ends the contour without closing it. This is only relevant if you are
    stroking contours.
 
-.. method:: glyphPen.addComponent(glyph_name, transform)
+.. method:: glyphPen.addComponent(glyph_name[, transform, selected])
 
    Adds a reference (a component) to the glyph. The PostScript transformation
-   matrix is a 6 element tuple.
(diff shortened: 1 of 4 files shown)


frank-trampe (approved):

frank-trampe:
Since the old documentation did not promise a limit on the length of the tuple, I think that we can slip this one through without any mailing list action.

PR #4664, opened 2021-03-13
GDraw and Charset cleanup

### LINKED ISSUE #4334: Removing the character builder
I'm looking into cleaning up the Unicode library, specifically currently into dropping the [charset conversion tables](https://github.com/fontforge/fontforge/tree/master/Unicode/charset), which are unused 99% of the time because iconv is used instead (and iconv is a hard dependency).

One of the remaining uses of the hand-rolled charset conversion tables is the existence of a character composer and insertion dialog, which is [documented here](https://fontforge.org/docs/ui/mainviews/metricsview.html#entering-non-ascii-characters), and looks like this:

image

I was contemplating switching it to use iconv if possible, but looking further, it turns out this functionality has been broken (read: unusable) since late 2012 (bisected to f5fbd93b1e8b653c515faacfdc48886fe950b8d9, with the introduction of the new hotkey system).

So considering:
* It hasn't been usuable for 8 years and no one has complained
* The GDK backend doesn't actually implement support to make it work anyway
* If moving to GTK/Qt this likely would not be ported over
* There exist other, dedicated tools to do something similar (charmap, gucharmap etc., alt codes etc.)

My preference is to remove it, rather than try to fix it up. Thoughts?

jtanx:
This includes multiple changes, probably best viewed commit by commit. Primary focus is:

* Removing unused code/functions from GDraw to make future porting easier
* Working towards making GDraw only do paint commands in an expose event. Far from complete, not sure it ever will. But this starts with the removal of GDrawSetDifferenceMode and changes to how the cursor is rendered - previously would XOR the display where the cursor was on a timer, it now requests an expose/redraw for the cursor area
* Getting rid of the custom charset conversions in favour if iconv (which is already a hard requirement). Shaves of ~2MB, for a rarely used feature(?) nowadays. I'm also fairly sure the conversion tables weren't always quite correct, from my ad-hoc testing.

There are a couple of gdraw related bug fixes too, related to keeping the palettes window visible when undocked.

Closes #4334 

### Type of change

- **Bug fix**
- **New feature**
- **Non-breaking change**


fontforge/parsettf.c
```diff
@@ -4241,71 +4241,73 @@ return( -1 );
...
-	}
-    } else if ( modtype==3 /* GB2312 offset by 0x8080, parse just like wansung */ ) {
-	if ( enc>0xa1a1 ) {
-	    enc -= 0xa1a1;
-	    enc = (enc>>8)*94 + (enc&0xff);
-	    enc = unicode_from_gb2312[enc];
-	    if ( enc==0 ) enc = -1;
-	} else if ( enc>0x100 )
-	    enc = badencoding(info);
+        // Apart from these, cp932 is a strict superset of sjis
```
jtanx, line 4244: This section may be the most contentious part, using  iconv for reading the cmap encoding. But at least for SJIS, I think this is actually better. There are some encoding differences for gb2312/big5/wansung/johab, but it's hard to say if it's for the better or worse.

From testing purely the conversion routines, our custom rolled conversion tables were spitting out garbage conversions for at least some slots, but I don't know if that had any effect in practice. Also I guess iconv support may not be equal everywhere. But I believe we're already using iconv for everything else as-is (particularly for re-encoding).

This was/is really hard to test, non-unicode encoded fonts are so rare today.

fontforgeexe/prefs.c
```diff
@@ -788,196 +784,17 @@ static char *PrefsUI_getFontForgeShareDir(void) {
...
-
-    enc = encmatch(loc,false);
-    if ( enc==e_unknown ) {
-	loc = strrchr(loc,'.');
-	if ( loc==NULL )
-return( e_iso8859_1 );
-	enc = encmatch(loc+1,true);
+    if (!SetupUCharMap(FindUnicharName(), charset, is_utf8)) {
+        fprintf(stderr, "Failed to set up unichar<->system local encoding, assuming utf-8 and trying again...\n");
+        if (!SetupUCharMap(FindUnicharName(), "UTF-8", true)) {
```
jtanx, line 793: Previously if iconv setup failed it would default to latin1, I now default to utf8, which I think is more appropriate today.

fontforge/parsettf.c
ctrlcctrlv: Using iconv is better because at least font developers know what iconv's choices are and what to look out for. Using our own tables, based just on George's opinions, makes things more confusing. So I agree with this change.

fontforgeexe/prefs.c
skef: I guess this isn't really related, but `default_encoding` in `startui.c` and `startnoui.c` is still built from `FindOrMakeEncoding("ISO8859-1")`. Given the difficulty you had finding non-unicode fonts, shouldn't we change this to BMP as well?

fontforge/parsettf.c
```diff
@@ -4241,71 +4241,73 @@ return( -1 );
...
+        if (enc == '\\')
+            enc = 0xa5; /* Yen */
+        else if (enc == 0x7e)
```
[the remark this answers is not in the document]
jtanx: See https://en.wikipedia.org/wiki/Code_page_932_(Microsoft_Windows)#Single-byte_character_differences for more info (+ #2820). From my sample size of 2, the Yen mapping is pretty common, and flip a coin for the tilde mapping.

At the least this keeps the same behaviour as before.

fontforgeexe/prefs.c
jtanx: If everyone's fine with changing the default encoding, I'm happy to change it, but maybe as a separate PR. ISO8859-1 is hardcoded in quite a few places.

fontforge/parsettf.c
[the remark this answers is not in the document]
skef: Sorry, the cropping on this was weird I meant what's the point of:
```
                else if (enc == 0x7e)
                     enc = 0x7e; /* Tilde */
```

fontforgeexe/prefs.c
skef: sounds reasonable

fontforge/parsettf.c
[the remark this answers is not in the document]
jtanx: Ah right, remnants of when I was using (iconv) SJIS instead of CP932. 

```
jeremy@ubuntu:~$ iconv -f CP932 -t UTF-8 test.txt
\~
jeremy@ubuntu:~$ iconv -f SJIS -t UTF-8 test.txt
¥‾
```

Guess it's more or less redundant now.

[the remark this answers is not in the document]
skef: Better to eliminate the guard if it's not needed or make the next one into `if (enc != 0x7e)`

skef:
Looks like we're having Windows build problems.

jtanx:
Yeah, I'm looking into it (https://github.com/msys2/MINGW-packages/pull/8186#issuecomment-807537301)

skef (approved):
Oh, thought I had approved this. Doing that now.

jtanx:
Looks like the cmake release is at least another week away so for now I've just pinned back to cmake 3.19 (https://github.com/fontforge/fontforgebuilds/commit/c0067e1af0bf859f6652b5a9b55a0a1992a96906).

Will rerun the other builds now.

PR #4680, opened 2021-03-26
Force getAutoDirName to have no arguments.

MaddTheSane:
This forces a _warning_ ("too many arguments in call to 'getAutoDirName'") encountered in Clang 12 in `DoAutoRecoveryExtended` to be an error. This also makes the  `getAutoDirName` call in said function to not pass any arguments.

### Type of change

- **Bug fix**


skef (approved):
lgtm

PR #4682, opened 2021-03-26
Improved python menu support

### LINKED ISSUE #4147: Python-based UI interface improvements
It seems like the project may be on the cusp of making more regular use of python scripts and potential user contributions. So I think I'll look into extending the interactive aspects of the scripting interface a bit more, in some ways we've already discussed in past issues and in other ways that are new (and easier). Here is a list:

1. [ ] Do the discovery thing discussed in #3663 so that you can install a python package in the normal way and have it add menu items. (Skip enable/disable for now, add that if and when people ask for it.)
2. [ ] Modify `registerMenuItem` so that if the `menu-function` argument is the string "separator" the "menu-name' item is replaced with a menu separator. 
3. [ ] Modify `registerMenuItem` so that if the `menu-function` argument is the string "link" the "data" field is treated as a URL, which is opened when the item is selected from the menu. (For good measure add an interface to the `fontforge` object for opening a web page in other contexts.) 
4. [ ] If possible, re-enable the `shortcut-string` (now more or less ineffectual) as a "suggestion", so that if the specified hotkey is not already "taken" it is enabled on that item. (The tricky thing here is whether one can tell that a hotkey is taken. I don't remember.) Junk the current syntax in favor of the hotkey file syntax (if they are different). Don't require repeating the name in the shortcut string. 
5. [ ] Write a short style guide, either in the docs or on the wiki (seems like docs would be preferable -- not many people know about the wiki), for scripts-for-other-people. Basically: encourage a single top-level menu item for each contribution with multiple sub-items. Then at the bottom a separator and a link to an "about" page. Also some suggestions for hotkey suggestions. Include the basics of how to write one module that does different things in different contexts. 

Unless discovery is trickier than I expect this shouldn't add up to much work.

### LINKED ISSUE #4148: Configurable menu design issues
If we're going to eventually have configurable menus we should think a bit about how `registerMenuItem()`, hotkeys, and configurability will all interact.

Setting hotkeys aside for the moment the problem boils down to this: When someone is laying out a menu what is the token that represents a particular item? For menu items built into fontforge this can be whatever we want it to be as long as something about the represents the window type (charview, fontview, etc). So maybe a shortish unique string. This would not be the string that appears in the menu because that should be configurable and localizeable. But perhaps there should be a default menu string and (therefore) a default set of localizations. 

In that case, what would represent a registerMenuItem item? We could make the author specify a string but we don't now and anyway that raises the problem of clashes, and therefore token "registration". So maybe the way to go is some concatenation of the sub-menu and menu names the item was registered with, preceded by or associated with a window identifier (CV or FV). Clashes are still possible then but unlikely enough that it's viable to punt that problem to the end-user. 

For Hotkeys I'm inclined to allow a default to be specified for each item as it is defined. If the default is already in use, or if the user has specified a different hotkey for the item, it is ignored. That way if you're laying out your own menu you only need to specify the hotkeys for to override that default. 

If and when it comes to it there should probably be a "fake" item that represents all or part of the Tools menu, so that a user has the option of inserting those items as a group rather than tracking them individually.

### LINKED ISSUE #4638: pythonui.c: InsertSubMenus() freeing memory still pointed to in fontview
I saw this with ASAN on when playing around with scripts that register menu items:

```
==1190545==ERROR: AddressSanitizer: heap-use-after-free on address 0x61200002ac40 at pc 0x5570e0758f3b bp 0x7ffd807f2ca0 sp 0x7ffd807f2c90
READ of size 8 at 0x61200002ac40 thread T0
    #0 0x5570e0758f3a in mb2FreeGetText ../fontforgeexe/windowmenu.c:312
    #1 0x5570e0758ef9 in mb2FreeGetText ../fontforgeexe/windowmenu.c:316
    #2 0x5570e043b8cc in FontViewFinish ../fontforgeexe/fontview.c:7034
    #3 0x5570e043b8eb in FontViewFinishNonStatic ../fontforgeexe/fontview.c:7039
    #4 0x5570e070e28e in fontforge_main ../fontforgeexe/startui.c:1441
    #5 0x5570e0063da8 in main ../fontforgeexe/main.c:33
    #6 0x7f2210465151 in __libc_start_main (/usr/lib/libc.so.6+0x28151)
    #7 0x5570e0063cbd in _start (/usr/bin/fontforge+0x160cbd)

0x61200002ac40 is located 0 bytes inside of 288-byte region [0x61200002ac40,0x61200002ad60)
freed by thread T0 here:
    #0 0x7f2212ac883a in __interceptor_realloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:164
    #1 0x5570e063b510 in InsertSubMenus ../fontforgeexe/pythonui.c:246
    #2 0x5570e063bd44 in PyFF_registerMenuItem ../fontforgeexe/pythonui.c:314
    #3 0x7f22113e82a1  (/usr/lib/libpython3.9.so.1.0+0x14c2a1)

previously allocated by thread T0 here:
    #0 0x7f2212ac883a in __interceptor_realloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:164
    #1 0x5570e063b510 in InsertSubMenus ../fontforgeexe/pythonui.c:246
    #2 0x5570e063bd44 in PyFF_registerMenuItem ../fontforgeexe/pythonui.c:314
    #3 0x7f22113e82a1  (/usr/lib/libpython3.9.so.1.0+0x14c2a1)

```

The relevant lines in `InsertSubMenus()` are:

```
        if ( *mn==NULL || (*mn)[j].ti.text==NULL ) {
            *mn = realloc(*mn,(j+2)*sizeof(GMenuItem2));
            memset(*mn+j,0,2*sizeof(GMenuItem2));
        }
```

So presumably when that realloc changes the pointer something still has a hold on the old one. 

`mb2FreeGetText()` takes the menu structure as an argument.

skef:
This does more or less what @frank-trampe, @ctrlcctrlv, and I discussed at the meeting a while back:

1. You can now suggest a hotkey to `registerMenuItem()` and it will be added if not already used.
2. You can add a mnemonic to a menu name and if the name is for a level lower than "Tools" it will be used.
3. At the Tools level the mnemonic is taken as a suggestion and added if not already used. If it is used the first 7bit character found that is not already used is added. If there are no such characters we start working through other possibilities, which are displayed after the name in parentheses.
4. You can now specify a name by just a string only, a two-tuple with a name and an identifier (for future use), or a three-tuple with a localized name, untranslated name, and identifier. The untranslated name is used to add the hotkey (both the "suggested" one from the python API or a line in the user's hotkeys file.
5. `registerMenuItem()` now has both a positional API and a keyword API. The former is forwards-compatible with previous versions of FontForge. There is also a short keyword-based API for adding a menu separator.
6. There's a fair amount of modified and additional documentation.

Adding hotkeys to Tools menu items continues to work, although now that names can be localized these hotkeys work a bit more like the others. If there is no localized/untranslated pair the string (minus mnemonic marks) is used in both contexts. I've verified that you can assign a hotkey to an item with a Japanese name.

As usual no Unit tests for UI stuff but I've done a fair amount of testing, including eliminating a number of problematic mnemonic key candidates.

It's a shame that you can't assign a hotkey with *either* English or a localized string but that would be tricky to fix because of mnemonics-related-issues.

Closes #4638
Closes #4148 
Closes #4147 (in conjunction with the already-merged #4642)


fontforgeexe/fontview.c
```diff
@@ -7047,7 +7055,17 @@ return;
 
     fontview_ready = true;
 
+// The tools menu handles its own translation. I would rather do this by testing
+// whether ti.text_untranslated is already null but there are hundreds of missing
+// initializers for that field in the menu layout code.
+#ifndef _NO_PYTHON
+    GMenuItem2 *t = FVGetToolsSubmenu();
+    FVSetToolsSubmenu(NULL);
+#endif
```
frank-trampe, line 7064: What side effect are we trying to bypass here?

skef: Ah -- The localization of the python sub-menu is handled differently and we don't want `mb2DoGetText()` to write over it for us.

I wanted to handle this by pre-setting `.ti.text_untranslated` to non-NULL and modifying `mb2DoGetText()` to only set the fields if `text_untranslated` is NULL. I believe the reason I couldn't do that is because all of the static `gtextinfo` initializers, including those in all the menu entries, poop out before the `text_untranslated` field (which was added later). Therefore there's often garbage in the field and you can't use its initial value for anything. \*sigh\*

skef: (maybe the menus aren't a problem because they're static, but I know I ran into it in some cases.)

fontforgeexe/pythonui.c
```diff
@@ -50,41 +51,215 @@
...
+    return ret;
+}
 
-static void py_tllistcheck(struct gmenuitem *mi,PyObject *owner,
-	struct python_menu_info *menu_data, int menu_cnt) {
+// Make sure there is no suffix when c==0 and exactly one suffix " (c)" when c!=0
+static char *SetMnemonicSuffix(const char*menu_string, unichar_t c) {
+    char *r = mncopy(menu_string, NULL, 4);
+    int l = strlen(r), i=l-1;
+
```
frank-trampe, line 214: Are you sure about the bounds-checking here? We can't count on a null terminator going backwards. If `r` is "", it seems like we might get a segmentation fault checking `*(r+i)` with `i = -1`. `r = " "` and other edge cases also seem likely to cause problems.

skef: I rewrote this a few times and am trying to remember why it wound up this way.

Clearly if menu_string is NULL you're already in trouble, so I guess I should return an `extra`-length zeroed out string in that case. That would also make `strlen()` happy. 

For the rest of it could I just add `i>=0 && ` to all the guards before the last one?

skef: I improved this some. `registerMenuItem()` rejects null name strings but I confirmed this works with empty strings and does something appropriate with all-space strings and trailing strings. It also passes my old test sample.

fontforgeexe/pythonui.c
```diff
@@ -50,41 +51,223 @@
...
+	if ( *src!='_' )
+	    ++l;
+	++src;
+    }
+
+    src = str;
+    ret = dst = malloc(l+1+extra);
+    for( ; *src; src++ ) {
+        if( *src == '_' ) {
+	    t = *(src+1);
```
jtanx, line 199: This is going to break if the following character is not ascii. Also noting the lack of the use of `utf82u_mncopy`

jtanx (commented):
So I did stop reviewing once I hit `mncopy`, but just posing a design question overall:

To confirm: on e.g. `"Annuler (4)";` - is that assigning it the mnemonic of 4 based on its position in the menu?

imo in general this feels like an over-complication of matters, in that I don't understand what's so bad about just doing a best effort of assignment of mnemonics (i.e. how much benefit there is over #4498 in terms of mnemonic selection) - and in particular, why this (python scripting menu items vs in-built menu items) needs to be treated so specially to everything else.

Also the ascii restriction feels like an over-restriction/simpification too. `u2utf8_mncopy` deals with extracting unicode mnemonics just fine, and gdraw not activating on such entries is possibly more of a gdk regression, or at least a result of comparing against keysyms instead of possible unicode values (I've just been delving into this, relates to #4712).

skef:
> To confirm: on e.g. `"Annuler (4)";` - is that assigning it the mnemonic of 4 based on its position in the menu?

No -- When the requested mnemonic is taken or there is no requested mnemonic the code just works through a hard-coded list of potential options, ordered roughly from least likely to most likely (to increase the chance that later entries get their requested mnemonic). It assigns the first one it finds free. 

> imo in general this feels like an over-complication of matters, in that I don't understand what's so bad about just doing a best effort of assignment of mnemonics (i.e. how much benefit there is over #4498 in terms of mnemonic selection) - and in particular, why this (python scripting menu items vs in-built menu items) needs to be treated so specially to everything else.

The answer to this last question is straightforward: All other menus in the project have, or are likely to have, static entries relative to the localized language (which can be adjusted as needed). This (likely) includes Tools *submenus*. That means either we or the submenu designer can lay out the mnemonics in advance in a way that makes sense. 

The Tools menu, in contrast, is composed of whatever the plugins and user have put in there, in whatever order they were added or the user picked. The developer of plugin Foo can't count on their requested mnemonic being available because they can't know what else has been added to the menu before their entry. Some users prefer mnemonics, and this was the way that I came up with that was "stable" (relative to the entries from the start, anyway) and that would assign a mnemonic to every entry. 

> Also the ascii restriction feels like an over-restriction/simpification too. `u2utf8_mncopy` deals with extracting unicode mnemonics just fine, and gdraw not activating on such entries is possibly more of a gdk regression, or at least a result of comparing against keysyms instead of possible unicode values (I've just been delving into this, relates to #4712).

I don't think non-ascii options work as accelerators, or at least when @frank-trampe, @ctrlcctrlv and I discussed this on the phone it didn't seem like an option. These are typically activated by Alt and accents and such are also often entered that way. (Or, however they're entered with the keyboard is likely to conflict.) The usual convention seems to be picking a salient letter in the case of other Latin-encoded languages and adding the English mnemonic in parentheses in non-Latin-encoded languages. If you start up FontForge in Korean (`LANGUAGE=ko fontforge`) or Japanese ('ja') you can see examples of that.

jtanx:
>  If you start up FontForge in Korean (LANGUAGE=ko fontforge) or Japanese ('ja') you can see examples of that.

Yep, I'm aware of that convention.

Having said that, reading the gdraw code I do believe it was done with the intention of supporting more than ascii, as per how `utf82u_mncopy` extracts unicode code points, as are mnemonics in terms of code points, and `GK_Special` used in some places to check if keysyms are unicode (a flawed check imo).

I was just playing around with it, and it can work: https://github.com/fontforge/fontforge/compare/master...jtanx:gms?expand=1

That also fixes #4712 (at least on KDE+Windows, Mac seems like something else is needed, not sure if gdk supports it). 

It comes down to the mnemonic checks being done against the `keysym`, which is not necessarily unicode (well especially for GDK, maybe this used to work for X11 - some comments certainly suggest this is the case), but the mnemonic being defined in terms of unicode

skef:
I think that URL is screwed up.

gdraw/hotkeys.c
```diff
@@ -266,36 +265,36 @@ void hotkeysLoad()
...
     char* fn = getHotkeyFilename(0);
     if( !fn ) {
 	return;
     }
-    loadHotkeysFromFile( fn, true, false );
+    loadHotkeysFromFile( fn, hk_user, false );
     free(fn);
 }
 
 static void hotkeysSaveCallback(Hotkey* hk,FILE* f) {
```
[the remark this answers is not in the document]
skef: I believe that `hotkeysSaveCallback()` was speculative generality anticipating a UI for changing hotkeys. It writes out the entries back to the user's hotkey file. I disabled it in #3606 because automatically writing over a file meant to be hand-edited is annoying.

jtanx:
> I think that URL is screwed up.

Basically this https://github.com/fontforge/fontforge/commit/a1e47320b6411d83e9e1b18710a4f3b38595c3d1

> It assigns the first one it finds free.

>The developer of plugin Foo can't count on their requested mnemonic being available because they can't know what else has been added to the menu before their entry. Some users prefer mnemonics, and this was the way that I came up with that was "stable" (relative to the entries from the start, anyway) and that would assign a mnemonic to every entry.

Yeah, I guess that's where I'd personally find this use questionable/overkill if the mnemonic assigned can vary depending on what's already present

gdraw/gtextinfo.c
```diff
@@ -928,14 +928,16 @@ return(0);
     }
 }
 
-void HotkeyParse( Hotkey* hk, const char *shortcut ) {
+int HotkeyParse( Hotkey* hk, const char *shortcut ) {
```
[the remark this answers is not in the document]
skef: Yeah -- false means (roughly) "couldn't parse".

skef:
If it's questionable/overkill what is the alternative? Is it preferable to have a mnemonic or not have one depending on what's already present?

jtanx:
Basically just try to set it and give up if it can't be set, same as what happens if you assign more than hard-coded mnemonic. That is as far as I'm aware, consistent with how Windows deals with duplicate/clashing mnemonics (and I assume a similar story for other toolkits).

skef:
> Having said that, reading the gdraw code I do believe it was done with the intention of supporting more than ascii, as per how utf82u_mncopy extracts unicode code points, as are mnemonics in terms of code points, and GK_Special used in some places to check if keysyms are unicode (a flawed check imo).

Do other programs do this? Do users of non-Latin scripts expect or want non-Latin accelerators? I was under the impression that the current FontForge convention follows normal practice.

jtanx:
I don't see why not, it is entirely possible to not have a Latin keyboard at all.

Just tested on Windows and it works fine using a Czech keyboard layout:
image

skef:
> Basically just try to set it and give up if it can't be set, same as what happens if you assign more than hard-coded mnemonic. That is as far as I'm aware, consistent with how Windows deals with duplicate/clashing mnemonics (and I assume a similar story for other toolkits).

In most cases a conflicting mnemonic is a configuration error that would be fixed by changing the sequence.

@frank-trampe, @ctrlcctrlv and I had an audio conference before I started on this work and we settled on this approach. The code may be more complicated but it's implemented. The user can still override the mnemonics if they want to. I don't see how its a sin to put one there otherwise.

skef:
> I don't see why not, it is entirely possible to not have a Latin keyboard at all.

So then it should be fairly easy to turn up a few examples, thus actually answering the question.

jtanx:
> I don't see how its a sin to put one there otherwise.

I guess my thinking is, apart from what I've already stated, this is now 100% FF specific behaviour that can't be found in any other toolkit, and also doesn't really follow the principle of least surprise (instead of assigning/trying to assign the given mnemonic, it chooses some other)

jtanx:
> So then it should be fairly easy to turn up a few examples, thus actually answering the question.

I just did a search on https://github.com/microsoft/vscode-loc with 

```
rg '\&\&[^a-zA-Z0-9]' --type json
```

And yes there definitely are. e.g. https://github.com/microsoft/vscode-loc/blob/main/i18n/vscode-language-pack-tr/translations/main.i18n.json#L1407

skef:
> I guess my thinking is, apart from what I've already stated, this is now 100% FF specific behaviour that can't be found in any other toolkit, and also doesn't really follow the principle of least surprise

1. plugin/script-contributed menu items are already part of FontForge (the latter for a long time) and those are quite rare already. I'd rather use such cases as the comparison class for additional functionality, as those are what face the same problems.

2. The posit of a violation of the principle of least surprise seems true on a technical level at best. Having your mnemonic go missing due to an order change seems surprising. Having the mnemonic change under the same circumstances also seems surprising. I suppose the latter may be slightly more surprising given the extra "content" but it seems dubious that that's what the principle is intended to adjudicate.

skef:
Wait, why are we using Czech as an example of non-Latin? OK: It's possible to use accents. That's not a central concern relative to our discussion. The question is whether there are Greek, Cyrillic, etc accelerators.

skef:
OK, the Russian example is better.

jtanx:
> 1. plugin/script-contributed menu items are already part of FontForge (the latter for a long time) and those are quite rare already. I'd rather use such cases as the comparison class for additional functionality, as those are what face the same problems.
> 2. The posit of a violation of the principle of least surprise seems true on a technical level at best. Having your mnemonic go missing due to an order change seems surprising...

Sure, but mnemonic functionality is not specific to FontForge, and as far as I know, nothing else will pick a new mnemonic for you if you have a clash. Having a clashing mnemonic should be fairly easy to see and reason about. If we're saying that such scripts are already quite rare, then I find it even more unlikely that there would be a clash in the first place.

skef:
I'm not saying that scripts are going to be rare -- the whole point of this plugin push is to increase the number of scripts available. I'm saying that programs with plugin interfaces that add menu items are rare. If you want to argue about these conventions then that is the proper comparison class.

skef:
I'm not going to have time to overhaul this PR in the short term. I'm not giving up on the design just because you see it as overwrought. It shouldn't be changed without reinvolving the interested parties (@ctrlcctrlv may not care anymore). So I guess you can feel free to stop your review.

skef:
>  Having a clashing mnemonic should be fairly easy to see and reason about.

For *who* to reason about? Not all users are the sort to read the docs and break out their ASCII editors to configure their own shortcuts and mnemonics. Not all users even know how to use an flat file ASCII editor, even those who prefer mnemonics. Your answer just punts to the end user.

jtanx:
> I'm saying that programs with plugin interfaces that add menu items are rare.

I mean, even by this logic, there shouldn't be all that many entries in the menu and so the chance of a collision is relatively rare.

> For who to reason about? Not all users are the sort to read the docs and break out their ASCII editors to configure their own shortcuts and mnemonics. Not all users even know how to use an flat file ASCII editor, even those who prefer mnemonics. Your answer just punts to the end user.

Maybe I'm missing something entirely, but is the script author not already choosing the mnemonic via the leading underscore? Yes, it is a punt, but no more so than what all other toolkits would do.

In terms of reasoning, like it's pretty clear to see that say this happened: 
image

You can see that both have the same mnemonic (o), and if you do activate the mnemonic, quite predictably, it chooses the first that appears (Reencode). That to me is easy to reason about.

> I'm not going to have time to overhaul this PR in the short term. 

Yep, I'm happy to leave this here for now.

skef:
> I mean, even by this logic, there shouldn't be all that many entries in the menu and so the chance of a collision is relatively rare.

I don't see how this follows, unless you're assuming global knowledge on the part of plugin authors. 

>  Yes, it is a punt, but no more so than what all other toolkits would do.

You're just continuing to ignore how this case is different from other cases.

jtanx:
> I don't see how this follows, unless you're assuming global knowledge on the part of plugin authors.
> You're just continuing to ignore how this case is different from other cases.

Is this wrt. not having a built-in default/not knowing in advance what mnemonics would be used?

OK, even assuming that there will be lots of menu entries, and potentially lots of clashes, I'd still say that assigning a random (to the user at least, no one is going to look into what algorithm it uses) mnemonic is going to cause more trouble than it's worth. - given it has the potential to vary between users and even depending on what plugins are used/enabled at the time.

At best, it is maybe usable *supposing* that key is available on the user's keyboard *and* supposing they keep exactly the same set of plugins/calls, and at worst, confusing if it differs between people(+enabled plugins), to unusable if the key doesn't even exist.

Outside of a complete redesign of everything, I still think a best-effort assignment remains the most practical. 

For comparison (I think I saw something similar), with vscode extensions (my memory is a little rusty), you assign a name to an action (like `extension.createTerminal`) and you can suggest a default shortcut, but is overridable in user-preferences (it doesn't auto-generate a new shortcut on clash though...).

skef:
>  is going to cause more trouble than it's worth

Trouble in the form of ... cognitive dissonance? Are you thinking of filed issues? Aren't missing mnemonics likely to result in filed issues as well? What is the form of this "trouble"? 

> At best, it is maybe usable supposing that key is available on the user's keyboard and supposing they keep exactly the same set of plugins/calls, and at worst

Let's at least attempt to keep the speculative apart from the current or expected. We're currently talking about Cyrillic, right? Because those are the only .po files we actually have that don't use Latin characters for the normal menus. So it's not like this is an obscure or unsupportable supposition. It would be pretty easy to special-case Cyrillic.

> I still think a best-effort assignment

Strange to use the term "best effort" to argue for removing effort. 

> For comparison (I think I saw something similar),

That is in the class, although given that it's a tool specifically for people who author and modify UTF-8 files for a living it's not the ideal reference.

skef:
> It would be pretty easy to special-case Cyrillic.

Indeed, given that this is the string: `"}{:[]&=~%^;?+/,#$-@*!.7869054321QXZVJWFYGKUBHPCDMLTNSIROEA"`, it would be even easier just to knock off the capitals at the end. Do common keyboards lack any of those characters. (Or, rather, which characters do some common keyboards lack, so I can remove those as well?)

jtanx:
> Trouble in the form of ... cognitive dissonance? Are you thinking of filed issues? Aren't missing mnemonics likely to result in filed issues as well? What is the form of this "trouble"?

I tried to expand this in the next paragraph, but to be clear:

* You usually associate muscle memory to a given mnemonic-action
* That the assigned mnemonic can change depending on what you load breaks that
* Comparing between users can become difficult (e.g. "I press alt+a" but for someone else it's "alt+b"

> It would be pretty easy to special-case Cyrillic.

That's digging an even deeper hole that I'm not sure I want to go to. 

> Strange to use the term "best effort" to argue for removing effort.

That's kind of twisting the meaning, it's aligning it with what I think is the more common expectation/pattern, where I don't want to see as much FF specific UI peculiarities.

> That is in the class, although given that it's a tool specifically for people who author and modify UTF-8 files for a living it's not the ideal reference.

I mean, it comes with a visual shortcut editor which I'm fairly sure anyone can use.

skef:
The first and second bullet are pretty much the same thing and the design already accounts for this. When new plugins are discovered they're added to the end of the list by default. The list determines initialization order. The initialization order determines the menu registration order (assuming tools follow the recommended convention of one Tools menu entry with a submenu -- if they don't and a tool adds a menu item it will come in that order). The assigned mnemonic is deterministic relative to a given sequence of menu registrations. So unless the user changes the ordering by hand the mnemonics won't change (if the suggested convention is followed). A user who changes the order and doesn't like the result can always change it back.

The third bullet seems like a stretch. Users with English and French locales can't compare either. They'll be the same when the requested mnemonic is assigned. On your suggestion one or another may be missing, which could leave these theoretical menu-comparing users just as confused.

> That's kind of twisting the meaning, it's aligning it with what I think is the more common expectation/pattern, where I don't want to see as much FF specific UI peculiarities.

I would say you just chose the wrong term. "Best effort" more or less means what it sounds like it means. 

> I mean, it comes with a visual shortcut editor which I'm fairly sure anyone can use.

OK, then that's something FontForge doesn't have.

jtanx:
Yes, I realise this algorithm will assign a deterministic mnemonic for a given load order. Regardless, that it now depends on said algorithm is exactly that which I have been arguing against.

> A user who changes the order and doesn't like the result can always change it back.

Who first has to realise that the cause is a clashing mnemonic.

> The third bullet seems like a stretch. Users with English and French locales can't compare either. They'll be the same when the requested mnemonic is assigned.

Yeah, but this is adding yet another dimension that I just don't think is worth - the exact set of plugins/scripts loaded at the time. Every other case is easy to infer by inspection, instead of running the algorithm you've come up with.

> I would say you just chose the wrong term. "Best effort" more or less means what it sounds like it means.

No, best effort describes it well. I meant in terms of comparing to 'removing effort'. Just because effort has been spent doesn't always mean it's worthwhile or the best approach (sunk cost fallacy anyone) - I've been guilty of this in the past myself.

frank-trampe:
@jtanx, I don't understand why the fallback mechanism is so objectionable. It may not be standard, but it seems like a useful feature for the sort of lightly curated ecosystem we're trying to support. I would guess that most type designers do their heavy work solely on one machine, so that sort of variation seems non-problematic to me. As for the possibility that installing a new script changes a hotkey, that's not significantly more problematic than it voiding a hotkey, and the fix has the same level of complexity.

jtanx:
The objection comes from having to support this highly specific semantics/code (only applicable to mnemonics on the tools menu) going forward, on something that should be fairly generic/have uniform behaviour across apps.

frank-trampe:
The mnemonic improvisation logic (almost completely contained in `FindMnemonic`) does not constitute a significant portion of the total complexity of the change. If it turns out to be problematic in real-world use, we can just disable the fallback mechanism. I still think that this provides a vastly better experience than forcing users to edit third-party user scripts. And I don't think that it's all that deviant. It does automatically what a menu designer would do by hand. The fact that there's not an actual person doing it does not change the validity of the process.

frank-trampe:
The L-1011 shipped with exterior doors that lifted up and out of the way (balanced by springs). The feature was unique at the time and never widely adopted, but it was handy and stylish. If we say that this feature is unique, I see it in the same light as the L-1011 doors. The few people who notice it are going to be mildly impressed rather than befuddled.

skef:
I pushed my proposed changes minus the preferences support: You should now be able to specify any sequence of Unicode characters as Tools menu mnemonics by internationalizing the default string. 

I would have liked to add this as a preference but the semantics of that system make it difficult. Ideally opening the window would display the default string for the current initialization but nothing would be saved unless the user changes its value. I can't figure out how to implement that in the current system and I'm not inclined to do yet more special-casing (e.g. with BDFFoundry now).

frank-trampe:
I remain in favor of merging this and believe that the increase in utility overwhelmingly offsets the increase in complexity. @jtanx?

jtanx:
I'll take a look at this again in the next couple of days

skef:
I've updated this with @jtanx's suggestions.

doc/sphinx/scripting/python/fontforge.rst
```diff
@@ -511,58 +514,130 @@ Not a very useful example.
...
+      the current interface.
 
-   .. object:: menu-name
+   .. object:: name
 
-      The name that will appear in the menu for this item. This will only affect
-      windows created after this command is executed. Normally the command will
-      be executed at startup and so it will affect all windows.
+      ``name`` can be a string but ideally it is a tuple of three strings
+      ``(localized_name, english_name, identifier_string)`` or of two strings
```
jtanx, line 585: Why is it necessary to specify the English name? The hotkey system (questionably) looks for shortcuts based on the untranslated text from the hotkeys file, so I can somewhat understand that, but all of these hotkeys are 'volatile' in that they are registered every time this function is called. Perhaps I'm missing something?

fontforgeexe/pythonui.c
```diff
@@ -50,41 +51,211 @@
...
+	if ( g_hash_table_contains(pmd->mn_avail, GUINT_TO_POINTER(alt)) ) {
+	    g_hash_table_remove(pmd->mn_avail, GUINT_TO_POINTER(alt));
+	    return alt;
+	}
+    }
+    return 0;
+}
 
-GMenuItem2 *cvpy_menu, *fvpy_menu;
+static unichar_t StripMN(unichar_t *str) {
```
jtanx, line 176: Feels like utf82u_mncopy should just be used instead

jtanx (commented):
Been a while since I last looked at this, but I still feel like the mnemonic stuff in particular is overcomplicating matters when https://github.com/fontforge/fontforge/pull/4498#issuecomment-869102611 would get you 95% of the way there, excluding the auto-substitution of a free mnemonic. 

I can get behind the rest, particularly the hotkey support, but just some questions on the need to provide both localised and non-localised strings

doc/sphinx/scripting/python/fontforge.rst
jtanx: Alternatively if it is somehow a requirement for hotkeys, why not just use the identifier (which imo, is what should have been done for all hotkeys; assigning hotkey identifiers instead of relying on untranslated strings)

skef: The hotkey specified at registration time is just a suggestion -- it has lowest priority compared with FontForge code hotkeys (which may be added to over time) and user-defined hotkeys.

The English string is for consistency in the user overriding (or just choosing -- I expect most registered menu items will skip the hotkey suggestion) the choice using their `hotkeys` file, which (for reasons that can be questioned but probably aren't changing anytime soon) are always overridden with the English/code-native strings.

skef: The reason not to use the identifier is consistency with the existing infrastructure, which doesn't use identifiers anywhere else.

jtanx: Given that the identifier is required to be provided, can that not be used instead?

skef: It would be an option to half-overhaul the hotkey system now, but that's not a road I went down. Do you have a suggestion with rough added code points, keeping in mind the Peculiarities of the existing hotkey system (like the initial View prefix)?

jtanx: I can understand elsewhere where there is no explicit need to set an identifier, because the untranslated string is implicitly the identifier, but in this case, you are providing both the untranslated string and an identifier - where the former is only ever used as an identifier anyway.

jtanx: Sorry, seems like I've missed these comments until I refreshed.

I'll take a closer look but I was under the impression that you could just sub the identifier for where it requires the untranslated text.

skef: I suppose I'm OK with getting rid of the identifier if necessary. I added it because it seemed preferable to eventually have a menu system based on identifiers and I thought it would be best to encourage plugin developers to add one for when we get there.

skef: Yeah - github doesn't seem to be refreshing like it used to.

jtanx: To be clear, I think having identifiers make sense, but I'm proposing to just specify `(text, identifier)`, instead of having `(localised_text, untranslated_text, identifier)`

skef: I understand, but I'd prefer (localized_text, untranslated_text) being merged to (text, identifier) being deferred to some future date, given that my adding `identifier` was just future-proofing.

jtanx: It's getting late and I'll spend more time trying this out over the next few days, but my initial impression is that it's pretty buggy.

It seems very hit and miss whether or not a hotkey will 'register', and there's absolutely no feedback if a hotkey is misregistered or anything in the process fails. I've been looking at it with the debugger attached and apparently I need to spend even more time to figure out how to use this, which isn't promising. Apparently a prerequisite to having hotkeys is to have an enable function? But even then half the time it doesn't work.

I've also reinforced my feeling that one of the untranslated/identifier has to go, and I'm very heavily preferring untranslated to go. As to why:

* Localisation is punted to the user - _they_ are the ones providing the localisation, not us. If someone barely knows English and only bothers to have the menu name in their language, why would they bother to translate that to English? More than likely I could see random text being inserted here just to satisfy the requirement
* Even more so, what's the point of providing a translation in two languages if one is never seen to the end user?
* Very little sanitisation is done on the untranslated text. It becomes trivial to write something that would mess up the way hotkey identifiers work (think: adding stray `.`s or `:`s into the string
* At the same time, you _could_ enforce that an identifier is alphanumeric plus maybe underscores/spaces and that seems like a perfectly valid restriction - the same of which cannot be said to what's meant to be a human readable string (even though it's only used for identifier purposes).

skef: Am I right that these issues are specific to the hotkey functionality? (So the "buggy" isn't a more general comment about the PR, but the code that is supposed to allow hotkeys to work again (as a suggestion)?)

I didn't have this sort of problem with hotkeys in general in my testing so I'll try to take another look soonish. 

As far at the identifiers-only route goes, I'm not against in principle but is the implementation clear? E.g. how will the menu item wind up with the hotkey printed if there's no association with the menu hierarchy?

jtanx: > Am I right that these issues are specific to the hotkey functionality?

No, just general feeling of registering/use of this functionality overall; I suspect some of it may be long-standing, but it makes testing/using this harder.

Had a bit more time to go through this - bugs (or surprising behaviour):
* (Apparently longstanding): 'Font' context menu items are not added/enabled until a new fontview is created (either by opening a new font or creating a new one)
* The 'Tools' menu is permanently in the 'Enabled' state, even when it has no items
  Old: image
  New: image
* Auto-assignment of mnemonics only happens at the first level, not the submenu
  image
* If I rerun the same registration code, I end up with two distinct entries instead of overwriting the first. Sample
  ```py
  fontforge.registerMenuItem(print, None, None, "Glyph", "Ctrl+Alt+6", "Sub_Menu", "Print Glyph Name")
  ```
* Assigning a hotkey only works if I omit mnemonic underscores, otherwise no such hotkey is assigned:
  ```py
  fontforge.registerMenuItem(mycb, None, None, "Glyph", "Ctrl+Alt+6", "SubMenu", "Print Glyph Name")
  ```
  (Compare this, which works, to the example above, which does not)
* Non-mnemonic parenthesised character is stripped:
  ```py
  fontforge.registerMenuItem(print, None, None, "Glyph", "Ctrl+Alt+7", "ABC (1)")
  ```
  image


> As far at the identifiers-only route goes, I'm not against in principle but is the implementation clear? E.g. how will the menu item wind up with the hotkey printed if there's no association with the menu hierarchy?

I don't think it's too bad, set the identifier as the untranslated text for the menu item

iterumllc: > (Apparently longstanding): 'Font' context menu items are not added/enabled until a new fontview is created (either by opening a new font or creating a new one)

The font menu context problem mirrors most of the rest of the UI: If you change things you typically don't see the result until the next time you open a window.

> The 'Tools' menu is permanently in the 'Enabled' state, even when it has no items

I noticed the tools menu thing. I'll try to fix that. 

> Auto-assignment of mnemonics only happens at the first level, not the submenu

The auto-assignment of mnemonics is intentionally limited to the first level, as noted in the earlier discussions. A sub-menu "collision" is of course possible but for the most part plugin developers will add their own independent submenus (and are encouraged in the plugin documentation to put additions in a submenu). When you know the contents of a submenu there's no possibility of collisions and therefore no need for auto-assignment. 

> If I rerun the same registration code, I end up with two distinct entries instead of overwriting the first. Sample

Having  unregister/reregister work better would be nice but seems unnecessary for this PR given that there's no change.. But I'll look at it.

> Assigning a hotkey only works if I omit mnemonic underscores, otherwise no such hotkey is assigned:

I think is how hotkey assignment works in general. I suppose we could remove underscores in the passed string to improve things, as long as we were confident they aren't being removed in names. (Or I guess we could pass the passed string through one of the `mncopy`s?)

> Non-mnemonic parenthesised character is stripped:

I can try to fix this.

I'm a bit busy before Wednesday and maybe next weekend so I may need until then to look at what I've said I will.

jtanx: > The font menu context problem mirrors most of the rest of the UI: If you change things you typically don't see the result until the next time you open a window.

Sure, I'm just saying that it's not intuitive or immediately apparent. But given it's longstanding it's fine to leave.

> The auto-assignment of mnemonics is intentionally limited to the first level, as noted in the earlier discussions.

My points from my prior comment remain the same

> Having unregister/reregister work better would be nice but seems unnecessary for this PR given that there's no change.. But I'll look at it

I'm not quite sure I understand this, but as is, there is no way to add more than one entry to a submenu, which is broken and needs fixing before this can be merged. It's at least a definite regression from behaviour prior to this PR

> I think is how hotkey assignment works in general. I suppose we could remove underscores in the passed string to improve things, as long as we were confident they aren't being removed in names. (Or I guess we could pass the passed string through one of the mncopys?)

I thought the mnemonic-stripped string was already being passed to the hotkey system, although admittedly I didn't look too deeply into the exact cause

iterumllc: > I thought the mnemonic-stripped string was already being passed to the hotkey system, although admittedly I didn't look too deeply into the exact cause

`loadHotkeysFromFile()` leaves `action` as-is and so does `hotkeySetFull()`, as far as I can tell. I think to set hotkeys at present you need to strip the underscores yourself. (I suppose we could also do something more clever than `strcmp()` in `HKActionMatchesFirstPartOf()`.)

skef: > The 'Tools' menu is permanently in the 'Enabled' state, even when it has no items

> If I rerun the same registration code, I end up with two distinct entries instead of overwriting the first.

> Non-mnemonic parenthesised character is stripped:

I think I've fixed the first two and greatly (greatly) reduced the chance of the third.

Take a look and also let me know where we are with this after these changes. The window-not-updating and hotkey-file-with-underscores issues seem consistent between normal and python menus. I think I've dealt with what I have to but you might disagree.

> >  As far at the identifiers-only route goes, I'm not against in principle but is the implementation clear? E.g. how will the menu item wind up with the hotkey printed if there's no association with the menu hierarchy?

> I don't think it's too bad, set the identifier as the untranslated text for the menu item

Menus exist in a hierarchy and identifiers don't. (I haven't thought much about submenu-pointing-entries and identifiers. But they're normally not needed or used that way.)

jtanx: > Menus exist in a hierarchy and identifiers don't. (I haven't thought much about submenu-pointing-entries and identifiers. But they're normally not needed or used that way.)

I'm not quite sure I get the implication, but ok.

jtanx:
So I still find some of the changes made questionable, and it is more complex than it needs to be, but functionally it works, and I'm not going to block the release for this. Worst case we can always change the interface down the line.

PR #4686, opened 2021-03-28
Don't add leading control points twice when reading glif

### LINKED ISSUE #4539: Another UFO glyph can't be roundtripped, this time a quadratic one
Arch Linux, fc1970174ba9e965bef9b676ee2afd679079a34e

I haven't yet solved #4327 on my own, and now have an example of a quadratic glyph that FontForge also can't generate a correct UFO of. :frowning: 

Non-FontForge programs also have problems with this UFO. For example `fontmake` can't convert it (crashes).

# SFD

2020-12-17-210237_1506x1392_scrot

# Conversion

```python3
import fontforge

f = fontforge.open("Issue4327Part2.sfd")
f.generate("{}.ufo".format(f.fontname))
```

```
$ fontforge Issue4327-Part2.ufo
```

# UFO

2020-12-17-211231_2560x1392_scrot

# Files

[Issue4327Part2.zip](https://github.com/fontforge/fontforge/files/5713253/Issue4327Part2.zip)

skef:
Closes #4539

### DIFF
--- a/fontforge/ufo.c
+++ b/fontforge/ufo.c
@@ -2909,7 +2909,7 @@ static SplineChar *_UFOLoadGlyph(SplineFont *sf, xmlDocPtr doc, char *glifname, 
 					wasquad = true;
 				if (ss->first == sp) {
 				  firstpointsaidquad = true;
-				}
+				} else {
 					if ( precnt>0 && precnt<=2 ) {
 						if ( precnt==2 ) {
 							// If we have two cached control points and the end point is quadratic, we need an implied point between the two control points.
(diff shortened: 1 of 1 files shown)


frank-trampe:
I'm having trouble thinking through this one without tests. Would it perhaps be more intuitive simply to expend the cached points by setting `precnt` to zero?

skef:
I guess you could *also* do that. But at least to the extent I've looked at it it does seem like there's a "do this if first, do the other if not first" branch here.

skef:
The code around line 3026 corresponds to the code around line 2913, you just need to make sure the right one runs and the wrong one doesn't.

frank-trampe:
Oh, I see.

frank-trampe (approved):

PR #4690, opened 2021-03-28
MacOS Big Sur App Icon

### LINKED ISSUE #4574: Update app icon for macOS Big Sur

### Important

- [ ] Screenshot

image

I would love for the Mac app to fit in Better with macOS big sur icons by switching the app icon from a circle to the macOS app shape (rounded rect).

JaydenIrwin:
Changed the macOS app icon (.icns) from a circle to the Big Sur app shape.

### Type of change
- **New feature**
- **Non-breaking change**

Closes #4574


skef:
If you try this in your App does it work well?

JaydenIrwin:
Not sure how. :(

skef:
This may have changed in more recent versions of MacOS but within the past few years you could "right-click" (secondary-click) on an app in the Finder and choose "Show Package Contents" and then you could descend into the App folder. The icon is in a location analogous to where it is in our source tree.

JaydenIrwin:
Looks great thanks!
Screen Shot 2021-03-28 at 15 30 56

skef (approved):
Don't see any reason not to approve this. I'll let it sit for a bit before merging in case anyone else has an opinion.

Thanks for this!

ctrlcctrlv (changes requested):
Why did you change the appearance of the icon? Now it doesn't look the same as on the splash screen, or on different OS's.

If you want to change it to be rectangular that's fine, but you should keep the shape of the internal icon as is.

image

skef:
Mac OS restyles their icons every few years and right now they're rounded rectangles. Apps often follow suit. You can see from the screenshot how the author has matched up the edges of the App icon with the other current icons.

The image is the same, I don't see any reason why the *border* has to be consistent between different OSes.

skef:
Oh, yes, I guess I agree with this. The illustration itself should be the same. 

@JaydenIrwin would it be possible for you to use either `desktop/tango/scalable/org.fontforge.FontForge.svg` or the pngs in the WxH directories in `desktop/tango` as the source material for the inside? I guess I should have pointed you towards those in the first place.

JaydenIrwin:
Yeah I can change it. I just used the previous Mac icon as the inner shape.

skef:
Ha, sure enough! Here we are yelling about "changing the appearance" when the Mac icon was like that in the first place.

JaydenIrwin:
Should be good now.

skef:
Looks good!

JaydenIrwin:
Screen Shot 2021-03-31 at 13 01 32

JaydenIrwin:
Screen Shot 2021-03-31 at 13 07 58

Issue #4691, opened 2021-03-29
Export of a glyph as a SVG file fails on Windows 10


StefanPeev:
I tried to export a glyph as a SVG file both through FontForge 20th Anniversary Edition and AppVeyor latest version and the results are:

- in Windows 7 - export is possible
- in Windows 10 - I receive a message "Save Failed"

What may be the reason for such results?

jtanx:
Please post a sample file. Just guessing, do you have permisions to save to the location you're trying? What if you try to save somewhere else?

skef:
Hard to say. The most likely candidate for the problem is the `fopen()` on line 392 of `fontforge/cvexport.c`. If you're in a position to compile you try changing that function to:

```
int ExportSVG(char *filename,SplineChar *sc,int layer, ExportParams *ep) {
    FILE *svg;
    int ret;

    svg = fopen(filename,"w");
    if ( svg==NULL ) {
        perror("SVG fopen error is:");
return(0);
    }
    ret = _ExportSVG(svg,sc,layer,ep);
    fclose(svg);
return( ret );
}
```
to see what it says.

StefanPeev:
> What if you try to save somewhere else?

The result is the same - Save Failed.

StefanPeev:
@skef No, I am not in position to compile... sorry.

skef:
> No, I am not in position to compile... sorry.

No shame in that—we aren't either. 

Once we get our ducks more in a row I'll try to remember to submit a draft PR and get you a build artifact to try out.

skef:
@StefanPeev See if you can unpack this (with archiver 7z) and make use of it: https://github.com/fontforge/fontforge/suites/2389141540/artifacts/50759798 . I added code analogous to`perror` above. If the problem reproduces we'll have a bit more information and I can add further diagnostics as needed. If not we can try to figure out why.

StefanPeev:
@skef Thank you! The ZIP archive MINGW64.zip is the decision. The problem is solved in the included in the ZIP archive version of FontForge.

skef:
That's not necessarily good news -- it could mean there's something about our usual packaging that's causing the problem. 

@jtanx any ideas? Should he try MINGW32.zip?

StefanPeev:
@skef I'm affraid that one of my students reported this error on Windows 10 even with FontForge in MINGW32.zip - see the picture...

[...]

PR #4695, opened 2021-03-31
Diagnostic perror for @StefanPeev

skef:
This is just a Draft PR to produce a diagnostic windows binary for #4691


ctrlcctrlv:
Interesting technique :-)

Issue #2570, opened 2015-11-14
Unicode "alternates" situation is a mess


mskala:
I'm continuing to analyse how FontForge uses its "unicode_alternates" array; this is part of the same effort that yielded #2566.

The unicode_alternates array is a multi-level lookup table that maps characters in the Unicode Basic Multilingual Plane (that is, between U+0000 and U+FFFF; the subset of Unicode that fits in 16 bits) into sequences of such characters.  This table lives in Unicode/unialt.c, which is roughly 9000 lines long and not designed to be maintained by hand.  Instead, one is supposed to run the makeutype.c program, which generates the code based mostly on the UnicodeData.txt file that can be downloaded from the Unicode Consortium.

For *some* characters, unicode_alternates apparently stores characters that George thought looked similar.  For instance, the entry for U+0391 (Α - that is a Greek capital alpha) consists of U+0041 (A - that is a Latin capital ay).  Several different kinds of official Unicode equivalence are merged into this concept by the program makeutype.c, but that program also overrides the Unicode tables with hardcoded extra rules.  Some of these rules are easier to defend than others; U+0391 to U+0041 are likely to really look the same and be interchangeable, but U+1D43 (ᵃ - that is a superscript "a") is mapped to U+0061 (a - that is an ordinary "a"), and substituting one for the other is unlikely to please users.

For *some* characters, this array instead stores a decomposition.  For instance, the entry for U+00C0 (À - capital ay with grave accent) is the two characters U+0041 U+0300 (A followed by a combining grave accent).  Again, some of these decompositions come from Unicode, but several different Unicode concepts are all merged into this single table and there are abundant exceptions and overrides hardcoded into unialt.c.  There is no way to tell the difference between the different kinds of decomposition, and the only way to tell the difference between decompositions and characters that look similar is to check how many characters are in the entry.

Although the unicode_alternates array by its design is not capable of representing any characters outside the BMP as indices, just a few of its entries mention non-BMP characters (U+FA6C maps to U+242EE, for instance).  That doubles the size of the whole thing by forcing the use of 32-bit instead of 16-bit integers.

[...]

PR #4714, opened 2021-04-26
Overhaul the unicode support library

### LINKED ISSUE #2566: Probable out-of-range in unicode_alternates array
From https://github.com/fontforge/fontforge/blame/master/fontforge/splinesaveafm.c lines 1294 to 1300:

``` c
if ( isdecompositionnormative(sc->unicodeenc) && sc->unicodeenc>=65536 &&
        unicode_alternates[sc->unicodeenc>>8]!=NULL &&
        (upt = unicode_alternates[sc->unicodeenc>>8][sc->unicodeenc&0xff])!=NULL ) {
    while ( *upt!='\0' )
        unicode[u++] = *upt++;
} else
    unicode[u++] = sc->unicodeenc;
```

Note that this only ever accesses the unicode_alternates array if sc->unicodeenc>=65536 is true, and in that case, the index is necessarily 256 or greater.  This array's valid range of indices is 0 to 255, so it can't possibly be right.  The code is from a March 2006 commit by George Williams.

At first glance it looks like that test needs to be inverted; it's asking whether there is valid data in the array and so it should be sc->unicodeenc<65536.  I'm hesitant to try to change it, though, because I don't really understand how the information from this array is intended to be used anyway.

jtanx:
This is a continuation of the charsets/iconv overhaul, extending to the code-genned stuff under the `Unicode` folder.

The original motivation was to shrink the size of the code-genned data (which used plain old arrays covering only the BMP), and also to make the process of updating this support layer not so cryptic/painful. To update now consists primarily of running one Python file (`makeutype.py`), and it will do everything for you, including fetching all required data.

But I think there are several more benefits that were picked up along the way as a consequence.

Primarily:

* The unicode functions now supports the full unicode range, not just the BMP(!!)
* It uses multi-stage lookup tables (same as what Python does) to reduce table sizes
* NamesList support is fully integrated
* There is minimal manual intervention required to update this layer
* Updated to Unicode 13.0.0

In terms of space savings, the complete layer, *including* the NamesList integration is around ~700kb, which comes in *under* the original size of the old Unicode implementation (~1.5-2mb), which excluded NamesList support. This is also all done with constant-time lookup tables, so negligible difference in speed (not that that was/is a concern anyway).

I was originally going to make the nameslist support optional (allowing to still use libuninameslist), but that ended up being more effort than it was worth, and this also aligns with something that has been discussed prior anyway (see also #3861). Basically if I can get this functionality in, in under the space used previously, I don't see why this should ever be disabled. This integration also makes sense in the fact that it actually matches up with the version of Unicode that FontForge has been built with.

-----

While this is all well and good, I have also removed some functions: namely those relating to ligatures/fractions. As to why: when you sit down and look at what it's doing, none of it is actually well defined in terms of normative Unicode properties. The decompositions provided by is_Ligature.c is primarily derived from the NFKD sequences, which are _already_ recorded in unialt.c. 

What remains are manual 'decompositions', most of which I find questionable - i.e. the [first one I come across](https://github.com/fontforge/fontforge/blob/a5dedb4010cd49a5fcfaeed5d188dd7942294005/Unicode/makeutype.c#L163), suggests that U+04a4 is composed of U+041d, and U+0413, but if you [read up](https://en.wikipedia.org/wiki/En-ghe) on what that character provides, you realise that it cannot actually be composed of either.

The exposed functions are also extremely low level, to the point of exposing the internal implementation details of how the data is organised (in tables, this is also true for the poorly named 'Names2' functions, which really only just extract the formal alias from the annotation).  Prove me wrong, but I have yet to see _any_ living example of scripts that make use of these functions.

-----

### Type of change

- **Bug fix**
- **New feature**
- **Breaking change**

Closes #2566

See also: #2570, although this doesn't really aim to address the issues laid out there.


khaledhosny:
Not sure if it is of any use to you, but HarfBuzz uses https://github.com/harfbuzz/packtab to generate compact Unicode data tables.

jtanx:
Ah, interesting. I think I saw the old packtab (I think glib uses something similar?) but didn't know about that one. I currently just use the one that Python uses, which does an ok job.

jtanx:
Any further comments? Otherwise I'd like to merge this over the coming days

ChiefMikeK:
> Any further comments?   
> Otherwise I'd like to merge this over the coming days
  
Just a FYI Unicode 14.0 final draft is out for public review and reliese soon 😁

jtanx:
Yep, I'll update it once it's actually released in September.

jtanx:
Final call for comments? Otherwise I'll merge tomorrow.

PR #4721, opened 2021-05-08
Initial fuzzing integration

DavidKorczynski:
Hi, 

I would like to set up continuous fuzzing of Fontforge by way of OSS-Fuzz. Essentially, OSS-Fuzz is a free service run by Google that performs continuous fuzzing of important open source projects. The only expectation of integrating into OSS-Fuzz is that bugs will be fixed. This is not a "hard" requirement in that no one enforces this and the main point is if bugs are not fixed then it is a waste of resources to run the fuzzers, which we would like to avoid.

In this PR https://github.com/google/oss-fuzz/pull/5723 I have done exactly that, namely created the necessary logic from an OSS-Fuzz perspective to integrate Fontforge. If you would like to integrate, could you please provide a set of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports, coverage reports and more stats. The emails should be linked to a Google account in order to view the detailed reports and notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file.

Let me know what you think

### Type of change

- **Non-breaking change**

### DIFF
--- a/tests/fuzzers/fuzz_pdf.c
+++ b/tests/fuzzers/fuzz_pdf.c
@@ -1,0 +1,25 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdint.h>
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+	// Create a file with the fuzz data
+	char filename[256];
+	sprintf(filename, "/tmp/libfuzzer.%d", getpid());
+	FILE *fp = fopen(filename, "wb");
+	if (!fp) {
+		return 0;
+	}
+	fwrite(data, size, 1, fp);
+	fclose(fp);
+
+    // parse the file
+	NamesReadPDF(filename);
+
+    // cleanup
+	unlink(filename);
+
+	return 0;
+}


skef:
> The only expectation of integrating into OSS-Fuzz is that bugs will be fixed. 

I doubt the project can currently meet this expectation. We don't address many of the lint/code analyzer bugs we get notices of and the way the PDF interpretation code is written would mean it would have to be almost rewritten to deal with arbitrary inputs. Unless you have someone up your sleeve to do that work it doesn't seem like a good idea, at least at this time.

ctrlcctrlv:
I agree, and would also encourage you to read [#D1](https://github.com/fontforge/fontforge/wiki/Community-guidelines#D1). I don't think our goals are aligned.

tshinnic:
Hmm, reading D1, I'm thinking that

"And FontForge does present any network interfaces..."

Is missing a "not"?

frank-trampe:
@tshinnic, that's probably never been fixed because it's so hard to say that it has or lacks network surface area. Now that we removed the OFL stuff, I think that we can say the latter, but I wouldn't bet my life on it.

JoesCat:
Hi @DavidKorczynski, I agree with @skef - there are still a lot of unanswered bugs that need addressing in FontForge. Between Coverity, lint, and this issues tracker, that number is well over 1000 - it is much too early to add OSSFuzzing to Fontforge itself.

Having looked at some other projects, I ran into this blog here:
https://blogs.gentoo.org/ago/2020/04/16/why-i-stopped-fuzzing-research/
To summarize the link above - from the point of view of FontForge - upstream would refer to all the libraries that FontForge depends on to make FontForge work. If these are not solid, then OSSfuzzing would show-up in FontForge when the true problem is a dependant library.
There have been issues (here) that were caused by libraries, eg gtk, cairo, pango.

To build a pyramid, you need a rock-solid foundation, and you build upwards, and likewise if you plan to build a rock solid program you need solid libraries too.
Looking at your Dockerfile for FontForge, it makes sense to get all those libraries checked first, but looking at your OSS fuzzer, some of them don't appear to be checked but probably should.

projects/fontforge/Dockerfile
wget - tested
libjpeg-dev - not sure - untested?
libtiff5-dev - not sure - untested?
libpng-dev - tested
libfreetype6-dev - not sure - untested?
libgif-dev - not tested
libgtk-3-dev - not tested
libxml2-dev - tested
libpango1.0-dev - not tested
libcairo2-dev - not tested
libspiro-dev - not tested
libuninameslist-dev - not tested
python3-dev - not sure - partially tested
ninja-build - tested?
cmake - tested
gettext - not tested
libc-dev - not tested?
liblzma-dev - tested
libffi-dev  - not tested
libselinux1-dev - not tested?
I think targeting these first, could help answer the upstream problem for FontForge

DavidKorczynski:
Thank you all for your comments! I think it's fair to say this is not something for Fontforg, but looking at [D1](https://github.com/fontforge/fontforge/wiki/Community-guidelines#D1) and in particular `The exception would be, if somehow, for example, something done outside of FontForge, e.g. in Firefox, can cause arbitrary code execution in FontForge, and this is somehow FontForge's fault and not the OS's or Firefox's.` it seems there is some threat model of the project and I am sure fuzzing could help here. However, the threat model might be difficult to develop in a pragmatic manner such that you won't be flooded with false positives. I will close this, but if you should see the need in the future for fuzzing then I would be happy to assist in the process.

DavidKorczynski:
> Hi @DavidKorczynski, I agree with @skef - there are still a lot of unanswered bugs that need addressing in FontForge. Between Coverity, lint, and this issues tracker, that number is well over 1000 - it is much too early to add OSSFuzzing to Fontforge itself.
> 
> Having looked at some other projects, I ran into this blog here:
> https://blogs.gentoo.org/ago/2020/04/16/why-i-stopped-fuzzing-research/
> To summarize the link above - from the point of view of FontForge - upstream would refer to all the libraries that FontForge depends on to make FontForge work. If these are not solid, then OSSfuzzing would show-up in FontForge when the true problem is a dependant library.
> There have been issues (here) that were caused by libraries, eg gtk, cairo, pango.
> 
> To build a pyramid, you need a rock-solid foundation, and you build upwards, and likewise if you plan to build a rock solid program you need solid libraries too.
> Looking at your Dockerfile for FontForge, it makes sense to get all those libraries checked first, but looking at your OSS fuzzer, some of them don't appear to be checked but probably should.
> 
> projects/fontforge/Dockerfile
> wget - tested
> libjpeg-dev - not sure - untested?
> libtiff5-dev - not sure - untested?
> libpng-dev - tested
> libfreetype6-dev - not sure - untested?
> libgif-dev - not tested
> libgtk-3-dev - not tested
> libxml2-dev - tested
> libpango1.0-dev - not tested
> libcairo2-dev - not tested
> libspiro-dev - not tested
> libuninameslist-dev - not tested
> python3-dev - not sure - partially tested
> ninja-build - tested?
> cmake - tested
> gettext - not tested
> libc-dev - not tested?
> liblzma-dev - tested
> libffi-dev - not tested
> libselinux1-dev - not tested?
> I think targeting these first, could help answer the upstream problem for FontForge

Thanks a lot for your summary. I think we have a few more projects in OSS-Fuzz, e.g. [cairo](https://github.com/google/oss-fuzz/tree/master/projects/cairo) [selinux](https://github.com/google/oss-fuzz/tree/master/projects/selinux) and I think [libuninameslist](https://github.com/fontforge/libuninameslist) is another Fontforge project?

JoesCat:
If you want to include libuninameslist and libspiro, these have reached stable states, complexity is pretty minimal and are pretty-much ready for testing your OSSfuzzing if you want to do so.
Libspiro is also known to be used with GIMP and IPE, so that is a good candidate worth trying.

Locations:
https://github.com/fontforge/libuninameslist
https://github.com/fontforge/libspiro

PR #4743, opened 2021-05-30
Update index.rst

sherbst:
Somehow these images are reversed from how they should be.

### Type of change

- **Bug fix**


jtanx:
thanks!

PR #4749, opened 2021-06-11
Korean translation corrected

### LINKED ISSUE #4748: Ongoing of Korean Translation Issue

### Important

- [ ] Screenshot

I've tried to correct Korean Translation of /po/ko.po because its result was so rough and weird. Although some people are helped, a part of results have used a machine translation result from Japanese translation. It seems to be correctly as soon as possible, I forked fontforge git from master then it fixed to my github location. ([here](https://github.com/hurss/fontforge/blob/master/po/ko.po))
If you want to approve it, you may can this file push to fontforge master.

ctrlcctrlv:
Opened PR at the request of @hurss in #4748.

Closes #4748.

### Type of change

- **Translation update**


jtanx:
The main source is on crowdin, so we'll need to find a way to sync it between here and there.

ctrlcctrlv:
I don't speak Korean but checked a few of the new strings with a dictionary/machine translator and this change seems very positive to me. @hurss is Korean so I trust their overall judgment, if they say that someone cheated previously and tried to pass off machine translation from Japanese as human translation I believe them. It happens sometimes unfortunately because maintainers don't speak every language on Earth so must trust submitters to a degree, which cannot be helped.

hurss:
Currently, I trying to correct Korean translation script of fontforge. When I have fixed it, I'll push it to my repository.

jtanx:
These changes have been incorporated into #4928.

PR #4750, opened 2021-06-11
IRC has moved to irc.libera.chat

ctrlcctrlv:
The Wiki has already been updated, but this should be too.

Users were requesting this on Twitter. Most other large channels have abandoned Freenode (#archlinux, #ubuntu…) due to what's becoming known as the [Andrew Lee incident](https://www.theregister.com/2021/05/19/freenode_staff_resigns/), which caused most Freenode network staff to resign in protest to a sale of Freenode. The new replacement is a 501(c)(3) organization with a Board of Directors, not a saleable entity.

I moved ahead with the change given the many requests and the fact that most users have abandoned Freenode, so it's not even practical to stay there even if we want to.

### Type of change

- **Documentation**


jtanx (approved):

frank-trampe:
Frank also approves.

PR #4753, opened 2021-06-14
gtextfield.c: fix length calculation for type 2 completions

### LINKED ISSUE #4751: Crash when going to a glyph by name
FontForge 20201107 in macOS 10.14.6 crashes if you follow these steps:

* Download [Catrinity.otf](http://catrinity-font.de/downloads/Catrinity.otf).
* Open it in FontForge.
* Click “View” in the menu bar.
* Click “Goto”.
* Type “A”.

After typing the letter “A”, FontForge immediately crashes. [Here is a crash report.](https://github.com/fontforge/fontforge/files/6644998/fontforge_2021-06-13.txt)

jtanx:
This part may cause an overestimation on first calculation:

    if ( ret[i][len]=='\0' )
        continue;

Fixes #4751

### Type of change

- **Bug fix**

### DIFF
--- a/gdraw/gtextfield.c
+++ b/gdraw/gtextfield.c
@@ -35,6 +35,7 @@
 #include "ustring.h"
 #include "utype.h"
 
+#include <assert.h>
 #include <math.h>
 
 GBox _GGadget_gtextfield_box = GBOX_EMPTY; /* Don't initialize here */
@@ -3085,8 +3086,10 @@ static void GTextFieldComplete(GTextField *gt,int from_tab) {
 		    } else if ( !doit )
 			ret2 = malloc((cnt+1)*sizeof(unichar_t *));
 		    else {
-			if ( type2 )
-			    cnt = type2;
+			if ( type2 ) {
+			    cnt = (cnt+MAXBRACKETS-1)/MAXBRACKETS;
+			    assert(cnt <= type2);
+			}
 			ret2[cnt] = NULL;
 		    }
 		}


gdraw/gtextfield.c
```diff
@@ -3086,7 +3086,7 @@ static void GTextFieldComplete(GTextField *gt,int from_tab) {
 			ret2 = malloc((cnt+1)*sizeof(unichar_t *));
 		    else {
 			if ( type2 )
-			    cnt = type2;
```
jtanx, line 3089: `type2` was the size calculated in the first loop, but the actual count may differ because there's a `continue` that only exists in the second loop

iterumllc: This is some serious spaghetti code with a limited change. Based on a brief review I have one concern, which is specifically how the following `ret2[cnt] = NULL;` is safe given the recalculation of `cnt`. Could you give a brief and/or partial explanation of that so I don't have to decipher the whole thing? (Note that I'm not calling its safety into question, just hoping to make the review easier.)

jtanx: Agreed that it's pretty indecipherable.

So like in a few other places, GWW has this pattern of calculating lengths in the first loop, then actually doing things in the second loop. Additionally, this code has two branches - one for if the number of suggestions is below some limit and another for if it's above. `type2` corresponds to the case where it's above. (in which case it tries to present a suggestion like `A[-abcd]...` which tells you that you can select `A` plus `a` or `b` or `c` or `d` etc. somewhat like a regex)

In the both loops, `cnt` counts the number of source elements being consumed (I think...). The index into the return value is computed from:

```c
int c2 = cnt/MAXBRACKETS, c3 = cnt%MAXBRACKETS;
...
ret2[c2] = ...
```

In the first loop, `type2` calculates the number of elements in the return value. But it overestimates, as in the second loop, it possibly skips over some elements:

```c
				if ( ret[i][len]=='\0' )
      		    continue</span>
```

`cnt` still counts the number of elements consumed, so it should be safe to just recalculate at the end of the loop to know where to null terminate it.

The alternative is you just calloc the buffers and recalculate the count at the end.

skef: `type2` is set to non-`false` in the `!doit` iteration when `cnt>=MAXLINES` and `ret2` is allocated immediately after based on that value of `type2`. Then a bunch of stuff happens that can invalidate (in some sense) the initial value of `type2`, so you've changed the code to recalculate `cnt` (the same way `type2` was calculated before). 

I guess the key here is the assertion that "`type2` ... overestimates", so if `type2` is non-`false` then before your modified assignment `(cnt+MAXBRACKETS-1)/MAXBRACKETS <= type2`, and therefore `ret2[cnt]` will refer to allocated memory. Could that be added as an `assert()` on that branch for clarity/verification?

jtanx: Done

iterumllc (approved):
Relative to the assert this seems like a safe an positive change. Approved.

skef (approved):
I meant me. Approved.

PR #4760, opened 2021-06-30
python.c: fix stroke flag checks

jtanx:
Fixes e.g.

    ../fontforge/python.c:4593:26: warning: & has lower precedence than !=; != will be evaluated first [-Wparentheses]
        si->removeinternal = ( f&1!=0 );
                                ^~~~~

### Type of change

- **Bug fix**


skef (approved):
lgtm

PR #4769, opened 2021-07-10
Try to fix missing setuptools from AppImage

### LINKED ISSUE #4761: [AppImage] Core python package 'pkg_resources' not found

### Important

- [x] Screenshot
  pic.1
  - Version: `FontForge-2021-07-01-d5593d0-x86_64.AppImage`
  - OS: `Debian 10.x buster`
  - There should be no missing files inside AppImage.
  1. Download latest nightly AppImage build;
  2. Launch it;
  3. See "Warnings" on startup.
  - Bundle needed Python package  `pkg_resources` into AppImage builds.

jtanx:
### Type of change

- **Bug fix**

Fixes #4761


jtanx:
tested, confirmed fixed.

skef:
Huh - I had thought `setuptools` was de facto core now, if not de jure -- if that's wrong there could be further issues after release.

Issue #4712, opened 2021-04-23
Shortcuts Copy-Paste 


Dimon-II:
The program does not support old Ctrl-Ins/Shift-Ins options, only Ctrl-C/Ctrl-V, but when the keyboard layout is Russian, copying is impossible, only through the menu.

Issue #3974, opened 2019-10-14
shortcut keys not working when keyboard language changes


aminabedi68:
shortcut keys not working when keyboard language changes to all except English(i have not tested for all) on windows (fontforge 20190801)

PR #4771, opened 2021-07-11
gdraw: better handling of mnemonics

jtanx:
Not quite ready yet as I still need to verify the mac changes. Just opening this now so I don't forget about it.

Follows from the discussions in #4682.

### Type of change

- **Bug fix**

### DIFF
--- a/fontforgeexe/fontview.c
+++ b/fontforgeexe/fontview.c
@@ -6841,12 +6841,14 @@ void FontViewRemove(FontView *fv) {
     FontViewFree(&fv->b);
 }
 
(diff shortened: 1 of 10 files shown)


skef:
This has been a draft since it was filed but I just went ahead and reviewed it and it looks good to me. Everything except the later content of `GGDKDrawShortcutKeyMatches()` is straightforward and that stuff looks plausible.

jtanx:
Alright, this works well enough. Turns out mnemonics aren't even a thing in 'native' mac apps, by the way.

jtanx:
Closes #3974, closes #4712

PR #4773, opened 2021-07-12
Fix for #4770 plugin initialization failure

### LINKED ISSUE #4770: Failure of plugin initialisation prevents script from running at all
* Have setuptools missing (or temporarily move `pkg_resources` from `site-packages`
* Run `fontforge -lang=py -c 'print("hello")'
* Expect it to print `hello`
* Instead it reports:
   ```
   PythonUI_Init()
   copyUIMethodsToBaseTable()
   Core python package 'pkg_resources' not found: Cannot discover plugins
   ModuleNotFoundError: No module named 'pkg_resources'
   ```

Relates to #4761

skef:
Closes #4770

### DIFF
--- a/fontforge/plugin.c
+++ b/fontforge/plugin.c
@@ -293,6 +293,7 @@ static bool DiscoverPlugins(int do_import) {
     PyObject *pkgres = PyImport_ImportModule("pkg_resources");
     if (pkgres == NULL || !PyObject_HasAttrString(pkgres, "iter_entry_points")) {
         LogError(_("Core python package 'pkg_resources' not found: Cannot discover plugins\n"));
+	PyErr_Clear();
         return false;
     }
     str = PyUnicode_FromString("iter_entry_points");


jtanx:
I think this is fine, but one thing that I wasn't sure about when looking at this is expected behaviour from when you manually try reloading from scripting - I think now this means that no error is raised?

skef:
This suppresses the redundant "ModuleNotFoundError: No module named 'pkg_resources'" message but not the first, so this PR shouldn't change much about the reload path.

But maybe the reload path is still wrong; can you say a bit more about the concern?

jtanx:
Suppose someone calls `fontforge.loadPlugins()` in the case where setuptools is missing (or where plugin loading fails), what is the desired behaviour in terms of exceptions raised, if any? Or is the error message printed to the console sufficient?

skef:
If a script is loading plugins after delaying with `-skippyplug` I would think it's hoping to have some specific plugin available, in which case it should look for it with `getPluginInfo()`. I can see some value in exposing infrastructure problems in the API but it seems marginal. Once the user tries to diagnose the problem they'll have the message.

jtanx (approved):

PR #4774, opened 2021-07-12
Skip invalid UTF-8 in utf82u_strncpy and make DlgCreate functions use it

### LINKED ISSUE #4731: Glyphs in multiple classes cause a hang
I think there are two issues here, but I'll keep it as one report for now.

**FontForge and operating system versions**
FontForge 20201107
Windows 10, Version 10.0.19043 Build 19043
Ubuntu 21.04

**Screenshot**
The bugs I am reporting occur within the _Kerning by Classes_ editor.
image

**Expected behaviour**
Add a character to a class in _kerning by classes_. If I add the same character to another class, FF should detect the duplication and offer me the choices of _remove from this class_ or _remove from other class_.
image

**Actual behaviour**
Whilst this occurs for every other character I've used so far, occasionally there is an error with the ă _abreve_ character. 
FF does not detect that it already exists in another class, and thus adds it to the new class. It does not display the character's name in the class list. (This did not happen *every* time, but in my recent tests, it has occurred every time.) 
Upon clicking the class entry, the _kerning by classes_ window hangs.
image
The window can be closed, and FF remains functional _unless_ I try to edit the data of the same lookup table again. If I click the _edit data_ button, the whole FF application hangs.

**Steps to reproduce the behaviour**
1. Create a series of classes.
2. One class must contain _ă (abreve)_.
3. Add _ă (abreve)_ to another class.
4. Popup dialog demanding the creator to choose which class does not appear.
5. Click on the list entry that you just added ă abreve to.
6. Interface should hang. You can close the window, and FF will still function.
7. Attempting to load the _kerning by classes_ editor again will cause FF to hang entirely.

**Another method, if the editor _doesn't_ allow you to add ă twice**
1. Open the .sfd file in a text editor. (Close FF.)
2. Find section _KernClass2:_
3. Add abreve into at least two classes.
4. Add 7 to the number at the beginning of the line. [This tells FF how many characters to parse for the class group.)
5. Save, and load the file in FF.

So, the two issues I think occur here are:
1. _Kerning by Classes_ character adding does not correctly/consistently filter/detect characters from existing classes. 
2. FontForge hangs when trying to load the class that the duplicate character was just added to, _instead_ of handling the situation gracefully. 

**Possible solution/fix/workaround**

1. Character detection & removal:
There could be a preliminary check before loading the class list, and a warning/error popup with the option buttons, just as already occurs when adding the characters to the class. This would prevent FF from reaching the stage of its process hanging. 

2. Duplicate character class membership hang:
I don't think FF needs to hang in this situation. I think it could be handled by prioritising the classes, either using a default fallback to alphabetical prioritisation as a temporary workaround; or with a priority based on class importance or recency. Ideally, there should be no duplicate characters, but on the event that FF fails to detect duplication, or the event that the source file contains duplicate characters, then FF should handle the situation more gracefully than just hanging out! 

According to [Wikipedia](https://en.wikipedia.org/wiki/%C4%82), it appears that ă abreve is only used in 3 languages and the transliteration of Bulgarian Cyrillic into Latin. 
Perhaps there is a deeper bug around the handling of this character, which may have got overlooked due to its uncommonness. 
The other Vietnamese characters in my font were handled fine, but most of them are just unixxxx character names so the processing of these names is probably simple and standard.

jtanx:
Fixes #4731

### Type of change

- **Bug fix**

### DIFF
--- a/Unicode/ustring.c
+++ b/Unicode/ustring.c
@@ -511,30 +511,27 @@ return( ubuf );
 }
 
 unichar_t *utf82u_strncpy(unichar_t *ubuf,const char *utf8buf,int len) {
-    unichar_t *upt=ubuf, *uend=ubuf+len-1;
-    const uint8 *pt = (const uint8 *) utf8buf, *end = pt+strlen(utf8buf);
-    int w, w2;
+    unichar_t *upt = ubuf;
+    int32 ch;
 
-    while ( pt<end && *pt!='\0' && upt<uend ) {
-	if ( *pt<=127 )
-	    *upt = *pt++;
-	else if ( *pt<=0xdf ) {
-	    *upt = ((*pt&0x1f)<<6) | (pt[1]&0x3f);
-	    pt += 2;
-	} else if ( *pt<=0xef ) {
-	    *upt = ((*pt&0xf)<<12) | ((pt[1]&0x3f)<<6) | (pt[2]&0x3f);
-	    pt += 3;
-	} else {
-	    w = ( ((*pt&0x7)<<2) | ((pt[1]&0x30)>>4) )-1;
-	    w = (w<<6) | ((pt[1]&0xf)<<2) | ((pt[2]&0x30)>>4);
-	    w2 = ((pt[2]&0xf)<<6) | (pt[3]&0x3f);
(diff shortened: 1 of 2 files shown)


skef (approved):
lgtm

PR #4783, opened 2021-07-26
savefontdlg.c: fix ParseBitmapSizes for non-English locales

### LINKED ISSUE #4779: Cannot generate bitmap fonts
When I try to generate a bitmap font, I get an error message about missing bitmaps. With BDF chosen as the output format, the error message is:

> Attempt to save a pixel size that has not been created (30@0)

With OTB, the error message is:

> The font database does not contain a bitmap of size 30 and depth 0

This is the same for existing and newly created bitmap fonts.

Fontforge versions tested: 20201107 and 21b5a6e65c5f3fb526615d02560640a10face615 on Void Linux

jtanx:
Fixes #4779

### Type of change

- **Bug fix**


skef (approved):
lgtm

Issue #4787, opened 2021-08-03
SplineCharDefaultNextCP and SplineCharDefaultPrevCP do not produce handles anymore


linusromer:
When drawing a spline with smooth points the current github version of FontForge produces a spline that is not smooth at all (upper part of the following picture). However, it used to produce something like the lower part of the following picture:
![fferror](https://user-images.githubusercontent.com/11213578/128070161-66de72bd-ce3f-43ce-be21-99aef54d282c.png)
This must have happened in the past month, because my version on https://github.com/linusromer/fontforge from 1. july still works as it used to be. 
This issue may be related to [https://github.com/fontforge/fontforge/pull/4685](https://github.com/fontforge/fontforge/pull/4685).

skef:
I'm too busy to look at this in any detail for quite some time but my guess is that the line https://github.com/fontforge/fontforge/pull/4685/commits/0f8fc11df7eaec55c75360db19494814d87ea33a#diff-78ff08502d7ca584fcdd8bc21ccc10e63b1eca0ed4406e0a1c8510b9e53a8712L3243 and its pair in `*DefaultPrevCP` was more substantive than I had realized in how it matches to  the change on (new line) 3385. I recommend putting something to match old line 3243/new line 3228 that actually alters the control point, such as setting `base->nextcp` to `base->me` + `unit`. It shouldn't matter *what* you set the nextcp to as long as it differs from `base.me`.

linusromer:
@skef I can confirm that your suggestion fixes the issue.

skef:
After thinking about this briefly a bit more it's probably more self-documenting to just maintain a function-local boolean in place of nonextcp to represent the need for the subsequent change. My suggestion works OK but is likely to mystify future developers (although maybe not with an appropriate comment?).

linusromer:
You mean a function-local boolean that would replace `( base->prevcp.x!=base->me.x || base->prevcp.y!=base->me.y )` ? This is probably preferable, indeed.

skef:
Yes -- checking the equality of ->prev/nextcp and ->me prior to calling SplineMake/Fixup is the general solution to the old nonextcp/noprevcp problems but in this case nonextcp was really being used as what amounted to a function-local boolean anyway. So better to just track that state with a local.

Issue #4763, opened 2021-07-01
Unsatisfying Point Type Change From Corner To Curve


linusromer:
The function `SPChangePointType()` in splinechar.c has a unsatisfying behaviour that I will try to describe in two GIFs:

1. If you change a single-linked corner point to a curve point, it will head to the next on-curve-node in the distance of the nice proportion = .39:
![smooth-to-corner0](https://user-images.githubusercontent.com/11213578/124164119-b3237500-daa0-11eb-9789-8058ebd1f6e1.gif)
2. If you change a double-linked corner point, the behaviour is basically the same but the direction is defined by the vector from the next and the previous on-curve-node:
![smooth-to-corner1](https://user-images.githubusercontent.com/11213578/124164688-3f359c80-daa1-11eb-871e-f7d71dad5386.gif)

This problem occurs both with the 2020-11-07 release as well as the current main branch of FontForge.

I have tried to fix the problem by changing line 699 of splinechar.c to something like

```
if( pointtype==pt_curve ) { 
	if ( sp->pointtype==pt_corner ) { 
		if ( sp->prev!=NULL && sp->next!=NULL && prevlen!=0 && nextlen!=0 ) {
			unitnext.x = .5*(unitnext.x-unitprev.x); /* average direction */
			unitnext.y = .5*(unitnext.y-unitprev.y); 
			sp->nextcp.x = sp->me.x + unitnext.x*nextlen;
			sp->nextcp.y = sp->me.y + unitnext.y*nextlen;
			sp->prevcp.x = sp->me.x - unitnext.x*prevlen; /* prev is antidirectional to next */
			sp->prevcp.y = sp->me.y - unitnext.y*prevlen;
		}
		if ( ( sp->prev==NULL && sp->next!=NULL && nextlen!=0 ) || (sp->next==NULL && sp->prev!=NULL && prevlen!=0) ) 
			makedflt = false; /* just do nothing */
	} 
    else makedflt = true; /* original behaviour */
}	
```
to change this behaviour but it did not work as desired. Maybe we should consider changing the function `SplineRefigureFixup()` as well? If someone has a good hint, I will try to figure out a patch and make a pull request.

skef:
Huh, I thought I had filed an issue on this a few years ago but apparently not. 

It depends on what you mean by "did not work as desired". Is it not taking effect or is the change just not desirable? 

I would say:

[...]

PR #4789, opened 2021-08-07
Improved Point Type Change From Corner To Curve and From Anything To Tangent 

linusromer:
This pull request results as an attempt to solve the issue [https://github.com/fontforge/fontforge/issues/4763](https://github.com/fontforge/fontforge/issues/4763). At the moment, details are explained there. But from now on, I think that changes are better described here. I have tested my code before and after [https://github.com/fontforge/fontforge/pull/4685](https://github.com/fontforge/fontforge/pull/4685) and while the overall behavior is the same, some things are different (one described at [https://github.com/fontforge/fontforge/issues/4787](https://github.com/fontforge/fontforge/issues/4787)). Therefore, my code may need some revision, when [https://github.com/fontforge/fontforge/issues/4787](https://github.com/fontforge/fontforge/issues/4787) is solved.

### DIFF
--- a/fontforge/splinechar.c
+++ b/fontforge/splinechar.c
@@ -45,6 +45,7 @@
 #include "tottf.h"
 #include "ttf.h"
 #include "ustring.h"
+#include "utanvec.h"
 #include "utype.h"
 
 #include <locale.h>
@@ -598,11 +599,39 @@ return( false );
 return( len2>=0 && len2<=len );
 }
 
+/* This is a fast computable measure for linearity. */
+/* First, we norm the spline: Scale and rotate the spline such that */
+/* one end lies on (0,0) and the other on (100,0). */
+/* Then the maximal y coordinate of the horizontal extrema is returned. */
+static bigreal Linearity(Spline *s) {
+    if ( s->islinear ) return 0;  
+    BasePoint ftunit = BPSub(s->to->me, s->from->me);
+    bigreal ftlen = BPNorm(ftunit);
+    if ( ftlen==0 ) return -1; /* flag for error, no norming possible */
+    ftunit = BPScale(ftunit, 1/ftlen);
+    BasePoint nextcpscaled = BPScale(BPSub(s->from->nextcp, s->from->me),100./ftlen);
+    if ( s->order2 ) return .5*fabs(BPCross(ftunit, nextcpscaled));
+    BasePoint prevcpscaled = BPScale(BPSub(s->to->prevcp, s->from->me),100./ftlen);
+    /* we just look a the bezier coefficients of the polynomial in t */
+    /* in y dimension (divided by 3): */
+    Spline1D ys1d;
+    ys1d.d = 0;
+    ys1d.c = BPCross(ftunit, nextcpscaled); /* rotate it to horizontal*/
+    ys1d.b = BPCross(ftunit, prevcpscaled)-2*ys1d.c;
+    ys1d.a = -ys1d.b-ys1d.c;
+    extended te1, te2;
+    SplineFindExtrema(&ys1d, &te1, &te2);
+    if ( te1==-1 && te2==-1 ) return 0;
+    if ( te2==-1 ) return 3*fabs(((ys1d.a*te1+ys1d.b)*te1+ys1d.c)*te1);
+return 3*fmax(fabs(((ys1d.a*te1+ys1d.b)*te1+ys1d.c)*te1), 
+fabs(((ys1d.a*te2+ys1d.b)*te2+ys1d.c)*te2));
+}
+
 void SPChangePointType(SplinePoint *sp, int pointtype) {
     BasePoint unitnext, unitprev;
     bigreal nextlen, prevlen;
     int makedflt;
-    /*int oldpointtype = sp->pointtype;*/
+    int oldpointtype = sp->pointtype;
 
     if ( sp->pointtype==pointtype ) {
 	if ( pointtype==pt_curve || pointtype == pt_hvcurve ) {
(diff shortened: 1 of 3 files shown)


fontforge/splinechar.c
```diff
@@ -620,25 +649,50 @@ return;
...
+		bigreal prevlinearity = Linearity(sp->prev);
+		bigreal nextlinearity = Linearity(sp->next);
+		if ( prevlinearity>=0 && nextlinearity>=0 ) {
+			if ( nextlinearity >= prevlinearity ) { /* make prev linear */
+				sp->prev->islinear = true;
+				sp->prev->from->nextcp = sp->prev->from->me;
+				sp->prevcp = sp->me; 
+				if ( sp->prev->from->pointtype!=pt_tangent) 
+					sp->prev->from->pointtype = pt_corner;
+				if ( sp->prev->order2 ) {
```
skef, line 662: Since review cycles are limited I'm dropping my "too much input to review" standard for this. 

My only question about the code at this stage is whether this part (with the `IntersectLines()`) is something already built into `SplineRefigureFixup()`, already used below in the curve case. Did you look at that?

linusromer:
I don't think that I have considered `SplineRefigureFixup()` when I wrote this part. Is your question heading to the question whether the following block (and its "sp->next" counterpart)
```
				if ( sp->prev->order2 ) {
					BasePoint inter;
					if ( IntersectLines(&inter,&sp->me,&sp->prev->from->me,&sp->nextcp,&sp->next->to->me) ) {
						sp->nextcp = inter;
						sp->next->to->prevcp = inter;
						SplineRefigure(sp->prev);
					} /* else just leave things as they are */
				}
```
could be replaced by `SplineRefigureFixup(sp->prev);` (and `SplineRefigureFixup(sp->next);` respectively)? If yes, I will gladly check this.

fontforge/splinechar.c
linusromer: I don't think that I have considered `SplineRefigureFixup()` when I wrote this part. Is your question heading to the question whether the following block (and its "sp->next" counterpart)
```
				if ( sp->prev->order2 ) {
					BasePoint inter;
					if ( IntersectLines(&inter,&sp->me,&sp->prev->from->me,&sp->nextcp,&sp->next->to->me) ) {
						sp->nextcp = inter;
						sp->next->to->prevcp = inter;
						SplineRefigure(sp->prev);
					} /* else just leave things as they are */
				}
```
could be replaced by `SplineRefigureFixup(sp->prev);` (and `SplineRefigureFixup(sp->next);` respectively)? If yes, I will gladly check this.

skef: Yes -- and to be clear I don't know the answer, it just seemed like a possibility given a scan of that code and its use elsewhere in the function.

linusromer: I have checked it and it did not work:
aa
left: original spline, middle: spline with top node converted to a tangent point with my original patch, right: using `SplineRefigureFixup(sp->prev);` and `SplineRefigureFixup(sp->next);`

skef: Alright - thanks for looking into this!

skef (dismissed):
I think this is in good shape and represents several substantial improvements.

linusromer:
I have realized that the quadratic splines are still not handled correctly. Consider the following contour where I want to make the lower left node smooth (image left: before point conversion; image right: after point conversion).
quadrff
The tactic I am using now is to take the average direction of the handles . However, in some cases there is no intersection of this average direction with the direction of the handles of the next nodes (or the intersection is on the wrong side). This should be fixed before merging.

linusromer:
Just to be clear: This is not a programming issue; I do not know how to handle this situation at all. Suggestions and hints are welcome!

skef:
@linusromer I just tried a similar shape in what amounts to the current code and the whole contour changes. There seem to be a lot of cases with quadratic splines where there's no good solution and what FontForge does isn't great. (I certainly wouldn't want my whole contour to change as a result of changing one point type.) 

This is a long-winded way of suggesting that the current solution may actually be fine. I suggest playing around with quadratic splines and the current heuristics to see if you convince yourself of something similar.

skef:
Oh, I guess it is a problem if the control points aren't coincident ...

I think it would be fine to leave the current algorithm for quadratics. It would also be fine to adjust the tangent angle to be the average (if that's not what already happens) and then to run the current algorithm. So much changes otherwise that the angle seems like a detail.

linusromer:
@skef 

> It would also be fine to adjust the tangent angle to be the average (if that's not what already happens)

Indeed, this happens already. I think there exists no proper solution for this problem (e.g. Fontlab struggles with my example contour, too). After thinking a bit I came to the conclusion that it would be best to leave the control points at their original position when this very specific problem occurs (the user has to interact anyway).

As far as I have checked, the patch should now be ready for merging.

linusromer:
After commit a0dc9d5 has been pushed, I have tested my patch again and I have discovered and fixed some minor curve updating issues after conversion to `tangent` and added better behaviour for conversion from `hvcurve` to `curve`.
There is a small remaining issue: After conversion to `tangent` in quadratic splines, FontForge does not make the adjacent line really to a line:
problem
I am not sure but I fear there is no `MakeLine()` function for splines and some kind of `CharChangedUpdate()` may be needed instead to fix this behaviour.

skef:
@linusromer I can look at the remaining problem but a quick set-of-steps-to-reproduce would be handy for me.

linusromer:
@skef I have made an example font with a contour made of quadratic bezier splines:
[quadr.sfd.zip](https://github.com/fontforge/fontforge/files/7874680/quadr.sfd.zip)
curve-update
The left part shows the original state of the contour and the right part shows the spline after two independent changes from me:
- I have made the spline DA to a line with "Make Line" from the menu. The control point is still displayed, which is a visual problem. However, it is also problematic when I want to move point B afterwards: Then the line DA is magically converted back to a curved spline.
- I have converted point C to a hv-smooth point. The displayed splines adjacent to C do not update afterwards (but they should of course).

skef:
The "Make Line" problem may be a little above my pay grade, as I'm not sure what the relevant issues are. The problem is that guard in SplineRefigure2 on line 1249 of `splineorder2.c` is failing because `from->nextcpindex` check. What's that about? As I understand it it's trying to preserve TTF point numbering.

This may be a simple local/global problem. If you eliminate the control point then you need to renumber all the points. SplineRefigure isn't really the level to try to do that sort of thing. There may also be TTF/quadratic spline font reasons to preserve the numbering. I really don't know much about quadratic spline font production.

This seems like something you may just have run across that isn't directly related to your changes here. Is it something we could file an issue about and leave for another time?

linusromer:
Yes, my example already showed that it has nothing to do with my changes because the two problems were already existant with current Fontforge versions. I just hoped the two problems could be fixed easily and the fix could be packed in this PR. But after your comment I agree that it is wiser to file a separate issue and merge my PR in the current state because it is absolutely independandt.

skef:
For the hvcurve problem: Isn't this just that the code between lines 725 and 745 that handles `pt_hvcurve` doesn't have any calls to `SplineRefigure()`? (For example, compare with the code below for `pt_curve`.) This diff seemed to work well for your test case:

```
diff --git a/fontforge/splinechar.c b/fontforge/splinechar.c
index 40bba8391..3a44fb3cf 100644
--- a/fontforge/splinechar.c
+++ b/fontforge/splinechar.c
@@ -737,11 +737,13 @@ return;
            sp->nextcp = ncp;
            if ( sp->next!=NULL && sp->next->order2 )
                sp->next->to->prevcp = ncp;
+           SplineRefigure(sp->next);
            pcp.x = sp->me.x + unitprev.x*prevlen;
            pcp.y = sp->me.y + unitprev.y*prevlen;
            sp->prevcp = pcp;
            if ( sp->prev!=NULL && sp->prev->order2 )
                sp->prev->from->nextcp = pcp;
+           SplineRefigure(sp->prev);
            makedflt = false;
        }
        if( pointtype==pt_curve ) { 
```

skef:
I just reviewed the code and it looks fine to me. I'll leave it for a few days in case @jtanx or someone else wants to take a look and comment and if not I'll approve it.

linusromer:
@skef Well, yes, this solves the problem. I have added your patch. (Some time ago SplineRefigure() was called anyway at another place but this call was cancelled by a commit in the last year.) 
And for the "make line"-problem: In the version 20190801 the "make line" command worked but the numbering problem was solved very improperly (all nodes kept their numbers but the numbers had gaps afterwards like 1,2,4,5). I guess, it's better to leave it in the way it is now to solve the numbering problem properly. With the knowledge I have now I would say that this PR can be merged and no additional issue should be filed.

skef:
There were many consistency problems with cubic splines and control points (or lack thereof) before #4685. I wouldn't be surprised to learn that there were related quadratic spline/point numbering problems as well. My guess is that that commit reactivated a consistency check in SplineRefigure2 that wasn't being applied before.

MakeLine should probably work in this case, as its changing the geometry so there's no need to preserve TrueType hinting (which depends on point numbering). The problem is that there isn't a consistent way of addressing point numbering in the code, at least that I can see. What's probably needed is a point numbering "philosophy" implemented with a commit similar in scope to #4685. Before that happens the current convention of leaving the point in SplineRefigure2 seems like the lesser of evils.

skef (approved):
LGTM

PR #4796, opened 2021-08-28
tottfgpos.c: ensure qsort determinism

kvtb:
When `qsort`'s comparator function returns `0`, the `qsort`'s result gets implementation-depended.

That makes the fonts not reproducible.

Within a single Linux distro there could be different binaries of, for example, `DejavuSans.ttf` for `i686`, `x86_64` and `aarch64`.
Not to mention the case when there is `musl` instead of glibc

In particular, `feat_alphabetize` often returns `0`, even on building such a widely used font as http://github.com/dejavu-fonts/dejavu-fonts
The rest `assert(0)` are just in case, they do not fix any real bug

### DIFF
--- a/fontforge/tottfgpos.c
+++ b/fontforge/tottfgpos.c
@@ -26,6 +26,7 @@
  */
 
 #include <fontforge-config.h>
+#include <assert.h>
 
 #include "tottfgpos.h"
 
@@ -677,6 +678,7 @@ static void dumpcoveragetable(FILE *gpos,SplineChar **glyphs) {
 
 static int sc_ttf_order( const void *_sc1, const void *_sc2) {
     const SplineChar *sc1 = *(const SplineChar **) _sc1, *sc2 = *(const SplineChar **) _sc2;
+    assert(sc1->ttf_glyph != sc2->ttf_glyph);
 return( sc1->ttf_glyph - sc2->ttf_glyph );
 }
 
@@ -1130,6 +1132,7 @@ struct sckppst {
 
 static int cmp_gid( const void *_s1, const void *_s2 ) {
     const struct sckppst *s1 = _s1, *s2 = _s2;
+    assert( ((int) s1->other_gid) != ((int) s2->other_gid) );
 return( ((int) s1->other_gid) - ((int) s2->other_gid) );
 }
 
@@ -1708,6 +1711,7 @@ return;
 
 static int orderglyph(const void *_sc1,const void *_sc2) {
     SplineChar * const *sc1 = _sc1, * const *sc2 = _sc2;
+    assert( (*sc1)->ttf_glyph != (*sc2)->ttf_glyph );
 
 return( (*sc1)->ttf_glyph - (*sc2)->ttf_glyph );
 }
@@ -2613,6 +2617,7 @@ static void AnchorsAway(FILE *lfile,SplineFont *sf,
 
 static int lookup_size_cmp(const void *_l1, const void *_l2) {
     const OTLookup *l1 = *(OTLookup **) _l1, *l2 = *(OTLookup **) _l2;
+    assert( l1->lookup_length != l2->lookup_length );
 return( l1->lookup_length-l2->lookup_length );
 }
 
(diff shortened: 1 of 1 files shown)


frank-trampe:
How much have you tested this? Are you sure we don't get assertion failures for your test cases on any of our platforms?

frank-trampe:
Let's add catch-all return statements to make static analysers happy.

kvtb:
> How much have you tested this? Are you sure we don't get assertion failures for your test cases on any of our platforms?

Not much, asserts are there exactly to test and hit
And if hit to add second parameter to compare. In some case it might require addling a field with sequential number to structure.

I expected your CI/CD to have some test suite; if not, i could run test on all fonts included into nixpkgs

kvtb:
There is another non-determinism: fontforge produces different font binaries if compiled with -mavx or -mavx2 (e.g. on Gentoo). So there are 3 different font binaries: without AVX, with AVX and with AVX2. It is still being investigated and might be related to this case. So I put a WIP

frank-trampe:
We really appreciate your work on this. Stable output matters to a lot of our users. I'm happy merging this as long as we are reasonably confident that there are no regressions.

kvtb:
AVX issue seems unrelated, there is something with float-to-int rounding: coordinates of some points differ by 1.

What will we do with this one?
Actual fix is only adding
https://github.com/fontforge/fontforge/blob/221de73c8f5b9a2dcd6cfa9e335ff81076c98fcc/fontforge/tottfgpos.c#L2920-L2923

And maybe wrapper for `qsort` which will crash or print warning on stability issues?
There are many more `qsort` comparators in other files

fontforge/tottfgpos.c
```diff
@@ -2613,6 +2617,7 @@ static void AnchorsAway(FILE *lfile,SplineFont *sf,
 
 static int lookup_size_cmp(const void *_l1, const void *_l2) {
     const OTLookup *l1 = *(OTLookup **) _l1, *l2 = *(OTLookup **) _l2;
+    assert( l1->lookup_length != l2->lookup_length );
```
frank-trampe, line 2620: How does this work? Surely there are lookups that have identical sizes?

jtanx (changes requested):
These changes don't make sense to me, you can't just convert returns to asserts and call it a day, what about release builds?

skef:
I'll summarize some of the concerns already expressed with this and add my own. If @kvtb is willing to update I'm willing to work with them towards merging. 

1. The "release" (no asserts) build needs to continue to work, even in cases with sorting indeterminism. So the `return 0`'s should be added back in after the `assert(0)`'s. 
2. `assert( l1->lookup_length != l2->lookup_length );`doesn't make much sense, as different lookups can have the same length. I'm not sure whether there's a reason other than determinism for arranging the lookups by length, let's assume there is. So something else would be needed to break ties between equal-length tables. One obvious candidate would be comparing the `subtable_name` of each of the `subtables` until one is different and sorting on that.
3. I don't like the general convention of your asserts, e.g. `assert(sc1->ttf_glyph != sc2->ttf_glyph);`, because it means you can't pass the same values into the order functions without getting an error. The code may never do that now but it should still be possible. So I recommend something more like "assert(sc1 == sc2 || sc1->ttf_glyph != sc2->ttf_glyph);`
4. indeterminism due to differences in floating point calculations may be a lost cause.

ferdnyc:
(Butting in on an older PR, apologies twice over...)

@kvtb 
> When qsort's comparator function returns 0, the qsort's result gets implementation-depended.
> 
> That makes the fonts not reproducible.

Be that as it may, `0` is how a comparison function expresses _equality_. It's a tri-state comparison, like the spaceship operator: `left <=> right` will be negative if `left < right`, will be positive if `right < left`, and will be `0` if, in the terms of the comparison function, `left == right`.

While it's true that the sorting for equal items in a list will be implementation-dependent (and affected by the CPU instruction set), the solution for that cannot be to make the comparison function **incapable** of comparing two members and determining them to be equivalent, if indeed they are. Its job description calls for tri-state returns.

If sorting is causing problems, then a solution would be... well, there are two actually:

1. Don't do the sort. Just eliminate the sort operation entirely, if doing without it won't cause other problems.
2. Somehow uniquely hash the items on the list so that the comparison will be the same on all systems, and no two items will ever compare equal. That way they become unambiguously, reproducibly orderable.
 
That second approach is not something that would be done from _inside_ the comparison function. Additional stable data needs to be added to the list items for the comparison function to examine.

skef:
@ferdnyc I agree with much of the spirit of your comment but I'm not sure one can be quite that definitive. It's best to have comparison functions return 0 for equal but they may only need to do that for "truly" identical values. If the function is never passed the same value in practice then asserting inequality is a good way of checking whether all aspects of identity are being taking into account.

Canonical serialization + a hash is one way of approaching identity but not the only one and it crucially depends on the "canonical" part and whether it captures the sort of ordering you're interested in in the first place. It can often be easier just to have cascading comparisons like the current code as a means of comparing.

PR #4809, opened 2021-09-14
Remove pre-gettext resource management functions

jtanx:
This removes old (and extremely obsolete) functionality to get string resources outside of gettext

### Type of change

- **Non-breaking change**


frank-trampe:
I don't like increasing our GNU lock-in, but at least it's not in libfontforge. I assume that you've tested the relevant codepaths?

frank-trampe (approved):
The changeset appears to me to do the following cleanly.
* Remove conditional execution based upon `_ggadget_use_gettext`.
* Remove conditional execution based upon the property `text_in_resource`.
* Remove functions that the passing tests confirm are usused.

jtanx:
The boat for not using gettext sailed decades ago, it's literally the only thing that works right now, so this PR is just removing dead code. GNU or not, no one would use GWW's method of internationalisation today.

PR #4810, opened 2021-09-14
Upgrade to Unicode 14.0.0

jtanx:
Looks like we never managed to release before it came out after all.

### Type of change

- **New feature**


jtanx:
Purely a codegen change.

KarenFarias2022:
`uni14`

KarenFarias2022:
### hi

PR #4833, opened 2021-10-10
Fix various source typos

luzpaz:
Follow up to d5227e60b04


skef (approved):
The only potential issue I see with these changes is that modifying to "PseudoColor" changes the API, but a) it's only in the X backend and b) that isn't likely to be widely used choice now even among those using this API at all. So it seems OK.

PR #4845, opened 2021-11-11
Fix for #4815 genericGlyphChange vMap segfault

### LINKED ISSUE #4815: genericGlyphChange vMap - always results in TypeError or Segmentation fault

Version: 20201107
 Based on sources from 2021-09-23 03:40 UTC-D.
 Based on source from git with hash: 21ad4a18fb3d4becfe566d8215eba4483b0ddc4b

Python 3.8.10 (default, Jun  2 2021, 10:49:15)
[GCC 9.4.0] on linux

Calling `genericGlyphChange` with a `vMap` keyword argument always causes either a TypeError with the message: `TypeError: vMap should be a tuple (or some other sequence type).` or a segmentation fault.

Open any font and try to run the `genericGlyphChange` function using a `vMap` argument:

```
>>> import fontforge
>>> f = fontforge.open('/path/to/anyfont')
>>> f['a'].genericGlyphChange( stemScale=1, hCounterScale=1, vMap=((300, 50, 200)) )
Segmentation fault
```

Or running the above as a script:

```
$ python test.py
  File "test.py", line 3, in <module>
    f['a'].genericGlyphChange( stemScale=1, hCounterScale=1, vMap=((300, 50, 200)) )
TypeError: vMap should be a tuple (or some other sequence type).
```

-[ ] \(optional) Possible solution/fix/workaround

skef:
Self-evident.

Closes #4815

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -17280,7 +17280,7 @@ static bool PyFFParse_genericGlyphChange(PyObject *args, PyObject *keywds,
 	    &rsbScale, &rsbAdd,
 /* Vertical counter info */
 	    &vCounterType,
-	    &vCounterScale, vCounterAdd,
+	    &vCounterScale, &vCounterAdd,
 	    &vScale,
 	    &vMap
 	    ) )


ctrlcctrlv:
`genericGlyphChange` has to be one of the strangest parts of the FontForge Python API. I'm sure it has a use, but I couldn't tell you what it is. :laughing:

PR #4850, opened 2021-11-13
Better font-level ascent and descent metrics

skef:
In the investigation of #4704 we discovered that the ascent and decent values set by `GDrawGetFontMetrics()` are routinely less than those for Latin alphanumeric strings. The Pango docs point out that there is no necessary relation between the font metrics and the "ink"-derived pixel sizes but there's code everywhere in FontForge that assumes the values returned by `GDrawGetFontMetrics()` are the "normal" case for pixel sizes.

This PR replaces the Pango statements with a call to DoText8 for the string "AMOfgjyql47" (which is easily changed or added to). Then the maximum of the fas,as (or fds,ds) from tf_arg.size is returned for as (ds).

### DIFF
--- a/gdraw/ggdkcdraw.c
+++ b/gdraw/ggdkcdraw.c
@@ -1285,24 +1285,18 @@ void GGDKDrawClipPreserve(GWindow w) {
 // PANGO LAYOUT
 
 void GGDKDrawGetFontMetrics(GWindow gw, GFont *fi, int *as, int *ds, int *ld) {
-    //Log(LOGDEBUG, " ");
+    struct tf_arg arg = {0};
 
-    GGDKDisplay *gdisp = ((GGDKWindow) gw)->display;
-    PangoFont *pfont;
-    PangoFontMetrics *fm;
-    PangoFontMap *pfm;
-
-    _GGDKDraw_configfont(gw, fi);
-    pfm = pango_context_get_font_map(gdisp->pangoc_context);
-    pfont = pango_font_map_load_font(pfm, gdisp->pangoc_context, fi->pangoc_fd);
-    fm = pango_font_get_metrics(pfont, NULL);
-    *as = pango_font_metrics_get_ascent(fm) / PANGO_SCALE;
-    *ds = pango_font_metrics_get_descent(fm) / PANGO_SCALE;
+    arg.first = true;
+    gw->ggc->fi = fi;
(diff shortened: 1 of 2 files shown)


skef:
It might be worth localizing the test string so that for other locales it includes a mix of code points for the script and Latin code points. That way if there are family fall-backs we'll get decent values for the size of a typical string.

skef:
@jtanx if you look at this it would probably be easier to play around by merging it with #4704

jtanx:
I think this seems somewhat sensible, although still looking into this. One thing I don't quite reconcile is that DoText8 _also_ uses `fas`/`fds` to compute `as`/`ds`, so this all seems a bit entwined...

skef:
If the baseline is consistent across both calculations that shouldn't be a problem, and if it isn't then we're probably screwed either way whenever the rendered height exceeds fas+fds.

jtanx:
Sorry, I wanted to check something against mac, but my PC that runs the mac vm is currently dead and will probably only be fixed next year.

jtanx:
After looking at this a bit more, I think the problem may more lie in how the ascent/descent is calculated off the ink rects provided by pango.

I could be wrong, but what I think that instead of

```c
// Why is fds even involved in this calc?
arg->size.as = ink.y + ink.height - arg->size.fds;
arg->size.ds = arg->size.fds - ink.y;
```

Perhaps this makes more sense:

```c
 int baseline = pango_layout_iter_get_baseline(iter) / PANGO_SCALE;
arg->size.ds = ink.y + ink.height - baseline;
arg->size.as = ink.height - arg->size.ds;
```

Layout ink rects have coordinates starting from the top left corner, for reference.

Just purely in testing, the latter seems to fix the issues noted without any of the other changes being required, although maybe this is purely coincidental with the fonts being used?

jtanx:
For comparison:

Old:
image
image


New:
image
image


Based off the res3 branch. You have to flip between the two to notice, but e.g. the `Testing` entry is aligned better with the icon to the left. Same too for the menu items

ctrlcctrlv:
Wow, look at the ‹g› in «Add Encoding Slots…». It no longer collides right with the baseline! That's a big change @jtanx, I certainly don't have to flip back and forth…then again, I am a type designer :laughing:

skef:
Alright -- this was a workaround for a problem that came up in #4704 for which there was no suggested fix at the time. If there is one now and it seems to work then we should likely go with that.

jtanx:
As discussed, superseded by f8bff5f

PR #4865, opened 2021-12-18
replace CMAKE_{SOURCE/BINARY}_DIR with CMAKE_CURRENT_*_DIR

andiwand:
importing your CMake with `add_subdirectory` will not work because `CMAKE_SOURCE_DIR` and `CMAKE_BINARY_DIR` will point to the wrong directory.

I am currently trying to get this library built and packaged with conan. While doing so I had to patch the source.
Maybe these patches are interesting for others too.

### Type of change

- **Bug fix**
- **Non-breaking change**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,19 +2,19 @@
 
 cmake_minimum_required(VERSION 3.5)
 
-# No in source builds
-if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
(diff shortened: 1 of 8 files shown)


jtanx:
One thing that I don't like about this is that it makes everything dependent on where you import the scripts from.

By the looks of it, you should be using PROJECT_SOURCE_DIR/PROJECT_BINARY_DIR instead.

andiwand:
@jtanx thank you for the suggestion - makes sense to me! I just changed it.

PR #4866, opened 2021-12-18
Fix 'Testing equality to None' issues

luzpaz:
Testing whether an object is 'None' using the `==` operator is inefficient and potentially incorrect.  
ref: https://lgtm.com/projects/g/fontforge/fontforge/?mode=tree&ruleFocus=7900090


ctrlcctrlv:
It's a bit strange we even distribute that file. First added by @davelab6 in 2014: https://github.com/fontforge/fontforge/commit/f6304c0e4cb09ac0d3124c3ee09304e33d8e04d2

Wonder why? It's not that well documented. "Specially prepared Inkscape SVG", not sure what that means. Why do we have this when FontForge can import SVG?

Issue #3174, opened 2017-10-07
Setting macStyle via python script not working


ffpyguy:
I don't understand why this isn't working using python. Here is my code:

```
import os, sys, fontforge

filename = sys.argv[1]
font = fontforge.open(filename)

print font.macstyle # outputs -1
font.macstyle = 2  # is Italic
print font.macstyle # outputs 2, but in final output font the macStyle hasn't changed, why?

font.generate('output.otf')
font.close()
```

I've tried several different input fonts but same results, macStyle doesn't change..

jtanx:
Since a9c48a6314b93cdded1b2bfa1312c2bb150b240c / #2145, you must ensure that the 'style map' parameter also matches (a subset of the OS2 fsSelection bits), e.g.:

```py
...
font.macstyle = 2  # is Italic
font.os2_stylemap = 1 # Italic - https://www.microsoft.com/typography/otspec/os2.htm#fss
...
```

That's enough such that you will have correctly overridden the style bit in the output. However, when you read the file back, it will still show as `-1`. I think this is by design;  normally the macstyle is inferred from the font subfamily name, so `-1` essentially indicates that it will be inferred on generation. (For future reference, `SFFillFromTTF` doesn't set `sf->macstyle` from `info->macstyle`, and leaves it at the default value of `-1`).

ffpyguy:
Thanks this worked! How come there is no mention of this **.os2_stylemap** property in the documentation? Perhaps it should be added and mentioned how to set the macstyle correctly.

PR #4877, opened 2022-01-05
doc: Add Python font.os2_stylemap

Finii:
[why]
The `font.os2_stylemap` is undocumented.

[how]
Add entry for `font.os2_stylemap` and reference `font.macstyle`.

No back reference added, albeit that might be useful,
considering Issue #3174.

See also https://github.com/fontforge/fontforge/issues/3174#issuecomment-335037255

### Type of change

- **Non-breaking change**

_Edit: Just formatting_

### DIFF
--- a/doc/sphinx/scripting/python/fontforge.rst
+++ b/doc/sphinx/scripting/python/fontforge.rst
@@ -3433,6 +3433,10 @@ This type may not be pickled.
 .. attribute:: font.os2_fstype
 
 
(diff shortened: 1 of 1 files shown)


Finii:
I believe `os2_stylemap` is kind of a misnomer. I wanted to set bit 8 (0x100) `WWS` which is a hard stretch to call 'style'; also people who want to access `fsSelection` will search for that word :grimacing:

PR #4880, opened 2022-01-06
Change label of `Encoding→Compact` for clarity

### LINKED ISSUE #130: Add 'View, Hide Unused Cells' to Font View
Deron suggested:

Add a "Hide unused cell rows" to the Font View window's View menu. 

This is similar to 'Encoding, Compact' but only affects the view and not the internals of the font at all.

It will also not change the order of the displayed glyphs, or cause any columns to shift around; and it would only hide completely empty rows in the view, and where they are hidden, show a thicker cell border line, to indicate that one or more rows are hidden.

ctrlcctrlv:
Changed it to `Encoding→Compact (hide unused glyphs)`

Closes #130.

### Type of change

- **Non-breaking change**

### DIFF
--- a/fontforgeexe/fontview.c
+++ b/fontforgeexe/fontview.c
@@ -5007,7 +5007,7 @@ return;
 
 static GMenuItem2 enlist[] = {
     { { (unichar_t *) N_("_Reencode"), NULL, COLOR_DEFAULT, COLOR_DEFAULT, NULL, NULL, 0, 1, 0, 0, 0, 0, 1, 1, 0, 'E' }, H_("Reencode|No Shortcut"), emptymenu, FVEncodingMenuBuild, NULL, MID_Reencode },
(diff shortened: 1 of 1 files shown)


frank-trampe:
Is "hide" the best term? Maybe "drop"?

ctrlcctrlv:
I think it's a good term. Nothing is really being dropped. It's a view-only flag, sfdnormalize even drops it. (Yes, it surprised me too, see https://github.com/alerque/sfdnormalize/issues/5, @alerque made a strong case for dropping it by default.)

frank-trampe:
Oh. I did not know that!

ctrlcctrlv:
Yes, although I argued with them in that thread hehe, this is one of those things that it took me some time to see the wisdom of…lots of things like that in fonts, honestly…

frank-trampe (approved):

PR #4889, opened 2022-01-21
Address all Wall warnings

jtanx:
Most of this work was done months ago and a lot of the fixes automated away with some scripting.

This addresses all the warnings at -Wall with my version of gcc. Would also like to tackle Wextra at some point, but this is plenty big as is. Having said that, the changes should all be straightforward, especially if observed commit by commit.

Fixes at least one case of UB.

### Type of change

- **Bug fix**
- **Non-breaking change**


fontforge/parsepdf.c
```diff
@@ -1565,6 +1562,8 @@ static void _InterpretPdf(FILE *in, struct pdfcontext *pc, EntityChar *ec) {
 		    tok==pt_fillstrokeeo || tok==pt_closefillstrokenz ||
 		    tok==pt_closefillstrokeeo )
 		ent->u.splines.fill.col = fore_fill;
+		// FIXME Where to set?
+		// memcpy(ent->u.splines.stroke.dashes,dashes,sizeof(dashes));
```
ctrlcctrlv, line 1566: Note: the bot above (@lgtm-com) is registering a new FIXME…I don't know how to use its website and don't care to learn so I checked out 2f86e366a58b97a36873253a1bc59869acebb209 of jtanx/fontforge cf2 branch and then compared it to HEAD. It's right here. I thought it was probably a bug but it's not, @jtanx really added a FIXME. Probably does not matter, would be good to fix in ideal world, which we don't live in. But, in case anyone else wants to know what bot is complaining about, this is it.

skef:
@jtanx can you give me a primer or a pointer on `BAD_CAST`? It doesn't seem to be defined in the repo and I haven't yet turned up a reference to it elsewhere. (`std::bad_cast` from c++ seems unrelated.)

jtanx:
It's a thing from libxml2, probably something to do with unicode support and preventing implicit casts of probably ansi byte strings to what should be utf-8

skef:
Ah, OK.

skef:
If/when you resolve the ggdkdraw.c conflict with the since-merged stuff I'll approve this. I didn't look at every line but I assumed the commit comments were accurate and looked at the commits that seemed more likely to introduce a problem. Looked fine to me.

frank-trampe:
That seems reasonable to me. I was going to try to do every line, but there are a lot of lines!

skef (approved):
LGTM

skef:
I approved this, but we may want to think about what order to merge it in compared with @linusromer's changes. Not sure we want to create a lot of merge/cleanup work for him.

fontforge/splineutil2.c
```diff
@@ -2911,7 +2911,7 @@ void SplineSetAddInflections(SplineChar *sc, SplineSet *ss, int anysel) {
     Spline *s, *first; 
     first = NULL;
     for ( s = ss->first->next; s!=NULL && s!=first; s = s->to->next ) {
-	    if ( !anysel || s->from->selected && s->to->selected )
+	    if ( !anysel || (s->from->selected && s->to->selected) )
```
jtanx, line 2914: @linusromer just confirming that this is the intended behaviour?

linusromer: Yes, either nothing is selected or both endpoints are selected. I thought that `a || b && c` is equivalent to `a || (b && c)` but I may be misled by other programming languages.

PR #4900, opened 2022-01-30
Add Points Of Inflection / Balance / Harmonize

linusromer:
As [intended](https://github.com/fontforge/fontforge/discussions/4879) I have implemented entries in the "Element Menu" named "Add Points Of Inflection", "Balance" and "Harmonize" for the font view, the char view and the metrics view. Corresponding scripting functions are included as well.
- "Add Points Of Inflection" relies on the computation that is already used to display the points of inflection but then additionally splits the spline there. Hotkey: Ctrl+Shift+Y (homophone Ynflection)
- "Balance" makes the line between the control points parallel to the chord such that the area is preserved. This is an improved version of the algorithm known as "tunnify". Hotkey: Ctrl+Shift+P (homophone Palance)
- "Harmonize" moves the smooth on-curve points between its adjacent control points such that the adjacent curvatures become equal (except for sign at points of inflection). Hotkey: Ctrl+Shift+Z (harmoniZe)

The translations to German are included. However, the translations to other languages will have to be done by others.

This makes my plug-in [https://github.com/linusromer/curvatura](https://github.com/linusromer/curvatura) more or less useless (because the "Harmonize handles" of curvature is probably not used by anybody). 

### Type of change
- **New feature**


Symbian9:
Awesome. FTR, Here are binary artifacts for this pull already:

- https://github.com/fontforge/fontforge/actions/runs/1769425722#artifacts

ctrlcctrlv:
Fantastic, excellent work. Finally my decision to make [MFEK's point type](https://github.com/MFEK/glifparser.rlib/blob/98ab5c6a30c85e159a79350b72f4b625a95a5820/src/point.rs#L135-L152) FontForge-similar (with `next` and `prev` points) (`a`, `b`), as well as having `PrevNext` trait on `Contour` impl's is going to be bear some fruit, as many of your algorithms here I can literally copy and paste and run a few `sed`'s on to bring them into MFEK. :-)

Thanks a lot also for using `BPAdd` and `BPCross` so often as it really helps me understand how the code works.

ctrlcctrlv:
Oh, and by the way, some code review as far as _this project_ goes—

① Why did you add the options to the MetricsView? Seems to clutter menus and not make much sense, can you rationalize it? https://github.com/fontforge/fontforge/pull/4900/files#diff-7dcb402a6783e5b71a16f0e5edf1387845b40574aaeea9faa90c172ee7832c05
② Regarding adding new hotkeys, is this allowable @skef? I think we said it might not be.
③ Please remove the font-level Python API functions. It's my opinion they're unnecessary clutter that can be had the same by iteration, but are very unlikely to lead to pleasing results when applied to whole font. I understand having the functions in the FontView, as it only is applied to selected glyphs, but not in Python API.
④ For the Python docs, please either— ⓐ _group_ your new functions together in each section (`layer`, `contour`, &c) with a link to your Curvatura docs on top, (breaking alphabetization is IMO ok, but if we think it's not then) ⓑ make a new docs section for these functions with images and link to that doc before the short blurb of each. 
⑤ Also for the Python docs, only write the short blurb once, for the others, write _please see contour.whatever_. `contour` is likely to be the most often used API.

linusromer:
@ctrlcctrlv 
① Well, that is a actually a question I have asked in this [discussion](https://github.com/fontforge/fontforge/discussions/4879). I reasoned that if "Add Extrema" has an entry in the metrics view then "my" three functions should probably have entries as well. But I will gladly remove them from the metrics view, because for me personally there is no reason for them.
② I think it's a pity when Harmonize and Balance do not get default hotkeys (however, I could live happily without a hotkey for Add Points Of Inflection). There are not many hotkeys I am using for my font designs but Harmonize and Balance are among them (besides Add Extrema, Simplify, Merge, Round, Save, Autohint, Open Metrics View). One reason for doing this pull request was that a new user can use these functions out of the box without adding additional files (like curvatura.py oder a personal hotkeys file).

I will implement your suggestions 1,3,4,5 as soon as I have an answer to your question 2.

skef:
I think there's an argument to be made that Add Points of Inflection should appear everywhere that Add Extrema appears, in contrast with Harmonize and Balance. The former are similar in that they're more about normalization/avoiding potential tool problems than about output geometry. So I would recommend leaving Add Points of Inflection in those places even if/when the other two are removed.

I'm not aware of any guidelines against adding hotkeys to new "core" menu items. User-chosen hotkeys override core hotkeys (I believe), so people using them for something else shouldn't be affected.

Symbian9:
> I'm not aware of any guidelines against adding hotkeys to new "core" menu items. User-chosen hotkeys override core hotkeys (I believe), so people using them for something else shouldn't be affected.

JFTR, Is there a way to fully customize FontForge hotkeys?

skef:
@Symbian9 Yes, this should work: https://fontforge.org/docs/ui/misc/HotKeys.html#hotkeys-hotkeyassign

linusromer:
@ctrlcctrlv 
After skef's comment I have implemented the following things regarding your five points:
① I have removed Balance and Harmonize from MetricsView but not Add Points Of Inflection.
② My shortcuts are still there.
③ I have removed the font-level Python API functions for Balance and Harmonize but not for Add Points Of Inflection.
④ I have chosen your option ⓐ with `.. seealso::` to link to the Curvatura documentation.
⑤ Done.

skef:
I'm gonna wait to review this until #4789 is in and this branch is rebased.

skef:
@linusromer Now that #4789 is merged can you rebase this off the current master. I think there are a bunch of common commits and doing that will make reviewing easier.

skef:
I've started reviewing this and so far the new code looks good, however, when reviewing with all commits it looks like some of #4789 was reverted. Could you take a look and either figure out how I or github is confused or fix the issue?

linusromer:
@skef Github is confused because of me. I think I have somehow messed up the commits of my "master" and "inflections" branches. But now I think I have solved the issue by cherry-picking the corresponding commits of my master branch. (In retrospective I probably should have done a rebase first to make things clearer.) 
I have tested everything again and looked at the "files changed" and now I am quite sure that nothing from https://github.com/fontforge/fontforge/pull/4789 is reverted.

fontforge/fontviewbase.c
```diff
@@ -929,6 +929,108 @@ void FVAddExtrema(FontViewBase *fv, int force_adding ) {
     ff_progress_end_indicator();
 }
 
+void FVAddInflections(FontViewBase *fv, int anysel) {
```
skef, line 932: I diffed these and they're identical except for the (reused) strings and the function call. The code base has that sort of thing but in this case I think its worth adding the string and function pointer to the signature of one of these and having the others be one-liners.

fontforge/splineutil2.c
```diff
@@ -2879,6 +2880,159 @@ void SplineCharAddExtrema(SplineChar *sc, SplineSet *head,enum ae_type between_s
...
+
+void SplineCharAddInflections(SplineChar *sc, SplineSet *head, int anysel) { 
+    SplineSet *ss;
+    for ( ss=head; ss!=NULL; ss=ss->next )
+	    SplineSetAddInflections(sc,ss,anysel);
+}
+
+/* Make the line between the control points parallel to the chord */
+/* such that the area is preserved (kind of an improved "tunnify") */
+Spline *SplineBalance(Spline *s) {
```
skef, line 2928: You're (obviously -- this is the point) modifying SplinePoint values in this function but I don't see calls to `SplineRefigure()` anywhere in the path. (Above that's taken care of by SplineSplit().)

fontforge/splineutil2.c
```diff
@@ -2879,6 +2880,159 @@ void SplineCharAddExtrema(SplineChar *sc, SplineSet *head,enum ae_type between_s
...
+
+void SplineCharBalance(SplineChar *sc, SplineSet *head, int anysel) { 
+    SplineSet *ss;
+    for ( ss=head; ss!=NULL; ss=ss->next )
+	    SplineSetBalance(sc,ss,anysel);
+}
+
+/* Make the splines adjacent to the SplinePoint sp G2-continuous */
+/* by moving sp between its adjacent control points */
+void SplinePointHarmonize(SplinePoint *sp) {
```
skef, line 2998: Same here, no `SplineRefigure()`s to be seen.

skef (commented):
Given the lack of refigures I'm wondering how extensively this PR was tested in the UI vs just confirming the math works.

fontforge/fontviewbase.c
linusromer: Your suggested generalized solution is surely preferable. I will implement it as soon as I have a little bit more time. This affects `FVHarmonize` and `FVBalance` as well.

fontforge/splineutil2.c
linusromer: Indeed, `SplineRefigure()` is not called ( I just have checked it) although it probably should be called. I HAVE tested my PR and I did not run into problems so far. But my tests were not more systematic than testing some problematic cases and run the functions on some huge fonts. I did not implement a separate torture test for the UI functions. Is there an existing torture test that I have missed (like trip and trap by Donald E. Knuth)?
Anyway, I will add `SplineRefigure()` calls as soon as I have a little bit more time.

skef: There is absolutely no good systematic method of UI testing, unfortunately. When one is adding or making significant modifications to UI-tied functionality it's best to just play around with it in the UI to see what happens. Making real-time changes in a CharView is probably the best way as its most likely to turn up updating problems (such as those caused by a lack of `SplineRefigure()`.

fontforge/fontviewbase.c
linusromer: @skef I am a little stuck here. I have tried your suggestion with a generalized function
`void _FVElementAction(FontViewBase *fv, int anysel, void (*func)(SplineChar*, SplineSet*, enum ae_type, int), const char* progressmsg)` that calls `func(sc, sc->layers[layer].splines, anysel, emsize)`. However my three functions `SplineCharAddInflections()`, `SplineCharBalance()` and `SplineCharHarmonize()` only take 3 parameters, whereas `SplineCharAddExtrema()` takes one additional parameter `emsize`. I think there is no elegant solution of optional parameters in C. There is of course a cheap solution: `void _FVElementAction(FontViewBase *fv, int anysel, const char* functionname)` that contains a `switch (functionname)`. Is there something I am obviously missing?

linusromer:
@skef I have abstracted all occuring AddInflections, Balance, Harmonize in python.c and in the FontView and CharView (the python font method for AddInflections excluded). And I have added `SplineRefigure()` as desired. My tests for this change were not very deep. I will test it more in short time.

linusromer:
@skef After implementing the asked changes I have now tested it more extensively (all Python functions, FontView, CharView, MetricsView) with different fonts (also with quadratic layers). I think it should work now.

iterumllc (approved):
Looks good to me!

po/de.po
```diff
@@ -598,31 +598,31 @@ msgstr "Kein CID mit dem Namen %s"
 
 #, c-format
 msgid "Unparseable include on line %d of %s"
-msgstr "Nicht parsebarer Include in Zeile %d in %s"
+msgstr "Unlesbares Einzufügendes in Zeile %d in %s"
 
 #, c-format
 msgid "Include filename too long on line %d of %s"
-msgstr "Include-Dateiname zu lang in Zeile %d in %s"
+msgstr "Name der einzufügenden Datei zu lang in Zeile %d in %s"
```
jtanx, line 605: This actually needs to be updated in crowdin, I'll find a way of getting it there

fontforge/fontviewbase.c
```diff
@@ -929,6 +929,52 @@ void FVAddExtrema(FontViewBase *fv, int force_adding ) {
...
+	}
+	SCCharChangedUpdate(sc,fv->active_layer);
+	if ( !ff_progress_next())
+    break;
+    }
+    ff_progress_end_indicator();
+}
+
+void FVAddInflections(FontViewBase *fv, int anysel) {
+	_FVElementAction(fv,anysel,&SplineCharAddInflections,"Adding points of inflection...");
```
jtanx, line 967: Due to how gettext works I'm pretty sure you need to use `_()` here and not in _FVElementAction

po/de.po
jtanx: I'll fix this once merged

fontforge/fontviewbase.c
jtanx: fixed

Typedesigners:
@linusromer I think "Add Points Of Inflection / Balance / Harmonize" are very good functions. But it could be extended by allowing you to control the curvature of a curve by entering numbers. For example, you could make an arc in all glyphs, which can be precisely defined with tunni lines. So one could get much more control over the shapes. Like in Fontlab 8, you can define the curvature very precisely by seeing numbers and entering them. Can you please programme this?

linusromer:
Are you talking about the "Edit tunni lines" feature of Fontlab 8? While I
probably could implement this feature as a raw algorithm to change a curve,
I see no easy way to implement a user interface to it.

Am Fr., 7. Apr. 2023 um 18:46 Uhr schrieb Sergej Lebedev <
***@***.***>:

Typedesigners:
@linusromer In FontLab, the tunni lines are given numbers. For example, for an ark shape 60 or 50. This allows you to determine how much an ark is curved. Whether it is flatter or rounder. This way you can create exactly the same round shapes for all letters. At the moment, FontForge does not have an exact way to easily control how much an arc is curved.

https://help.fontlab.com/fontlab-vi/Tunni-Lines/

Typedesigners:
@linusromer Smart Corners - It might be possible to implement a similar function in FontForge.
https://help.fontlab.com/fontlab-vi/Using-Smart-Corners/

linusromer:
The things you wish are clearly a matter of user interface. I am the wrong
person to implement this or other things that have to be displayed on the
canvas. Someone else might be the right person...

Sergej Lebedev ***@***.***> schrieb am Mo., 10. Apr. 2023,
10:34:

Typedesigners:
@linusromer Maybe as a plugin?
It would also be nice to have a function in FontForge that allows you to centre elements horizontally and vertically. Such alignment functions are a matter of course in any graphics programme such as Illustrator or Affinity Designer.

linusromer:
Horizontal and vertical alignment could be solved by a plugin similar to
github.com/linusromer/curvatura. However, I will not work on it (I am not
part of the FontForge team). You could write the plugin on your own. Before
programming you should think of the interface: Should only complete
contours be aligned or only selected nodes? If only complete contours
should be aligned how would the selection work? To which place should it be
aligned? What are the menu entries ("align horizontally" and "align
vertically" ?).

Am Mo., 10. Apr. 2023 um 13:46 Uhr schrieb Sergej Lebedev <
***@***.***>:

Typedesigners:
I am a type designer, not a programmer, so I cannot program a plug-in.

PR #4907, opened 2022-02-04
Compare UTanVec equality by the same standard

### LINKED ISSUE #4884: Quite strange degenerate output from a simple oval
I seem to have somehow drawn an oval FontForge cannot tolerate. Twice, actually, first in `a` and then `d`.

image

No tricks here of any sort, of that I promise you—this is the glyph:

image

[strokebug20220113.zip](https://github.com/fontforge/fontforge/files/7867454/strokebug20220113.zip)

Bug triggered thus:

image

Original glif file:

image

```xml
<glyph name="a" format="2">
  <advance width="461"/>
  <unicode hex="61"/>
  <outline>
    <contour>
      <point x="372" y="392" type="move"/>
      <point x="372" y="392"/>
      <point x="372" y="388.43213"/>
      <point x="372" y="382" type="curve"/>
      <point x="372" y="25" type="line"/>
    </contour>
    <contour>
      <point x="371" y="212" type="move"/>
      <point x="371" y="116"/>
      <point x="319" y="26.000015"/>
      <point x="228" y="26" type="curve"/>
      <point x="137" y="26.000008"/>
      <point x="81" y="116"/>
      <point x="81" y="212" type="curve"/>
      <point x="81" y="308"/>
      <point x="137" y="392"/>
      <point x="228" y="392" type="curve"/>
      <point x="319" y="392"/>
      <point x="371" y="308"/>
      <point x="371" y="212" type="curve"/>
    </contour>
  </outline>
</glyph>

```

Not sure if @skef still cares. Really, not sure why I do—except holding out hope or a nib-based libskef which seems to be more and more extinguished as of late. Reported all the same.

skef:
#4884 was caused by comparing utanvec (a normalized vector representing an angle) equality sometimes using UTMARGIN (1e-7) and sometimes using INTRASPLINE_MARGIN (1e-8). After this change the former is used consistently in the expand stroke code. 

This duplicates one inline function in two C files because I didn't want to change the minimal inclusion in utanvec.h or make the comparison non-inline. (This is negotiable.) 

Closes #4884

### DIFF
--- a/fontforge/splinestroke.c
+++ b/fontforge/splinestroke.c
@@ -69,6 +69,11 @@ static inline int BPNear(BasePoint bp1, BasePoint bp2) {
     return BPWithin(bp1, bp2, INTRASPLINE_MARGIN);
 }
 
+// Not called "BPNear" because this is specific to UTanVecs
+static inline int UTNear(BasePoint bp1, BasePoint bp2) {
+    return BPWithin(bp1, bp2, UTMARGIN);
+}
+
 enum nibtype { nib_ellip, nib_rect, nib_convex };
 
 // c->nibcorners is a structure that models the "corners" of the current nib
@@ -634,7 +639,7 @@ static void BuildNibCorners(NibCorner **ncp, SplineSet *nib, int *maxp,
 	    // Flatten potential LineSameSide permissiveness
 	    nc[i].utv[NC_OUT_IDX] = nc[i].utv[NC_IN_IDX];
 	if (    UTanVecGreater(nc[i].utv[NC_IN_IDX], max_utanangle)
-	     || BPNear(nc[i].utv[NC_IN_IDX], max_utanangle) ) {
+	     || UTNear(nc[i].utv[NC_IN_IDX], max_utanangle) ) {
 	    max_utan_index = i;
 	    max_utanangle = nc[i].utv[NC_IN_IDX];
 	}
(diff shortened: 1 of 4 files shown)


fontforge/splinestroke.c
```diff
@@ -69,6 +69,11 @@ static inline int BPNear(BasePoint bp1, BasePoint bp2) {
     return BPWithin(bp1, bp2, INTRASPLINE_MARGIN);
 }
 
+// Not called "BPNear" because this is specific to UTanVecs
+static inline int UTNear(BasePoint bp1, BasePoint bp2) {
+    return BPWithin(bp1, bp2, UTMARGIN);
```
jtanx, line 74: Seems like BPWithin is only used in splinestroke.c and utanvec.c, why not move BPWithin to utanvec.h, which apparently has the rest of the utility functions on BasePoints already? Then you can move UTNear into that header too.

skef: This would work in the way I would prefer it to if you'll spot me a separate extern declaration of RealWithin.

jtanx: Seems fine to me to include splineutil2.h from utanvec.h, the former is just a bunch of function declarations which seems pretty light, and utanvec.h is already including splinefont.h

skef: OK, I'll take a look.

jtanx (approved):

PR #4909, opened 2022-02-05
Choose a better default background color when no theme is present

jtanx:
Follow-up from https://github.com/fontforge/fontforge/pull/4704#issuecomment-1018253917

### Type of change

- **Bug fix**


skef (approved):
LGTM

PR #4912, opened 2022-02-05
Fix a couple of regressions

jtanx:
See the commit messages for more info

### Type of change

- **Bug fix**

Fix 1: Make these format strings work again (on Windows)
image

Fix 2: Restore this entry
image

### DIFF
--- a/Unicode/memory.c
+++ b/Unicode/memory.c
@@ -29,6 +29,7 @@
 
 #include "ustring.h"
 
+#include <ffglib.h>
 #include <stdio.h>
 #include <string.h>
 
@@ -74,7 +75,7 @@ char *vsmprintf(const char *fmt, va_list args) {
     int len;
 
     va_copy(args2, args);
-    len = vsnprintf(NULL, 0, fmt, args2);
+    len = g_vsnprintf(NULL, 0, fmt, args2);
     va_end(args2);
 
     if (len < 0) {
@@ -88,7 +89,7 @@ char *vsmprintf(const char *fmt, va_list args) {
         return NULL;
     }
 
-    len = vsnprintf(ret, len + 1, fmt, args);
+    len = g_vsnprintf(ret, len + 1, fmt, args);
 
     if (len < 0) {
         free(ret);
--- a/gdraw/gfilechooser.c
+++ b/gdraw/gfilechooser.c
@@ -28,6 +28,7 @@
 #include <fontforge-config.h>
 
 #include <assert.h>
+#include <dirent.h>
 #include <errno.h>
 
 #include "ffglib.h"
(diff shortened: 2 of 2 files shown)


Unicode/memory.c
```diff
@@ -74,7 +75,7 @@ char *vsmprintf(const char *fmt, va_list args) {
     int len;
 
     va_copy(args2, args);
-    len = vsnprintf(NULL, 0, fmt, args2);
+    len = g_vsnprintf(NULL, 0, fmt, args2);
     va_end(args2);
 
     if (len < 0) {
```
skef, line 81: From some internet searching it seems as though part of the difference with (more recent versions of) `g_vsnprintf()` is that it doesn't return -1, it returns the number of bytes that would have been written if they were available. If that's right, we probably shouldn't be leaving the `len < 0` check. (And same with line 93 below.)

jtanx: These checks look ok to me, at line 81 we expect `len` to be equal to the size of the buffer required, so a negative value is unexpected.

Again at line 92, a negative return is unexpected if the first run on line 78 returned the size required and we have a buffer of the correct size.

skef: I guess this boils down to what "unexpected" means. I agree the check is harmless if a negative values is never returned.

jtanx: [The docs](https://docs.gtk.org/glib/func.vsnprintf.html) mention this:

> In versions of GLib prior to 1.2.3, this function may return -1 *if the output was truncated*, and the truncated string may not be nul-terminated. In versions prior to 1.3.12, this function returns the length of the output string.

> The return value of g_vsnprintf() conforms to the vsnprintf() function as standardized in ISO C99. Note that this is different from traditional vsnprintf(), which returns the length of the output string

So:
1. There is a caveat on behaviour on ancient versions of glib not conforming to standard
2. On non-ancient versions, it conforms to C99

So I take it as it behaving semantically the same as the version provided by the C99/Posix standard .

From the manpage of vsnprintf: https://linux.die.net/man/3/vsnprintf

> Upon successful return, these functions return the number of characters printed (excluding the null byte used to end output to strings). 

> If an output error is encountered, *a negative value is returned*. 

So it seems like the behaviour of checking for negative return is correct.

ctrlcctrlv: I don't agree @jtanx, I think @skef is right.

Look at glib source code. Everywhere it's used internally it's unchecked. _Cf_ gmessages.c `g_printf_string_upper_bound`, which returns `gsize` which is `unsigned long`.

Or `g_logv` in the same file which calls it multiple places unchecked for < 0.

I don't know why we need to be more paranoid than the developers of GLib.

skef: There could be older glib versions out there. (Whether they would link in the first place is questionable.) It's harmless, it's fine.

ctrlcctrlv: Was only noting for future—don't particularly care if it's merged as-is, but the check seems useless and of higher paranoia than glib's own developers, so that's suspect. :)

jtanx: That it's not done in various places in glib means little to me, it's relying on its internal implementation which it can make assumptions on. Take `g_printf_string_upper_bound`, it's returning:

```
return _g_vsnprintf (&c, 1, format, args) + 1;
```

Which is clearly not the same as `g_vsnprintf`. You can prove this empirically with this:

```c
#include <glib.h>
#include <stdio.h>

void aa(const char *c, ...) {
    va_list ap;
    va_start(ap, c);
    printf("%d\n", g_vsnprintf(NULL, 0, c, ap));
    va_end(ap);
}

int main() {
    aa("%J %s", 1234, "blah");
}
```

`g_vsnprintf` returns -1 -- because of the invalid format string. So clearly it an return a negative value, so I don't think these checks are 'paranoid'.

My takeaway is that I just don't see the point in trying to treat these any differently to the standard version; they are designed to solve portability issues,  and _improve_ compatibility with the spec, which is why I switched.

Issue #4937, opened 2022-03-09
new release bugs in Windows OS


aminabedi68:
1- in GUI Generate Fonts, format of the name doesn't change with changing font type anymore and if you change font type more than once, fontforge missing the directory:

https://user-images.githubusercontent.com/25493297/157380540-0d78c3bd-fd6d-4e9e-9515-86f8abbc91fe.mp4

2- moving triangle point with garbing ctrl doesn't work like before. this is now:

https://user-images.githubusercontent.com/25493297/157381114-7ab17a14-8952-43e2-af0a-cb68f202b83b.mp4

and this is before:

https://user-images.githubusercontent.com/25493297/157393433-0830cebf-5ad7-49a2-b3d8-e83f9a628015.mp4

jtanx:
I'll look into the first issue.

For that second issue, @skef I'm guessing this might be related to #4685?

skef:
> For that second issue, @skef I'm guessing this might be related to #4685?

Eh, the other control point is still *moving* during most of the drag in that first clip, so #4685 isn't the most obvious candidate. Without casting any aspersions (I reviewed it and it looked great to me) I would consider something in #4789 first. The control-grab functionality might have been relying on certain side-effects in how that function worked. I dunno, I didn't even know that control-grab functionality existed for tangent points.

skef:
Oh, well, maybe I see your point looking at `SplineCharDefaultNextCP()`. I'll take a look at this at some point relatively soon but it may not be in the next few days.

skef:
> I wanted to open an issue for that as a feature request: it's possible to move triangle point with ctrl in the opposite side? i see this need kind of a lot.

I'm not sure I have a clear picture of what you mean.

aminabedi68:
this side:
![bitmap](https://user-images.githubusercontent.com/25493297/157607589-e7514f66-1dcd-4f74-ae01-3df5da839723.png)

PR #4828, opened 2021-10-04
Remove remaining gio remnants

jtanx:
Obsoleted since the removal of the FTP layer; all file system actions
are synchronous so no need to interface through George's gio layer

Mime implemenation was loosely based on
https://github.com/fontforge/fontforge/blob/v20120731-b/gutils/giomime.c

### Type of change

- **Non-breaking change**

PR #4939, opened 2022-03-09
gfilechooser.c: fix behaviour when changing file type

jtanx:
Fixes the first issue raised in #4937.

Regression from #4828 - restores this code: https://github.com/fontforge/fontforge/blob/21ad4a18fb3d4becfe566d8215eba4483b0ddc4b/gdraw/gfilechooser.c#L405-L409

### Type of change

- **Bug fix**


skef (approved):
lgtm

PR #4941, opened 2022-03-09
cmake: Install fonttools and pycontrib

jtojnar:
Restore the files that were installed to fontforge-extras Debian package prior to the FontForge 2020 March Release.

The [patch this is based on](https://salsa.debian.org/fonts-team/fontforge/-/blob/e5764cc23f4aaf2b84d454430aff78328bf75645/debian/patches/0001-add-extra-cmake-install-rules.patch) is also used by the [NixOS package](https://github.com/NixOS/nixpkgs/pull/163439).

### Type of change

- **Non-breaking change**

cc @anthonyfok who created this patch.

### DIFF
--- a/contrib/fonttools/CMakeLists.txt
+++ b/contrib/fonttools/CMakeLists.txt
@@ -18,3 +18,27 @@ target_link_libraries(dewoff PRIVATE ZLIB::ZLIB)
 target_link_libraries(pcl2ttf PRIVATE MathLib::MathLib)
 target_link_libraries(ttf2eps PRIVATE fontforge)
 target_link_libraries(woff PRIVATE ZLIB::ZLIB)
+
+install(
+  TARGETS
+    acorn2sfd
+    dewoff
+    findtable
+    pcl2ttf
+    pfadecrypt
+    rmligamarks
+    showttf
+    stripttc
+    ttf2eps
+    woff
+  RUNTIME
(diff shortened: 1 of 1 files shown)


pycontrib/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
 # Distributed under the original FontForge BSD 3-clause license
 
 install(FILES excepthook.py DESTINATION ${CMAKE_INSTALL_DATADIR}/fontforge/python)
+install(DIRECTORY simple DESTINATION ${CMAKE_INSTALL_DATADIR}/fontforge/python)
```
jtanx, line 4: I don't see any reason for this to be bundled

jtojnar: I can only see that it was installed by the autotools build system, no idea what it is for:

https://github.com/fontforge/fontforge/blob/71cde275d083dbea87b1e8d0a6baaf7144189ea3/pycontrib/simple/Makefile.am#L29

jtanx: It's a bunch of toy scripts, but they're basic/half-broken and are not easily discoverable, so yeah, no reason to bundle it

jtojnar: Ok, dropped that.

PR #4949, opened 2022-03-14
Fix drawing outside of expose calls

jtanx:
This fixes all the cases that I can find where drawing occurs outside of expose calls.

All modern toolkits force you to draw within expose events, and even now, only somewhat works with more than a few hacks in the GDK code.

Fixing this allows some of those hacks to be removed (and at times, removes the need for some triple-buffering)

### Type of change

- **Bug fix**
- **Non-breaking change**


jtanx:
Unless there's any objections I'll be merging this in the next few days

skef (commented):
I looked at these changes and although I don't think I've done a "real" review I get the basic idea and it looks OK to me.

PR #4956, opened 2022-03-20
Fix resolution of program root

### LINKED ISSUE #4953: Some menu not working on MAC
I just installed the latest FF version (20220308) on a 13 'Mac Book Pro with Monterey 12.3.
 I note that:
1. in the Preferences menu it is impossible to select the menu items, both the pull-down menus and the on / off buttons.
2. The same situation for many other checkboxes. For example, Show -- Show H. Metrics, where options cannot be set.
Thank you for your attention

jtanx:
From #4827, the app bundle's entrypoint is now a symlink to the real binary instead of a startup script, which throws off the resolution of the program root.

This fix checks to see if the determined root is valid, and if not, tries to resolve symlinks. Similar to #4626 but preserves existing behaviour if everything else is also symlinked properly.

### Type of change

- **Bug fix**

Fixes #4953


skef (approved):
I feel like you would lecture me about the added statics but I ain't got no problem with 'em. LGTM.

jtanx:
If you're referring to the statics in fsys.c, I'm pretty sure I reduced the net amount

skef:
@jtanx you might not have meant to leave the TRACE in `fsys.c` in this uncommented.

PR #4960, opened 2022-03-21
Fix name extraction for Type3 fonts from pdfs

aidos:
Fixes a bug when extracting Type3 fonts from PDFs. In the existing code, `name` is pointed to the correct place in the font dictionary, but `pdf_getcharprocs` reads the Charprocs dictionary before the name is copied. This causes the later name copy step to read a random section of the CharProcs dict instead. This change copies the name out _before_ loading CharProcs.

### Type of change
 **Bug fix**

### DIFF
--- a/fontforge/parsepdf.c
+++ b/fontforge/parsepdf.c
@@ -1942,6 +1942,8 @@ static SplineFont *pdf_loadtype3(struct pdfcontext *pc) {
 return( sf );
 
   fail:
+    if ( name!=NULL )
+        free(name);
     LogError( _("Syntax errors while parsing Type3 font headers") );
 return( NULL );
 }
--- a/fontforge/parsepdf.c
+++ b/fontforge/parsepdf.c
@@ -1942,8 +1942,7 @@ static SplineFont *pdf_loadtype3(struct pdfcontext *pc) {
 return( sf );
 
   fail:
-    if ( name!=NULL )
-        free(name);
+    free(name);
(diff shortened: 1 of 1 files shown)


fontforge/parsepdf.c
```diff
@@ -1893,7 +1893,9 @@ static SplineFont *pdf_loadtype3(struct pdfcontext *pc) {
 
     name=PSDictHasEntry(&pc->pdfdict,"Name");
     if ( name==NULL )
-	name=PSDictHasEntry(&pc->pdfdict,"BaseFont");
+        name=PSDictHasEntry(&pc->pdfdict,"BaseFont");
+    if ( name!=NULL )
+        name = copy(name+1);
```
jtanx, line 1898: Can you add a `free` of name at the `fail` label

aidos: Done

fontforge/parsepdf.c
```diff
@@ -1942,6 +1942,8 @@ static SplineFont *pdf_loadtype3(struct pdfcontext *pc) {
 return( sf );
 
   fail:
+    if ( name!=NULL )
```
jtanx, line 1945: You can remove the `if` check, a free of NULL is fine

aidos: Gotcha. Done. I appreciate you taking the time to guide me through this - I know teaching the fundamentals of C wasn't on your todo list for today!

PR #4961, opened 2022-03-21
Support ToUnicode extraction from PDFs for Type3 fonts

aidos:
For the most part, fontforge honours ToUnicode maps within PDFs, but not so for Type3 fonts. This change applies the cmaps for Type3 fonts if they're available (they're allowed within the spec and are often supplied).

### Type of change
 ** New feature**

### DIFF
--- a/fontforge/parsepdf.c
+++ b/fontforge/parsepdf.c
@@ -1962,8 +1962,12 @@ static SplineFont *pdf_loadfont(struct pdfcontext *pc,int font_num) {
     if ( !pdf_findobject(pc,pc->fontobjs[font_num]) || !pdf_readdict(pc) )
 return( NULL );
 
-    if ( (pt=PSDictHasEntry(&pc->pdfdict,"Subtype"))!=NULL && strcmp(pt,"/Type3")==0 )
-return( pdf_loadtype3(pc));
+    if ( (pt=PSDictHasEntry(&pc->pdfdict,"Subtype"))!=NULL && strcmp(pt,"/Type3")==0 ) {
+        sf = pdf_loadtype3(pc);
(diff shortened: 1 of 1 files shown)


fontforge/parsepdf.c
```diff
@@ -1962,8 +1962,12 @@ static SplineFont *pdf_loadfont(struct pdfcontext *pc,int font_num) {
     if ( !pdf_findobject(pc,pc->fontobjs[font_num]) || !pdf_readdict(pc) )
 return( NULL );
 
-    if ( (pt=PSDictHasEntry(&pc->pdfdict,"Subtype"))!=NULL && strcmp(pt,"/Type3")==0 )
-return( pdf_loadtype3(pc));
+    if ( (pt=PSDictHasEntry(&pc->pdfdict,"Subtype"))!=NULL && strcmp(pt,"/Type3")==0 ) {
+        sf = pdf_loadtype3(pc);
+        if ( sf != NULL && pc->cmapobjs[font_num] != -1 )
```
jtanx, line 1967: Why not just do this in pdf_loadtype3?

aidos: Makes sense. Will update tomorrow.

aidos: `pdf_loadtype3` doesn't currently take `font_num`, so pushing the work down there would require changing the signature, is that ok to do? Given that the other call to `pdf_getcmap` also happens nearby in `pdf_loadfont`, it seems reasonable to leave it where it is, but happy to go with whichever approach you think makes the most sense.

jtanx: Right, I see. Yep in that case given the other call in the same function happy to leave it here.

jtanx:
Thanks!

PR #4963, opened 2022-03-23
Fix normalisation of absolute paths

jtanx:
#4956 depends on the path being normalised, but the fallback path for older glib versions doesn't do anything if the path is absolute. This changes it to always normalise paths. Also fixes potential buffer overflows on long path names.

### Type of change

- **Bug fix**

### DIFF
--- a/gutils/fsys.c
+++ b/gutils/fsys.c
@@ -217,19 +217,19 @@ char *GFileGetAbsoluteName(const char *name) {
     g_free(abs);
     return GFileNormalizePath(ret);
 #else
-    char buffer[1000];
+    char *buffer, *pt, *spt, *rpt, *bpt;
 
-     if ( ! GFileIsAbsolute(name) ) {
-	char *pt, *spt, *rpt, *bpt;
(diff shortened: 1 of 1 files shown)


gutils/fsys.c
```diff
@@ -217,19 +217,19 @@ char *GFileGetAbsoluteName(const char *name) {
...
 
 	if ( dirname_[0]=='\0' ) {
 	    getcwd(dirname_,sizeof(dirname_));
 	}
-	strcpy(buffer,dirname_);
-	if ( buffer[strlen(buffer)-1]!='/' )
-	    strcat(buffer,"/");
-	strcat(buffer,name);
+
+	buffer = smprintf("%s/%s", dirname_, name);
```
skef, line 229: I'm assuming getcwd won't return a trailing slash, hence removing that check.

skef (approved):
lgtm

gutils/fsys.c
jtanx: More like the code below already deals with removing redundant slashes, so there's no point checking for it here

PR #4966, opened 2022-03-25
Fix menu item “Hints” can show up in wrong language

iclukas:
The menu item “Hints” is incorrectly untranslated in the German translation, since the term ”Hints“ is used in German as well. When there is a second locale active, untranslated items can show up in the secondary language.
I found it also missing in the French translation.

### Type of change

**Bug fix** see screenshot
fontforge-hints


jtanx:
Please set these here: http://crowdin.com/project/fontforge

ctrlcctrlv:
Closing as incorrect filing location.

PR #4980, opened 2022-04-03
Add flags to Python font.transform()

### LINKED ISSUE #4976: Python font.transform(psMat.scale(FACTOR)) does not give the same result as Scale Uniformly in the GUI.
- [x] Screenshot

**Fontforge version (on Arch Linux 5.17.1)**

```
fontforge -v
Copyright (c) 2000-2022. See AUTHORS for Contributors.
 Version: 20220308
 Based on sources from 2022-03-29 01:43 UTC-ML-D-GDK3.
 Based on source from git with hash: 7fec181549677e62cfb16101972f479183082c29
fontforge 20220308
build date: 2022-03-29 01:43 UTC
```

As titled. I would like to be able to perform a Scale Uniformly on all glyphs in a font via Python scripting, but so far I have not been able to replicate the same functionality via font.transform() - which, as far as I'm aware, the only method that I can use to replicate it. In particular, I've noticed the MATH table does not get scaled in Python, but it is (correctly, IMO) when done via the GUI.

**Expected Behaviour**
`font.transform(psMat.scale(FACTOR)` does the same thing as a GUI `Scale Uniformly...` with all glyphs selected (and checkboxes checked. Since font.transform() can only take in one argument, I think it's appropriate it is applied to the categories specified in the checkboxes. Either this, or more options should be exposed via font.transform()).

Note that "by all checkboxes checked", I mean this:

<details>
   <summary>Screenshot</summary>

image
</details>

**Actual Behaviour**
The two methods are not equivalent - in particular, the MATH table does not get scaled in Python.  I do not know what else has not been missed in Python, but I suspect there may be more than just the MATH table that was not scaled.

<details>
   <summary>Screenshots</summary>

image
image
</details>

**Concrete steps to reproduce**
1. Open a font with an OpenType MATH table in the GUI and in Python (i.e. a math font. You can download the `Garamond-Math.otf` that I was working with from here: https://ctan.org/tex-archive/fonts/garamond-math)
2. Scale the font in the GUI with all glyphs selected (`Ctrl+A`) and options as seen in the screenshot above. Scaling factor can be 110% for simplicity
3. Do the same thing in Python i.e.
```
f.selection.all() # f is the variable we loaded the font into
f.transform(psMat.scale(1.10))
```
4. Save both versions (GUI and Python) and open them both up in the GUI. Notice that in particular the MATH table has not been scaled in Python.

**Possible solution/fix/workaround**

Expose a font.scaleUniformly() function in Python or add more parameters to font.transform(), allowing me to choose all the same categories in the GUI for transformation.

skef:
Closes #4976 

This is specific enough that I don't think it needs automated tests. I did basic checks on each of the flags.

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -17967,9 +17967,9 @@ return( NULL );
 return( NULL );
     // Flip fvt_alllayers back
     if ( flags&fvt_alllayers )
-	flags |= fvt_alllayers;
-    else
 	flags &= ~fvt_alllayers;
+    else
+	flags |= fvt_alllayers;
     for ( i=0; i<6; ++i )
 	t[i] = m[i];
     bvts[0].func = bvt_none;
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -9358,7 +9358,7 @@ return( Py_BuildValue( "i", SCValidate(sc,layer,force)) );
 }
 
 /* Transformation flags: see 'enum fvtrans_flags' in baseviews.h */
-static struct flaglist trans_flags[] = {
+static struct flaglist g_trans_flags[] = {
     { "partialRefs", fvt_partialreftrans },
     { "round", fvt_round_to_int },
     FLAGLIST_EMPTY /* Sentinel */
@@ -9375,7 +9375,7 @@ static PyObject *PyFFGlyph_Transform(PyObject *self, PyObject *args) {
     if ( !PyArg_ParseTuple(args,"(dddddd)|O",&m[0], &m[1], &m[2], &m[3], &m[4], &m[5],
 	    &flagO) )
 return( NULL );
-    flags = FlagsFromTuple(flagO,trans_flags,"transformation flag");
+    flags = FlagsFromTuple(flagO,g_trans_flags,"transformation flag");
     if ( flags==FLAG_UNKNOWN )
 return( NULL );
     flags |= fvt_alllayers;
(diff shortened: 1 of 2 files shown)


fontforge/python.c
```diff
@@ -17937,22 +17937,43 @@ return (NULL);
...
     fv = self->fv;
-    if ( !PyArg_ParseTuple(args,"(dddddd)",&m[0], &m[1], &m[2], &m[3], &m[4], &m[5] ) )
+    if ( !PyArg_ParseTuple(args,"(dddddd)|O",&m[0], &m[1], &m[2], &m[3], &m[4], &m[5], &flagO ) )
 return( NULL );
+    flags = FlagsFromTuple(flagO,f_trans_flags,"transformation flag");
+    if ( flags==FLAG_UNKNOWN )
+return( NULL );
+    // Flip fvt_alllayers back
+    if ( flags&fvt_alllayers )
+	flags |= fvt_alllayers;
```
jtanx, line 17970: Why is this needed? Isn't this a no-op?

skef: The intention was to have fvt_alllayers in the python flags be opposite to that in the flags passed down (as its supposed to be on by default). I'll take a look later.

skef: It's going to be Wed evening at the earliest before I could do something about this properly, but I believe the line should read `flags &= ~fvt_alllayers;` . I just goofed it, apparently. If that makes sense to you and you want to integrate early, go ahead.

skef: OK, had time to actually look at and revise this. I just had the two bit-flipping lines in the wrong positions.

PR #4994, opened 2022-04-20
Change which lookup types are suggested for jamo features

dscorbett:
The jamo features 'ljmo', 'tjmo', and 'vjmo' are meant to replace single jamo glyphs with contextual variants. They are not meant for ligating. I have therefore changed which lookup types they are suggested for.

### Type of change
- **Non-breaking change**


ctrlcctrlv:
Right, thanks :)

PR #5044, opened 2022-06-13
Add a preference for saving editor state

khaledhosny:
### Type of change
- **New feature**

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -81,6 +81,7 @@
 
 int UndoRedoLimitToSave = 0;
 int UndoRedoLimitToLoad = 0;
+int SaveEditorState = 1;
 
 static const char *joins[] = { "miter", "round", "bevel", "inher", NULL };
 static const char *caps[] = { "butt", "round", "square", "inher", NULL };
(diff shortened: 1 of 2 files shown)


fontforge/sfd.c
```diff
@@ -2813,17 +2816,19 @@ static int SFD_Dump( FILE *sfd, SplineFont *sf, EncMap *map, EncMap *normal,
 	for ( remap = map->remap; remap->infont!=-1; ++remap )
 	    fprintf(sfd, "Remap: %x %x %d\n", (int) remap->firstenc, (int) remap->lastenc, (int) remap->infont );
     }
-    if ( sf->display_size!=0 )
+    if ( sf->display_size!=0 && SaveEditorState )
 	fprintf( sfd, "DisplaySize: %d\n", sf->display_size );
     if ( sf->display_layer!=ly_fore )
```
jtanx, line 2821: how is this not part of editor state?

khaledhosny: IIRC, this affects the exported font. The display layer is the layer that will be used for exported outlines.

Issue #5059, opened 2022-07-04
Command Injection due to lack of sanitisation of .bdf filename


firmianay:
hi, great project!

I found a vulnerability that seems to exist. Due to lack of proper validation of .bdf filename, this vulnerability allows to inject arbitrary commands and execute them.

bImport -> FVImportBDF -> _SFImportBDF -> system(buf)

```c
static BDFFont *_SFImportBDF(SplineFont *sf, char *filename,int ispk, int toback, EncMap *map) {
    int i;
    char *pt, *temp=NULL;
    char buf[1500];
    BDFFont *ret;

    pt = strrchr(filename,'.');
    i = -1;
    if ( pt!=NULL ) for ( i=0; compressors[i].ext!=NULL; ++i )
	if ( strcmp(compressors[i].ext,pt+1)==0 )
    break;
    if ( i==-1 || compressors[i].ext==NULL ) i=-1;
    else {
	sprintf( buf, "%s %s", compressors[i].decomp, filename );
	if ( system(buf)==0 )
	    *pt='\0';
	else {
```

firmianay:
In addition, there may be some overflow problems, such as the getFontForgeUserDir function, because the length of the environment variable is not checked, which may cause the stack overflow of buffer in getPfaEditEncodings

```c
static char *getPfaEditEncodings(void) {
    static char *encfile=NULL;
    char buffer[1025];
    char *ffdir;

    if ( encfile!=NULL )
        return encfile;
    ffdir = getFontForgeUserDir(Config);
    if ( ffdir==NULL )
        return NULL;
    sprintf(buffer,"%s/Encodings.ps", ffdir);
    free(ffdir);
    encfile = copy(buffer);
    return encfile;
}
```

skef:
https://github.com/fontforge/fontforge/wiki/Community-guidelines#D1

firmianay:
> https://github.com/fontforge/fontforge/wiki/Community-guidelines#D1

ok, sorry to bother you guys, I didn't mean any harm.

ctrlcctrlv:
Yes, no problem at all. Thanks for the report. :)

ctrlcctrlv:
@firmianay Is it okay with you if I consider this bug to involve only the `system(buf)` exploit in `_SFImportBDF`, and you can open another one at your option for the buffer overflow in `getPfaEditEncodings` if you still care about it, which in my mind does indeed fall under the general §D1 guideline and so I'm not interested in patching it?,

[...]

PR #5060, opened 2022-07-05
Shell-quote command sent to compressor in BDF

ctrlcctrlv:
Fixes security issue in #5059.

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -3269,7 +3269,7 @@ int SFDWriteBakExtended(char* locfilename,
 
 
 int SFDWriteBak(char *filename,SplineFont *sf,EncMap *map,EncMap *normal) {
-    char *buf=0, *buf2=NULL;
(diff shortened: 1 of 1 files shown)


fontforge/sfd.c
```diff
@@ -3326,7 +3326,9 @@ int SFDWriteBak(char *filename,SplineFont *sf,EncMap *map,EncMap *normal) {
     if ( ret && sf->compression!=0 ) {
 	unlink(buf2);
 	buf = malloc(strlen(filename)+40);
```
skef, line 3328: Probably best to switch this to:

```
    qbuf = g_shell_quote(filename);
    buf = malloc(strlen(qbuf)+40);
```

given that `40` is presumably sized against `compressors` and `g_shell_quote` doesn't document how much it may extend the string length.

skef (approved):
LGTM

Issue #5093, opened 2022-08-25
Bugs in Splinefit, and ideas for improvements


Alatius:
In my current font project I do extensive simplification of autotraced splines, and have generally been very pleased with the result of @linusromer's implementation of Raph Levien's curve fitting algorithm. However, it has also lead to me discovering some bugs and undesireable behaviour. Please consider the following font file, which contains a series of short curve segments, each of which demonstrates a problem: [splinefit-bugs.sfd.zip](https://github.com/fontforge/fontforge/files/9428495/splinefit-bugs.sfd.zip)

The first one simplifies to a weird, extremely compressed S-shape, which traces the original curve multiple times, by going up and down and then up again. The solution seems straightforward: when calculating the minimum quadratic distance sum, I think that the shortest distance mapping from the mid points to the approximation should be forced to be monotonic. My fix can be seen in commit 2a0b4a631940f78fefa72c0f4adf65967f594e24 to my fork.

The second curve (in character "2") ends in a point which has a control point handle of length 0. This triggers the second bug: the tangent at the end is by mistake estimated based on the slope towards the end of the _first_ spline, not the last. I correct this in my second commit, 15d5ca921ceb1adf2178a0babd6c1613b9f8967b. (I should point out though that I do not fully understand the code segment at lines 751-755. Under what situation, if any, does this occur? Is the value of the temp variable correct and relevant, considering my change? Should both from->next and to->prev be set to NULL?)

That second fix also saves the third curve from becoming heavily distorted. But its simplified curve is far from ideal. Curves 4 and 5 are still distorted. However, I have found that all of these will become much improved (perfect for my needs), if I allow the handles to extend beyond the Tunni point (i.e., by not limiting aMax and bMax). (Doing so also enables the exact simplification of the constructed curve 6.) However, I acknowledge that removing this restriction generally may introduce many unwanted inflection points. I hope that it should be possible to come up with sensible criteria for when longer handles should be permissible -- perhaps it is as simple as checking if the original curve segment has an inflection point, and allow it then? But I have not thought this over in detail, and I am not sure how to implement that in the best way. What are your thoughts on this?

In any case, i.e., even if the solution isn't to twiddle with aMax and bMax, the distortion currently seen in the simplification of 4 and 5 would have to be amended somehow, of course.

linusromer:
> I think that the shortest distance mapping from the mid points to the approximation should be forced to be monotonic

This is probably a good idea.

[...]

PR #5095, opened 2022-08-30
splinefit.c: Improvements to merge/simplify

Alatius:
**Bug fix** for problems discussed in #5093. My changes are described and explained there.

### DIFF
--- a/fontforge/splinefit.c
+++ b/fontforge/splinefit.c
@@ -748,17 +748,16 @@ return( SplineMake3(from,to));
 	fromunit.x = from->me.x-from->prevcp.x; fromunit.y = from->me.y-from->prevcp.y;
     }
     flen = sqrt(fromunit.x*fromunit.x + fromunit.y*fromunit.y);
-    if ( tlen==0 || flen==0 ) {
-	if ( from->next!=NULL )
-	    temp = *from->next;
-	else {
-	    memset(&temp,0,sizeof(temp));
-	    temp.from = from; temp.to = to;
-	    SplineRefigure(&temp);
-	    from->next = to->prev = NULL;
-	}
+    if ( (tlen==0 || flen==0) && (from->next==NULL || to->prev==NULL) ) {
+	memset(&temp,0,sizeof(temp));
+	temp.from = from; temp.to = to;
+	SplineRefigure(&temp);
+	from->next = to->prev = NULL;
     }
     if ( tlen==0 ) {
+	if ( to->prev!=NULL ) {
+	    temp = *to->prev;
+	}
 	if ( (to->pointtype==pt_curve || to->pointtype==pt_hvcurve) &&
 		to->next && !to->nonextcp ) {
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
@linusromer: Thoughts? :)

linusromer:
I have tested the new "merge" and "simplify" with some fonts and some special situations: I believe that [https://github.com/fontforge/fontforge/pull/5095/commits/2a0b4a631940f78fefa72c0f4adf65967f594e24](https://github.com/fontforge/fontforge/pull/5095/commits/2a0b4a631940f78fefa72c0f4adf65967f594e24) is a good idea - it may even speed up things slightly. 
I am not that sure about [https://github.com/fontforge/fontforge/pull/5095/commits/15d5ca921ceb1adf2178a0babd6c1613b9f8967b](https://github.com/fontforge/fontforge/pull/5095/commits/15d5ca921ceb1adf2178a0babd6c1613b9f8967b) but for me it seems to work and fix the bug. This may need a bit more testing with expanding strokes.

linusromer:
I have tested the new "merge" and "simplify" with some fonts and some special situations: I believe that [https://github.com/fontforge/fontforge/pull/5095/commits/2a0b4a631940f78fefa72c0f4adf65967f594e24](https://github.com/fontforge/fontforge/pull/5095/commits/2a0b4a631940f78fefa72c0f4adf65967f594e24) is a good idea - it may even speed up things slightly. 
I am not that sure about [https://github.com/fontforge/fontforge/pull/5095/commits/15d5ca921ceb1adf2178a0babd6c1613b9f8967b](https://github.com/fontforge/fontforge/pull/5095/commits/15d5ca921ceb1adf2178a0babd6c1613b9f8967b) but for me it seems to work and fix the bug. This may need a bit more testing with expanding strokes.

Alatius:
> I am not that sure about [15d5ca9]

Is there anything in particular that makes you hesitate? The change to use `to->prev` when `tlen==0` seems to me to be quite straight-forward. To the extent that I myself have doubts, it boils down to me not quite understanding under what circumstances the code block starting at line 752 is even supposed to be executed, i.e., I don't know how it can occur that `from->next == NULL` (or `to->prev == NULL`). Lacking input data to test on where this is the case, I have never seen this code being run, and thus I have been unable to test the effect. But in any case, I don't see how it would have any adverse effects (supposing that it is ever executed), compared to the result of the code before this commit.

linusromer:
> Is there anything in particular that makes you hesitate?

No. Many things in FontForge depend on `splinefit.c` and sometimes it is hard to predict how it will affect things in FontForge (e.g. rounding problems of near zero values seem to have been quite a struggle). I would not have commented on this particular commit if not ctrlcctrlv had asked me about my thoughts. So I am just saying: I am not sure about this commit.

ctrlcctrlv:
I've merged it as it also looks fine to me, but you are indeed right that it's hard to predict. I have a ton of fonts that use Expand Stroke though and none broke, sometimes the only way we can get feedback on it is to merge. Wouldn't be upset if it's reverted of course.

PR #5110, opened 2022-09-23
Upgrade to Unicode 15.0.0 and fix and expand the script lists

dscorbett:
This upgrades the Unicode files to version 15.0.0 and the script tags to the latest script tag registry. It also fixes the set of Devanagari code points, which were accidentally a copy of the set of Deseret code points.

### Type of change
- **Bug fix**
- **Non-breaking change**


iandoug:
Is the Unicode version used, shown somewhere?  e.g. on the Unicode Ranges  panel?

Thanks.

PR #5120, opened 2022-10-07
Fix logic for CharString double-movetos after PR 4685

skef:
See discussion at the end of #4685


frank-trampe (approved):
This looks inherently safe (to the extent that there's no meaning in a previously set control point at `rmoveto`) and makes sense in the context of the "3270Medium.ttf" use case in #4685.

Issue #3084, opened 2017-06-09
please delete Adobe-Identity cidmap, or treat it specially.


HinTak:
I think in the past or even current, Adobe Acrobat Reader, and also the (x)dvipdfmx/XeTeX people treat(ed) the Adobe-Identity cidmap as a direct cid/gid to Unicode mapping, for the purpose of text extraction and what not.

Well, the Adobe Source CJK fonts, and Google Noto CJK fonts definitely not - in their usage, it is just "because I said so" custom encoding and carries no meaning whatsoever in relation to Unicode.

Unfortunately fontforge takes the former interpretation, and this causes the corruption of the encoding vector during Re-encode ( https://github.com/fontforge/fontforge/issues/3080 ). Simply deleting the file and forces fontforge not to treat Adobe Identity as direct cid->Unicode mapping, forces fontforge to treat the cmap properly and fixes the problem seen in https://github.com/fontforge/fontforge/issues/3080 .

So I suggest either simply deleting the cidmap, or at least provide an scripting API to selectively disable its use. Otherwise you'll not be able to process Adobe Source CJK / Google Noto CJK properly.

HinTak:
This is also the main cause of https://github.com/fontforge/fontforge/issues/3079 .

HinTak:
I think only cid-key fonts embedded in pdf's should be treated as unicode direct - standalone opentype fonts with CFF outlines should use the cmap, and only the cmap, for coding purposes.

HinTak:
One suggestion I might make would be to make it a user preference.

Anyway, my solution was to rename the file temporarily so that fontforge cannot find it. 

This comes about because Ubuntu ships it separate as extra, and can convert the font somewhat more correctly than Fedora's. Fedora ships fontforge complete as one package. See the entire May/June traffic ( http://lists.nongnu.org/archive/html/cjk-list/ ) .

HinTak:
See also comment from @kenlunde http://typedrawers.com/discussion/comment/28483#Comment_28483

kenlunde:
[...]

PR #5131, opened 2022-10-21
Restore `Adobe-Identity-0.cidmap` for those who need it.

ctrlcctrlv:
This file was deleted by @dscorbett in #4993.

I am cognizant of his and @HinTak's thought in #3084 that the file is not useful to most font developers due to how FontForge handles it.

Nevertheless I need it sometimes. Further, the way it was deleted would keep it around due to the way `ninja install` works.

This adds a CMake build flag called `INSTALL_IDENTITY_CIDMAP`, default `OFF`.

### Type of change

- **Non-breaking change**


HinTak:
I think the file itself is fairly standard, and there is a file of equivalent content (due to copyright/license issues, not verbatim copy) in the Adobe family and elsewhere. I think the important issue is correcting how fontforge treats it, actually. (Or get fontforge to ignores it internally, for the time being).

HinTak:
I think the file itself is fine, and all Adobe products, and some non-Adobe font software capable of dealing with CID fonts too, bundles a version/equivalent of it. It is Fontforge's way of using of it that's the problem...

ctrlcctrlv:
_Sometimes_ a problem. Forcing CID to GID has some uses in Python scripting, etc. Using the Identity CIDMap is the only way really to get the CID for a glyph.

ctrlcctrlv:
Either way, this PR doesn't change the default behavior you've established so arguing it doesn't really have a purpose :)

Users have to pass `-DINSTALL_IDENTITY_CIDMAP` to cmake to get this cidmap installed. It's off by default.

contrib/fontforge/cidmap/CMakeLists.txt
```diff
@@ -14,15 +17,34 @@ if(ENABLE_MAINTAINER_TOOLS)
...
+  list(REMOVE_ITEM CIDMAPS Adobe-Identity-0.cidmap)
+endif()
+
+install(
+  FILES ${CIDMAPS}
+  DESTINATION "${CIDMAP_DIR}"
   COMPONENT cidmaps
 )
+
+if(NOT INSTALL_IDENTITY_CIDMAP)
```
jtanx, line 42: I don't see why this is special cased like so, especially when none of the other installed files do this.

If you don't want it there do a clean install.

contrib/fontforge/cidmap/CMakeLists.txt
```diff
@@ -14,15 +17,34 @@ if(ENABLE_MAINTAINER_TOOLS)
...
+    Adobe-Identity-0.cidmap
     Adobe-Japan1-5.cidmap
     Adobe-Japan1-6.cidmap
     Adobe-Japan1-7.cidmap
     Adobe-Japan2-0.cidmap
     Adobe-Korea1-2.cidmap
-  DESTINATION ${CMAKE_INSTALL_DATADIR}/fontforge/cidmap
+)
+
+if (NOT INSTALL_IDENTITY_CIDMAP)
```
jtanx, line 32: I'd invert this and append if the option is enabled

CMakeLists.txt
```diff
@@ -18,11 +18,11 @@ if(${CMAKE_VERSION} VERSION_LESS "3.15.7")
 endif()
 
 # Include any required modules
+include(GNUInstallDirs)
```
jtanx, line 21: Why is this reordered?

contrib/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # Distributed under the original FontForge BSD 3-clause license
 
-add_subdirectory(cidmap)
+add_subdirectory(fontforge)
```
jtanx, line 3: Why is this reordering needed?

contrib/fontforge/cidmap/CMakeLists.txt
ctrlcctrlv: Oh, just thought it'd help fix bugs on some people's machines. I don't mind removing it.

contrib/CMakeLists.txt
ctrlcctrlv: Hi. I did it while I was here because I thought the eventual plan was to have directories under contrib/ that are actually… well, contrib. Not project-authored, but useful third party scripts. It felt very confusing at the time, but I submitted this so long ago I have no real opinion anymore Jeremy, so if you think it sucks I can move it all back, but that'll be more work lol.

jtanx: This is fine I guess

contrib/fontforge/cidmap/CMakeLists.txt
ctrlcctrlv: > Nevertheless I need it sometimes. Further, the way it was deleted would keep it around due to the way ninja install works.

Please confirm whether you think this explanation is good enough. I don't think that the directory will get deleted even by a clean install especially if it contains other user-provided CIDMaps, so I don't think a clean install is really a good solution..

Issue #5137, opened 2022-11-10
Assigning to fontforge.font.bitmapSizes broken/impossible?


nabijaczleweli:
### When reporting a bug/issue:
- [x] Screenshot: 
    ```
    >>> fontforge.font().bitmapSizes = (16, 32, 64, 128)
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
    SystemError: new style getargs format but argument is not a tuple
    ```
- [x] The FontForge version and the operating system you're using: 8ce3216d492d4a249ffec163c005d098df28e58c on sid (also 1:20201107\~dfsg-4 (bullseye) and 1:20220308\~dfsg-1 (sid))
- [x] The behavior you expect to see, and the actual behavior: as [per spec](https://fontforge.org/docs/scripting/python/fontforge.html#fontforge.font.bitmapSizes): to be able to assign a tuple there, and actually I'm not
- [x] Steps to reproduce the behavior: see p. 1

To ensure I wasn't holding it wrong I also did:
```
>>> fontforge.open('Untitled1.sfdir').bitmapSizes
(16, 32, 64, 128)
>>> fontforge.font().bitmapSizes = (16, 32, 64, 128)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
SystemError: new style getargs format but argument is not a tuple
>>> fontforge.font().bitmapSizes = fontforge.open('Untitled1.sfdir').bitmapSizes
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
SystemError: new style getargs format but argument is not a tuple
```
so idk.

What may be relevant and what the template doesn't ask for is the Python version: 3.10.6-1 and 3.9.2-3

iorsh:
Fixed by #5138

PR #5138, opened 2022-11-10
Don't require individual tuple encapsulation in fontforge.font.bitmapSizes setter

nabijaczleweli:
Documentation specs that this should take the same format as font.bitmapSizes returns: it doesn't;
```
>>> f.bitmapSizes = ((1,), (2,))
```
yields
```
>>> f.bitmapSizes
(1, 2)
```

Which is fun, but what actually happens is the user gets
```
>>> f.bitmapSizes = (1, 2)
  File "<stdin>", line 1, in <module>
SystemError: new style getargs format but argument is not a tuple
```
and gives up, as in #5137. This is also, obviously, not reversible, unlike specced.

The new behaviour is congruent with the documentation (error cases included at the end):
```
(sid-with-nab ff)root@tarta:/srv/fontforge/build# bin/fontforge -
Copyright (c) 2000-2022. See AUTHORS for Contributors.
 Version: 20220308
 Based on sources from 2022-11-10 01:12 UTC-ML-D-GDK3.
 Based on source from git with hash: 8ce3216d492d4a249ffec163c005d098df28e58c
>>> f=fontforge.font()
>>> f.bitmapSizes = (16, 32, 64, 128)
Number out of range: 2.14748e+09 in type2 output (must be [-65536,65535])
>>> f.bitmapSizes
(16, 32, 64, 128)
>>> f.bitmapSizes = ()
>>> f.bitmapSizes
()
>>> f.bitmapSizes = (48,)
Number out of range: 2.14748e+09 in type2 output (must be [-65536,65535])
>>> f.bitmapSizes
(48,)
>>>
>>>
>>> f.bitmapSizes = (48,())
  File "<stdin>", line 1, in <module>
TypeError: 'tuple' object cannot be interpreted as an integer
>>> f.bitmapSizes = 48
  File "<stdin>", line 1, in <module>
SystemError: ../Objects/tupleobject.c:172: bad argument to internal function
```

idk what the ERANGEs are but they're unrelated

### Type of change
- **Bug fix**: #5137

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -12003,7 +12003,8 @@ return(-1);
 return( -1 );
     sizes = malloc((cnt+1)*sizeof(int));
     for ( i=0; i<cnt; ++i ) {
-	if ( !PyArg_ParseTuple(PyTuple_GetItem(value,i),"i", &sizes[i])) {
+	sizes[i] = PyLong_AsLong(PyTuple_GetItem(value,i));
+	if (PyErr_Occurred() && !PyArg_ParseTuple(PyTuple_GetItem(value,i),"i", &sizes[i])) {
 	    free(sizes);
 return( -1 );
 	}


fontforge/python.c
```diff
@@ -12003,7 +12003,8 @@ return(-1);
 return( -1 );
     sizes = malloc((cnt+1)*sizeof(int));
     for ( i=0; i<cnt; ++i ) {
-	if ( !PyArg_ParseTuple(PyTuple_GetItem(value,i),"i", &sizes[i])) {
+	sizes[i] = PyLong_AsLong(PyTuple_GetItem(value,i));
```
iorsh, line 12006: This fixes the issue, but will break the existing scripts for those who managed to figure out this peculiarity.

On one hand, preserving existing state has merits, but on the other hand we do break things from time to time, like when migrating to Python3 or updating AGL for new fonts.

@skef, what do you think?

skef: Yeah, I'm fine with standardizing on flat ints but the code should still accept the singleton tuples in case scripts have relied on that behavior. It's not that hard to code that way.

nabijaczleweli:
Updated to allow both
```
>>> f=fontforge.font()
>>> f.bitmapSizes = (16, 32, 64, 128)
>>> f.bitmapSizes = ((32,), (64,), (128,), (256,))
```

skef (approved):
lgtm

PR #5163, opened 2022-12-30
Fix painting behaviour of the ruler linger window and layer change dialog

### LINKED ISSUE #5119: GGDKDraw assertion failure when using ruler
@jtanx when I use the ruler tool in the current master (Arch Linux, debug build) I get an assertion failure just when the tool snaps to the second point: `fontforge: /home/skef/src/ff3/gdraw/ggdkcdraw.c:52: _GGDKDraw_CheckAutoPaint: Assertion 'false && "No window should be drawing outside of expose calls"' failed.`

I've only tried with a couple files (always cubic so far) but it seems to happen every time.

jtanx:
Fixes #5119

### Type of change

- **Bug fix**

### DIFF
--- a/fontforgeexe/cvruler.c
+++ b/fontforgeexe/cvruler.c
@@ -577,7 +577,7 @@ static void RulerLingerPlace(CharView *cv, GEvent *event) {
     GDrawSetVisible(cv->ruler_linger_w,true);
 }
 
-static void RulerLingerMove(CharView *cv) {
+void CVRulerLingerMove(CharView *cv) {
     CharViewTab* tab = CVGetActiveTab(cv);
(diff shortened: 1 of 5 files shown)


fontforgeexe/cvruler.c
```diff
@@ -1003,6 +1003,5 @@ return;
 	    }
 	    prev_rect = rect;
 	}
-	RulerLingerMove(cv);	/* in case things are moving or scaling */
```
jtanx, line 1006: Instead of calling this from the expose event, I call it in response to the charview doing a scroll or zoom.

skef (approved):
lgtm

skef:
This second commit looks OK too, given that the timer is set just before.

PR #5166, opened 2022-12-31
ufo.c: emit guidelines key

### LINKED ISSUE #4951: fontmake crashes with new fontforge ufo output
Hi
fontmake crashes during font build process with new fontforge, it seems that the only difference is related to new entry `<key>guidelines</key>` in `fontinfo.plist` file.

it's a fontforge issue or fontmake?

jtanx:
Fixes #4951

### Type of change

- **Bug fix**


frank-trampe (approved):
This looks fine. The existing behavior is broken enough that I see no reason to warn people about the change.

PR #5167, opened 2022-12-31
fvmetrics.c: partial revert of 1033bc6

### LINKED ISSUE #4955: Wrong behavior of: Metrics - Set Both Bearing
I'm using last official FF Appimage (FontForge-2022-03-08-582bd41-x86_64.AppImage) on Debian Testing.
I start from a glyph that has _n_ pt of bearing on both sides. So I resort to **Metrics - Set Both Bearing** by setting them to _n + x_, LSB is correctly set to _n + x_, but RSB il set to _n + 2x_. It is as if the increase is applied twice to the right.
The bug is not specific to the operating system, as the same problem occurs in the MAC version.
This problem does not appear, from my tests, if instead of increasing the measure I reduce it.

jtanx:
Fixes #4955

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/fvmetrics.c
+++ b/fontforge/fvmetrics.c
@@ -160,7 +160,7 @@ return;
 	    SCSynchronizeWidth(sc,width,sc->width,fv);
 	}
 	if ( transform[4]!=0 ) {
-	    FVTrans(fv,sc,transform,NULL,fvt_alllayers);
+	    FVTrans(fv,sc,transform,NULL, fvt_dontmovewidth | fvt_alllayers);
 	    bvts[0].x = transform[4];
 	    for ( bdf = fv->sf->bitmaps; bdf!=NULL; bdf=bdf->next ) if ( bdf->glyphs[sc->orig_pos]!=NULL )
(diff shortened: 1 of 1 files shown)


frank-trampe:
@ctrlcctrlv, would you review this one? It's going to take either of us a while to process the relevant code, but you have seen more of it than I have and probably have better test cases.

ctrlcctrlv:
Yes, I can.

ctrlcctrlv:
God, this part of the code is such a slog and has such a complicated history, but I think I understand it. The most important commit is bbde2a58b350dfde29d2700a3606b6fb9c6752c2 as that's where George tried to fix this originally in 2012 and is likely a good version, but it predates certain additions like `fvt_addlayer`. The version @jtanx proposes seems to match in the ways that matters.

To be certain, though, I did the following runtime tests:
* lbearing to 200 image
* both bearings to 113 image
* rbearing to 90 image
* width to 500 image
* and then both back to 200 image

Given all that, I think this is fine to merge. It's certainly less broken than master, and I found no issues.

PR #5171, opened 2022-12-31
lookups.c: fix variable used in ApplyAppleStateMachine

### LINKED ISSUE #4985: Crash from the Metrics window after adding classes to Mac Contextual State Machine
- [x] [YouTube video link](https://youtu.be/Abj9fVZwPWo)
FontForge-2022-03-08-74e2eca.app
OS Version:            Mac OS X 10.15.7 (19H1824)

Selecting a lookup in the Metrics window applies or removes that lookup to the text in the window.

Selecting a default Mac Contextual Substituion lookup does nothing.

Selecting a Mac Contextual Substituion with an added glyph class causes FontForge to crash.

1. Add a Mac Contextual Substituion lookup
2. Add a subtable
3. Add a glyph to a new glyph class in the subtable. OK out of all the windows to apply the new lookup.
4. Open a new Metrics window.
5. Select the Mac Contextual Substitution lookup

jtanx:
Maybe fixes #4985

Not entirely sure tbh but this seems to be more correct based on the context of the original commit https://github.com/fontforge/fontforge/commit/1827b9dfe169348f860fff1b524eaeae61007b2d

And it does fix a crash that I observed

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/lookups.c
+++ b/fontforge/lookups.c
@@ -2996,7 +2996,7 @@ static void ApplyAppleStateMachine(OTLookup *otl,struct lookup_data *data) {
 		class = 0;
 	    else {
 		for ( class = sm->class_cnt-1; class>3; --class )
-		    if ( GlyphNameInClass(data->str[i].sc->name,sm->classes[class]) )
+		    if ( GlyphNameInClass(data->str[pos].sc->name,sm->classes[class]) )
 		break;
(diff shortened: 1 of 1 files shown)


skef (approved):
lgtm

jtanx:
Merge seems to have glitched, I see this on master now...

PR #5199, opened 2023-02-10
Remove duplicate namelist entry (Rfractur 0x211C)

12Me21:
`namelist.c` has a duplicate entry in `psaltnames`:
line 20085: `{ "Rfractur", 0x211c, 0 },`
line 20712: `{ "Rfractur", 0x211C, 0 },`

### Type of change
- **Non-breaking change**

### DIFF
--- a/fontforge/namelist.c
+++ b/fontforge/namelist.c
@@ -20709,7 +20709,6 @@ static struct psaltnames psaltnames[] = {
 	{ "circlecopyrt", 0xa9, 0 },
 	{ "smile", 0x263a, 0 },
 	{ "Ifractur", 0x2111, 0 },
-	{ "Rfractur", 0x211C, 0 },
 	{ "Omegainv", 0x2127, 0 },
 	{ "mho", 0x2127, 0 },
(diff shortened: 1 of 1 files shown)


iorsh:
Quite a lot of names are appearing twice in `psaltnames`. For example, `Shcha`. I agree this is suboptimal, but does it cause some real issue? Glyph name selector in "Glyph Info" dialog seems to filter out duplicate entries.

iorsh:
This change doesn't seem useful to me. There are many duplications in `psaltnames`, and they don't seem to cause any undesired effects.

skef (commented):
I have mixed feelings about this sort of PR. On the one hand merging this change would probably be fine. On the other, this is not a systematic solution to duplication (there are other duplicated values), and there could be a (perhaps dubious) rationale for the duplication: the comment indicates these are copied from another source, and maybe it makes sense to keep the copied values grouped together, even at the cost of some duplication. 

I think for now I'd rather leave things as they are.

iorsh:
Closing. The entry duplication is an artifact of combining several sources and should be preserved for reference.

PR #5221, opened 2023-03-23
[Python] fixes for generateTtc,  os2_unicoderanges, and font.os2_codepages

IsaacOscar:
A couple of very simple fixes for the Python API.

First I make [`font.generateTttc`](https://fontforge.org/docs/scripting/python/fontforge.html#fontforge.font.generateTtc) default to using `layer=font.activeLayer`. This is the same behaviour as [`font.generate`](https://fontforge.org/docs/scripting/python/fontforge.html#fontforge.font.generate).
Secondly, I allow setting [`font.os2_unicoderanges`](https://fontforge.org/docs/scripting/python/fontforge.html#fontforge.font.os2_unicoderanges) and [`font.os2_codepages`](https://fontforge.org/docs/scripting/python/fontforge.html#fontforge.font.os2_codepages) to `None`, thus causing FontForge to automatically generate them (this is why clicking the "Default" checkbox in the Font-Info GUI does).

In particular, before my change:
```
font = fontforge.open(....)
...

font.generate("hello.ttf") // same as font.generate("hello.ttf", layer=font.activeLayer)
// the above works as expected

font.os2_unicoderanges = None // raises SystemError: new style getargs format but argument is not a tuple
font.os2_codepages= None // ditto
font.generateTtc("hello.ttc") // same as font.generateTttc("hello.ttc", layer=0)
// the above will raise a "ValueError: Layer is out of range"
```
And after my change:
```
...
print(font.os2_unicoderanges) // prints  something like (-452980993, 1342203387, 0, 0)
font.os2_unicoderanges = None
print(font.os2_unicoderanges) // Now prints something different, like (-1, -65537, -469762049, 134217663)

print(font.os2_codepages) // prints something like (536871359, 0)
font.os2_codepages= None
print(font.os2_codepages) // Now prints something different, like (1613693439, -65536)

font.generateTtc("hello.ttc") // same as font.generateTttc("hello.ttc", layer=font.activeLayer)
// the above now works as expected
```

Working out why my `generateTtc` calls where failing took quite a while, as I had no idea wherethe layer that was "out of range" came from. The [documentation](https://fontforge.org/docs/scripting/python/fontforge.html#fontforge.font.generateTtc) did not mention this difference.

Before you could get the `= None` behaviour if you could somehow do `= NULL`, where `NULL` is a `C` null-pointer, but I don't think that was possible.

### Type of change
- **Bug fix** (I havn't bothered to create an Issue for this)

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -13852,7 +13852,7 @@ static int PyFF_Font_set_os2codepages(PyFF_Font *self,PyObject *value, void *UNU
     if ( CheckIfFontClosed(self) )
 return (-1);
     sf = self->fv->sf;
-    if ( value == Py_None ) {
+    if ( value == NULL ) {
         sf->pfminfo.hascodepages = false;
 return( 0 );
     }
@@ -13883,7 +13883,7 @@ static int PyFF_Font_set_os2unicoderanges(PyFF_Font *self,PyObject *value, void 
     if ( CheckIfFontClosed(self) )
 return (-1);
     sf = self->fv->sf;
-    if ( value == Py_None ) {
+    if ( value == NULL ) {
         sf->pfminfo.hasunicoderanges = false;
 return( 0 );
     }


iorsh:
The change for `font.generateTtc()` is fine, it aligns the actual behavior with the documentation.

Allowing `None` in `os2_codepages` and `os2_unicoderanges` is misleading. For example, `font.os2_codepages` is defined as tuple of two values, and the proper way to reset it is to assign `font.os2_codepages = (0, 0)`

PR #5242, opened 2023-05-28
Always set `usDefaultChar` to 0 (.notdef)

dscorbett:
`usDefaultChar` is the code point to use for a default glyph for an unsupported code point. If it is 0, the font uses glyph 0 (.notdef), which is always what you want instead of FontForge’s current behavior of defaulting to U+0020 SPACE.

### Type of change
- **Non-breaking change**

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -3509,8 +3509,6 @@ docs are wrong.
         os2->capHeight = (caph >= 0.0 ? caph : 0);
     }
 	os2->defChar = 0;
-	if ( format==ff_otf || format==ff_otfcid )
-	    os2->defChar = ' ';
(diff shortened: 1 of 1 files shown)


iorsh:
@dscorbett, could you please rebase to the latest `fontforge:master`?

This generally looks safe to me. I couldn't see any specific Linux software incorrectly using "space" as a default character, but I recall such behavior in Windows. I agree that ".notdef" is better than "space", since it allows missing characters to stand out when displayed, instead of being quietly replaced with space.

iorsh:
Per GW (https://github.com/fontforge/fontforge/commit/224315a) is was a guess, so probably safe to sacrifice.

skef (approved):
This seems good

PR #5246, opened 2023-06-12
Fix crashes on importing Mac-style dfont/NFNT bitmap fonts with some legal (but unusual?) values 

robhagemans:
This PR contains two commits fixing different crashes due to out-of-bounds array access when reading certain dfont bitmap font files.
- the first crash occurs when a `FOND` resource has a nonzero `startChar` value in its header, and a family glyph-width table is present. The out-of-bounds access is caused as the array index is not adjusted for the start value in the loop.
- the second crash occurs when a `FOND` resource as a Style Mapping table with a Style Index Table with any zero-valued entries. This leads to accessing array index -1.

### Type of change

- **Bug fix** - fixes a bug that is not presently reported as an Issue


winlover32 (approved):

iorsh (approved):
Looks good

PR #5257, opened 2023-07-07
Fix errors in French and Italian translations

yselkowitz:
With gettext-0.22, these format mismatches trigger errors:

`'msgstr' is not a valid C format string, unlike 'msgid'. Reason: In the directive number 2, the argument size specifier is invalid.`

### Type of change
- **Bug fix**

Fixes: #5251


ctrlcctrlv:
I'm sorry, but merging this would cause issues because FontForge translations are managed via Crowdin. Please make your changes there. _Cf._ https://github.com/fontforge/fontforge/wiki/Updating-translations-from-crowdin. I will add a note to [Community guidelines](https://github.com/fontforge/fontforge/wiki/Community%20guidelines) page.

PR #5262, opened 2023-07-17
make harmonization robust and avoid zero handles after harmonization

linusromer:
**Non-breaking change** 
After my presentation at the TUG Conference 2023, Jan Šustek correctly made me aware of the fact, that my harmonization formula is not numerically stable for some cases and that I should use the [citardauq formula](https://en.wikipedia.org/wiki/Quadratic_formula) instead. Which I have done now in the code.

Furthermore harmonization is now no longer applied if it would produce a zero handle (because then the curvature would be either infinitely large, so it wouldn't be continuous, or it would be zero, so harmonization would not change anything).

I have also written a paper, that describes the whole process of harmonization in more detail here (the reference implementation is in METAPOST, but the math is the same)
[romer-curvetools.pdf](https://github.com/fontforge/fontforge/files/12072466/romer-curvetools.pdf)


skef:
@linusromer could you rebase?

linusromer:
@skef I guess merging into master was enough? Or do you really need a rebase?

skef:
Naw, either works.

skef (approved):
lftm

PR #5265, opened 2023-07-21
python: Fix bug in getter/setter for glyph bearing

jforberg:
The getter would return a float value (same as the internal storage type) but the setter expected an int value. This caused a mysterious exception on modification:

    glyph.right_side_bearing += 0
    ^^^^^^^^^^^^^^^^^^^^^^^^
    TypeError: 'float' object cannot be interpreted as an integer

The fix is to use the same type (float) for both getter and setter.


frank-trampe:
It may be worthwhile to discuss whether to standardize on integers or floats. FontForge stores bearings and widths as 16-bit integers, and the OpenType specification also has them as 16-bit integers. On the other hand, point positions that are likely to be in the same mathematical expression as the bearings and widths (including `maxx`) are double-precision floats. Neither solution is perfect.

The current code is a bit confusing. The current setter takes an integer, converts it to a float (via addition with a float), converts it back to an integer (via `rint`) and passes the integer to `SCSynchronizeWidth`, which accepts a float. `SCSynchronizeWidth` then stuffs the float into an integer field (`sc->width`) without a cast. `SCSynchronizeWidth` makes no use of the added precision of the floating point value at any point as far as I can tell. I propose that, whatever we do with the Python setter, we add to `SCSynchronizeWidth` a cast and a comment that the value is an integer for all practical purposes.

jforberg:
I have no opinion on this, I just noticed that the internal storage type was float. If the idea is to rewrite a bigger part FF to use int instead of float, that is probably a task for someone more familiar with the codebase.

frank-trampe:
What is the internal floating point value to which you refer?

jforberg:
Ah, nevermind.

PR #5300, opened 2023-10-17
doc: added missing sudo to installation instructions

gekkedev:
Without `sudo`, permission errors would occur. I guess a lot of people just ran it all with admin permissions or figured it out on their own. Also corrected some grammar below in the file.

### Type of change

- **Non-breaking change**

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -17,7 +17,7 @@ mkdir build
 cd build
 cmake -GNinja ..
 ninja
-ninja install
+sudo ninja install
 ```
 
 For even more details on the build system itself, see [this wiki page](https://github.com/fontforge/fontforge/wiki/CMake-guide-for-FontForge).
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -74,10 +74,10 @@ If your machine doesn't have them and you want them they are available from:
     It used to be disabled by default because of some [patents granted to Apple](http://freetype.org/patents.html).
     Now that they have expired, you no longer need to worry about this, unless your setup happens to use an old library version.
     Then you may enable the interpreter by setting the appropriate macro in `*.../include/freetype/config/ftoption.h*` before you build the library (see the README.UNX file on the top level of the freetype distribution).
-    To enable the truetype debugger, FontForge needs to have the freetype source directories available when it is built (there are some include files there which it depends on.)
--   libintl Is standard on most unixes. It is part of the fink package on the mac. Handles UI localization.
+    To enable the truetype debugger, FontForge needs to have the freetype source directories available when it is built (there are some included files there which it depends on).
```
skef, line 77: "include files" is a term of art WRT the C language, so however ungrammatical it sounds the old text was correct in this case.

iorsh: @gekkedev, if this really bugs you, you can write with markup `#include` instead

skef (approved):
LGTM

PR #5305, opened 2023-10-27
Remove `psaltnames` for multi-code-point names

dscorbett:
`psaltnames` contains glyph names. Some of the names correspond to code point sequences but are only mapped to the first element in the sequence. For example, the Adobe Glyph List maps `qoftsere` to <U+05E7, U+05B5>, but `psaltnames` maps it to U+05E7. Because FontForge no way to represent named sequences, I just removed them all.

### Type of change
- **Bug fix**


iorsh:
For reference, the original Adobe Glyph List is maintained at https://github.com/adobe-type-tools/agl-aglfn/blob/master/glyphlist.txt. It looks like `psaltnames` just picks the first code point from AGL, no matter what.

Looks good to me.

iorsh:
@dscorbett, did you use some `sed`-like script for this or just updated everything manually? There are some other names in the list which look dubious, e.g. "Ifraktur" is called "Ifractur".

dscorbett:
I probably used some kind of automation but I don’t remember the exact commands.

`Ifractur` appears under the comment “From Richard Kinch's TeX list of glyph aliases” and is attested in various glyph name lists. I think it’s fine to keep old names, as long as they aren’t incorrect like the ones I removed here.

myoresh:
The issue of `Ifractur` vs `Ifraktur` bothers me because `Ifraktur` does appear in AGL, but is somehow missing altogether from `psaltnames`. If you can recall your automation, it would be great to add it as a comment in the code. I can re-develop it if you are short on time at the moment.

Overall, your change is definitely in place, but there are a lot of interdependent lists here, and their logic is [for me] very elusive.

iorsh:
The issue of I`fractur` vs `Ifraktur` bothers me because `Ifraktur` does appear in AGL, but is somehow missing altogether from `psaltnames`. If you can recall your automation, it would be great to add it as a comment in the code. I can re-develop it if you are short on time at the moment.

Overall, your change is definitely in place, but there are a lot of interdependent lists here, and their logic is [for me] very elusive.

dscorbett:
`Ifraktur` appears in the `agl_sans_p0_b21` array, which is part of the `NameList`-based system of glyph names, where each glyph name standard (e.g. AGL) gets an array of arrays of glyph names, with one string per code point (or `NULL` if there is no name for that code point). `psaltnames`, on the other hand, is a miscellaneous collection of alternative names like `Ifractur` that don’t fit elsewhere.

iorsh:
`grep -v ';....$' glyphlist.txt | grep -v ^#`
produces 81 lines, corresponding to the number of removed lines in this PR.

I checked the other names in AGL, and they are all present in `namelist.c` in some place, so I think we are good to go.

iorsh:
@dscorbett, could you please merge with the latest to trigger the CI?

skef (approved):
lgtm

PR #5306, opened 2023-10-27
Fix several segfaults

ttxine:
Fixed segfaults in:
- `SplineFontFromPSFont()`
If `fd->wascff` was `true`, then `sf` was set to `fd->sf`, which is always `NULL`. And `sf->multilayer` check caused a segfault.
- `readcffglyphs`
`readcfffontnames` may return `NULL`.
- `readcfffontnames`
`(offsets[i+1]-offsets[i])` could be equal to `UINT32_MAX`, which led to `malloc(0)` call, but the loop exit condition remained `j < UINT32_MAX`.

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/parsettf.c
+++ b/fontforge/parsettf.c
@@ -2735,6 +2735,14 @@ return( NULL );
 		++i;
 	    }
 	    --i;
+	} else if ( (offsets[i+1]-offsets[i])==UINT32_MAX ) {
+		LogError(_("CFF name %u too large\n"), i );
+		if ( info!=NULL ) info->bad_cff = true;
+	    while ( i<count ) {
+		names[i] = copy("");
+		++i;
+	    }
+	    --i;
 	} else {
 	    names[i] = malloc(offsets[i+1]-offsets[i]+1);
 	    for ( j=0; j<offsets[i+1]-offsets[i]; ++j )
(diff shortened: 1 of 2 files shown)


fontforge/parsettf.c
```diff
@@ -2735,6 +2735,14 @@ return( NULL );
 		++i;
 	    }
 	    --i;
+	} else if ( (offsets[i+1]-offsets[i])==UINT32_MAX ) {
```
skef, line 2738: This is a weirdly specific problem. Normally one would worry about a more general class of strings that are too large. Here you're finding the offset difference is exactly UINT32_MAX. That strongly implies (to me, anyway) that some other code is causing the difference to be that value, and adding this fix would just disguise that problem.

Before we do this we should at least try to figure out why we're seeing such a specific symptom. Do you have an example file?

fontforge/splineutil.c
```diff
@@ -3014,7 +3014,7 @@ SplineFont *SplineFontFromPSFont(FontDict *fd) {
 	SplineFontMetaData(sf,fd);
 	if ( fd->wascff ) {
 	    SplineFontFree(sf);
-	    sf = fd->sf;
+return ( NULL );
```
skef, line 3017: We're not doing the unindented returns anymore. Some of those are left over from the old days. 

Still, this may not be quite right. I assume the problem is that `fd->sf` can be NULL, and then (for example) the sf->multilayer check below will fail? That's true, but in that case shouldn't we first check if fd->sf is NULL, return if it is, and otherwise do what's there now?

winlover32 (approved):

winlover32 (approved):

winlover32 (approved):

PR #5317, opened 2023-11-26
msgstr formatting discrepancies

iorsh:
Fix formatting discrepancies between msgid and msgstr. Windows build fails because of them.


muelli:
this looks like it could fix https://github.com/fontforge/fontforge/issues/5251

skef:
I've never been much involved in this part of the project, but I believe FontForge uses CrowdIn for localization here: https://crowdin.com/project/fontforge , and then a script is run to pull any changes down prior to a release. If the strings aren't fixed there there will be an ongoing synchronization problem. Have you checked into that?

iorsh:
> I've never been much involved in this part of the project, but I believe FontForge uses CrowdIn for localization here: https://crowdin.com/project/fontforge , and then a script is run to pull any changes down prior to a release. If the strings aren't fixed there there will be an ongoing synchronization problem. Have you checked into that?

Yes, Fredrick addresses it in https://github.com/fontforge/fontforge/pull/5257. I will kill this PR and see what can be done with CrowdIn.

skef:
Looks like at least one message is fixed in Crowdin as of six months ago. It's likely they all are. Let me see about synching.

PR #5328, opened 2023-12-28
Update GinHub CI runners

iorsh:
Partially fixed CI pipelines.
1. Linux ubuntu-release now passes
2. MacOS now runs to the point where msgfmt fails for some localization messages
3. Windows not changed, also runs till msgfmt failure

The localization messages need to be addressed separately - see https://github.com/fontforge/fontforge/pull/5257


skef (approved):
After looking at this PR relative to the current CI failures, I'm approving as a step towards resolving those problems.

Issue #5326, opened 2023-12-21
Segfault during Open() when LANG=C and metadata contain Unicode chars


gioele:
Original report: https://bugs.debian.org/1030842 

## Problem description

fontforge segfaults while opening a TTF file that contains Unicode chars in its metadata fields when `LANG=C`.

To reproduce:

```
$ wget https://salsa.debian.org/fonts-team/fonts-femkeklaver/-/raw/master/femkeklaver.ttf
$ echo 'Open($1)' > open.pe
$ export LANG=C
$ fontforge -script open.pe femkeklaver.ttf 
Program root: /usr
Copyright (c) 2000-2023. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Version: 20230101
 Based on sources from 2023-01-18 18:05 UTC-D.
Segmentation fault
```

Backtrace provided by Bernhard Übelacker

```
$ LANG=C rr record fontforge -script /home/benutzer/source/fonts-femkeklaver/git/fonts-femkeklaver/debian/clear-fstype.pe /home/benutzer/source/fonts-femkeklaver/git/fonts-femkeklaver/femkeklaver.ttf
$ rr replay -o-q
...
Breakpoint 3, iconv (cd=cd@entry=0x55b4437f72d0, inbuf=inbuf@entry=0x7fff48abee88, inbytesleft=inbytesleft@entry=0x7fff48abee80, outbuf=outbuf@entry=0x7fff48abeea0, outbytesleft=outbytesleft@entry=0x7fff48abee98) at ./iconv/iconv.c:32
32      {
(rr) finish
Run till exit from #0  iconv (cd=cd@entry=0x55b4437f72d0, inbuf=inbuf@entry=0x7fff48abee88, inbytesleft=inbytesleft@entry=0x7fff48abee80, outbuf=outbuf@entry=0x7fff48abeea0, outbytesleft=outbytesleft@entry=0x7fff48abee98) at ./iconv/iconv.c:32
0x00007fd0fc8a0b69 in do_iconv (cd=0x55b4437f72d0, inbuf=<optimized out>, inbuf@entry=0x7fff48abeef0, incount=<optimized out>, inunitsize=inunitsize@entry=1, outunitsize=outunitsize@entry=1) at ./Unicode/ucharmap.c:101
[...]

PR #5329, opened 2023-12-29
Handle failed iconv conversion. Unhandled execution path was UB, causing a segfault for me

ViliusSutkus89:
### Type of change
- **Bug fix**

Hello,

FontForge does some locale conversions before printing errors. `NOUI_IError` and `NOUI__LogError` (in `nouiutil.c`) calls `utf82def_copy` (in `ucharmap.c`), which calls `do_iconv`.

`do_iconv` handles failed `iconv` call and returns `NULL`, which propagates back to `NOUI_IError` and `NOUI__LogError` functions and is not handled there.

`NOUI_IError` and `NOUI__LogError` are calling `printf` with `NULL` as an argument to `"%s"`, which is undefined behaviour and that UB is crashing my Android devices.

Yes, I do understand, that I should fix the `iconv` call to not fail, but until then, I wouldn't mind fixing this UB, if that looks acceptable.

Regards,
Vilius


skef:
This looks good to me, but could you rebase @ViliusSutkus89 ? I don't want to override the checks on all of the back-log.

fontforge/nouiutil.c
```diff
@@ -42,21 +42,29 @@ static void NOUI_IError(const char *format,...) {
...
-    if ( str[strlen(str)-1]!='\n' )
-	putc('\n',stderr);
-    free(str);
+    if (NULL != str) {
+        fprintf(stderr,"%s",str);
+        if ( str[strlen(str)-1]!='\n' )
+            putc('\n',stderr);
+        free(str);
+    } else {
+        fprintf(stderr, "utf82def_copy failure!\n");
```
iorsh, line 51: Throwing the error out because `iconv()` has opinions doesn't seem right to me. Maybe add `fprintf(stderr,"%s",buffer);` here, just to print whatever we have, even if badly encoded?

ViliusSutkus89: Only issue I could see would be malformed characters being interpreted as terminal control characters

iorsh: @ViliusSutkus89, do you have a sample font + steps to reproduce this?

ViliusSutkus89: I have a broken iconv that's causing the issue :D But it is still UB. 
The original error message is "The PostScript font name "BGDGBW弫Roboto-Bold" is invalid.\nIt should be printable ASCII,\nmust not contain (){}[]<>%/ or space\nand must be shorter than 63 characters"

ViliusSutkus89: Broken iconv would mean that any error message would trigger UB here

iorsh: I understand that we use `iconv()` mainly for two things: file names and error messages to the terminal. Maybe we can add separate `iconv_t` descriptors for errors using `ASCII//TRANSLIT` as a fallback? In that case your error message would become "The PostScript font name "BGDGBW?Roboto-Bold...", which is still informative.

iorsh:
Related issue: #5326

iorsh:
This patch fixes #5326 (which calls `utf82def_copy()` elsewhere) and should fix this segfault too:
[utf82def_copy_safe.patch.gz](https://github.com/fontforge/fontforge/files/13840782/utf82def_copy_safe.patch.gz)

@ViliusSutkus89, could you please apply and verify with your font?

ViliusSutkus89:
@iorsh , yes! this works for me. I still get "utf82def_copy failure!", but that's because of my broken iconv

iorsh:
> @iorsh , yes! this works for me. I still get "utf82def_copy failure!", but that's because of my broken iconv

@ViliusSutkus89 Great, could you please commit the patch into your PR? I don't have access.

ViliusSutkus89:
Sure thing, @iorsh . I've also pulled commits from current master branch, for a cleaner merge.

skef (approved):
lgtm

ViliusSutkus89:
Sweet, thanks :)

PR #5335, opened 2024-01-02
Fix CI for Windows GitHub runner

iorsh:
Duplicate libfontforge.dll for "py" and "pyhook" tests.

Both `fontforge.exe` (used in embedded Python tests) and `psMat.pyd, fontforge.pyd` (used in Python hook tests) are looking for `libfontforge.dll` strictly in their respective directories. Both `PATH` environment variable and Python's `sys.path` seem to be ignored.

### DIFF
--- a/fontforge/CMakeLists.txt
+++ b/fontforge/CMakeLists.txt
@@ -222,6 +222,11 @@ set_property(SOURCE splineoverlap.c APPEND PROPERTY COMPILE_OPTIONS ${FONTFORGE_
 set_property(TARGET fontforge PROPERTY VERSION 4)
 if(BUILD_SHARED_LIBS)
   if(WIN32 OR CYGWIN)
+    # Both fontforge.exe and *.pyd Python modules look for libfontforge.dll in
+    # their directory, so we need to produce a second copy.
+    add_custom_command(TARGET fontforge POST_BUILD
+      COMMAND cp $<TARGET_FILE:fontforge> ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
+    )
     target_link_options(fontforge PRIVATE -Wl,--export-all-symbols) # FIXME
   endif()
 endif()


skef:
On the PR content: This extra binary gets us through the tests, but does it result in something that works? It doesn't appear (just from the changes) that we're installing that copy of the library in the respective module directory. If python modules are more limited in how they bind, we need to confirm we're within those limitations in the built product.

iorsh:
Yes, that's tricky. Unfortunately, I don't have Windows machine to try this and see what works and what doesn't. I think the situation is somewhat similar to Mac, where I aligned everything to the same Python version (including subversion) to satisfy tests, but it would be ridiculous if a binary compiled with Python 3.11.7 actually breaks when installed on a system with 3.11.8. 

I don't have any solution for this except sweeping everything under the rug and waiting until somebody a) complains and b) agrees to help tuning the build with their environment.

skef:
@iorsh File a new issue on the Windows build question and cross-link it here and I'll merge this. 

I found the stale status check and updated it to 20.04.

skef:
Note that:

1. This praise-worthy revival of the repo will be almost all for naught if we can't do a release, which includes doing a Windows release
2. I've seen evidence that the Jan 2023 Windows release is corrupt -- specifically that the wrong appearance configuration may have been included. If that's true then doing a new, corrected release to fix the Windows problem should be a project priority.

JoesCat:
> @skef, the "linux (ubuntu-18.04, Release)" check is probably stalled because ubuntu-18.04 runner no longer exists

[Ubuntu release cycle](https://ubuntu.com/about/release-cycle) shows 18.04 has gone through 5 years. maybe consider bumping up to 20, and 22?

skef:
@JoesCat I took care of that, the updated ones are "20.04" and "latest"

skef (approved):
lgtm

PR #5338, opened 2024-01-03
Fix lookup flags parsing

### LINKED ISSUE #5337: Lookup flags parsed incorrectly in an OT feature file import
FontForge ignores second lookup flag when merging feature info file with multiple flags.

Example feature file:
```
lookup Kern_HEBR {
  lookupflag RightToLeft IgnoreMarks;
    pos \afii57645 \afii57681 -10;
    pos \afii57645 \afii57664 -10;
} Kern_HEBR;

feature kern {

 script hebr;
     language dflt ;
      lookup Kern_HEBR;
} kern;
```
After the import the lookup metadata shows just the first flag set:
lookup_flags_bug

The issue seems to be in `fea_ParseLookupFlags()`

iorsh:
Fixes parsing of lookup flags in an OT feature file import.

In a line like `lookupflag RightToLeft IgnoreMarks;` the first flag is parsed in line
featurefile.c:3192, and the next token is parsed on featurefile.c:3213. The problem is that after the second token is verified, there is another parsing command at featurefile.c:3222, which effectively discard the second token and picks yet another one.

Closes #5337

### DIFF
--- a/fontforge/featurefile.c
+++ b/fontforge/featurefile.c
@@ -3219,7 +3219,6 @@ static void fea_ParseLookupFlags(struct parseState *tok) {
 		fea_skip_to_semi(tok);
 	break;
 	    }
-	    fea_ParseTok(tok);
 	}
 	if ( tok->type != tk_char || tok->tokbuf[0]!=';' ) {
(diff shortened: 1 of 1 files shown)


iorsh:
@skef, could you please take a look?

skef (approved):
Yeah, seems like a typically Georgean artifact. Approved

PR #5345, opened 2024-01-15
nltransform of anchor points

### LINKED ISSUE #5320: Feature Request: Apply non-linear transformation to anchors
When applying a non-linear transformation, the anchors don't change: I tried to create obliques by applying the transformation x → x + y*0.2, but this ruins diacritic placement. (Of course, the skew transform works as an alternative in this case.)

mf2vec-dev:
Fixes #5320

Apply non-linear transformations to anchor points while taking selection into account. This should work in GUI, Python and scripting. I tested in GUI and added a Python test.

This is my first contribution to a C project. Please check my changes carefully.

### Type of change
- **Bug fix**
- **New feature**


skef (approved):
LGTM

skef:
Ah, sure enough, SCPreserveLayer handles the anchor points if the layer is ly_fore.

PR #5352, opened 2024-01-25
fix segfault triggered by Python `del c[i:j]`

mf2vec-dev:
This PR adresses a segmentation fault when running the following Python code:
```py
import fontforge
c = fontforge.contour()
c.moveTo(0, 0).lineTo(10, 0).lineTo(20, 0).lineTo(30, 0)
del c[1:3]
```

With this change, a Python `TypeError` is raised instead of triggering the segmentation fault.

This is only a problem for contours. Layers (`del l[i]`) and Layer Info Arrays (`del font.layers[i]`) don't allow slicing.

### Type of change
- **Bug fix**


iorsh:
Could it be better to just support the `del c[i:j]`?
https://github.com/fontforge/fontforge/blob/master/fontforge/python.c#L2699 already has the code for deleting a single point, deleting a range shouldn't be much different.

fontforge/python.c
```diff
@@ -2852,6 +2852,10 @@ static int PyFFContour_SubAssign( PyObject *self, PyObject *key, PyObject *val )
     if ( PyLong_Check(key)) {
         return PyFFContour_IndexAssign(self, PyNumber_AsSsize_t(key, PyExc_IndexError), val);
     }
+    if ( val==NULL ) {
+        PyErr_Format(PyExc_TypeError, "Deleting multiple points is not supported.");
```
iorsh, line 2856: Deletion with any non-numeric key (even just a mistaken argument) would trigger this message. Maybe differentiate:
- valid range --> delete range
- invalid range --> invalid range error
- not range --> type error for key argument

tests/test1021.py
```diff
@@ -0,0 +1,19 @@
...
+c = cont.dup()
+del c[1]
+assert len(c) == 3
+assert c[0].x == 0
+assert c[1].x == 20
+assert c[2].x == 30
+
+c = cont.dup()
+del c[1:3]
+assert len(c) == 2
```
iorsh, line 17: Something looks strange for me here. `del c[1:3]` is supposed to delete the elements 1, 2. But the remaining contour has length 2 and not 1. Could you please check?

iorsh:
Looks good to me. Appveyor is broken, hopefully it will be back in a few hours. I'll check it.

skef (approved):
LGTM

PR #5354, opened 2024-01-27
add `font.style_set_names` attribute to Python API

### LINKED ISSUE #5263: “StyleSet Names” cannot edit by Python API
I want to edit “StyleSet Names” by API, but I think it does not exist in [doc](https://fontforge.org/docs/scripting/scripting.html). I can edit them from GUI: Element → Font Info → StyleSet Names.

   - FontForge 20230101 from Homebrew
   - macOS Ventura 13.4.1

mf2vec-dev:
Closes #5263.

This PR adds the `style_set_names` attribute to the `font` type of the Python API to get and set the style set names (equivalent to `Element` -> `Font Info...` -> `StyleSet Names` in the GUI), including input validation, a Python test and documentation.

The last non-empty test case (name: `"😀𝄞"`) passes in the Python interface. But when generating an sfd and loading it with the GUI, the name shown in the GUI is different (`턞`). I am not sure if this is a bug in my PR or an issue with `SFDDumpUTF7Str`/`SFDReadUTF7Str` or the GUI. 
I would appreciate it if someone could double check this.

### Type of change
- **New feature**


fontforge/python.c
```diff
@@ -14066,6 +14066,140 @@ return (NULL);
...
+
+        PyObject *tag_str_py = PyTuple_GetItem(ss_names_tuple, 1);
+        if ( !PyUnicode_Check(lang_str_py) ) {
+            PyErr_Format(PyExc_TypeError, "The second element of a style set specification must be a string." );
+            return( -1 );
+        }
+        tag_str = copy(PyUnicode_AsUTF8(tag_str_py));
+        is_valid = 0;
+        if ( strlen(tag_str)== 4 && tag_str[0]=='s' && tag_str[1]=='s' ) {
+            int ss_num = 10*(tag_str[2]-'0') + tag_str[3]-'0';
```
dscorbett, line 14142: The last two characters are not verified to be digits, so, for example, `"ss0D"` is detected as valid and equivalent to `"ss20"`.

fontforge/python.c
```diff
@@ -14066,6 +14066,142 @@ return (NULL);
...
+            return( -1 );
+        }
+        if ( PyTuple_Size(ss_names_tuple)!=3 ) {
+            PyErr_Format(PyExc_TypeError, "Style set name specification must have 3 elements (language, style set and name)." );
+            return( -1 );
+        }
+
+        PyObject *lang_str_py = PyTuple_GetItem(ss_names_tuple, 0);
+        if ( !PyUnicode_Check(lang_str_py) ) {
+            PyErr_Format(PyExc_TypeError, "The first element of a style set specification must be a string." );
```
dscorbett, line 14122: The language should also be allowed to be specified as a number, as in the attribute `sfnt_names`.

iorsh:
I tried to run your branch, but could see no data in Font Info -> StyleSet Names at all. Saved SFD font also has no entries for StyleSets. What is your system?

iorsh:
The issue with "😀𝄞" is a bug in `SFDDumpUTF7Str`/`SFDReadUTF7Str`. You can observe it regardless of your PR, by setting any name (for example in "Font Info" --> "TTF Names") with characters in a supplementary plane.

iorsh:
> @iorsh Correct, but I'm not 100% sure if it's `SFDDumpUTF7Str`/`SFDReadUTF7Str` or the GUI, because saving the font as sfd via Python and loading it with Python gives the correct language name `"😀𝄞"`. So these Unicode characters seem to "survive" the sfd in this case. Can you reproduce this? (Perhaps a new issue would be a good place for this.)

I couldn't reproduce this. Could you please specify the exact course of action or script?

iorsh:
Ah, you mean the assertion, which passes and indicates correctly loaded string. That's interesting, I'll take a look.

iorsh:
If you add `font.close()` after saving, the assertion will fail. Probably `fontforge.open()` doesn't read file from disk if it's already open. So this is indeed a separate issue with `SFDDumpUTF7Str`/`SFDReadUTF7Str`.

Overall, I think we are good to go with this PR.

fontforge/python.c
```diff
@@ -14066,6 +14066,150 @@ return (NULL);
...
 
+static PyObject *PyFF_Font_get_style_set_names(PyFF_Font *self, void *UNUSED(closure)) {
+    int cnt, i;
+    SplineFont *sf = self->fv->sf;
+    struct otffeatname *fn;
+    struct otfname *on;
+    PyObject *ss_names_tuple, *ss_name_spec;
+    char *lang_str, tag_str[5];
+
+    if ( CheckIfFontClosed(self) )
```
skef, line 14077: If we need to check this, then presumably the `self->fv->sf` access above isn't safe.

fontforge/python.c
```diff
@@ -14066,6 +14066,150 @@ return (NULL);
...
+        return( NULL );
+
+    for ( cnt=0, fn=sf->feat_names; fn!=NULL; fn=fn->next )
```
skef, line 14080: You don't need to change this, but I think it's generally clearer to initialize variables used this way (accessed after the loop ends) before the loop rather than in it.

fontforge/python.c
```diff
@@ -14066,6 +14066,150 @@ return (NULL);
...
+        for ( on=fn->names; on!=NULL; on=on->next, ++cnt );
+    if ( cnt==0 )
+        return( PyTuple_New(0) );
+    ss_names_tuple = PyTuple_New(cnt);
+    for ( cnt=0, fn=sf->feat_names; fn!=NULL; fn=fn->next ) {
+        for ( on=fn->names; on!=NULL; on=on->next, ++cnt ) {
+            lang_str = copy(NOUI_MSLangString(on->lang));
+            ss_name_spec = PyTuple_Pack(3, PyUnicode_FromString(lang_str), TagToPythonString(fn->tag, false), PyUnicode_FromString(on->name));
```
skef, line 14088: Doesn't `PyUnicode_FromString()` leave its argument unaffected (e.g. the string is copied into a different form rather than referenced)? If so, aren't we leaking the copy() in the previous line?

fontforge/python.c
```diff
@@ -14066,6 +14066,150 @@ return (NULL);
...
+static int PyFF_Font_set_style_set_names(PyFF_Font *self, PyObject *value, void *UNUSED(closure)) {
+    int i, j, rows, is_valid;
+    PyObject *ss_names_tuple, *lang_py;
+    char *tag_str, *name_str;
+    uint32_t tag, lang;
+    SplineFont *sf = self->fv->sf;
+    struct otffeatname *fn;
+    struct otfname *on;
+
+    if ( CheckIfFontClosed(self) )
```
skef, line 14104: Same worry here as above

fontforge/python.c
```diff
@@ -14066,6 +14066,150 @@ return (NULL);
...
+            PyErr_Format(PyExc_TypeError, "The first element of a style set name specification must be a string or an integer.");
+            return( -1 );
+        }
+
+        PyObject *tag_str_py = PyTuple_GetItem(ss_names_tuple, 1);
+        if ( !PyUnicode_Check(tag_str_py) ) {
+            PyErr_Format(PyExc_TypeError, "The second element of a style set name specification must be a string.");
+            return( -1 );
+        }
+        tag_str = copy(PyUnicode_AsUTF8(tag_str_py));
```
skef, line 14144: What's the `copy()` for? Seems like we're just leaking here.

fontforge/python.c
```diff
@@ -14066,6 +14066,150 @@ return (NULL);
...
+        if ( !PyUnicode_Check(PyTuple_GetItem(ss_names_tuple, 2)) ) {
+            PyErr_Format(PyExc_TypeError, "The third element of a style set name specification must be a string.");
+            return( -1 );
+        }
+    }
+
+    /* check for duplicates */
+    for ( i=0; i<rows; ++i ) {
+        for ( j=i+1; j<rows; ++j ) {
+            if ( PyTuple_GetItem(PyTuple_GetItem(value, i), 0) == PyTuple_GetItem(PyTuple_GetItem(value, j), 0) &&
```
skef, line 14166: I'd be surprised (but not flabbergasted) if these comparisons work. They're pointer comparisons, and if they work it implies that Python is doing a very good job making sure that tuples are unique internally. 

Might `PyObject_RichCompareBool` with `PyEQ` be what we want here?

fontforge/python.c
```diff
@@ -14066,6 +14066,150 @@ return (NULL);
...
+    char *tag_str, *name_str;
+    uint32_t tag, lang;
+    SplineFont *sf = self->fv->sf;
+    struct otffeatname *fn;
+    struct otfname *on;
+
+    if ( CheckIfFontClosed(self) )
+        return( -1 );
+
+    if ( !PyTuple_Check(value) ) {
```
skef, line 14107: Would be good to test these checks, or at least the non-trivial ones.

fontforge/python.c
```diff
@@ -14066,6 +14066,150 @@ return (NULL);
...
+            lang = FindFlagByName(sfnt_name_mslangs, PyUnicode_AsUTF8(lang_py));
+        } else {
+            lang = PyLong_AsLong(lang_py);
+        }
+
+        tag = StrObjToTag(PyTuple_GetItem(ss_names_tuple, 1), false);
+
+        name_str = copy(PyUnicode_AsUTF8(PyTuple_GetItem(ss_names_tuple, 2)));
+
+        for ( fn=sf->feat_names; fn!=NULL && fn->tag!=tag; fn=fn->next );
```
skef, line 14197: Note that this code appears to use the existing `feat_names` entry for the tag, or (below) adds a new entry for a missing tag.

fontforge/python.c
```diff
@@ -14066,6 +14066,150 @@ return (NULL);
...
+        if ( fn==NULL ) {
+            fn = chunkalloc(sizeof(*fn));
+            fn->tag = tag;
+            fn->next = sf->feat_names;
+            sf->feat_names = fn;
+        }
+        on = chunkalloc(sizeof(*on));
```
skef, line 14204: Here we're just always adding a new chunk. That means that if you use this interface multiple times you'll just accumulate new entries, right?

I think what's needed here is to look for the `lang` in the list and replace it, and add a new entry only if it's not found, just like we do above for `tag`.

tests/test1020.py
```diff
@@ -0,0 +1,24 @@
+#Test getter and setter of font.style_set_names
```
skef, line 1: As noted or implied above, a bunch more tests for error conditions and sanity when changing the list multiple times would be good here.

fontforge/python.c
```diff
@@ -14201,11 +14216,14 @@ static int PyFF_Font_set_style_set_names(PyFF_Font *self, PyObject *value, void
...
-        fn->names = on;
-        on->lang = lang;
-        on->name = name_str;
+        for ( on=fn->names; on!=NULL && on->lang!=lang; on=on->next );
+        if ( on==NULL ) {
+            on = chunkalloc(sizeof(*on));
+            on->lang = lang;
+            on->next = fn->names;
+            fn->names = on;
+        }
```
skef, line 14229: Think we need an `else free(on->name);` here.

skef (approved):
LGTM

skef:
@iorsh do you have any further comment on the recent changes before I merge this?

myoresh:
> @iorsh do you have any further comment on the recent changes before I merge this?

Looks good to me

myoresh (approved):

iorsh:
> @iorsh do you have any further comment on the recent changes before I merge this?

Looks good to me

PR #5359, opened 2024-01-30
Fix the lists of Windows language IDs

dscorbett:
This updates the lists of Windows language IDs in fontforgeexe/fontinfo.c (the GUI) and fontforge/python.c (the Python API) to match [the 'name' table specification](https://learn.microsoft.com/en-us/typography/opentype/spec/name#windows-language-ids).

The first commit adds missing IDs. Existing IDs were not modified, except for Galician, Brazilian Portuguese, and European Portuguese, which had incorrect numbers.

The second commit removes undefined IDs from the GUI. These IDs are valid Windows locale IDs in other contexts, but they are not defined in OpenType. I didn’t remove them from the Python API, to avoid breaking scripts.

### Type of change
- **Bug fix**
- **Non-breaking change**


skef (approved):
Trusting you on this ...

PR #5365, opened 2024-02-02
Bulk tester

iorsh:
This is a BASH script, which can perform an operation on a large number of SFD files and compare its output to the reference FontForge version.

Specifically, for each SFD file in the list the following actions are performed:
1. Call `/usr/local/bin/fontforge` with some script and keep the output
2. Call a development version of `fontforge` with the same script and keep the output separately
3. Compare the outputs byte-to-byte and report failure if they differ.

### Type of change
- **New feature - regression testing**

To verify this script actually works:
1. Go to https://github.com/fontforge/fontforge/blob/eff88fc98a5801190ddf6adefa67c9f78b65fd1a/fontforge/sfd.c#L7668 and modify this line to `sf->fullname = copyn(val, 23);`. This change cuts long font names.
2. Compile the project and run the script
3. Observe some fonts (with longer names) start to fail. For example, _Homerton.Medium.Oblique.sfd_


skef (approved):
This would be more portable if it were written in Python but the type of testing its doing is less likely to vary by platform, and anyway where bash isn't available by default one can figure things out, so this seems fine.

PR #5367, opened 2024-02-06
Fix Splinefont shell invocation

pkvanca:
This patch addresses security vulnerabilities arising from how FontForge opens certain files. The code has been changed to use glib's `g_spawn_sync/async` functions that do not execute commands through a shell. This patch should still retain all of the original functionality and introduce no breaking changes, having passed all tests as well.

As discussed with the maintainers via email, opening a public PR here.

Addresses CVES: 
- CVE-2024-25081 - Command Injection from Malicious Filenames
- CVE-2024-25082 - Command Injection from Malicious Archives / Compressed Files

### Type of change
- **Bug fix**

### DIFF
--- a/fontforge/splinefont.c
+++ b/fontforge/splinefont.c
@@ -788,11 +788,14 @@ return( name );
 
 char *Unarchive(char *name, char **_archivedir) {
     char *dir = getenv("TMPDIR");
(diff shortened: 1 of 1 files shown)


skef (approved):
LGTM

PR #5372, opened 2024-02-10
Compare vertical metrics check when generating TTC

### LINKED ISSUE #5371: Vertical metrics not saved properly when generating a TTC file
I have five fonts, in which some glyphs are identical across all fonts but have different vertical metrics in one of them. I have generated them as one .ttc file with the most recent version of FontForge (2023-01-01).

When I open, from the .ttc file, the font in which they are different, the vertical metrics have been altered to that of the other fonts.

For example, take （ U+FE35 and ）U+FE36. Originally, across the fonts,
（　has different vector data and different metrics,
 ） has the same vector data and different metrics.

In the generated ttc,
（ has correct metrics.
） has mistaken metrics.

I used the ttx command from fonttools to inspect the .ttc file, and the vertical metrics table (vmtx) shows:

    <mtx name="uniFE35" height="4096" tsb="2899"/>
    <mtx name="uniFE35#1" height="2048" tsb="850"/>
    <mtx name="uniFE36" height="4096" tsb="119"/>
    <mtx name="uniFE37" height="4096" tsb="2737"/>
    <mtx name="uniFE37#1" height="2048" tsb="688"/>

The character map table (cmap) shows:

    <map code="0xfe35" name="uniFE35#1"/>
    <map code="0xfe36" name="uniFE36"/>
    <map code="0xfe37" name="uniFE37#1"/>

There should be a uniFE36#1 in (vmtx) and (cmap) should link to it...

I think the TTC generation should not only check glyph similarity but also vertical metrics!

Yukinoroh:
Fixes #5371

### DIFF
--- a/fontforge/tottf.c
+++ b/fontforge/tottf.c
@@ -6363,6 +6363,8 @@ static int glyphmatches(SplineChar *sc,SplineChar *sc2,int layer) {
 
     if ( sc->width!=sc2->width )
 return( false );
+    if ( sc->vwidth!=sc2->vwidth )
+return( false );
     if ( sc->ttf_instrs_len != sc2->ttf_instrs_len )
 return( false );
     if ( sc->ttf_instrs_len!=0 && memcmp(sc->ttf_instrs,sc2->ttf_instrs,sc->ttf_instrs_len)!=0 )


iorsh:
Looks good to me.

Did some superficial code check, but couldn't see similar missing vwidth checks in other places.

skef (approved):
lgtm

PR #5376, opened 2024-02-17
Set window type dialog for dialogs and transient windows

serhiyhelp:
Now dialogs and transient windows have a correct window type so a window manager can better understand how to display them. (In my particular case, dwm checks window type and displays dialogs in floating mode)

### DIFF
--- a/gdraw/ggdkdraw.c
+++ b/gdraw/ggdkdraw.c
@@ -476,8 +476,13 @@ static GWindow _GGDKDraw_CreateWindow(GGDKDisplay *gdisp, GGDKWindow gw, GRect *
         }
 
         attribs.title = nw->window_title;
-        attribs.type_hint = (nw->is_popup || (wattrs->mask & wam_palette)) ?
-                            GDK_WINDOW_TYPE_HINT_UTILITY : GDK_WINDOW_TYPE_HINT_NORMAL;
+        if (nw->is_popup || (wattrs->mask & wam_palette)) {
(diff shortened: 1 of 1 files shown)


jtanx:
Example screenshots?

serhiyhelp:
I don't know how to show this in a screenshot so I've recorded a video: https://youtu.be/embWmXLwOXk

iorsh:
Tested in Gnome, KDE and windows:

In **Gnome** the Font Info dialog is non-modal, but is kept above all other windows, which is inconsistent with other window managers and probaly undesired.

On project:
ff_gnome_proj

On system:
ff_gnome_sys

iorsh:
In **Windows** the Font Info dialog ceased to participate in tiling arrangement, which is the desired behavior.

skef (approved):
This seems to be fine

iorsh:
I'll take a deeper look at it, but it doesn't seem to be easy thing. We are talking about very sensitive z-order / modality stuff which would need to be tested across many environments. X11 / Wayland / Windows / Gnome / Kde Plasma all have their peculiarities.

skef:
Given our crimes with the Point Info dialog that may be a venal sin.

skef:
So it sounds like any benefits don't currently outweigh the drawbacks.

Do we have any evidence that there is relevant OS/stacks-specific behavior, or as far as we know now is this just a matter of some windows (e.g. the font info dialog) getting the wrong type_hint? And if the latter, is the issue that `is_dlg` shouldn't imply modality?

PR #5379, opened 2024-02-25
Treat FT_PIXEL_MODE_MONO as 2 grey levels

### LINKED ISSUE #5208: Grit Fit view monochrome broken

### Important

- [x] Screenshot

### Setup
* Fontforge 20230101 (self bilt from git master on Linux)
* Fontforge 20230101 (prebuild Windows from release page)
* Fontforge 20220308 (Ubuntu 22.04 package)
* Font: https://github.com/microsoft/cascadia-code/releases/download/v2111.01/CascadiaCode-2111.01.zip
* File `ttf/static/CascadiaCode-Regular.ttf`

### Problem / Reproduce
* Open the font (it happened with all fonts I tried, the one I actually tried to debug the instructions was the one references above)
* Open one glyph in outline window
* View -> Grit Fit ->Show Git Fit
* xy-same, 96 dpi, size 15, **Mono**

While **Anti-Aliased** produces a reasonable result, the **Mono** rendering seems broken:

grafik

_Left: anti-aliased mode, right: mono mode_

It looks like it renders it multiple times, squashed?

grafik

_mono mode, very high size_

jaklein:
### Type of change

- **Bug fix**
Fixes #5208

### DIFF
--- a/fontforge/freetype.c
+++ b/fontforge/freetype.c
@@ -718,7 +718,7 @@ return( NULL );
     ret->bytes_per_row = slot->bitmap.pitch;
     ret->as = slot->bitmap_top;
     ret->lb = slot->bitmap_left;
-    ret->num_greys = slot->bitmap.pixel_mode == ft_pixel_mode_mono ?  2 : slot->bitmap.num_grays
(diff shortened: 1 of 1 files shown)


iorsh:
I confirmed the issue and verified the fix. Looks good to me.

skef (approved):
lgtm

PR #5386, opened 2024-03-06
Update mm.c

slichtzzz:
Corrected the situation, when substituted part of the string remains untranslated.

### Type of change

- **Bug fix**
- **New feature**
- **Breaking change**
- **Non-breaking change**

### DIFF
--- a/fontforge/mm.c
+++ b/fontforge/mm.c
@@ -1129,14 +1129,14 @@ return( false );
 		    if ( !HintsMatch(sf->glyphs[i]->hstem,mm->instances[j]->glyphs[i]->hstem)) {
 			if ( complain ) {
 			    FVChangeGID( sf->fv,i);
-			    ff_post_error(_("Bad Multiple Master Font"),_("The horizontal hints in glyph \"%2$.30s\" in font %3$.30s do not match those in %4$.30s (different number or different overlap criteria)"),
+			    ff_post_error(_("Bad Multiple Master Font"),_("The horizontal hints in glyph \"%1$.30s\" in font %2$.30s do not match those in %3$.30s (different number or different overlap criteria)"),
 				    sf->glyphs[i]->name,sf->fontname, mm->instances[j]->fontname);
 			}
 return( false );
 		    } else if ( !HintsMatch(sf->glyphs[i]->vstem,mm->instances[j]->glyphs[i]->vstem)) {
 			if ( complain ) {
 			    FVChangeGID( sf->fv,i);
-			    ff_post_error(_("Bad Multiple Master Font"),_("The vertical hints in glyph \"%2$.30s\" in font %3$.30s do not match those in %4$.30s (different number or different overlap criteria)"),
+			    ff_post_error(_("Bad Multiple Master Font"),_("The vertical hints in glyph \"%1$.30s\" in font %2$.30s do not match those in %3$.30s (different number or different overlap criteria)"),
 				    sf->glyphs[i]->name,sf->fontname, mm->instances[j]->fontname);
 			}
 return( false );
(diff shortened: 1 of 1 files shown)


fontforge/mm.c
```diff
@@ -1130,14 +1130,14 @@ return( false );
 			if ( complain ) {
 			    FVChangeGID( sf->fv,i);
 			    ff_post_error(_("Bad Multiple Master Font"),_("The %1$s hints in glyph \"%2$.30s\" in font %3$.30s do not match those in %4$.30s (different number or different overlap criteria)"),
-				    "horizontal", sf->glyphs[i]->name,sf->fontname, mm->instances[j]->fontname);
+				    _("horizontal"), sf->glyphs[i]->name,sf->fontname, mm->instances[j]->fontname);
```
iorsh, line 1133: Splitting localizable strings is a bad practice, as they cannot be reliably translated due to syntax differences. This string should be as follows:
`_("The horizontal hints in glyph \"%1$.30s\" in font %2$.30s do not match those in %3$.30s (different number or different overlap criteria)"), sf->glyphs[i]->name,sf->fontname, mm->instances[j]->fontname`

The other string below should be changed accordingly.

slichtzzz: Oh, yes! Especially in this case substitution makes no sense.

fontforge/mm.c
```diff
@@ -1110,8 +1110,8 @@ return( false );
 		} else if ( !mm->apple && !KernsMatch(sf->glyphs[i],mm->instances[j]->glyphs[i])) {
 		    if ( complain ) {
 			FVChangeGID( sf->fv,i);
-			ff_post_error(_("Bad Multiple Master Font"),_("The glyph %1$.30s in font %2$.30s has a different set of kern pairs than in %3$.30s"),
-				"vertical", sf->glyphs[i]->name,sf->fontname, mm->instances[j]->fontname);
+			ff_post_error(_("Bad Multiple Master Font"),_("The glyph vertical in font %2$.30s has a different set of kern pairs than in %3$.30s"),
```
iorsh, line 1113: This string looks buggy at the first place - note that it used to have 3 parameters and 4 arguments. I think `"vertical"` should be just removed, the parameter `%1$.30s` is about glyph name, and it should be kept.

slichtzzz: Thank you so much! Didn't notice that.

iorsh:
Looks good to me now

skef (approved):
lgtm

skef:
The checks aren't starting on this for some reason

skef:
@slichtzzz can you push some trivial change to this branch? The checks are stuck and probably won't trigger without a commit. (You could just add two commits that make a change and then reverse it, for example.)

iorsh:
macos 11 CI is dead
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

https://github.com/fontforge/fontforgebuilds needs to be synced and https://github.com/fontforge/fontforge/pull/5423 merged to update the CI

fontforge/mm.c
```diff
@@ -1129,15 +1129,15 @@ return( false );
 		    if ( !HintsMatch(sf->glyphs[i]->hstem,mm->instances[j]->glyphs[i]->hstem)) {
 			if ( complain ) {
 			    FVChangeGID( sf->fv,i);
-			    ff_post_error(_("Bad Multiple Master Font"),_("The %1$s hints in glyph \"%2$.30s\" in font %3$.30s do not match those in %4$.30s (different number or different overlap criteria)"),
-				    "horizontal", sf->glyphs[i]->name,sf->fontname, mm->instances[j]->fontname);
+			    ff_post_error(_("Bad Multiple Master Font"),_("The horizontal hints in glyph \"%2$.30s\" in font %3$.30s do not match those in %4$.30s (different number or different overlap criteria)"),
```
jtanx, line 1132: Still looks broken to me? Variable indexes don't start at 1 now

jtanx (dismissed):
Fix format strings

skef (approved):
Re-approving

iorsh (approved):

PR #5399, opened 2024-04-11
Check for NULL value of SplineChar parent

### LINKED ISSUE #5130: FontForge crashes when a font has anchor classes, and you perform bulk operations on the font (such as changing the EM size)
This caused me to lose a significant amount of work on what's basically my first font project, because at some point one of the bulk operations happened while saving the file, which corrupted it. I believe it was trying to auto-hint glyphs that were composed of references to other glyphs, though I'm not 100% sure of that.

To reproduce it with the current `master` branch, simply open up any popular multilingual font (Roboto is the one I mostly tested on, but this happens with Noto Sans and several others as well; basically, any font that supports Unicode's 'combining diacritics', with anchor marks set up so they work on a variety of characters), open up 'Element' → 'Font Info...', go to the 'General' page, and then change the EM size (I typically tried doubling or halving the EM size; results were the same). Make sure that 'Scale Outlines' is enabled/checked, and click 'OK' at the bottom of the window to experience the crash.

If you have a glyph that has both its own anchor classes, as well as references to other glyphs with those same anchor classes defined (or maybe just anchor classes in general, but I didn't test that), the crash happens much more often. Simply opening the 'Anchor Control' window for an anchor class, and then within it choosing to view any other glyph (doesn't matter if it's a mark or base glyph), will trigger the same crash. The backtrace in that case is usually longer, but the lines of code causing the crash are the same.

On the topic of backtraces, here's a couple of examples:

```gdb
0x00007ffff7bba9f9 in AnchorPointsDuplicate (sc=0x55555638e4c0, base=0x5555562d8420) at /home/tynach/Software/Code/fontforge/fontforge/fvfonts.c:318
318             for ( ac=sc->parent->anchor; ac!=NULL; ac=ac->next )
(gdb) backtrace
#0  0x00007ffff7bba9f9 in AnchorPointsDuplicate (sc=0x55555638e4c0, base=0x5555562d8420) at /home/tynach/Software/Code/fontforge/fontforge/fvfonts.c:318
#1  SplineCharCopy (sc=sc@entry=0x555555f18a00, into=into@entry=0x0, mc=mc@entry=0x0) at /home/tynach/Software/Code/fontforge/fontforge/fvfonts.c:542
#2  0x00007ffff7cc3e99 in SFDDumpUndo (sfd=sfd@entry=0x555555963eb0, sc=sc@entry=0x555555f18a00, u=u@entry=0x55555600b360, keyPrefix=keyPrefix@entry=0x7ffff7e0effe "Undo", idx=idx@entry=0) at /home/tynach/Software/Code/fontforge/fontforge/sfd.c:980
#3  0x00007ffff7cc4f98 in SFDDumpChar (sfd=sfd@entry=0x555555963eb0, sc=0x555555f18a00, map=map@entry=0x555555e5ced0, newgids=newgids@entry=0x0, todir=todir@entry=0, saveUndoes=1) at /home/tynach/Software/Code/fontforge/fontforge/sfd.c:1637
#4  0x00007ffff7cd9f82 in SFAutoSave (sf=sf@entry=0x555555e5dcf0, map=0x555555e5ced0) at /home/tynach/Software/Code/fontforge/fontforge/sfd.c:9508
#5  0x00007ffff7b5be1f in _DoAutoSaves (fvs=<optimized out>) at /home/tynach/Software/Code/fontforge/fontforge/autosave.c:173
#6  DoAutoSaves () at /home/tynach/Software/Code/fontforge/fontforge/autosave.c:179
#7  0x00005555557272c5 in splash_e_h (gw=0x555555b586d0, gw@entry=<error reading variable: value has been optimized out>, event=0x7fffffffcec0, event@entry=<error reading variable: value has been optimized out>)
    at /home/tynach/Software/Code/fontforge/fontforgeexe/startui.c:443
#8  0x0000555555758aae in _GGDKDraw_CallEHChecked (gw=0x555555b586d0, event=<optimized out>, eh=<optimized out>) at /home/tynach/Software/Code/fontforge/gdraw/ggdkdraw.c:347
#9  0x0000555555758f20 in _GGDKDraw_ProcessTimerEvent (user_data=0x555555da2bd0) at /home/tynach/Software/Code/fontforge/gdraw/ggdkdraw.c:863
#10 0x00007ffff7006be8 in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#11 0x00007ffff700604e in g_main_context_dispatch () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#12 0x00007ffff7006400 in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#13 0x00007ffff70064a3 in g_main_context_iteration () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#14 0x00005555557578bd in GGDKDrawEventLoop (gdisp=0x555555a46aa0) at /home/tynach/Software/Code/fontforge/gdraw/ggdkdraw.c:2131
#15 0x000055555572845e in fontforge_main (argc=<optimized out>, argv=<optimized out>) at /home/tynach/Software/Code/fontforge/fontforgeexe/startui.c:1032
#16 0x00007ffff6c97083 in __libc_start_main (main=0x5555555b08c0 <main>, argc=1, argv=0x7fffffffd688, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffd678) at ../csu/libc-start.c:308
#17 0x00005555555b08fe in _start ()
```

```gdb
0x00007ffff7bba9f9 in AnchorPointsDuplicate (sc=0x555555e2e750, base=0x555555f20f50) at /home/tynach/Software/Code/fontforge/fontforge/fvfonts.c:318
318             for ( ac=sc->parent->anchor; ac!=NULL; ac=ac->next )
(gdb) backtrace
#0  0x00007ffff7bba9f9 in AnchorPointsDuplicate (sc=0x555555e2e750, base=0x555555f20f50) at /home/tynach/Software/Code/fontforge/fontforge/fvfonts.c:318
#1  SplineCharCopy (sc=sc@entry=0x555555f20c80, into=into@entry=0x0, mc=mc@entry=0x0) at /home/tynach/Software/Code/fontforge/fontforge/fvfonts.c:542
#2  0x00007ffff7cc3f41 in SFDDumpUndo (sfd=sfd@entry=0x5555560aeeb0, sc=sc@entry=0x555555f20c80, u=u@entry=0x5555560b30d0, keyPrefix=keyPrefix@entry=0x7ffff7e0effe "Undo", idx=idx@entry=0) at /home/tynach/Software/Code/fontforge/fontforge/sfd.c:995
#3  0x00007ffff7cc4f98 in SFDDumpChar (sfd=sfd@entry=0x5555560aeeb0, sc=0x555555f20c80, map=map@entry=0x555555e5d110, newgids=newgids@entry=0x0, todir=todir@entry=0, saveUndoes=1) at /home/tynach/Software/Code/fontforge/fontforge/sfd.c:1637
#4  0x00007ffff7cd9f82 in SFAutoSave (sf=sf@entry=0x555555e5bb90, map=0x555555e5d110) at /home/tynach/Software/Code/fontforge/fontforge/sfd.c:9508
#5  0x00007ffff7b5be1f in _DoAutoSaves (fvs=<optimized out>) at /home/tynach/Software/Code/fontforge/fontforge/autosave.c:173
#6  DoAutoSaves () at /home/tynach/Software/Code/fontforge/fontforge/autosave.c:179
#7  0x00005555557272c5 in splash_e_h (gw=0x555555b54f60, gw@entry=<error reading variable: value has been optimized out>, event=0x7fffffffadc0, event@entry=<error reading variable: value has been optimized out>)
    at /home/tynach/Software/Code/fontforge/fontforgeexe/startui.c:443
#8  0x0000555555758aae in _GGDKDraw_CallEHChecked (gw=0x555555b54f60, event=<optimized out>, eh=<optimized out>) at /home/tynach/Software/Code/fontforge/gdraw/ggdkdraw.c:347
#9  0x0000555555758f20 in _GGDKDraw_ProcessTimerEvent (user_data=0x555555da2ab0) at /home/tynach/Software/Code/fontforge/gdraw/ggdkdraw.c:863
#10 0x00007ffff7006be8 in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#11 0x00007ffff700604e in g_main_context_dispatch () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#12 0x00007ffff7006400 in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#13 0x00007ffff70064a3 in g_main_context_iteration () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#14 0x00005555555b66a7 in AnchorControl (sc=<optimized out>, ap=<optimized out>, layer=<optimized out>) at /home/tynach/Software/Code/fontforge/fontforgeexe/anchorsaway.c:1448
#15 0x00005555555b683a in AnchorControlClass (_sf=<optimized out>, ac=<optimized out>, layer=<optimized out>) at /home/tynach/Software/Code/fontforge/fontforgeexe/anchorsaway.c:1501
#16 0x00005555556aeae2 in AnchorClassD_ShowAnchors (e=<optimized out>, g=<optimized out>) at /home/tynach/Software/Code/fontforge/fontforgeexe/lookupui.c:2400
#17 AnchorClassD_ShowAnchors (g=<optimized out>, e=<optimized out>) at /home/tynach/Software/Code/fontforge/fontforgeexe/lookupui.c:2378
#18 0x00005555557462f7 in GButtonInvoked (b=b@entry=0x555556093310, ev=ev@entry=0x7fffffffc2f0) at /home/tynach/Software/Code/fontforge/gdraw/gbuttons.c:252
#19 0x000055555574678a in gbutton_mouse (g=0x555556093310, event=0x7fffffffc2f0) at /home/tynach/Software/Code/fontforge/gdraw/gbuttons.c:480
#20 0x000055555574bfe8 in _GWidget_Container_eh (gw=gw@entry=0x55555607fe60, event=event@entry=0x7fffffffc2f0) at /home/tynach/Software/Code/fontforge/gdraw/gcontainer.c:294
#21 0x000055555574c60a in _GWidget_TopLevel_eh (event=0x7fffffffc2f0, gw=0x55555607fe60) at /home/tynach/Software/Code/fontforge/gdraw/gcontainer.c:614
#22 _GWidget_TopLevel_eh (gw=0x55555607fe60, gw@entry=<error reading variable: value has been optimized out>, event=0x7fffffffc2f0, event@entry=<error reading variable: value has been optimized out>) at /home/tynach/Software/Code/fontforge/gdraw/gcontainer.c:562
#23 0x0000555555758aae in _GGDKDraw_CallEHChecked (gw=0x55555607fe60, event=<optimized out>, eh=<optimized out>) at /home/tynach/Software/Code/fontforge/gdraw/ggdkdraw.c:347
#24 0x00005555557593e2 in _GGDKDraw_DispatchEvent (event=<optimized out>, data=<optimized out>) at /home/tynach/Software/Code/fontforge/gdraw/ggdkdraw.c:1234
#25 0x00007ffff72fff69 in  () at /usr/lib/x86_64-linux-gnu/libgdk-3.so.0
#26 0x00007ffff73330f6 in  () at /usr/lib/x86_64-linux-gnu/libgdk-3.so.0
#27 0x00007ffff700617d in g_main_context_dispatch () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#28 0x00007ffff7006400 in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#29 0x00007ffff70064a3 in g_main_context_iteration () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#30 0x00005555556abc5f in AnchorClassD (sf=sf@entry=0x555555e5bb90, sub=sub@entry=0x555555e5c270, def_layer=def_layer@entry=1) at /home/tynach/Software/Code/fontforge/fontforgeexe/lookupui.c:2678
#31 0x00005555556b7faf in _LookupSubtableContents (sf=0x555555e5bb90, sub=0x555555e5c270, sd=sd@entry=0x0, def_layer=1) at /home/tynach/Software/Code/fontforge/fontforgeexe/lookupui.c:6248
#32 0x000055555565ecc7 in LookupSubtableContents (gfi=gfi@entry=0x555555efff80, isgpos=isgpos@entry=1) at /home/tynach/Software/Code/fontforge/fontforgeexe/fontinfo.c:5970
#33 0x0000555555678753 in LookupMouse (event=0x7fffffffceb0, isgpos=1, gfi=0x555555efff80) at /home/tynach/Software/Code/fontforge/fontforgeexe/fontinfo.c:7365
#34 lookups_e_h (gw=<optimized out>, event=0x7fffffffceb0, isgpos=<optimized out>) at /home/tynach/Software/Code/fontforge/fontforgeexe/fontinfo.c:7409
#35 0x000055555574c178 in _GWidget_Container_eh (gw=0x55555606eb50, gw@entry=<error reading variable: value has been optimized out>, event=0x7fffffffceb0, event@entry=<error reading variable: value has been optimized out>)
    at /home/tynach/Software/Code/fontforge/gdraw/gcontainer.c:395
#36 0x0000555555758aae in _GGDKDraw_CallEHChecked (gw=0x55555606eb50, event=<optimized out>, eh=<optimized out>) at /home/tynach/Software/Code/fontforge/gdraw/ggdkdraw.c:347
#37 0x00005555557593e2 in _GGDKDraw_DispatchEvent (event=<optimized out>, data=<optimized out>) at /home/tynach/Software/Code/fontforge/gdraw/ggdkdraw.c:1234
#38 0x00007ffff72fff69 in  () at /usr/lib/x86_64-linux-gnu/libgdk-3.so.0
#39 0x00007ffff73330f6 in  () at /usr/lib/x86_64-linux-gnu/libgdk-3.so.0
#40 0x00007ffff700617d in g_main_context_dispatch () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#41 0x00007ffff7006400 in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#42 0x00007ffff70064a3 in g_main_context_iteration () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#43 0x00005555557578bd in GGDKDrawEventLoop (gdisp=0x555555a48140) at /home/tynach/Software/Code/fontforge/gdraw/ggdkdraw.c:2131
#44 0x000055555572845e in fontforge_main (argc=<optimized out>, argv=<optimized out>) at /home/tynach/Software/Code/fontforge/fontforgeexe/startui.c:1032
#45 0x00007ffff6c97083 in __libc_start_main (main=0x5555555b08c0 <main>, argc=1, argv=0x7fffffffd698, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffd688) at ../csu/libc-start.c:308
#46 0x00005555555b08fe in _start ()
```

The culprit seems to be that within `AnchorPointsDuplicate()`, `sc->parent` is sometimes set to `NULL`, causing `sc->parent` to be at address `0x300`, which causes a segfault when an attempt is made to set `ac` too its value. I've studied the code around it, and *think* that what's happening is that only `base` is tested for a `NULL` value, because it's assumed that if it's *not* `NULL`, then `sc->parent` will also be valid.

I've tried changing the outer `for` loop to read:

```C
for ( ; base!=NULL && sc->parent!=NULL; base = base->next ) {
```

And this seems to fix this particular crash.. But I don't know if it could potentially lead to some data not being copied over when copying from one font to another. I've done some testing, and an error dialog box pops up if I try to just paste glyphs from one font into another font, but if I remake the anchor classes within the new font and then try pasting again it appears to work fine.

Trying to use 'Edit' → 'Copy Lookup Data' in one font, and then using 'Paste' on another font, crashes still... But it crashed before, still crashes, and does not seem to be a related crash.

I'd open a pull request, except I'm not confident in my fix. I don't know this codebase nearly well enough to know exactly what `sc->parent` is usually used for, or how it's supposed to be set.

Within the calling function `SplineCharCopy()`, there's the line `nsc->parent = into;` - and when I have a font that **doesn't** have glyphs containing both anchor marks *and* references to glyphs with those same anchor marks, I've used breakpoints to confirm that in that case both `nsc->anchor` *and* `into` (and thus `nsc->parent`) are set to `NULL` when that function calls `AnchorPointsDuplicate()`. This leads me to suspect that it's quite likely that a *real* fix would involve making sure that the `sc` passed to `SplineCharCopy` has `sc->anchor` set to `NULL` in the situations where this is happening... But I don't know enough about this codebase to have any idea where to look for that.

Tynach:
When bulk operations are performed on a font, FontForge runs `AnchorPoint *AnchorPointsDuplicate(AnchorPoint *base, SplineChar *sc)`. Early in the function there is a `for` loop which runs while `base` isn't NULL, but it later accesses `sc->parent->anchor` without checking if `sc->parent` itself is NULL first.

This seems to be because the function assumes it will only be called when the duplication of anchor points is actually desired, and thus there should be a valid place to copy the anchor points to. Instead, it's being called from within
`SplineChar *SplineCharCopy(SplineChar *sc, SplineFont *into, struct sfmergecontext *mc)`, which is itself being called with `into` set to 0 within 2 different cases handled by
`void SFDDumpUndo(FILE *sfd, SplineChar *sc, Undoes *u, const char *keyPrefix, int idx)`, and also once within
`Undoes *SFDGetUndo(FILE *sfd, SplineChar *sc, const char *startTag, int current_layer)`.

`SFDDumpUndo()` seems to only call `SplineCharCopy` as a means toward some of its side effects. It appears to make a new empty SplineChar copy the SplineChar it's working on into the new SplineChar, extract and dump hints from it, and then delete the SplineChar it created. Likewise, `SFDGetUndo()` seems to also only call `SplineCharCopy()` to do some temporary managing of hint information.

Since these appear to be the only scenarios where `SplineCharCopy` is called with `into` set to NULL, and the only thing that is being done in those situations is do some stuff relating to font hints, I've determined that it's safe to simply not duplicate the AnchorPoints when sc->parent is set to NULL.

This fixes #5130, and indeed is just the same fix I proposed back then. At the time I wasn't confident this was safe, but I've extensively used FontForge with this change made and had no further AnchorPoint crashes. I also further investigated the root causes, which I describe above.

A more 'proper' fix would probably include systematic changes to how hint undoes are saved and loaded from SFD files, but that's beyond my skill and, quite honestly, not something I even care about that much. The code appears to work despite it's oddities, and simply checking if `sc->parent` is NULL before doing anything with the AnchorPoints allows all the hint-related stuff that `SplineCharCopy()` does continue to work for the sake of both `SFDDumpUndo()` and `SFDGetUndo()`.

### Type of change

- **Bug fix**
    Fixes #5130

### DIFF
--- a/fontforge/fvfonts.c
+++ b/fontforge/fvfonts.c
@@ -311,7 +311,7 @@ static AnchorPoint *AnchorPointsDuplicate(AnchorPoint *base,SplineChar *sc) {
     AnchorPoint *head=NULL, *last=NULL, *cur;
     AnchorClass *ac;
 
-    for ( ; base!=NULL; base = base->next ) {
+    for ( ; base!=NULL && sc->parent!=NULL; base = base->next ) {
 	cur = chunkalloc(sizeof(AnchorPoint));
 	*cur = *base;
 	cur->next = NULL;


Tynach:
For the record, I was tempted to fix up the indentation in that function, but figured I'd start out by just making the one small change and keeping it formatted as much like the rest of the file as possible.

fontforge/fvfonts.c
```diff
@@ -311,7 +311,7 @@ static AnchorPoint *AnchorPointsDuplicate(AnchorPoint *base,SplineChar *sc) {
     AnchorPoint *head=NULL, *last=NULL, *cur;
     AnchorClass *ac;
 
-    for ( ; base!=NULL; base = base->next ) {
+    for ( ; base!=NULL && sc->parent!=NULL; base = base->next ) {
```
skef, line 314: I haven't looked at the conditions that lead to this problem but just looking at the loop itself I think I would be more comfortable with

```
    for ( ; base!=NULL; base = base->next ) {
        if (sc->parent==NULL)
            continue;
```
That way you still cover all the `base`s (as it were).

Tynach: What would that solve? The entire loop can be skipped over because `sc` and `sc->parent` are never changed within the body of the loop. It's pointless to iterate over and do that check a ton of times when we know that if it's NULL at the start, it will always be NULL the duration of the function.

Tynach: Actually, on that note, I suppose I could just change it to:

```c
    if (sc->parent == NULL)
        return head;
    for ( ; base!=NULL; base = base->next ) {
```

But having the check be part of the loop body seemed more elegant to me, because then there's still only one `return` statement at the bottom of the function.

Walking through a linked list and explicitly not doing anything with any of the elements seems a bit pointless.. Though I suppose with my proposed change, it does *both* checks for every element (when the function is called correctly without `sc->parent` being NULL), when it doesn't actually need to. That's a point in favor of having a check before the loop that adds another `return` statement.

Tynach: Ah, I suppose I *could* just wrap the whole `for` loop in an `if` statement that checks if it's *not* NULL. That would both ensure the check only happens once, *and* keeps only a single `return` statement.

However, that would break Linux kernel guidelines, which state to not indent more than 3 times.. I've found that rule to be bizarrely very good at fixing bugs for no discernible reason, so I try to stick with it. The function already breaks it with 4-levels of indentation at times, but this would be bumping that up to 5 levels of indentation. Besides, I've heard that it's better to avoid wrapping things in `if` statements like that if you can just have a small `if` statement that returns early instead.

I could force things by instead of an extra `return` statement, have a `goto` that jumps to the return statement at the bottom.. But `goto` is discouraged for good reason and I feel like that's a rather silly use for one.

What are your thoughts, @skef?

Tynach: ... Actually, come to think of it, maybe this is the wrong place to change *any*thing. Maybe it'd be best to just.. Remove the function call. That is, make sure that the function isn't called if `sc->parent` is NULL to begin with. The function is only called in a single place (within `SplineCharCopy()`), so we could just wrap the function call in an `if` statement.

That would be more elegant anyway, because that's *also* where `sc->parent` gets set to NULL to begin with, currently on line 531 of fontforge/fvfonts.c:

```c
    nsc->parent = into;
```

We could just change line 542 (the call to `AnchorPointsDuplicate()`) to be:

```c
    if (into)
        nsc->anchor = AnchorPointsDuplicate(nsc->anchor,nsc);
```

And that should be just as good. After all, if we're detecting that we're copying things into NULL, we should just.. Not do that.

Tynach: Turns out there's another line in that function that gives special handling for when `into` is set to `NULL` or `0`: line 505:

```c
    nsc->layer_cnt = into==NULL?2:into->layer_cnt;
```

Indeed, if I try to *just* wrap that `nsc->anchor = ...` line in an `if` statement, trying to open any file causes a crash. It has to be set to `NULL` instead of being left as-is.

The fix that works is changing line 542 to this:

```c
    nsc->anchor = into==NULL?NULL:AnchorPointsDuplicate(nsc->anchor,nsc);
```

skef: I reviewed this in a hurry and didn't think my advice entirely through. I think it's fine to return NULL before the loop if sc->parent is NULL. That makes it clear why the flow doesn't progress through the loop.

Tynach:
@skef, what are your thoughts about skipping the function call entirely instead? I haven't tested that as thoroughly, but I don't see any reason why skipping the function call would be any different from skipping the loop, as long as `nsc->anchor` is set to `NULL` in the calling function (`SplineCharCopy()`) as an alternative.

If that's deemed as a preferred way of handling this, should I revert my initial commit in this pull request, then commit a new change? Or should I just open a new pull request?

skef:
It's best for a function to (try to) ensure it's own correctness when it comes to data corruption. So if a function will dereference some data member it should either assert it is non-null (when there is a conceptual chance it will be null but the developer believes it can't be in that context) or check and handle the problem itself, possibly emitting an error message.

So if you believe the function shouldn't be called in this case in the first place (and is properly being called in other cases), you have the option of adding an `assert()` and avoiding the call. From your initial description of the bug it sounds like that may be the case. If so, it's fine with me.

You can re-use this PR by force-updating or just adding new commits to the branch or make a new one, either is fine.

Tynach:
I've decided to submit a different pull request to fix this. I don't like needlessly overwriting other people's work, since it messes up `git blame`; makes it more difficult to tell what was changed and when.

New pull request is #5405

Tynach:
I've decided to submit a different pull request to fix this. I don't like needlessly overwriting other people's work, since it messes up `git blame`; makes it more difficult to tell what was changed and when.

New pull request is https://github.com/fontforge/fontforge/pull/5405

PR #5423, opened 2024-05-24
Use sysconfig for Python module locations

iorsh:
[distutils](https://docs.python.org/3.10/library/distutils.html#module-distutils) is deprecated with removal planned for Python 3.12.

- Replace distutils with a similar functionality from sysconfig to avoid CI failure
- Update macos CI runner to 12

This change is necessary to update CI to macos-12, which doesn't have Python 3.10.

**TODO:** Revert `iorsh/fontforgebuilds` repo before merge and after https://github.com/jtanx/fontforgebuilds/pull/21 is merged into fontforgebuilds

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -130,7 +130,8 @@ jobs:
       - name: Checkout FontForgeBuilds
         uses: actions/checkout@v3
         with:
-          repository: fontforge/fontforgebuilds
+          repository: iorsh/fontforgebuilds
+          ref: py_sysconfig
           path: fontforgebuilds
       - name: Setup MSYS2
         uses: msys2/setup-msys2@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,7 +74,7 @@ jobs:
           CROWDIN_API_KEY: ${{ secrets.CROWDIN_API_KEY }}
         run: java -jar $DEPSPREFIX/crowdin-cli.jar -c .crowdin.yml upload
   mac:
-    runs-on: macos-11
+    runs-on: macos-12
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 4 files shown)


iorsh:
Ok, appveyor is failing. I'll need to look into this with a separate dummy PR

iorsh:
Pointed appveyor to `iorsh/fontforgebuilds` repo, tests now pass

.github/workflows/main.yml
```diff
@@ -106,7 +106,7 @@ jobs:
           # MacOS runner has way too much preinstalled Python versions. We must
           # use the exact one discovered by CMake.
           PYTHON=`cat build/CMake_Python3_EXECUTABLE`
-          PYTHONPATH=$PYTHONPATH:$PREFIX/$($PYTHON -c "import distutils.sysconfig as sc; print(sc.get_python_lib(prefix='', plat_specific=True,standard_lib=False))")
+          PYTHONPATH=$PYTHONPATH:$PREFIX/$($PYTHON -c "import sysconfig as sc; print(sc.get_path('platlib', sc.get_preferred_scheme('user'), vars={'userbase': '.'}))")
```
jtanx, line 109: Same comment as on the fontforgebuilds pr, I think this should be

```suggestion
          PYTHONPATH=$PYTHONPATH:$PREFIX/$($PYTHON -c "import sysconfig as sc; print(sc.get_path('platlib', sc.get_preferred_scheme('prefix'), vars={'platbase': '.'}))")
```

iorsh: Please see https://github.com/jtanx/fontforgebuilds/pull/21#issuecomment-2131031460

> Unfortunately `'prefix'` key is inconsistent across platforms. On MacOS the preferred scheme for `prefix` is `osx_framework_library`, and its platlib entry looks like
> 
> ```
> 'platlib': '/usr/local/{platlibdir}/python{py_version_short}/site-packages'
> ```
> 
> This forces absolute path (which is discouraged by CMake) and the variable names are different for each platform. You can check this at https://github.com/iorsh/fontforge/actions/runs/9223726161/ - see the "Run pyhook smoke test" entry for Mac and Linux

iorsh:
@jtanx, do you want any additional changes here? Could you please merge if this is ok with you?

skef (approved):
This is a little scary but I suppose since it's CI stuff the results should be self-evident. If you give me the go-ahead I'll merge.

jtanx:
Woops, forgot to reply to this one. I think there may have still been something but I've forgotten now, so if it builds, send it

PR #5431, opened 2024-06-14
This partially fixes building on Python 3.13.

limburgher:
Additional work is needed as distutils is removed; In Fedora we work around this by specifying PYHOOK_INSTALL_DIR at build time.

**Bug fix**
**Non-breaking change**


iorsh:
See https://github.com/fontforge/fontforge/pull/5423 - should be working, still being discussed in https://github.com/jtanx/fontforgebuilds/pull/21

limburgher:
It that fixes it, I can close this.

PR #5451, opened 2024-07-24
Only install GUI-specific files if ENABLE_GUI is set

cjmayo:
The desktop menu entry and MIME association are not usable if the GUI is not installed.
There are also hundreds of unused image files.

---

This reduces the number of files installed for me from 412 to 40.

### Type of change

- **Non-breaking change**


skef:
This is reasonable but it scares me. I'll approve if someone (e.g. @iorsh) looks into it and thinks it's OK.

iorsh:
Looks good to me. We also have NoUI CI pipeline, so if it doesn't break, we would be good to go.

iorsh (approved):

skef (approved):

PR #5456, opened 2024-08-08
Update start.c

flagsoft:
fixes a crash on macOS file, open, .svg inkscape own font design

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/start.c
+++ b/fontforge/start.c
@@ -77,7 +77,8 @@ void InitSimpleStuff(void) {
     initrand();
     initadobeenc();
 
-    setlocale(LC_ALL,"");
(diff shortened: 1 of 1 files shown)


iorsh (changes requested):
It's a very general change, and its implications across platforms and across locales are difficult to predict. Please provide the root cause for the crash you encountered, preferably with stack. Please limit your fix to the root cause.

flagsoft:
I have no idea why this is so, but it also works if user sets environment variables like so:

This will fix the macOS fontforge file open svg font crash.

```
% export LANG="en_US.UTF-8"
% export LC_ALL="en_US.UTF-8"
% /Volumes/DATA/Applications/FontForge.app/Contents/Resources/opt/local/bin/fontforge

```

flagsoft:
Please fix it within your source code. use a locale independent version of strtod().

PR #5457, opened 2024-08-13
Updated python api to allow importOutlines without using filesystem

### LINKED ISSUE #5454: python feature request: importOutlines from memory or stream with format
there is a method `font.importOutlines(path)`, which takes a file path, guesses the format by the extension, and read outlines from the file. but if your data is not all in separate files, you have to create a temporary file with an extension (e.g. /tmp/asdf.svg), write data to it, and pass it to this function. this may cause worse performance (it's copying data a lot, /tmp/ may be ram, may be ssd, may be hdd, since an extension is required memfd cannot be used), and also is just more code

it would be nice to have something like `font.importOutlineData(bytes(svg_doc), format="svg")`, or to have the existing method accept any stream/file-like object: `font.importOutlines(BytesIO(bytes(svg_doc)), format="svg")`

zacharyvmm:
### Type of change

Allows glyph to import outlines with file-like object instead of just filename.

- **Bug fix**
- **New feature**
resolves #5454
- **Breaking change**
- **Non-breaking change**


iorsh (commented):
I have two major suspicions here. I'm away from computer, so can't verify myself, but...
1. A keyword-style first argument would fail for legacy scripts, like `glyph.importOutlines(filename="foo.svg")`. Better add `file` keyword without killing the `filename`.
2. `fmemopen()` doesn't seem to be available on Windows. I'd go for `std::istream` as an opaque C pointer with appropriate wrappers replacing `FILE` calls

zacharyMcSweenManickchand:
@iorsh

I have researched `std::istream`, while it does read the buffer as a stream. It does not seem to be compatible with FILE.

Instead, I chose to use a temporary file, `tmpfile()`.

This is because:
1. It avoids the complexities of working with Windows in-memory streams.
2. It's relatively fast considering in the current implementation (master, not my pr), the Python script has to create a temporary file.
3. This solution can be upgraded in the future if need be (Here are some solutions I found):
    - https://github.com/Arryboom/fmemopen_windows/
    - https://github.com/Snaipe/fmem/

iorsh:
I'm ok with temp file approach, indeed `std::istream` would have required a lot of refactoring.

fontforge/python.c
```diff
@@ -8605,44 +8649,130 @@ static PyObject *PyFFGlyph_import(PyObject *self, PyObject *args,
...
+        return NULL;
     }
-    else if ( strcasecmp(pt,".svg")==0 ) {
-	SCImportSVG(sc,((PyFF_Glyph *) self)->layer,locfilename,NULL,0,false,ipp);
+
+
+    if ( strcasecmp(filetype,"eps")==0 || strcasecmp(filetype,"ps")==0 || strcasecmp(filetype,"art")==0 ) {
+        if ( locfilename==NULL ){
+            FILE *psfile = NULL;
+            #ifdef __unix__
```
iorsh, line 8672: Please extract this ifdef...else...endif thing into a separate function and place it in, maybe, `fsys.c`.

zacharyMcSweenManickchand: Done

gutils/gimagereadras.c
```diff
@@ -328,13 +321,25 @@ GImage *GImageReadRas(char *filename) {
...
     }
 
 errorGImageReadRas:
-    fprintf(stderr,"Bad input file \"%s\"\n",filename );
+    fprintf(stderr,"Bad input file\n");
     GImageDestroy(ret);
-    fclose(fp);
     return( NULL );
 }
+
```
skef, line 332: You don't need all of these, and I'd prefer to see them refactored a bit:

1. Make one function "GImageRead_Wrapper(char * filename, GImageReadFunc *func) ...
2. Typedef `GImageReadFunc` as something like `GImage *(*GImageReadFunc)(File *, int *);`, where the second argument (when not NULL) returns non-zero on error and 0 on success.
3. Modify your _ functions to take the second argument. If it is null, print the error, If it is non-null, set the success condition and return without printing.
4. Modify the (now single) wrapper to call the passed function pointer with a non-null second argument (e.g. `int success;` and then `&success` as the argument. Check for success, and on error print the error message with the filename.
5. Modify the calls to the _ functions in GImageReadBuf to pass NULL as the second argument. 

This way there is less code and you don't loose the filename in the error message when it is available.

skef:
@iorsh I tried to fix up the pattern I asked for. The code passes the tests but that doesn't give me much confidence. We need to consider:

* Additional testing: preferably automated but failing that manual.
* Indentation: These changes are really inconsistent with the existing functions and I didn't help things.

What do you think?

inc/gimage.h
```diff
@@ -155,18 +155,20 @@ extern Color GDrawColorBrighten(Color col, int by);
 extern int GImageWriteGImage(GImage *gi, char *filename);
 extern int GImageWrite_Bmp(GImage *gi, FILE *fp);
 extern int GImageWriteBmp(GImage *gi, char *filename);
-extern GImage *GImageRead_Bmp(FILE *file);
+extern GImage *GImageRead_Bmp(FILE *file, int *success);
```
iorsh, line 158: As far as I can see, all the specific image readers used to signal success/failure by non-NULL or NULL return value. Is there any benefit from explicitly setting `success` flag?

tests/test1022.py
iorsh, line 1: Please add this test to `tests/CMakeLists.txt`

gutils/fsys.c
```diff
@@ -1018,6 +1018,83 @@ static int mime_comp(const void *k, const void *v) {
...
+    FILE *file = NULL;
+    #ifdef __unix__
+    file = fmemopen(buffer, size, mode);
+    #else
+    file = tmpfile();
+    if ( file==NULL ) {
+        fprintf(stderr, "Failed to create temporary file");
+        return NULL;
+    }
+    fwrite(buffer, size, 1, file);
```
iorsh, line 1031: `fwrite()` puts the file-position indicator at the end of the buffer, which is probably undesirable. You would need to call `rewind()` or `fseek()` after it.

iorsh:
> @iorsh I tried to fix up the pattern I asked for. The code passes the tests but that doesn't give me much confidence. We need to consider:
> 
>     * Additional testing: preferably automated but failing that manual.
> 
>     * Indentation: These changes are really inconsistent with the existing functions and I didn't help things.
> 
> 
> What do you think?

I can live with the indentation. There is already a lot of mess with it, and I wouldn't bother to be overly picky about it.

The new unit test is nice, let's see how it runs once enabled. I'll think about performing more testing with minimum pain.

inc/gimage.h
skef: I guess we could change `success` to a boolean (along with its name). My concern was that the previous version of the code wound up printing error messages without the filename whereas the code before this PR would include it. Assuming this rough arrangement we need some way of telling the called function whether to print its own message or not.

skef:
> I can live with the indentation. There is already a lot of mess with it, and I wouldn't bother to be overly picky about it.

I think that's fine with me.

skef:
test1022.py seems to be screwed up. One issue is unreliable relative paths to the doc directory. But more generally it's not clear what is supposed to add the files to `/tmp` so that `generic_file_stream_testing` can read them.

PR #5469, opened 2024-09-14
Fix bug in font.appendSFNTName

### LINKED ISSUE #2635: appendSFNTName doesn't always replace existing 'Family' value
If you have a font that has a style component in it's 'Family' SFNT variable (e.g. `('English (US)', 'Family', 'Source Code Pro Medium')`), then trying to shorten this name by removing the style component fails:
`Font.appendSFNTName('English (US)', 'Family', "Source Code Pro")` doesn't overwrite the existing 'Family' variable.
However, if you change the new variable slightly, even add one space or anything, e.g. `Font.appendSFNTName('English (US)', 'Family', "Source Code Pro ")` or
`sourceFont.appendSFNTName('English (US)', 'Family', "Source Code P")`
then it will overwrite existing value

The only solution so far seems to be to clear this value with
`Font.appendSFNTName('English (US)', 'Family', '')` and then doing what you originally wanted to
`Font.appendSFNTName('English (US)', 'Family', "Source Code Pro")`

System: OS X 10.11.3
FontForge (installed via Homebrew): stable 20150824

siroccal:
This unneeded if-statement made it impossible to change e.g. the 'Family' SFNT data from 'Sans Bold' to 'Sans'. Fixes #2635.

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -11870,11 +11870,6 @@ return( 1 );
 	Py_XDECREF(val);
 return( 0 );
     }
-    if ( lang==0x409 && english!=NULL && english->names[strid]!=NULL &&
-         strcmp(string,english->names[strid])==0 ) {
-	Py_DECREF(val);
-return( 1 );	/* If they set it to the default, there's nothing to do */
-    }
 
     if ( names==NULL ) {
 	names = chunkalloc(sizeof( struct ttflangname ));


skef (commented):
I'm confused by this -- `strcmp` shouldn't return 0 when comparing "Sans" to "Sans Bold". Are you sure you have the analysis of your problem right?

siroccal:
> I'm confused by this -- `strcmp` shouldn't return 0 when comparing "Sans" to "Sans Bold". Are you sure you have the analysis of your problem right?

You can try it out yourself and see it doesn't work properly.
E.g. download [Source Sans Pro](https://www.fontsquirrel.com/fonts/source-sans-pro), extract 'SourceSansPro-Semibold.otf' and try to use `appendSFNTName` to change the name from 'Source Sans Pro Semibold' to 'Source Sans Pro'. And you'll see it doesn't work:
```
$ python Code/font-name-fix.py -f 'Source Sans Pro' .local/share/fonts/adobe-source-sans/SourceSansPro-Semibold.otf
$ fc-list | grep -i 'Source Sans Pro'
/home/user/.local/share/fonts/adobe-source-sans/SourceSansPro-Semibold.otf: Source Sans Pro,Source Sans Pro Semibold:style=Semibold,Regular
```
The font name stays 'Source Sans Pro,Source Sans Pro Semibold'.

But after applying this patch it works properly:
```
$ python Code/font-name-fix.py -f 'Source Sans Pro' .local/share/fonts/adobe-source-sans/SourceSansPro-Semibold.otf
$ fc-list | grep -i 'Source Sans Pro'
/home/user/.local/share/fonts/adobe-source-sans/SourceSansPro-Semibold.otf: Source Sans Pro:style=Semibold,Regular
```
Now the font name is just 'Source Sans Pro', as intended.

Here's the python script I used: [font-name-fix.txt](https://github.com/user-attachments/files/17284526/font-name-fix.txt) (just change extension to .py)

siroccal:
> It looks like the solution is slightly different. `SetSFNTName()` is not called with the real English name values, but rather with _default_ values produced earlier by `DefaultTTFEnglishNames()`. So when the current value differs from default, but the target value is the default, `strcmp()` does receive two identical strings and produces `0`.
> 
> So for the English names the logic should be as follows:
> 
>     1. The new name is empty ==> delete entry
> 
>     2. **The new name equals to default ==> delete entry**
> 
>     3. The new name differs from default ==> add or overwrite entry
> 
> 
> Instead of deleting the `if` clause entirely we should do
> 
> ```
> 	if ( names!=NULL ) {
> 	    free( names->names[strid] );
> 	    names->names[strid] = NULL;
> 	}
> 	Py_DECREF(val);
> ```

Yeah, that does also seem to work for me.

Removing the if-statement entirely or doing it your way both seem to work for fixing #2635, so either approach is fine with me. If your way is implemented, then this pull request can be closed.

skef:
@iorsh Looks like the author would rather we implement the suggested fix rather than changing this PR. What do you think?

iorsh:
> @iorsh Looks like the author would rather we implement the suggested fix rather than changing this PR. What do you think?

I'm ok with that. Superseding this PR with #5494

PR #5476, opened 2024-10-02
Fix resource leak in unParseTTInstrs

ideasman42:
PySequence_GetItem adds a reference count to the argument it returns. Use PySequence_Fast() to simplify access.

### Type of change
- **Bug fix**

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -952,32 +952,43 @@ return( binstr );
 }
 
 static PyObject *PyFF_UnParseTTFInstrs(PyObject *UNUSED(self), PyObject *args) {
-    PyObject *tuple, *ret;
+    PyObject *seq_any, *ret;
     int icnt, i;
     uint8_t *instrs;
     char *as_str;
 
-    if ( !PyArg_ParseTuple(args,"O",&tuple) )
+    if ( !PyArg_ParseTuple(args,"O",&seq_any) )
 return( NULL );
-    if ( !PySequence_Check(tuple)) {
+    if ( !PySequence_Check(seq_any)) {
 	PyErr_Format(PyExc_TypeError, "Argument must be a sequence" );
 return( NULL );
     }
-    if ( PyBytes_Check(tuple)) {
+
+    if ( PyBytes_Check(seq_any)) {
 	char *space; Py_ssize_t len;
-	PyBytes_AsStringAndSize(tuple,&space,&len);
+	PyBytes_AsStringAndSize(seq_any,&space,&len);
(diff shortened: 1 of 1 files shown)


fontforge/python.c
```diff
@@ -952,32 +952,43 @@ return( binstr );
...
+    if ( PyBytes_Check(seq_any)) {
 	char *space; Py_ssize_t len;
-	PyBytes_AsStringAndSize(tuple,&space,&len);
+	PyBytes_AsStringAndSize(seq_any,&space,&len);
 	instrs = calloc(len,sizeof(uint8_t));
 	icnt = len;
 	memcpy(instrs,space,len);
     } else {
-	icnt = PySequence_Size(tuple);
+	PyObject *seq_fast = PySequence_Fast(seq_any,__func__);
```
skef, line 974: Seem like returning NULL here will result in an error message that is just the name of this C function. That's a little opaque -- how will the caller know what to do?

ideasman42: The possibility of `PySequence_Fast(..)` returning an error that uses `__func__` as an error is (I believe) so unlikely that it's difficult to justify a comprehensive message. I would say it's _ball-park_ similar likelihood of the `malloc` failing or other rare-but-possible things that could be checked for.

That error is used when: `PyObject_GetIter(seq)` raises a `PyExc_TypeError`.
Since the object has already passed a `PySequence_Check`, a type error is likely to be caused by some internal error where the sequence is unable to create and iterator the iterator function happens to raise a TypeError.
Note that this isn't an error accessing the items from the iterator (which could happen for any number of reasons), the error is only caused by creating the iterator instance.

This can be explained in a message which might read: `"unParseTTInstrs: error creating an iterator for the  truetype-instruction sequence"`

... even if this were added, any errors accessing the sequence items (a much more likely) error, would simply set the sequence error and not provide any context to the script author.

For improved error messages, it would be necessary to use:

`PyArg_ParseTuple` 's  `"O&"` formatting, which moves the validation into the argument parsing. Then any error generated will have the context of the argument that caused the error.

skef: Fair enough

skef (approved):
Tentatively approved

skef:
@iorsh any thoughts?

iorsh:
I'll review this and other stuff later today.

iorsh:
I'm not quite proficient in C-level Python reference counting. The change looks reasonable to me.

iorsh (approved):

skef:
I'm just going to override the checks rather than try to get the mac build working.

PR #5483, opened 2024-10-25
Avoid crashes in Python scripts when objects are accessed in invalid state

### LINKED ISSUE #4615: Python GlyphPen "changed" heuristic greatly increases chances of crash
The python API has a number of crash scenarios that would be time consuming to systematically fix. If you `close()` a font at the python level there can still be any number of `fontforge.glyph` objects (for example) that, if subsequently used, will cause crashes. This isn't great, but because such scenarios are rare and avoidable it isn't the end of the world.

Unfortunately `PyFF_GlyphPen_dealloc()` has the following code:

```
static void PyFF_GlyphPen_dealloc(PyFF_GlyphPen *self) {
    if ( self->sc!=NULL ) {
        if ( self->changed )
            SCCharChangedUpdate(self->sc,self->layer);
        self->sc = NULL;
    }
    Py_TYPE(self)->tp_free((PyObject *) self);
}
```

Changed is set to `True` by almost every operation and never set back to `False`.

I think the idea here is to avoid calling `SCCharChangedUpdate()` after every change for performance reasons and therefore to do it on deallocation as an alternative. That's not *great* because the pen might hang around because there's still a reference to it, but maybe it's acceptable. Unfortunately it means a GlyphPen dangling after a font is closed (or, presumably, after the underlying `SplineChar` is deleted) will call `SCCharChangedUpdate()` on freed memory even when it is never used again. 

The proper fix for these problems would be to keep lists of python objects with their respective objects and "invalidate" them as needed. Short of that I'm not sure what to do about this problem. One option is to add a "delay update" parameter to each pen method and call `SCCharChangedUpdate()` when it isn't specified. Then folks who don't want to pay the performance penalty could add the parameter and everyone else could just live with it.

### LINKED ISSUE #5479: Python crashes when font-related objects are accessed after the font has been closed
After `fontforge.font` object is closed, and its memory relased, any operation with its glyphs etc. causes segfault.

Sample script:
```
import sys, fontforge

font=fontforge.open(sys.argv[1])

gA = font["A"]
font.close()
l = gA.layers[1]
```
(run it with any Latin-enabled font, e.g. `fontforge -script test.py Ambrosia.sfd`)

Expected result: the script shall throw `RuntimeError` instead.

See also discussion in #5348.

iorsh:
When the font is closed or a specific glyph is deleted in a Python script, there could still be Python objects around which reference the deleted C object. This causes segmentation faults in many scenarios.

We add checks and references which allow to check C object validity before performing operations.

Fixes #4615, fixes #5479.


iorsh:
The diff in `python.c` is infinite, but the most interesting stuff is as follows:

- `PyFF_FreeFV()` - line 20343, invalidating the unique PyFF_Font object by setting its `SplineFont *fv` pointer to `NULL`. 
- `PyFF_FreeSC()` - line 20357, invalidating the unique PyFF_Glyph object by setting its `SplineChar *sc` pointer to `NULL`
- `splineuti.c:6001` - invalidate the Python object when SplineChar is destructed.
- `PyFF_Glyph_GetSC()` - line 479, this function is the only way to get `SplineChar*` from `PyFF_Glyph`, and it raises exception whenever the glyph is invalid.
- `struct glyph_accessors_` - line 8093, introduces new internal getter/setter calling sequence for `PyFF_Glyph`, which receives `SplineChar*` as a **guaranteed** non-NULL argument.
- `PyFF_Glyph_set/get_wrapper()` - line 10155, wrappers which implement Python standard getter/setter calling sequence, verify `PyFF_Glyph` and call internal accessors only when the glyph is valid.
- `setup_glyph_type()` - line 10172, finalizes the PyFF_GlyphType with wrapped getters/setters.

iorsh:
Ready for review now.

skef:
Sorry, I want to confirm the reference counting in this code and it's going to take me a while.

skef (approved):
I didn't check this for completeness but what's here looks good to me and barring any bugs that should be easy to fix will make things better. Good work!

PR #5492, opened 2024-11-08
Warning rollup from clang trunk

nabijaczleweli:
### Type of change
- **Non-breaking change**

These were found in a default build on clang trunk.

I prioritised retaining actual behaviour. There are probably one or two spots (the tautological comparisons, first_cnt most likely) where I actually found and enshrined a bug! Don't apply any of these without validating!

Also:
```
[66/280] Building C object fontforge/CMakeFiles/fontforge.dir/encoding.c.o
/home/nabijaczleweli/uwu/fontforge/fontforge/encoding.c:2269:28: warning: variable 'refs' used in loop condition not modified in loop body [-Wfor-loop-analysis]
 2269 |             for ( bref=bfc->refs; refs!=NULL; bref = brnext ) {
      |                                   ^~~~
1 warning generated.
```
this is an infinite loop if `refs!=NULL`. So `refs == NULL` always at this point. Someone who understands this code should either remove the loop or fix the condition.

I *ignored* -Wmaybe-uninitialized. There's *so many*. Someone who understands the code should build with Clang and see!

### DIFF
--- a/fontforge/stemdb.c
+++ b/fontforge/stemdb.c
@@ -2362,8 +2362,7 @@ return( 0 );
     /* fonts or fonts with quadratic splines). */
     /* But do that only for colinear spline segments and ensure that there are  */
     /* no bends between two splines. */
-    if ( !tp && ( !fp || t > 0.5 ) &&
-	topd->colinear && &other->to->next != NULL ) {
+    if ( !tp && ( !fp || t > 0.5 ) && topd->colinear ) {
 	testpt = topt->next->to; 
 	testpd = &gd->points[testpt->ptindex];
 	BasePoint *initdir = &topd->prevunit;
@@ -2378,8 +2377,7 @@ return( 0 );
 	}
 	if ( tp ) t_needs_recalc = true;
     }
-    if ( !fp && ( !fp || t < 0.5 ) &&
-	frompd->colinear && &other->from->prev != NULL ) {
+    if ( !fp && ( !fp || t < 0.5 ) && frompd->colinear ) {
 	testpt = frompt->prev->from; 
 	testpd = &gd->points[testpt->ptindex];
 	BasePoint *initdir = &frompd->prevunit;
--- a/fontforge/tottfvar.c
+++ b/fontforge/tottfvar.c
@@ -572,26 +572,12 @@ return;
 	}
 	/* Now output the corresponding deltas for those points */
 	for ( j=0; j<pcnt; ) {
-	    if ( deltas[i][j]>0x7f || deltas[i][j]<0x80 ) {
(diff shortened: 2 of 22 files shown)


fontforge/stemdb.c
```diff
@@ -2362,8 +2362,7 @@ return( 0 );
     /* fonts or fonts with quadratic splines). */
     /* But do that only for colinear spline segments and ensure that there are  */
     /* no bends between two splines. */
-    if ( !tp && ( !fp || t > 0.5 ) &&
-	topd->colinear && &other->to->next != NULL ) {
```
skef, line 2366: Change it to `topd->next != NULL` rather than deleting it. That's the likely original intention and it can't hurt.

fontforge/stemdb.c
```diff
@@ -2378,8 +2377,7 @@ return( 0 );
 	}
 	if ( tp ) t_needs_recalc = true;
     }
-    if ( !fp && ( !fp || t < 0.5 ) &&
-	frompd->colinear && &other->from->prev != NULL ) {
+    if ( !fp && ( !fp || t < 0.5 ) && frompd->colinear ) {
```
skef, line 2380: Make this one `frompt->prev != NULL`.

fontforge/tottfvar.c
```diff
@@ -572,26 +572,12 @@ return;
 	}
 	/* Now output the corresponding deltas for those points */
 	for ( j=0; j<pcnt; ) {
-	    if ( deltas[i][j]>0x7f || deltas[i][j]<0x80 ) {
```
skef, line 575: These changes obviously don't affect the behavior but they make me nervous. But this is a MultiMaster thing and while that code may be revived at some point for VF support it's effectively dead now.

fontforge/stemdb.c
nabijaczleweli: Applied

fontforge/stemdb.c
nabijaczleweli: Applied

fontforge/tottfvar.c
nabijaczleweli: I tried to make some sense of this but it appears whole-sale in 6243f733fb327c99b91987be95a70712e9b69a3c with the initial addition of this file in 2004 and hasn't changed since, so if it behaved correctly when tested then, then clearly this was meant to be a dead branch?

skef: It's hard to judge because the multi-master code was almost never used. The implementation is at best incomplete.

skef:
@iorsh this tentatively looks good to me. Any thoughts?

PR #5496, opened 2024-11-14
Add contour draw option to H.Metrics.

vasilky3:
I'm working with cnc fonts. And it's hard to navigate dew to lack of glyph splines visualisation in the MainWindow.
So i added one. 
image
### Type of change
- **New feature**

### Code quality 
I would say it's low. But somewhat consistent with surrounded code. 

My main problem  new method `fontview.c :: FVDrawOutlineOnly()` probably should be placed in `gdrawP.h :: displayfuncs`.
But i was confused trying to figure out how to achieve this. 

Sorry for not putting spacing refactor in a separate commit.


fontforgeexe/fontview.c
```diff
@@ -245,180 +245,268 @@ extern BDFFont *FVSplineFontPieceMeal(SplineFont *sf, int layer, int ptsize, int
...
-    GRect box, old2;
-    int feat_gid;
-    SplineChar *sc;
-    struct _GImage base;
-    GImage gi;
-    GClut clut;
-    int i,j;
-    int em = fv->b.sf->ascent+fv->b.sf->descent;
-    int yorg = fv->magnify*(fv->show->ascent);
+void FVDrawOutlineOnly(GWindow pixmap, SplinePointList *set,
```
myoresh, line 248: Spaces vs tabs garble the diff. Please adjust indents (tab stands for double indent, 4 spaces for single indent where present), at least in FVDrawGlyph().

iorsh, line 248: Spaces vs tabs garble the diff. Please adjust indents (tab stands for double indent, 4 spaces for single indent where present), at least in FVDrawGlyph().

vasilky3: done 

> at least in FVDrawGlyph()

fontforgeexe/fontview.c
```diff
@@ -245,7 +245,71 @@ extern BDFFont *FVSplineFontPieceMeal(SplineFont *sf, int layer, int ptsize, int
     return new;
 }
 
-static void FVDrawGlyph(GWindow pixmap, FontView *fv, int index, int request_expose ) {
+void FVDrawOutlineOnly(GWindow pixmap, SplinePointList *set,
+                       Color fg, enum outlinesfm_flags strokeFillMode, float xoff, float yoff, double scale) {
+	if (!set) {
+		return;
+	}
+	for (SplinePointList *spl = set; spl != NULL; spl = spl->next) {
```
iorsh, line 253: It looks like this code mostly duplicates a part of `CVDrawSplineSetOutlineOnly()`. Could you please extract it to a common function?

fontforge/CMakeLists.txt
```diff
@@ -227,7 +227,7 @@ if(BUILD_SHARED_LIBS)
     # Both fontforge.exe and *.pyd Python modules look for libfontforge.dll in
     # their directory, so we need to produce a second copy.
     add_custom_command(TARGET fontforge POST_BUILD
-      COMMAND cp $<TARGET_FILE:fontforge> ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
+            COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:fontforge> ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
```
iorsh, line 230: I'm generally ok with this change, but does it fix a specific bug or just improves style?

vasilky3: QOL change.
It's for win compilation. cp not supported in win. While "Cmake copy" is cross platform.

fontforgeexe/fontview.c
vasilky3: you right i get it from `CVDrawSplineSetOutlineOnly()`

for some reason `CVDrawSplineSetOutlineOnly()` code needs extra attributes. And Y coordinates are inverse. Also a grid drawing code. 

`cy2 = -tab->yoff + cv->height - cy2*tab->scale - dy;`
vs
`cy2 = yoff - spline->to->prevcp.y * scale - dy;`

When i trying to extract function from  `CVDrawSplineSetOutlineOnly()` it has extra dependencies and functions i don't have in FV scope.
`void PixmapDrawSplineSet(CharView *cv, GWindow pixmap, SplinePointList *set, Color fg, enum outlinesfm_flags strokeFillMode, SplinePointList *spl, int activelayer, CharViewTab *tab)`
I was not able to find this in FV scope:
`CharView *cv, int activelayer, CharViewTab *tab`

With my current knowledge of code i wont risk refactor existing code.
If you insist, I will.

iorsh: My intent was to extract only the lines 1533-1586 from CVDrawSplineSetOutlineOnly(), and this is effectively what you did.
I think you can extract them to a function like
```
CVDrawSplinePointList(CharView *cv, GWindow pixmap, SplinePointList *set,
    Color fg, enum outlinesfm_flags strokeFillMode,
    float xoff, float yoff, real scale, int height)
```

You would be able to call that function both from CVDrawSplineSetOutlineOnly() and from font view with appropriate values for xoff/yoff/scale, and from fontview with `cv=NULL, height=0`. 

`rpt()` would need to be modified to NULL-check the `cv` argument. My only concern is about `rint()` calls you removed. Did you do that on purpose?

vasilky3: > My only concern is about rint() calls you removed. Did you do that on purpose?
I was trying to remove as much deps as possible from function. In order to make it work. 
I didn't understood what rint() and rpt() are for. So i will look into it.

I need couple days to make it. Will notify then done.

vasilky3: @iorsh, done

fontforgeexe/charview.c
```diff
@@ -408,7 +408,7 @@ int CVCountSelectedPoints(CharView *cv) {
  * shapes not to cross axes multiple times while scaling.
  */
 static double rpt(CharView *cv, double pt) {
-    return cv->snapoutlines ? rint(pt) : pt;
+    return cv && cv->snapoutlines ? rint(pt) : pt;
```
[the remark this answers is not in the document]
iorsh: Looks like "snap outline" is off by default, so we shouldn't round unless there is a good reason for it. Your code is ok here.

fontforgeexe/fontview.c
```diff
@@ -33,6 +33,7 @@
 #include "autowidth2.h"
 #include "bitmapchar.h"
 #include "bvedit.h"
+#include "charview_private.h"
```
iorsh, line 36: This header is exclusive for `cvpalettes.c`, please put `CVDrawSplinePointList()` into `views.h`

vasilky3: done

iorsh (approved):
Looks good to me

iorsh:
@skef, please take a look.

fontforgeexe/charview.c
```diff
@@ -1495,103 +1495,122 @@ void CVDrawSplineSet(CharView *cv, GWindow pixmap, SplinePointList *set,
...
-
-    if( strokeFillMode == sfm_fill ) {
-    	GDrawFillRuleSetWinding(pixmap);
-    }
-
-    for ( spl = set; spl!=NULL; spl = spl->next ) {
+    int height = 0;
+    for (spl = set; spl != NULL; spl = spl->next)
+    {
+        Color fc = spl->is_clip_path ? clippathcol : fg;
```
skef, line 1505: I haven't tried this, but from the code it seems like you turn on the metrics option and you get the outline in the foreground color? Is that always going to make sense? What if you have a mix of closed and open contours (maybe you're in the middle of developing a stroked font) and want to see the contrast?

skef: I guess from the attached image it's already a different color ...

vasilky3: The image is corresponds to this code.
fg is passed as function argument.

In fontview.c it's called with green color:
````CVDrawSplinePointList(NULL, pixmap, sc->layers[1].splines, fvmetadvancetocol,
		        sfm_stroke, x0, y0, (double) (box.width - 1) / sc->vwidth);
		}```

Comment (skef) [reply fontforgeexe/charview.c:1505]: Hmm, so we're reusing a metrics color: "Color used to draw the (horizontal or) vertical line from the origin to the advance when that metric is selected in View". That's not ideal. Could we maybe add a color to fontview_re?

Comment (skef) [reply fontforgeexe/charview.c:1505]: (I admit this is hairy, as it will also involve changes to the appearance editor ...)

Comment (vasilky3) [reply fontforgeexe/charview.c:1505]: I was trying to create a new color in code. And failed it.
Moreover I have no idea what is `fontview_re`?
If you ask I can do this. Please specify the DOD. So I can figure out that to look for.

Comment (iorsh) [reply fontforgeexe/charview.c:1505]: @vasilky3, the `static struct resed fontview_re[]` is located at fontview.c:7283 in your branch. It's a resource for all font-view related color definitions and fonts.

You can use `fvmetorigincol` as an example:

1. Add new static variable like `static Color fvmetorigincol = 0xc00000;` with the default color of your liking near line 136
2. Add a line for it in `fontview_re` as appropriate 
3. Use it instead of the currently used `fvmetadvancetocol`

You should also be able to change the color in UI from "File --> Appearance Editor --> Font View":
image

Comment (iorsh) [reply fontforgeexe/charview.c:1505]: As far as I can see, there is no need to mess with Appearance Editor specifically, the new color definition should show up there automagically.

Comment (vasilky3) [reply fontforgeexe/charview.c:1505]: @iorsh  thx for info. Helped me alot.
> there is no need to mess with Appearance Editor specifically
it's true.

Comment (skef) [reply fontforgeexe/charview.c:1505]: One more quick thing: Make an entry for this using the case-sensitive string you added in fontview_re ("MetricsGlyphContourColor") around here: https://github.com/fontforge/fontforge/blob/2e5bbdc933e125332ee0cdc6d74dabf1f90bb0f4/fontforgeexe/pixmaps/tango/resources.in#L22 . ("Tango" is the current resource directory, I don't think it's necessary to mess with 2012 as well but you're free to.) For the PR you can use 0x3030c0 like you did as the internal program default as long as that makes sense with the other colors, as in:

```
fontforge.FontView.MetricsGlyphContourColor: #3030C0
```

but I would also advise changing it and restarting the program to ensure the value is getting picked up properly. 

@vasilky3 from your original screenshots it looks like you maybe didn't the resource directory configured for your system, but the more recent screenshot of the appearance editor looks more like what I would expect. Did you change something/get that working?

vasilky3: Done. Here is the result. 
image

Previously i was trying to build fontforge from this repo by my own. 

This time I did read wiki. 
Now I'm using FontForgeBuild repo for building. 

Got some problems with understanding how to force it to build my branch. But got it finally.

skef (approved):
I think this looks OK to me. @iorsh should we merge it?

iorsh:
> I think this looks OK to me. @iorsh should we merge it?

Yes, looks good to me too

PR #5506, opened 2024-12-05
Modernize fixed pitch flag computation

### LINKED ISSUE #5489: Ufo3 export of monospace fonts does not set postisfixedpitch to true
When a sfd file is exported to ufo, if the font is selected as monospace, fontinfo.plist file should contain following key
```
<key>postscriptIsFixedPitch</key>
                <true/>
```
However the generated ufo do not contain such key.

iorsh:
This PR allows monospace fonts with modern features
* Zero-width glyphs are allowed, such as control characters or diacritical marks
* Dual-width glyphs are alowed, such as emojis or kanij

Fixes #5489

### DIFF
--- a/fontforge/splinesave.c
+++ b/fontforge/splinesave.c
@@ -1796,48 +1796,86 @@ static void SplineFont2FullSubrs1(int flags,GlyphInfo *gi) {
 #endif	/* FONTFORGE_CONFIG_PS_REFS_GET_SUBRS */
 }
 
-int SFOneWidth(SplineFont *sf) {
-    int width, i;
+typedef enum font_pitch { pitch_unknown, pitch_variable, pitch_fixed, pitch_dual } FontPitch;
 
-    width = -2;
-    for ( i=0; i<sf->glyphcnt; ++i ) if ( SCWorthOutputting(sf->glyphs[i]) &&
-	    (strcmp(sf->glyphs[i]->name,".notdef")!=0 || sf->glyphs[i]->layers[ly_fore].splines!=NULL)) {
+static FontPitch SFComputePitch(SplineFont *sf, int *p_width) {
+    FontPitch pitch = pitch_unknown;
+    int width = -1;
+
+    for ( int i=0; i<sf->glyphcnt; ++i) {
 	/* Only trust the width of notdef if it's got some content */
 	/* (at least as far as fixed pitch determination goes) */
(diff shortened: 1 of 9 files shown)


iorsh:
CI is a gift that keeps on giving... Libspiro is broken on Windows.

@JoesCat, do you know just in case what could have happened?

JoesCat:
On December 6, 2024 12:06:24 AM Maxim Iorsh wrote:
> Looks like the CI issue is not about Spiro, but rather broken `pacman` -
> see https://github.com/jtanx/fontforgebuilds/pull/22

repology shows several distros haven't upgraded to libspiro 2024 so there 
will likely be more faults reported due to libspiro 2022 (which is faulty).

JoesCat:
mingw may also be gradually sliding away from 32bit. I noted reading a request to upgrade something to help with gimp3 which wasn't built for 32bit, but was for 64bit.

fontforge/splinesave.c
```diff
@@ -2074,7 +2112,7 @@ struct pschars *SplineFont2ChrsSubrs(SplineFont *sf, int iscjk,
 	dummynotdef.parent = sf;
 	dummynotdef.layer_cnt = sf->layer_cnt;
 	dummynotdef.layers = calloc(sf->layer_cnt,sizeof(Layer));
-	dummynotdef.width = SFOneWidth(sf);
+	dummynotdef.width = CIDOneWidth(sf);
```
skef, line 2115: If we're using this function in these updated cases, should the final function name be SFOneWidth again?

iorsh: Agree, renamed.

Finii:
> This PR allows monospace fonts with modern features
> 
>     * Zero-width glyphs are allowed, such as control characters or diacritical marks
> 
>     * Dual-width glyphs are alowed, such as emojis or kanij

Well, the different-from-monospaced-width widths are not written out to a `ttf` file; `otf` works.
`tottf` does not allow different width for individual glyphs if `isFixedPitch` is active.

`tottf` exports only the widths of the first 4 glyphs, beginning with the 3 magic glyphs.

So even if the 4th glyph has 'normal' width, there will be no zero and no dual width glyphs in `ttf`s.
The PR above handles the more severe case with the 4th glyph has zero width :grimacing: and of course also the general case.

_Edit: Add last paragraph_

Issue #5508, opened 2024-12-08
Generating TTC segfaults


Yukinoroh:
I am using .sfd files from February to make a ttc file.

I open them all at once, and select Generate TTC. After clicking Generate, Fontforge hangs and segfaults.

I could generate those ttc files in February, and I am using the same version of FontForge as I was using in February (FontForge-2023-01-01-a1dad3e-x86_64.AppImage).

I have tried with LinuxMint's old 2020 version as well, same issue.
I have tried compiling from source in this repository, same issue.

dmesg output:

[11034.905260] fontforge[35540]: segfault at 8 ip 00007fb07657b010 sp 00007ffc4e00b350 error 6 in libfontforge.so.4[7fb07630b000+2a3000]
[11034.905269] Code: 89 38 4a 8b 14 e9 48 89 50 08 4a 89 04 e9 49 63 47 18 e9 ad f4 ff ff 0f 1f 00 85 db 0f 85 01 fc ff ff 48 89 ef e8 50 73 d9 ff <c7> 04 25 08 00 00 00 00 00 00 00 0f 0b 31 c0 e9 35 fb ff ff 41 0f

Looks like some new version of an external library installed between February and now is causing libfontforge to be unhappy. I really don't see why this can happen otherwise..

Yukinoroh:
Adding two blank sfd files with different font names, impossible to generate TTC in official release or binary compiled from repository.

Yukinoroh:
Adding two blank sfd files with different font names, impossible to generate TTC in official release or binary compiled from repository.
[test.zip](https://github.com/user-attachments/files/18051691/test.zip)

Yukinoroh:
Similar issue as #4417

Yukinoroh:
Currently debugging this. I need it for my project.

Yukinoroh:
I found the problem. Patch coming!

Yukinoroh:
Fixed in pull request #5509

Issue #4417, opened 2020-08-03
Crash when generating OTC


WolfGangsta:
I'm trying to use FontForge's font collection generation capability to make an OpenType Collection out of a set of OTF files, as mentioned in #3543, but FontForge always fails, leaving only an empty (0 bytes) file in the target directory with the name I wanted to give it (emmentaler.otc).

When I try to run the Python generateTtc function on the files with the "cff" and "merge" ttcflags set, execution stops almost immediately. This was tested on Windows 10 (with FF 20190801) via FontForge interactive console and on LilyDev 2, a fork of Debian (with FF 20170812), via regular bash. On LilyDev, the crash is reported as a segfault, while no error message appears on Windows.

When I use the "Generate TTC..." GUI dialog (again, with the "Merge tables across fonts" and "As CFF fonts" options checked), FF exits immediately, having hit, I presume, the same segfault. I tested the GUI on both Windows and LilyDev, same as above.

The issue is present even when only attempting to merge two of the eight fonts, so I'm attaching a couple of the fonts that fail the merge, along with a stripped-down version of the generation script that still demonstrates the crash: [emmentaler.zip](https://github.com/fontforge/fontforge/files/5018232/emmentaler.zip)

(N.B. The files are a WIP version of Emmentaler, the standard music font for GNU LilyPond, so there's no issue with them going public. Also, since it's a music font, most of the glyphs reside in the PUA after 0xE000.)

skef:
I appreciate the effort put into isolating this. Unfortunately the script in the provided archive refers to `emmentaler-11.otf` and `emmentaler-13.otf` but only contains `emmentaler.otf`. If possible could you update the archive with the needed files?

jtanx:
@WolfGangsta do you still have the files required to reproduce this issue?

skef:
Closing for lack of information.

ceztko:
The attached [FoxitSerifCFFSet.zip](https://github.com/user-attachments/files/17788065/FoxitSerifCFFSet.zip), taken from [pdf.js](https://github.com/mozilla/pdf.js/tree/master/external/standard_fonts) (but was first found in pdfium) crashes FontForge 20230101 when trying to create a collection "As CFF fonts". Please re-open.

Yukinoroh:
I just debugged this. It took two days.
Incomplete "don't generate FFTM" implementation was the culprit.
Fixed in pull request #5509.

PR #5509, opened 2024-12-10
Segfault fix and complete implementation of "Don't generate FFTM tables"

Yukinoroh:
The "Don't generate FFTM tables" option was not only not working, its partial implementation was causing a segfault when user aksed not to include FFTM and generated a TTC file.
See https://github.com/fontforge/fontforge/issues/5508 and https://github.com/fontforge/fontforge/issues/4417

The issue was:

1) WriteTTC (tottf.c) called ttc_prep (tottf.c), which called ttf_fftm_dump (ttfspecial.c), and there the check "if ( at->gi.flags & ttf_flag_noFFTMtable )" succeeded, function returned false and no FFTM table was generated.
2) The line "tab->offset = ftell(ttc);" after "tab = findtabindir(&all[cnt].tabdir,CHR('F','F','T','M'));" in ttc_dump (tottf.c) caused a segfault, due to the pointer pointing at a place outside of the file.

There was, surprisingly, no FFTM check at all at all places where FFTM is mentionned in tottf.c, is to say in:
	* ttc_perfonttables, called by ttc_dump (which is called by WriteTTC)
	* buildtablestructures, called by:
	**	initTables (which is called by _WriteTTFFont)
	**	ttc_dump (two places) (which is called by WriteTTC)
	* ttc_dump, called by WriteTTC
	* ttc_prep, called by WriteTTC
	* initTables, called by _WriteTTFFont

I fixed this issue by first propagating the flags already present in _WriteTTFFont and WriteTTC, down to initTables, ttc_dump, then to buildtablestructures, and performing the necessary check in ttc_perfonttables, buildtablestructures, ttc_dump, ttc_prep and initTables. This includes before calling ttf_fftm_dump, thus I moved the FFTM check outside of ttf_fftm_dump (we shouldn't get in there if no FFTM is to be generated) and reformed the function.

Since all of these FFTM-related stuff were entangled in _WriteTTFFont and WriteTTC, this patch should fix the issue I reported and also those related to FFTM being generated against users' will.

I tested generating TrueType, TrueType TTC, OpenType CID, all of them with FFTM checked and unchecked, Mac checked and unchecked, and everything seems to work.


Yukinoroh:
Looks like some Windows checks are failing due to remote configuration. What should I do?

iorsh:
> Looks like some Windows checks are failing due to remote configuration. What should I do?

It should be ok now, please push empty commit to trigger CI run.

Yukinoroh:
> > Looks like some Windows checks are failing due to remote configuration. What should I do?
> 
> It should be ok now, please push empty commit to trigger CI run.

How do I do that?

iorsh:
```
git commit --allow-empty -m "trigger CI"
git push
```

Yukinoroh:
> ```
> git commit --allow-empty -m "trigger CI"
> git push
> ```

Hmm, I've been struggling with git. Is there anyway to do this with github?

Yukinoroh:
Conversely, I could have accessed fftm option through at->gi.flags everywhere (exporting options are saved in glyphinfo?!?), but this seems less reliable than what's passed from the UI down to _WriteTTFFont and WriteTTC. That is, in both these functions there is an alltabs variable, but it is generated there, not passed down. And it is again re-generated in ttc_prep, I suppose from the output of all open fonts. At each of these regenerations of alltabs we'd have to make sure that the flags are copied properly.

Yukinoroh:
P.S.: I did some additional testing on the repository's original code. When generating a TTC file, the value of at->gi.flags does not change no matter if you check FFTM or not. It does change if you generate a single font, though. So maybe FontForge gets confused as to which opened font's glyphinfo flags has to use (maybe picks up the last one?). I think it's better to use the value that is passed down from the UI; in my code it works perfectly.

skef (commented):
I haven't checked to see if any fftm references were missed, but what is here looks good. @iorsh thoughts?

iorsh:
Looks like everything is covered here. Looks good to me.

PR #5515, opened 2024-12-25
python: use PyConfig_InitIsolatedConfig to initialize the embedded Python

philiptaron:
## Motivation for change

Reduce surprising and security-sensitive loading of Python modules based on the current working directory where the `fontforge` program is run. After this change, the Python scripts which are explicitly loaded based on `fontforge`'s command line flags, plus hard-coded ones like `pkg_resources`, are the only ones used.

This is a breaking change!

From the [docs](https://docs.python.org/3/c-api/init_config.html#init-isolated-conf):

> `PyConfig_InitIsolatedConfig()` creates a configuration to isolate Python
> from the system. This is ideal to embed Python into an application.
>
> This configuration ignores global configuration variables, environment
> variables, command line arguments (`PyConfig.argv` is not parsed) and
> user site directory. The C standard streams (ex: stdout) and the
> `LC_CTYPE` locale are left unchanged. Signal handlers are not installed.
>
> Configuration files are still used with this configuration to determine
> paths that are unspecified. Ensure `PyConfig.home` is specified to avoid
> computing the default path configuration.

This will change the default behavior of Python in FontForge from loading whatever is in the current working directory to loading what's installed as part of the system Python or the venv Python, which ever gets loaded as the shared library.

### **Breaking changes**

- Python modules in the current working directory will no longer be loaded preferentially ([aka disabled `safe_path`](https://docs.python.org/3/c-api/init_config.html#c.PyConfig.safe_path)). The files processed by `LoadFilesInPythonInitDir` will still be loaded.
- Locale will no longer be [auto-configured](https://docs.python.org/3/c-api/init_config.html#c.PyPreConfig.configure_locale). 
- C locale will [no longer be possibly coerced](https://docs.python.org/3/c-api/init_config.html#c.PyPreConfig.coerce_c_locale).
- C input streams will no longer [be mutated by Python initialization](https://docs.python.org/3/c-api/init_config.html#c.PyConfig.configure_c_stdio).
- Environment variables will [no longer influence behavior](https://docs.python.org/3/c-api/init_config.html#c.PyPreConfig.use_environment).
- Python signal handlers will [no longer be installed.](https://docs.python.org/3/c-api/init_config.html#c.PyConfig.install_signal_handlers) 
- UTF-8 mode will no longer [be possibly set](https://docs.python.org/3/c-api/init_config.html#c.PyPreConfig.utf8_mode).
- Command line parameters understood by Python will [no longer be read](https://docs.python.org/3/c-api/init_config.html#c.PyConfig.parse_argv). 
- The user site directory will not [be added to the configuration](https://docs.python.org/3/c-api/init_config.html#c.PyConfig.user_site_directory).

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -20155,7 +20155,7 @@ void FontForge_InitializeEmbeddedPython(void) {
 
     PyConfig config;
     PyStatus status;
-    PyConfig_InitPythonConfig(&config);
+    PyConfig_InitIsolatedConfig(&config);
 
     status = PyConfig_SetBytesString(&config, &config.program_name,
                                      "fontforge");


philiptaron:
Some of the tests failed with [Python module errors](https://ci.appveyor.com/project/fontforge/fontforge/builds/51225934): 👀 

```
The following tests FAILED:
	 58 - test1001_py (Failed)
	 64 - test1001c_py (Failed)
	 68 - test1003_py (Failed)
	104 - test926_py (Failed)
	116 - test932_py (Failed)
	122 - test_fea_hyphens_py (Failed)
	124 - test_sfd_suppl_plane_py (Failed)
	126 - test_fea_context_sub_py (Failed)
```

skef:
It seems like this raises a lot of potential issues and would require some communication at a minimum. The main benefit would be a slightly decreased attack surface but FontForge is already incredibly insecure. 

Our guideline about untrusted input isn't directly related but I'm tempted to refer to it anyway: https://github.com/fontforge/fontforge/wiki/Community-guidelines#D1

philiptaron:
> Our guideline about untrusted input isn't directly related but I'm tempted to refer to it anyway: https://github.com/fontforge/fontforge/wiki/Community-guidelines#D1

Actually, that was an incredibly helpful reference. Plus the set of failing tests, it made me realize the impact here for making this the _default_ is... not worth the squeeze.

For nixpkgs's packaging of FontForge, I might pursue this approach, since it makes much more sense in that context. In NixOS/nixpkgs, there is an assumption about hermeticity that's not present in the broader Python ecosystem. There are also simple ways of constructing a Python "system" environment with a crafted set of packages just for FontForge. None of those hold true in the general case.

Given all that, I'm closing this PR. Thank you for your time and attention.

PR #5517, opened 2024-12-27
Make SmallCaps() create symbols

cspiel:
**Bug fix**

The script function `SmallCaps` does not create symbols, only letters.
However, the GUI as well as the Python interface allow to generate
smallcaps symbols.

The fix for this problem is trivial as it boils down to a flag that must be set.

The documentation of `SmallCaps` is extended to mention both the
`.sc` and `.taboldstyle` namespaces of letters and symbols, respectively.


skef (approved):
I think we officially discourage updates to the native scripting stuff in the FAQ but I don't have an issue with this kind of fix. Seems fine to me.

PR #5519, opened 2024-12-29
Fix function PyFFFont_addSmallCaps.

cspiel:
**Bug fix**

(1) The initialization of `struct genchange` was incomplete.
I transfered the values from native scripting function bSmallCaps
to fill the gaps.
  
(2) Components `stem_width_scale` and `stem_height_scale`
were not reliably initialized.


cspiel:
The documentation of Python method `addSmallCaps` does _not_ mention
the parameters `hscale` and `vscale`.  Is this done intentionally?

I'd suggest to document these parameters and point out that they
act as overrides of the calculated values for the given selection.

skef (approved):
This looks fine to me -- @iorsh any thoughts?

iorsh:
I'm fine with this. I have some vague suspicions, that the same initial values should also appear somewhere in UI, but this change by itself is sufficient.

iorsh (approved):

PR #5522, opened 2025-01-09
Support Harfbuzz in Metrics View

### LINKED ISSUE #3733: RTL kerning reversed in the metrics window
צילום מסך מ־2019-06-13 15-36-46

The kerning on the left (Scribus) is how it looks after exporting. But Fontforge (right) shows the spacing on the wrong side.

- [x] Screenshot
- [x] The FontForge version and the operating system you're using : 14 jan 2019
- [x] Steps to reproduce the behavior : create rtl kerning and open the metrics window.

### LINKED ISSUE #3837: FontForge fails to understand Aldhabi.ttf
Microsoft distributes an interesting font which they call &laquo;Al Dhabi&raquo;.

It's similar to Noto Nastaliq Urdu in how it works. The `curs` table is most important.

Let's use as an example the text **النوع المفتوح**.

Below image shows Noto Sans Arabic, Noto Nastaliq Urdu and Aldhabi rendered correctly.

2019-07-30-151319_872x1083_scrot

FontForge however cannot understand Aldhabi at all!

2019-07-30-151302_1876x911_scrot

The `curs` table does appear in `GPOS` though, and to my (perhaps ignorant) eyes even looks complete.

At the risk of angering Microsoft, (they distribute this font for free with their operating system,) I upload it also here to be sure if anyone looks into this that they are considering the same version of Aldhabi that I've considered. 
[aldhabi.zip](https://github.com/fontforge/fontforge/files/3445592/aldhabi.zip)

### LINKED ISSUE #4158: Metrics View should just use HarfBuzz
&ldquo;My font works in the Metrics View but then looks different in Chrome&rdquo; is such a common type of bug report. Sometimes the reverse is also complained about.

HarfBuzz is the new Uniscribe. It is the most common shaper out there. I am tired of always having to explain to people...&ldquo;no, the Metrics View is just an interpretation by one possible shaper...not one anyone actually uses, just fontforge...it is perfectly possible to make fonts in fontforge which fontforge's own shaper cannot interpret, and in fact some of my own fonts fit this bill.&rdquo;

This would close #3837 if implemented, and probably more issues.

What do you all think? Is it worth taking the time to do this?

### LINKED ISSUE #4511: Implementation of Universal Shaping Engine (USE)
I'm developing font of Indonesian scripts such as Sundanese and Javanese. They need to be rendered correctly using open type features and Universal Shaping Engine (USE). Unfortunately, I did not see the USE has been implemented into the Fontforge. Some open-source projects have implemented this feature like Firefox, LibreOffice, and Chromium.  

There is documentation of **[Special thought for special script](https://fontforge.org/docs/tutorial/scriptnotes.html)** for Fontforge but it does not cover the use of USE yet. If Fontorge can implement the USE, it will more useful for developing special scripts registered for the USE (45 more scripts). 

You might be aware of this before, but this documentation can be helpful:
- [Creating and supporting OpenType fonts for the Universal Shaping Engine - Microsoft Typography](https://docs.microsoft.com/en-us/typography/script-development/use)
- [Why do I need a shaping engine? - HarfBuzz](https://harfbuzz.github.io/why-do-i-need-a-shaping-engine.html) 
- [Making fonts for the Universal Shaping Engine - TYPO Labs](http://tiro.com/John/Universal_Shaping_Engine_TYPOLabs.pdf)

I'm using Fontforge version 20200314. The captured images are using the font from the Google Noto project [GitHub repository](https://github.com/googlefonts/noto-fonts). 

Noto Sans Sundanese in Fontforge
gambar
(not expected)

Noto Sans Sundanese in LibreOffice
gambar
(expected)

Noto Sans Javanese in Font Forge 
gambar
(not expected)

Noto Sans Javanese in Libre Office 
gambar
(expected)

Please check this on Mozilla Firefox, the results should appear like the expected above.  
ᮊ + ᮦ = ᮊᮦ
ꦗ + ꦺ = ꦗꦺ

They only simple position reordering samples of USE implementation. Hopefully, the next version of Fonforge will cover this feature. 

Best wishes, 
Ilham

- [x] Screenshot

### LINKED ISSUE #5343: Metrics window preview of arabic fonts doesn't join characters correctly
The 4 forms of Arabic characters; initial, medial, final, and isolated should be used according to character position in the word, and according to the joining property of the character and the joining property of the surrounding characters ['isol' feature](https://learn.microsoft.com/en-us/typography/opentype/otspec181/features_fj#isol)

However, when using the metrics window of fontforge, it seems that it is determining the form, based on character position only, without looking at characters' joining property [Unicode Arabic Shaping](https://www.unicode.org/Public/UCD/latest/ucd/ArabicShaping.txt)

Steps to reproduce:

1. Open arabic font with fontforge (I opened trado.ttf which I copied from C:\Windows\Fonts\Traditional Arabic)
2. Open Metrics window and type the arabic character ("alef" U+0627), then ("lam" U+0644)
3. You see that the second glyph "lam" is getting replaced with its final form (U+FEDE "Lam Final Form") instead of its isolated form (U+FEDD "Lam Isolated Form")

Screenshot1

Expected behaviour:
- Since the first character ("alef U+0627) has an R joining type (doesn't join with following characters), the next character should be in its isolated form (U+FEDD "Lam Isolated Form")
- If we add a third character which has a D joining type, such as ("kaf" U+0643), we can see in the metrics window that the second character "lam" is getting replaced with its medial form (U+FEE0 "Lam Medial Form"), instead of its initial form (U+FEDF "Lam Initial Form")

Screenshot2

OS: Windows 10 Pro 22H2
FontForge version: 20230101

### LINKED ISSUE #5513: Combining marks break kerning
I'm at the end of my wits. I have a mark feature that positions two marks above characters. I have kern feature with Ignore Combining Marks selected. When I have kern between two characters, a mark changes position too!

Expected behavior: the mark is positioned according to the anchor, and kern is applied between two base characters. The same font is tested with harfbuzz, and is displayed correctly:
image

Actual behavior: the kern is applied between the base character and its mark in the FontForge Metrics Window:
image

FontForge version: 20230101
OS: Windows 11

iorsh:
This PR introduces [HarfBuzz](https://harfbuzz.github.io/) shaping in Metrics view.

1. The current shaping functionality is preserved and can be invoked from View menu as "Built-in shaper". It's very complex, and I'm pretty sure I didn't manage to fully support everything it provides, so some people might find it better in some aspects.
2. The new shaper comes with some generic infrastructure which can be eventually expanded to support more shapers (e.g. Uniscribe under Windows)
3. The new code is written mostly in C++17. This version seems reasonably conservative to me.
4. Editing of kerning values, bearings, and width is supported.
5. The font is generated for HarfBuzz use every time the Metrics View is opened or shaper engine is switched (in the View menu). This could be slow for large fonts. 

**After:**
aldhabi-hb
**Before:**
aldhabi-old
**Menu:**
hb_menu

 Fixes #3733, fixes #3837, fixes #4158, fixes #4511, fixes #5343, fixes #5513


iorsh:
CI is broken.
- Linux: ubuntu-latest moved from 22 to 24
- Windows - pacman issues again.

I'll take a look.

Yukinoroh:
iorsh, for your information HarfBuzz is shutting off all OpenType features if Apple Advanced Typography (AAT) tables are present in a font, even for features unique to OpenType. I would really proceed with caution by incorporating HarfBuzz into FontForge.

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -0,0 +1,349 @@
...
+        glyphs_after_gpos[i] = glyph_out;
+
+        // Fill unscaled metrics in font units
+        hb_position_t h_advance =
+            hb_font_get_glyph_h_advance(hb_ttf_font, glyph_info.codepoint);
+
+        hb_glyph_extents_t extents;
+        hb_bool_t res = hb_font_get_glyph_extents(
+            hb_ttf_font, glyph_info.codepoint, &extents);
+        assert(res);
```
khaledhosny, line 111: Calculating glyph extents can be slow (especially for CFF fonts), and I don’t see extents used anywhere so it seems wasteful.

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -0,0 +1,349 @@
...
+            hb_font_get_glyph_name(hb_ttf_font, glyph_info.codepoint,
+                                   glyph_name, sizeof(glyph_name) - 1);
+        SplineChar* glyph_out =
+            context_->get_glyph_by_name(context_->sf, -1, glyph_name);
+
+        glyphs_after_gpos[i] = glyph_out;
+
+        // Fill unscaled metrics in font units
+        hb_position_t h_advance =
+            hb_font_get_glyph_h_advance(hb_ttf_font, glyph_info.codepoint);
```
khaledhosny, line 106: You can get the original width from splinechar, shaping does not change it.

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -0,0 +1,349 @@
...
+    hb_buffer_t* hb_buffer = hb_buffer_create();
+    hb_buffer_add_utf8(hb_buffer, utf8_str, -1, 0, -1);
+
+    // Set script and language
+    hb_script_t hb_script = hb_script_from_iso15924_tag((uint32_t)script);
+    hb_buffer_set_script(hb_buffer, hb_script);
+    hb_language_t hb_lang = hb_language_from_string((const char*)lang, -1);
+    hb_buffer_set_language(hb_buffer, hb_lang);
+
+    bool rtl = !u_vec.empty() && isrighttoleft(u_vec[0]);
```
khaledhosny, line 268: Checking first character is not enough, for example if text starts with punctuation. HarfBuzz has [`hb_buffer_guess_segment_properties`](https://harfbuzz.github.io/harfbuzz-hb-buffer.html#hb-buffer-guess-segment-properties) which works slightly better here (it checks the first character with strong direction).

Ideally, there should be a UI to set the direction, since it is possible the text is entirely composed of unencoded glyphs which will be given PUA encoding and all PUA codepoints has strong LTR direction.

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+namespace ff::shapers {
+
+HarfBuzzShaper::HarfBuzzShaper(std::shared_ptr<ShaperContext> context)
+    : context_(context) {
+    FILE* ttf_file = GFileTmpfile();
+
+    _WriteTTFFont(
+        ttf_file, context_->sf, ff_ttf, NULL, bf_ttf,
+        ttf_flag_otmode | ttf_flag_oldkernmappedonly | ttf_flag_fake_map,
```
khaledhosny, line 43: I don’t think `ttf_flag_oldkernmappedonly` does anything unless `ttf_flag_oldkern` is also set. But you don’t want old kern table anyway.

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -0,0 +1,349 @@
...
+        context_->get_enc_map(context_->sf), ly_fore);
```
khaledhosny, line 44: You can probably speed things up by skipping generating tables that HarfBuzz does not use, e.g. glyph outlines (glyf and CFF table) which is usually the larges font tables and the slowest to build.

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -0,0 +1,349 @@
...
+    hb_buffer_set_language(hb_buffer, hb_lang);
+
+    bool rtl = !u_vec.empty() && isrighttoleft(u_vec[0]);
+    hb_direction_t hb_dir = vertical ? HB_DIRECTION_TTB
+                            : rtl    ? HB_DIRECTION_RTL
+                                     : HB_DIRECTION_LTR;
+    hb_buffer_set_direction(hb_buffer, hb_dir);
+
+    // Shape the text
+    hb_shape(hb_ttf_font, hb_buffer, NULL, 0);
```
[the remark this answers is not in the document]
khaledhosny: In general, default features are always applied and selecting/unselecting features on the features list has no effect.

khaledhosny:
Selecting builtin shaper always crashes for me.

iorsh:
Windows CI fixed with https://github.com/jtanx/fontforgebuilds/pull/23

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -0,0 +1,349 @@
...
+    SplineChar** glyphs_after_gpos = extract_shaped_data(hb_buffer);
+    int glyph_count = metrics.size() - 1;
+
+    std::vector<int> width_deltas =
+        compute_width_deltas(hb_buffer, glyphs_after_gpos);
+
+    if (rtl) {
+        // HarfBuzz reverses the order of an RTL output buffer
+        std::reverse(glyphs_after_gpos, glyphs_after_gpos + glyph_count);
+        std::reverse(width_deltas.begin(), width_deltas.end());
```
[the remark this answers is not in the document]
iorsh: This doesn't work for me. The RTL buffer before the reversing contains glyphs tail-to-head, and `hb_buffer_get_glyph_positions()` returns values with geometrically correct left-to-right advances and offsets.

In the case of mixed base glyphs (positive advance) and marks (positive offset, since the mark comes *before* the glyph), the function `hb_buffer_reverse()` correctly reorders the glyphs, but `hb_buffer_get_glyph_positions()` would just return the same values in reverse order, and this breaks the geometric relation.

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -0,0 +1,349 @@
...
+        char glyph_name[64];
+        hb_glyph_info_t& glyph_info = glyph_info_arr[i];
+        hb_glyph_position_t& glyph_pos = glyph_pos_arr[i];
+
+        // Warning: after the shaping glyph_info->codepoint is not a Unicode
+        // point, but rather an internal glyph index. We can't use it in our
+        // functions.
+        hb_bool_t found =
+            hb_font_get_glyph_name(hb_ttf_font, glyph_info.codepoint,
+                                   glyph_name, sizeof(glyph_name) - 1);
```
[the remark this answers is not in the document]
iorsh: It looks to me that FF does enforce unique names, at least this is what I got when trying to rename a glyph to existing name: 
image
Do you have in mind some specific scenario where this doesn't hold?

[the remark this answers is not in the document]
khaledhosny: Here is an SFD file with duplicate glyph names (two glyphs both named test), FontForge will open it without any complaint and it will also generate a font with the duplicate glyph name: [test.sfd.zip](https://github.com/user-attachments/files/18550394/test.sfd.zip)

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -69,6 +69,17 @@ HarfBuzzShaper::~HarfBuzzShaper() {
...
 
+std::vector<hb_feature_t> HarfBuzzShaper::hb_features(
+    const std::vector<Tag>& feature_list) const {
+    std::vector<hb_feature_t> hb_feature_vec;
+    for (const Tag& feature_tag : feature_list) {
+        hb_feature_t hb_feat{feature_tag, 1, HB_FEATURE_GLOBAL_START,
+                             HB_FEATURE_GLOBAL_END};
+        hb_feature_vec.push_back(hb_feat);
+    }
+    return hb_feature_vec;
```
khaledhosny, line 80: Some work is needed to handle features that are enabled by default. Features like `init`, `medi`, etc. are enabled by default and HarfBuzz applies them selectively based on text analysis, but when enabled manually they will be applied unconditionally which will break their intended use.

What should happen is that if a default feature is selected in the UI, it should not be in the features list, but if it is unselected it should be in the features list with value set to 0 (disable).

HarfBuzz does not have an API that tells which features are enabled by default, so you will have to check the HarfBuzz sources. The list of default features depend on the script, I don’t know how FontForge decides which features to enable by default, but I think you will have to sync that with HarfBuzz so that features pre-selected in the UI match what HarfBuzz does.

Some applications use a tri-state checkbox for features, the default is neither selected or unselected what omits the feature from the features list, and selected set the feature to 1 and unselcted sets it to 0.

skef: The HarfBuzz sources are still the best reference, but here's a quick guide based on recent research on which features are considered enabled by default "in general": https://w3c.github.io/IFT/Overview.html#feature-tag-list

iorsh:
> Selecting builtin shaper always crashes for me.

@khaledhosny, I tested it with a few fonts, but couldn't reproduce. Could you please attach a crash stack, or the font file if possible?

fontforgeexe/shapers/harfbuzz.cpp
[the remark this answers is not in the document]
khaledhosny: If the current code works for you, then fine. `hb_shape()` returns the glyphs in visual order (that is, for RTL text the buffer is reversed), so that code that is drawing the text wouldn‘t need to worry about the buffer direction. `hb_buffer_reverse()` undoes that. Either case the individual glyph advances will be the same since there are no RTL glyph advances in OpenType.

I don’t know what FontForge expects here, but it is probably some FontForge-specific thing, not HarfBuzz that is being unexpected.

khaledhosny:
> > Selecting builtin shaper always crashes for me.
> 
> @khaledhosny, I tested it with a few fonts, but couldn't reproduce. Could you please attach a crash stack, or the font file if possible?

It happens with any font, when `ff::shapers::IShaper::~IShaper()` is called, this warning seems to be relevant:

```
/fontforge/fontforgeexe/shapers/shaper_shim.cpp:75:5: warning: delete called on 'ff::shapers::IShaper' that is abstract but has non-virtual destructor [-Wdelete-abstract-non-virtual-dtor]
   75 |     delete shaper;
```

This patch makes the warning and the crash go away:
```diff
diff --git a/fontforgeexe/shapers/i_shaper.hpp b/fontforgeexe/shapers/i_shaper.hpp
index 1884b05c1..f44e6f3cb 100644
--- a/fontforgeexe/shapers/i_shaper.hpp
+++ b/fontforgeexe/shapers/i_shaper.hpp
@@ -36,6 +36,8 @@ namespace ff::shapers {

 class IShaper {
  public:
+    virtual ~IShaper() = default;
+
     // The internal shaper name is non-localizable and serves to identify the
     // shaper in the system.
     virtual const char* name() const = 0;
```

fontforgeexe/shapers/harfbuzz.cpp
iorsh: I added support for default feature list (copied from https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-subset-input.cc#L82), but couldn't see any differentiation by script. @khaledhosny, is there any additional place for default features per script?

khaledhosny: This is the code of the subsetter, not the shaper. For shaping, start here: https://github.com/harfbuzz/harfbuzz/blob/6d8035a99c279e32183ad063f0de201ef1b2f05c/src/hb-ot-shape.cc#L285, where you see the features that applies to all scripts (some features depend on the direction here).

Per-script features are defined in the respective shapers, here, for example, Arabic shaper:
https://github.com/harfbuzz/harfbuzz/blob/6d8035a99c279e32183ad063f0de201ef1b2f05c/src/hb-ot-shaper-arabic.cc#L184

Which shaper is used for which script(s) is handled here:
https://github.com/harfbuzz/harfbuzz/blob/6d8035a99c279e32183ad063f0de201ef1b2f05c/src/hb-ot-shaper.hh#L176

So features depend on the direction and script.

khaledhosny: Some shaper also *disable* some of the common features, e.g. Hangul:
https://github.com/harfbuzz/harfbuzz/blob/6d8035a99c279e32183ad063f0de201ef1b2f05c/src/hb-ot-shaper-hangul.cc#L67

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -0,0 +1,505 @@
...
+    } else if (script == "Beng" || script == "Deva" || script == "Gujr" ||
+               script == "Guru" || script == "Knda" || script == "Mlym" ||
+               script == "Orya" || script == "Taml" || script == "Telu") {
+        return features_indic;
+    } else if (script == "Khmr") {
+        return features_khmer;
+    } else if (script == "Mymr") {
+        return features_khmer;
+    } else {
+        return features_use;
```
iorsh, line 479: @khaledhosny, I tried to duplicate HarfBuzz logic here, but I'm confused about `_hb_ot_shaper_default` vs `_hb_ot_shaper_use`. I chose the latter by solely considering _the more the merrier_, but I'm not quite sure this is the way to go. Could you please take a look?

khaledhosny: USE (Universal Shaping Engine) is used for most new scripts in Unicode. Default is the default shaper that is used for scripts that do not have a dedicated shaper (e.g. Latin). HarfBuzz might chose default shaper for scripts that should go through USE shaper if the font does not declare a script (i.e. all features are under `dflt` script).

khaledhosny:
I’m working on a HarfBuzz API that returns the features enabled on a shape plan, see https://github.com/harfbuzz/harfbuzz/pull/5047. Can you try this API and see if it works for your case? (you can see examples of how to use it in the unit tests).

iorsh:
> I’m working on a HarfBuzz API that returns the features enabled on a shape plan, see [harfbuzz/harfbuzz#5047](https://github.com/harfbuzz/harfbuzz/pull/5047). Can you try this API and see if it works for your case? (you can see examples of how to use it in the unit tests).

Sure, I'll check that. I noticed that in unit tests you are supplying a default `" "` buffer, and I'm planning to do the same. As far as I understand, supplying actual buffer could lead to changes in default feature list, and Metric View UI is not built to accommodate them. With the default buffer, the displayed feature list would only depend on direction and script.

khaledhosny:
> > I’m working on a HarfBuzz API that returns the features enabled on a shape plan, see [harfbuzz/harfbuzz#5047](https://github.com/harfbuzz/harfbuzz/pull/5047). Can you try this API and see if it works for your case? (you can see examples of how to use it in the unit tests).
> 
> Sure, I'll check that. I noticed that in unit tests you are supplying a default `" "` buffer, and I'm planning to do the same. As far as I understand, supplying actual buffer could lead to changes in default feature list, and Metric View UI is not built to accommodate them. With the default buffer, the displayed feature list would only depend on direction and script.

To be honest, it didn’t occur to me that the text content can change the enabled features. I used an empty string in the unit tests because they are not concerned about the actual shaping.

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -502,4 +505,48 @@ std::set<Tag> HarfBuzzShaper::default_features(Tag script,
...
+
+    hb_shape_plan_t* shape_plan =
+        hb_shape_plan_create(hb_ttf_face, &props, NULL, 0, NULL);
+    hb_bool_t ret =
+        hb_shape_plan_execute(shape_plan, hb_ttf_font, buffer, NULL, 0);
+
+    // dummy call to check the total projected number of tags
+    unsigned int zero = 0;
+    unsigned int count =
+        hb_ot_shape_plan_get_feature_tags(shape_plan, 0, &zero, nullptr);
```
khaledhosny, line 531: You can pass nullptr instead of the dummy &zero.

iorsh:
@khaledhosny, I think I have addressed all your comments. Could you please review again?

khaledhosny:
Looks good, thanks! 

A few observations:
* Opening the metrics window still takes a long time for some fonts (e.g. Amiri):
* This happens every time the metrics window is opened even if the font did not change at all. I guess the temporary font is not cached, but may be it should.
* Guessing direction does not still work for unencoded glyphs. If the glyph string is all unencoded glyphs, the direction remains LTR even if I select Arabic script in the UI

fontforgeexe/shapers/harfbuzz.cpp
skef: HarfBuzz kinda-sorta has an API to tell which features are enabled by default, if I'm remembering right. If you initialize subset input object (`hb_subset_create_or_fail`) and then call `hb_subset_input_set(input, HB_SUBSET_SETS_LAYOUT_FEATURE_TAG);` on it, you'll get the `hb_set_t` of features as initialized by HarfBuzz. This is the features HarfBuzz retains by default, which is something like the features it expects it may need by default. I don't necessarily recommend using this API, however, not least because HarfBuzz can be installed without subsetting support.

khaledhosny:
> HarfBuzz does lots of caching internally, though, so you might want to destroy the face and re-create it when one of the tables changes, but you can still cache the unchanged blobs independent of the HarfBuzz face, so creating a new face will be a light operation.

Actually, there is `hb_font_changed()` to invalidate the internal caches.

skef:
Ah, right, I forgot `hb_face_create_for_tables`.

Unfortunately I'm not sure the on-demand aspect of that API, driven from within HarfBuzz, will help much. I suppose it could help simplify some of the changes: You keep the overall refresh operations the same, and arrange for `hb_font_changed` to be called whenever the window would be refreshed. Then if you've hooked up the callback for hb_face_create_for_tables to the right generators the problem boils down to determining what is stale, which means you can keep the staleness indicator close to the structure rather than having to signal it back to some more central place.

fontforge/splinechar.h
```diff
@@ -0,0 +1,654 @@
+/* Copyright (C) 2000-2012 by George Williams */
```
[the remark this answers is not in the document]
iorsh: This is a sort of cleanup, which I might have skipped, but `splinefont.h` has structures with `class` and `new` fields, which are incompatible with C++. I could rename them, but this would have just polluted the PR with a lot of unrelated changes.

fontforgeexe/shapers/harfbuzz.cpp
iorsh: Yes, I initially used the feature set from subsetting, but following @khaledhosny's advice, changed it to an explicit collection of features per script. @khaledhosny also provided a new API (https://github.com/harfbuzz/harfbuzz/pull/5047), which kicks in automatically once FontForge is built with supporting HarfBuzz version (10.3+).

fontforgeexe/metricsview.c
```diff
@@ -3948,6 +3950,7 @@ static ShaperContext* MVMakeShaperContext(MetricsView *mv) {
     context->get_enc_map = SFGetMap;
     context->get_char_width = MVCharWidth;
     context->get_kern_offset = MVGetKernOffset;
+    context->script_is_rtl = ScriptIsRightToLeft;
```
khaledhosny, line 3953: HarfBuzz has `hb_script_get_horizontal_direction()` which might be more up to date (`hb_buffer_guess_segment_properties()` will call it if the direction is not set or set to `HB_DIRECTION_INVALID`). But I still think a separate UI control to set direction is needed since a few scripts can be written horizontally in both directions.

iorsh: ScriptIsRightToLeft() is passed for use in the legacy shaper, HarfBuzz won't use it. Regarding the UI, I have some enhancements in mind, but for now I prefer to keep UI changes at the necessary minimum. GDraw poses quite a few challenges in its workings.

fontforge/splinechar.h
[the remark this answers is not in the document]
skef: Oh, right, that makes particular sense.

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -132,7 +133,10 @@ SplineChar** HarfBuzzShaper::extract_shaped_data(hb_buffer_t* hb_buffer) {
         // Warning: after the shaping glyph_info->codepoint is not a Unicode
         // point, but rather an internal glyph index. We can't use it in our
         // functions.
-        SplineChar* glyph_out = ttf_map_[glyph_info.codepoint];
+        auto ttf_map_it = ttf_map_.find(glyph_info.codepoint);
+        SplineChar* glyph_out = (ttf_map_it != ttf_map_.end())
+                                    ? ttf_map_it->second
+                                    : get_notdef_glyph();
```
khaledhosny, line 139: There is `hb_buffer_set_not_found_glyph()` that allow setting any arbitrary glyph index to be used instead of 0, you can, say, set it to a special value that is guaranteed not to be a valid glyph index like `INT32_MAX` and intercept it here. In case it makes things any simpler/faster.

iorsh: Yes, I tried it, but Ubuntu 22 has HarfBuzz 2.7.4, and `hb_buffer_set_not_found_glyph()` was introduced in 3.1.0. Besides, I need to check the glyph index anyway, it would be a crash otherwise.

iorsh:
The high-level HarfBuzz shaper structure is as follows:

- When opening the "Metrics View" window, the shaper is initialized with TTF file generated on-fly.
- While the window is active, it is responsive to the changes in glyph kerning using `HarfBuzzShaper::initial_kerning_`. The initial kerning is preserved, and once its value changes, the delta is applied internally over the metrics computed by HarfBuzz.
- In a similar manner changes to glyph width are also addressed with `HarfBuzzShaper::initial_width_`, without the need to regenerate the TTF.
- In case of more complex changes, such as in GSUB/GPOS tables, the user is expected to perform manual refresh, and I added "Refresh Shaper" menu item for that purpose. During the refresh the TTF file is fully regenerated.
- In total, the shaper is regenerated whenever
  - A new "Metrics View" is opened
  - The user hits "Refresh Shaper"
  - The uses changes the shaper ("Built-in" to "HarfBuzz" or vice versa)
 
The building of HarfBuzz shaper is heavy, and eventually we shall make an effort to improve its performance. Currently the TTF is written to a temporary file, and as far as I can see, it's not even guaranteed to reside in RAM. In Windows it's always on the disk (see `GFileTmpfile()`), and in Linux it probably depends on the mount method of `/tmp`.

For now I would prefer to keep the scope of this PR as is, but I can see several possibilities to improve the performance in future:
- Refactoring of `_WriteTTFFont()`, so that it actually uses a generic buffer which can be either kept in memory or dumped onto disk.
- To employ the granular approach with `hb_face_create_for_tables()`, it would be necessary to refactor the change tracking. Currently there is a bunch of `changed_???` flags scattered across multiple structures, and it would be good to harmonize them with some unified approach.
- Inherent FF deficiencies, e.g. the names hashtable `struct glyphnamehash` with 257 baskets, which is totally inadequate for fonts with thousands of glyphs.

skef:
I'll take another look at this review-wise, but I suggest that @iorsh should start thinking about documentation.

fontforge/views.h
```diff
@@ -399,6 +397,7 @@ typedef struct metricsview {
     int16_t cmax, clen;
     SplineChar **chars;		/* Character input stream */
     struct opentype_str *glyphs;/* after going through the various gsub/gpos transformations */
+    ShapeMetrics *metrics;      /* One for each glyph above */
```
iorsh: Comment on difference between ShapeMetrics (metrics_core) and metricchar (metrics_ui)

fontforgeexe/metricsview.c
```diff
@@ -996,21 +1011,24 @@ static void MVRemetric(MetricsView *mv) {
...
 
     // Regenerate glyphs for the selected characters according to features, script, and resolution.
     free(mv->glyphs); mv->glyphs = NULL;
     sf = mv->sf;
     if ( sf->cidmaster ) sf = sf->cidmaster;
-    mv->glyphs = ApplyTickedFeatures(sf,feats,script, lang, mv->pixelsize, mv->chars);
+    mv->glyphs = shaper_apply_features(mv->shaper, mv->chars, feats,
+        				script, lang, mv->pixelsize, mv->vertical);
+    if (mv->glyphs == NULL) {
+	mv->glyphs = calloc(1, sizeof(struct opentype_str));
```
iorsh, line 1037: Review this

fontforgeexe/metricsview.c
```diff
@@ -1021,6 +1039,10 @@ static void MVRemetric(MetricsView *mv) {
 	mv->perchar = realloc(mv->perchar,mv->max*sizeof(struct metricchar));
 	memset(mv->perchar+oldmax,'\0',(mv->max-oldmax)*sizeof(struct metricchar));
     }
+
+    // TODO: Move metrics calculations completely into shapers, and make MetricsView::metric const.
+    mv->metrics = (ShapeMetrics*)shaper_metrics(mv->shaper);
```
iorsh: Return this from shaper_apply_features()

fontforgeexe/shapers/shaper_shim.cpp
```diff
@@ -0,0 +1,143 @@
...
+    static ShaperDef shaper_defs[] = {{"builtin", N_("Built-in shaper")},
+#ifdef ENABLE_HARFBUZZ
+                                      {"harfbuzz", N_("HarfBuzz")},
+#endif
+                                      {NULL, NULL}};
+
+    return shaper_defs;
+}
+
+const char* get_default_shaper() {
```
iorsh, line 64: Add user preference for default shaper.
Retain current shaper per session, which would be initialized from preference, but would not affect it.

fontforgeexe/shapers/builtin.cpp
```diff
@@ -0,0 +1,114 @@
...
+        metrics[i].dy = y;
+        if (vertical) {
+            metrics[i].kernafter = rint(iscale * ots_arr_[i].vr.v_adv_off);
+            y += metrics[i].dheight + metrics[i].kernafter;
+        }
+        metrics[i].scaled = true;
+    }
+}
+
+std::set<Tag> BuiltInShaper::default_features(Tag script, Tag /*lang*/,
```
iorsh, line 94: Why this was not deleted in legacy code? If still in use, put the reference in comment

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -0,0 +1,591 @@
...
+        // [khaledhosny] Some OpenType features like init, medi, etc. are
+        // enabled by default and HarfBuzz applies them selectively based on
+        // text analysis, but when enabled manually they will be applied
+        // unconditionally which will break their intended use.
+        //
+        // If a default feature is selected in the UI, it should not be in the
+        // features list, but if it is unselected it should be in the features
+        // list with value set to 0 (disable).
+        const std::set<Tag> default_feats =
+            default_features(script, lang, vertical);
```
iorsh: Move outside the cycle

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -0,0 +1,591 @@
...
+        for (const auto& [idx, sc] : ttf_map_) {
+            if (name == sc->name) {
+                notdef_glyph_ = sc;
+                return notdef_glyph_;
+            }
+        }
+    }
+
+    // No special glyph found, create it.
+    notdef_glyph_ = context_->get_or_make_char(context_->sf, -1, ".notdef");
```
iorsh, line 591: What happens in legacy shaper?

fontforgeexe/metricsview.c
iorsh: By convention, output of ApplyTickedFeatures() is terminated by a zeroized structure and is never NULL (see source). As we move to shaper interface, this convention should hold, but we put this allocation here out of caution.

fontforgeexe/shapers/builtin.cpp
iorsh: Duplicates `StdFeaturesOfScript()`, which is still in use

fontforgeexe/shapers/harfbuzz.cpp
iorsh: In the legacy code we act similarly, calling
`MVTextChanged()`
` WordlistEscapedInputStringToParsedDataComplex()`
`  SFGetOrMakeChar()`

brigazvi:
THIS IS HUGE!!!!
@iorsh thank you so much for all the hard work you put into this. for RTL designers like me it's makes such a difference.

in my opinion it justifies a new release.

fontforge/splinefont_enums.h
```diff
@@ -64,6 +64,7 @@ enum ttf_flags {
     ttf_flag_dummyDSIG         = 1 << 15,
     ttf_native_kern            = 1 << 16, // This applies mostly to U. F. O. right now.
     ttf_flag_fake_map          = 1 << 17, // Set fake unicode mappings for unmapped glyphs
+    ttf_flag_no_outlines       = 1 << 18, // HarfBuzz: Skip "glyf" and "CFF " tables for performance
```
tuffnatty: @iorsh Hi! Here you're adding `ttf_flag_no_outlines` which is accidentally equal to `ps_flag_nohints`. So, after this change, a `font.generate()` call from Python with a `('no-hints')` flags passed generates a TTF without outlines, which is quite annoying and took some time to bisect and debug.
Probably `ttf_native_kern` (which is quite old) and `ttf_flag_fake_map` (which belongs to this changeset) which are conflicting with `ps_flag_nohintsubs` and `ps_flag_noflex`, also need to be addressed. And we're running out of bit space, so these values can't be just moved after 1 << 30. Probably `ps_flag_nohints` does not mean much in TTF building context, I am not sure; but changing the existing behaviour so that a corrupted file is generated seems to me a thing that should be avoided.

PR #5531, opened 2025-01-22
Fix CI for Ubuntu 24

iorsh:
`ubuntu-latest` CI runner now defaults to Ubuntu 24.04 (see https://github.com/actions/runner-images/issues/10636). This PR fixes the issues which arose with the new version.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,8 +9,6 @@ on:
 jobs:
   linux:
     runs-on: ${{ matrix.os }}
-    env:
-      PYTHON: python3.10
     strategy:
       matrix:
         os: ["ubuntu-latest"]
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,7 +50,12 @@ jobs:
         run: ninja -C repo/build dist
       - name: Run pyhook smoke test
         if: matrix.target != 'NoUI'
-        run: $PYTHON repo/tests/pyhook_smoketest.py
+        working-directory: repo
+        run: |
+          # There can be several Python versions installed. We must use the exact one discovered by CMake.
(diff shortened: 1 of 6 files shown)


iorsh:
Mac is failing with `hdiutil: create failed - Resource busy` error, which I've been observing coming and going for a few months now. Probably some instability, please ignore.

iorsh:
Windows CI is pending https://github.com/jtanx/fontforgebuilds/pull/23

.github/workflows/main.yml
```diff
@@ -52,7 +50,12 @@ jobs:
         run: ninja -C repo/build dist
       - name: Run pyhook smoke test
         if: matrix.target != 'NoUI'
-        run: $PYTHON repo/tests/pyhook_smoketest.py
+        working-directory: repo
+        run: |
+          # There can be several Python versions installed. We must use the exact one discovered by CMake.
+          PYTHON=`grep -m1 'Python3_EXECUTABLE.*=' build/CMakeCache.txt | cut -d= -f2`
```
jtanx, line 56: But the original version it selects is influenced by the PYTHON environment variable set above?

iorsh: Which one do you mean? I removed `PYTHON: python3.10` at line 13 above.

`PYTHON=python3` in `setup_linux_deps.sh` only serves the installation. Whatever is installed there would be later discovered by CMake and used accordingly.

iorsh:
@skef, @jtanx, could you please review? This PR should finalise the CI issues for the time being.

iorsh:
In an unrelated issue, the standalone `typing` package uses deprecated `setuptools` environment, and is now broken. We don't need it, since `typing` module is a part of standard Python since 3.5.

For more information: https://github.com/pypa/setuptools/pull/4870.
I also erroneously [blamed](https://github.com/sphinx-doc/sphinx/issues/13451) `sphynx` for this, but its maintainers kindly pointed me in the right direction.

PR #5545, opened 2025-03-14
GTK3-based Font View

iorsh:
This is a technology preview for GTK3-based UI. It's not meant to be merged right away (although it can), but it demonstrates the possibility to gradually move the FontForge UI to GTK, without breaking everything on the way. Based on infrastructure work by @Omnikron13.

The demonstration includes the Font View, Glyph Set Selection dialog, and Kerning Format dialog.

Tested on Linux (KDE, Gnome under X, Wayland), Windows 64-bit. Binaries for testing are available at https://github.com/iorsh/fontforge/actions/runs/13828445022 (scroll to the end of the page).

**Known issues:**
- Not tested on Mac
- Customized appearance (UI colors and fonts) not supported (support possible, but not implemented yet)
- Top menu bar doesn't collapse, so the window can't be narrower than the total menu bar width.

fv
sel_glyph
kf
See also #5324


BastienSANTE:
Hello!
I tried running your appimage on Debian Bookworm, but I could not get it to start for the moment : 

``` user@pc$ ./FontForge-2025-03-13-abb6b8a-x86_64.AppImage
Copyright (c) 2000-2025. See AUTHORS for Contributors.
 Version: 20230101
 Based on sources from 2025-03-13 06:28 UTC-ML-TtfDb-D-GDK3.
 Based on source from git with hash: abb6b8aa85cd62d7b7ce74d7ffe1e2d4ab54e38d
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
Python path configuration:
  PYTHONHOME = (not set)
  PYTHONPATH = (not set)
  program name = 'fontforge'
  isolated = 0
  environment = 1
  user site = 1
  import site = 1
  sys._base_executable = '/usr/bin/fontforge'
  sys.base_prefix = '/usr'
  sys.base_exec_prefix = '/usr'
  sys.executable = '/usr/bin/fontforge'
  sys.prefix = '/usr'
  sys.exec_prefix = '/usr'
  sys.path = [
    '/usr/lib/python38.zip',
    '/usr/lib/python3.8',
    '/usr/lib/lib-dynload',
  ]
Python initialization failed: failed to get the Python codec of the filesystem encoding
```

Your project to move the UI to GTK is great, And I think fontforge is in dire need of a new interface.

iorsh:
@BastienSANTE, thank you for the feedback! Please try an updated appimage at https://github.com/iorsh/fontforge/actions/runs/14563244522.

skef:
Since this has some activity: 

It doesn't seem like this serves the role of a technology demonstration as much now given that we have other GTK windows being integrated. And there are some limitations with this (such as the menu size) that the gdraw widgets don't have. So I wonder whether *as this is currently implemented* it makes much sense to merge. 

More specifically: The FontView is kind of crappy. The fact that it's basically either collapse (where it's tricky to add new glyphs) or not-collapsed (where might be **huge** gaps between non-empty glyphs) is inconvenient and hokey. Moving to GTK makes it much more viable to do something better. But maybe before merging this we should think about and implement the first steps in that direction, rather than eating the instability of the change for little benefit first, and then eating the instability of changing it later.

PR #5546, opened 2025-03-26
Simple GTK-based dialog with CSS appearance support

iorsh:
Initial support for GTK3 / GTKMM3-based UI. Only "Add Encolding Slots" dialog was ported, to keep stability while allowing for initial testing. Based on infrastructure work by @Omnikron13.

The customized UI appearance is supported with CSS - see screenshots.

Tested on Linux (KDE, Gnome under X, Wayland), Windows 64-bit, macOS. Binaries for testing are available at https://github.com/iorsh/fontforge/actions/runs/14070685506 (scroll to the end of the page).

Known issues:

- On macOS the dialog is shown in the corner instead of middle of the screen.

See also discussion #5324.

New GTK-based UI (default FontForge apperance):
gtk_dlg_default
Legacy GDraw-based UI (default FontForge apperance):
gdraw_dlg_default
---
New GTK-based UI (customized FontForge apperance):
gtk_dlg
Legacy GDraw-based UI (customized FontForge apperance):
gdraw_dlg


iorsh:
Currently the GTK window doesn't block its parent legacy window from receiving mouse events, and this can lead to crashes or issues. We plan to fix it later, once we have initial feedback of GTK usability and stability.

frank-trampe (approved):
We talked through this, and it makes sense. It is impossible to test all cases on all platforms. We will fix anything that breaks in testing at a later time. It is a small change, so any problems ought to be minimally disruptive.

.appveyor.yml
```diff
@@ -30,7 +30,7 @@ clone_depth: 10
 #cache:
 #  - 'C:\msys64\var\cache\pacman\pkg'
 install:
-  - git clone --depth=1 --branch=master https://github.com/fontforge/fontforgebuilds.git
+  - git clone --depth=1 --branch=gtk_win https://github.com/iorsh/fontforgebuilds.git
```
iorsh: Branch to "master_local"  in fontforge organization

PR #5547, opened 2025-03-31
Performance fixes for GSUB/GPOS dumps

iorsh:
This PR improves the performance of saving large OpenType GPOS/GSUB tables. 

@khaledhosny, on my machine the performance of saving Amiri font improved roughly from 68s to 10s, with comparable HarfBuzz metrics view (#5522) improvement. This is still suboptimal, but some Amiri kerning features are loaded as subtable with 6.8M (yes, six million) kerning pairs. I don't know what causes it, as the original font doesn't have anything like that.

### Type of change
- **Non-breaking change**

### DIFF
--- a/fontforge/CMakeLists.txt
+++ b/fontforge/CMakeLists.txt
@@ -57,6 +57,7 @@ set(FONTFORGE_NOINST_HEADERS
   splineutil.h
   splineutil2.h
   start.h
+  std_maps.hpp
   svg.h
   tottf.h
   tottfaat.h
@@ -199,6 +200,7 @@ add_library(fontforge
   splineutil2.c
   start.c
   stemdb.c
+  std_maps.cpp
   svg.c
   tottf.c
   tottfaat.c
--- a/fontforge/lookups.c
+++ b/fontforge/lookups.c
@@ -709,44 +709,51 @@ return( false );
 return( true );
 }
 
-SplineChar **SFGlyphsWithPSTinSubtable(SplineFont *sf,struct lookup_subtable *subtable) {
-    uint8_t *used = calloc(sf->glyphcnt,sizeof(uint8_t));
-    SplineChar **glyphs, *sc;
-    int i, k, gid, cnt;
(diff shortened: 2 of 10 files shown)


khaledhosny:
> This is still suboptimal, but some Amiri kerning features are loaded as subtable with 6.8M (yes, six million) kerning pairs. I don't know what causes it, as the original font doesn't have anything like that.

FontForge has limited support for class based kerning, it supports only certain value record formats, and for the rest it explodes the class kerning into kern pairs:
https://github.com/fontforge/fontforge/blob/072208fae782cd85e34bc09905e9133d8ded555c/fontforge/parsettfatt.c#L556-L562
https://github.com/fontforge/fontforge/blob/072208fae782cd85e34bc09905e9133d8ded555c/fontforge/parsettfatt.c#L662-L675

khaledhosny:
This is because `KernClass` uses an array of 16-bit integers for kerning values (`offsets`):
https://github.com/fontforge/fontforge/blob/072208fae782cd85e34bc09905e9133d8ded555c/fontforge/splinefont.h#L593-L609

But too fully represent GPOS PairPosFormat2 subtable, it would probably need two `valuerecord` arrays, and all code that interacts with kerning classes (including the UI) will need to be updated/extended to handle this.

That is one of the severe limitations of FontForge when working with RTL fonts, as it means it basically can’t handle RTL class-based pair positioning as it requires adjusting both the offset and advance width of the glyph.

It might be possible to detect RTL kerning and store and store only one value and special flag that it is RTL so that when writing GPOS back, the value is used for both offset and advance. This might not be perfect, but it will at least handle the common RTL kerning case while not needing too much changes to existing code.

iorsh:
Run bulk test

iorsh:
Bulk tests passed with the following environment:
```
FF_PY_SCRIPT_LINE="f=fontforge.open(argv[1]); f.generate(argv[2], flags='no-FFTM-table')"
FF_OUTPUT_FILE="Output.ttf"
```

PR #5558, opened 2025-05-08
Fix scaling of references in CharView

### LINKED ISSUE #5555: scaling composite glyph vertically fails
[Based on sources from 2025-05-01 04:54 UTC-ML-TtfDb-D-GDK3.
 Based on source from git with hash: 770356c9b52c003939a36ed3df711b08805efb3c]

Please try the following (captured in [this video](https://github.com/user-attachments/assets/91f9b172-94c4-47c1-9f88-6e87f321da35)).

* Call
  ```sh
  LANG=C fontforge DejaVuSansMono.ttf
  ```
  (I've tried with font version 2.25)
* Open editor window for glyph 'eogonek', which is a composite glyph.
* Use the mouse to completely select the base character so that the bounding box appears.
* Select the upper edge of the bounding box and move the mouse up and down (while pressing the mouse button).  The expected behaviour is that the bounding box line stays attached to the cursor so that you can scale the box up and down.  However, this doesn't work correctly: even smallest movements of the mouse cause extreme scalings, and if the scaling reaches zero, it stays as a horizontal line.  More mouse movements to undo this (which doesn't work) usually lead to a crash.

It seems there is some continuous multiplication done with a scaling factor instead of resetting the scaling to apply a new value after the mouse was moved.

iorsh:
Preserve `reflayer::splines` for Undo operation

### Type of change
- **Bug fix** - fixes #5555

_Note: it's not clear why `RefChar` holds a copy of layers in `struct reflayer`. It looks like an unnecessary and buggy data duplication. It could be good to remove it altogether, but at the moment such a refactoring would probably be too big and risky._

### DIFF
--- a/fontforge/cvundoes.c
+++ b/fontforge/cvundoes.c
@@ -93,6 +93,11 @@ return( NULL );
 	free(new->layers);
 	*new = *crefs;
 	new->layers = calloc(new->layer_cnt,sizeof(struct reflayer));
+	for ( layer=0; layer<new->layer_cnt; ++layer ) {
+	    new->layers[layer] = crefs->layers[layer];
+	    new->layers[layer].splines = SplinePointListCopy(crefs->layers[layer].splines);
+	    new->layers[layer].images = NULL;
+	}
 	new->next = NULL;
 	if ( last==NULL )
 	    head = last = new;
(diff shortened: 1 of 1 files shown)


skef (approved):
LGTM

PR #5562, opened 2025-05-24
Fix TTF validation on load for fixed pitch fonts

### LINKED ISSUE #5198: fontlint: bug: incorrect warnings for double-width fixed pitch truetype fonts
- [x] Screenshot

```
The advance width of space (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of exclam (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of quotedbl (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of numbersign (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of dollar (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of percent (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of ampersand (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of quotesingle (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of parenleft (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of parenright (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
```

(and then another twenty-five thousand lines, approximately, of similar warnings)

```
 Version: 20230101
 Based on sources from 2023-01-18 18:05 UTC-ML-D-GDK3.
```

Debian `bookworm`.

These warnings that appear aren't expected - the font is a double-width fixed pitch font, so a glyph width of `500` is acceptable for the advance width limit of `1000`.

```
fontlint iosevka-thin.ttf
```
([`iosevka` v18.0.0](https://github.com/be5invis/Iosevka/releases/tag/v18.0.0))
 
If it's possible to identify double (or other multiple) width fonts, then perhaps the logic here could be adjusted to use that information during lint checks: https://github.com/fontforge/fontforge/blob/20230101/fontforge/parsettf.c#L6304-L6306

Thank you!

iorsh:
When checking glyph width validity, consider zero, full width and double width for fixed-pitch fonts.

### Type of change

- **Bug fix** - Fixes #5198

### DIFF
--- a/fontforge/parsettf.c
+++ b/fontforge/parsettf.c
@@ -6304,9 +6304,13 @@ static SplineFont *SFFillFromTTF(struct ttfinfo *info) {
 			LogError(_("A point in %s is outside the font bounding box data.\n"), sc->name );
 			info->bad_cff = true;
 		    }
-		    if ( info->isFixedPitch && i>2 && sc->width!=info->advanceWidthMax )
-			LogError(_("The advance width of %s (%d) does not match the font's advanceWidthMax (%d) and this is a fixed pitch font\n"),
+		    if ( info->isFixedPitch ) {
+			bool valid = (sc->width==0 || sc->width==info->advanceWidthMax || sc->width==info->advanceWidthMax / 2);
+			if (!valid) {
+			    LogError(_("The advance width of %s (%d) does not match the font's advanceWidthMax (%d) and this is a fixed pitch font\n"),
 				sc->name, sc->width, info->advanceWidthMax );
+			}
+		    }
 		}
 	    }
 	    ++k;


fontforge/parsettf.c
```diff
@@ -6304,9 +6304,13 @@ static SplineFont *SFFillFromTTF(struct ttfinfo *info) {
 			LogError(_("A point in %s is outside the font bounding box data.\n"), sc->name );
 			info->bad_cff = true;
 		    }
-		    if ( info->isFixedPitch && i>2 && sc->width!=info->advanceWidthMax )
```
jayaddison, line 6307: What is the reason for the `i>2` condition in the existing code?  Are the first two glyphs special/ignorable for some reason?

fontforge/parsettf.c
```diff
@@ -6304,9 +6304,13 @@ static SplineFont *SFFillFromTTF(struct ttfinfo *info) {
...
-			LogError(_("The advance width of %s (%d) does not match the font's advanceWidthMax (%d) and this is a fixed pitch font\n"),
+		    if ( info->isFixedPitch ) {
+			bool valid = (sc->width==0 || sc->width==info->advanceWidthMax || sc->width==info->advanceWidthMax / 2);
```
jayaddison, line 6312: Is the `sc->width==0` test here for the same reasons (hidden characters) -- and/or others? -- outlined in: https://github.com/fontforge/fontforge/blob/770356c9b52c003939a36ed3df711b08805efb3c/fontforge/splinesaveafm.c#L3064-L3065

jayaddison: (and either way, maybe worth adding a comment about the reason here?)

fontforge/parsettf.c
iorsh: This is a very old code, so the first glyph is probably `.notdef`, and the second also used to have some special meaning back then, maybe `\n`. I tend to kill unexplained cryptic stuff, especially when it only affects a warning message.

fontforge/parsettf.c
iorsh: Modern Unicode fonts tend to have lots of zero-width characters like control stuff, e.g. `LRM` or diacritics. These characters don't have an advance width and shouldn't affect pitch computation. 

Good idea about comment.

jayaddison: Gotcha.  Thank you!

jayaddison (changes requested):
@iorsh I approved this without testing it, but then thought today: I really should have applied the patch and checked the results first.

So, I applied this branch as a patch against Debian's packaged version of the `20230101` (`1:20230101~dfsg-4`, specifically), and rebuilt that.  Unfortunately the warning messages continue to appear:

```
The advance width of space (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of exclam (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of quotedbl (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of numbersign (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of dollar (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of percent (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of ampersand (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of quotesingle (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of parenleft (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of parenright (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of asterisk (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of plus (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of comma (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
The advance width of hyphen (500) does not match the font's advanceWidthMax (1000) and this is a fixed pitch font
...
```

I'm not certain why they continue to appear despite the changes; I'll try to figure that out.

jayaddison:
> @iorsh I approved this without testing it, but then thought today: I really should have applied the patch and checked the results first.
... 
> I'm not certain why they continue to appear despite the changes; I'll try to figure that out.

I think this may be a testing error by me; I think that although I ran the `fontlint` script from the patched build, the unpatched `fontforge` binary and/or library were invoked to perform the TTF parsing.  I'm re-checking my testing process to try to confirm that theory.

jayaddison:
> I think this may be a testing error by me; I think that although I ran the `fontlint` script from the patched build, the unpatched `fontforge` binary and/or library were invoked to perform the TTF parsing. I'm re-checking my testing process to try to confirm that theory.

My apologies; yes, after an isolated rebuild of the package with this branch (at e9fa9a3de7e136b9bc8df10878c01d65d86f0ea7) applied as a patch, the warnings have disappeared.  The resulting output of the `fontlint` is concise and appears useful:

```sh
$ fontlint ~/Iosevka-Thin.ttf 2>&1 | more
Copyright (c) 2000-2024. See AUTHORS for Contributors.
 Version: 20230101
 Based on sources from 2024-09-06 01:07 UTC-ML-D-GDK3.
PythonUI_Init()
copyUIMethodsToBaseTable()
Program root: /usr
Attempt to unget two characters
CHECKING     /root/Iosevka-Thin.ttf
ERROR      2 Self-intersecting glyph
ERROR      5 Missing points at extrema
ERROR      9 Bad glyph name
FAIL         /root/Iosevka-Thin.ttf
```

Thank you for the fix!

PR #5575, opened 2025-06-29
Update po files from crowdin translations

skef:
Also add (currently minimal) Thai translations in th_TH.po

I'm not the person who has usually done this in the past. What I did was:

1. Download the updated po files from Crowdin
2. Put them in a subdirectory called `updates` in the `po` directory
3. Configured a cmake build of FontForge
4. Used that build to generate a `FontForge.pot` file
5. Copied a different existing .po file to `th_TH.po`, edited the headers and blank top message, and deleted the remaining contents
6. Added `th_TH` to LINGUAS
7. With the `po` directory as CWD, ran `python ./update.py ../build/po/FontForge.pot`

Or, at least, that's what I think I did. 

I'll attach the CrowdIn zip in case other things need to happen
[FontForge (translations).zip](https://github.com/user-attachments/files/20970071/FontForge.translations.zip)

---
Changes from skef/fontforge#1:
- Stripped trailing `\n` from all (most) `LogError()` messages
- Renamed Crowdin `msgid`s to stripped strings to retain provenance of translations
- Removed approvals from all Crowdin translations, this feature is useless without dedicated translation team which reviews translations in their native languages. In some languages this caused a backlog of proposed translations to be automatically accepted.
- No line wrapping in `.po` files, for stability and facilitation of automatic changes


iorsh:
There seem to be some errors in .po compilation, I'll take a look at them.

skef:
Lots of newline mismatch complaints. It would be best to correct all of of this in the Crowdin sources, of course. Hmm.

iorsh:
I think the root cause here is that translatable messages should not end with `\n` at all. I can see that lots of them belong to `LogError()`, which doesn't really need a trailing newline. It resolves to either `NOUI_LogError()`, which adds `\n` to `stderr` anyway, or `UI_LogError()`, which always removes it.

These newlines are translators' nightmare - they don't show up properly in Crowdin, and can only be spotted through an obscure and unreliable "QA" mark
image

iorsh:
Looks like Crowdin ignores whitespace changes (such as added or removed newline) when uploading translations. The only thing I could do so far is deleting the string altogether and re-uploading it again. Unfortunately it deletes the provenance, which is not good.

skef:
Maybe add the newlines back as a post-processing step?

iorsh:
I'm trying to automatically remove as much trailing newlines as I can, this is quite a pain. I'll see if this reduces the number of error to a manageable amount

JoesCat:
Open-up KDE Kwrite, it's a pretty good text editor. Set the settings to remove trailing spaces on entire document. Load the po file, add a space, to update, then save, the entire document should have all trailing spaces removed.
I think Notepad+ (or Notepad++) does similar if you run windows.

iorsh:
The problem is unfortunately much worse than that. We have `msgid`s (source strings) which are harvested from the source code. Some of them have trailing `\n`, and some of them don't. Those which do have trailing newline, must be translated with a trailing newline too. Those which don't have it, must be translated without. Otherwise the `gettext` compiler complains.

Even if I do align all this stuff manually, the problem will recur. The Crowdin interface doesn't emphasize the trailing newline. One must be pretty technical to notice the newline presence or absence and spot this additional line break which is only manifested by cursor placement.

Basically I'm trying to strip as much trailing newlines as possible, as most of them are already meaningless - the software normalizes them anyway.

iorsh:
Also, stripping them in our source doesn't help. Crowdin doesn't recognize this change and silently discards translations whose `msgid` was changed by removing thew newline.

iorsh:
A summary of fixing procedure: https://github.com/iorsh/fontforge/wiki/Crowdin-fix . We are destined to suffer...

JoesCat:
...while making updates on designwithfontforge, I was running into some wild behaviour with Farsi's right to left, so that's yet another thing to look forwards too ;-P

perhaps ssome thinking outside the box is needed here....
I just finished contributing to another project which had some different way of writing text, it seemed a bit verbose, but might help here in this case....

What I was seeing looked something like this:
```
printf("blah, blah, blah\n"
       "\n"
       "blah, blah, blah\n"
       "\tblah, blah, blah\n"
       );
```
so perhaps adapting to consider crowdin and newline issues....
```
/* string below is:
 * blah, blah, blah value1 value2
 *
 * blah, blah, blah
 *          string1 blah, blah, blah
printf(_("blah, blah, blah %d1 %d2"),value1, value2);
printf("\n\n");
printf(_("blah, blah, blah"));
printf("\n");
printf(_("\t%s blah, blah, blah", string1));
printf("\n");
```
This turns into a crazy verbose, and sort of hard to read string as a developer, but probably low (or lower) maintenance in the long run as you wouldn't need to come revisit and remodify per update.
You'd need to include tabs inside the translation, because Farsi would go right-to-left.

maybe something to think about?

iorsh:
Newlines and inline formatting are generally considered a bad practice in localization.
What I would do is
```
/* Comment for readability... */
char title[100], *paragraph = _("blah, blah, blah"), tab[100];

sprintf(title, _("blah, blah, blah %d1 %d2"),value1, value2);
sprintf(tab, _("%s blah, blah, blah", string1));

printf("%s\n\n%s\n\t%s\n", title, paragraph, tab);
```

Otherwise you end up with twenty different approaches to each formatting challenge, as different languages are usually translated by different people.

iorsh:
Btw, @JoesCat, feel free to poke me about RTL issues at designwithfontforge. That stuff is not new to me :)

JoesCat:
You're almost there.... I ran into issues earlier (with plugin-gimp-fourier and gimp-fix-ca) and discovered you'd also need to handle xgettext with one set of data, and gettext with the printf() stage. What worked, was to split it in 2 stages. My solution for the 1st stage was to do something like this:
```
/* Comment for readability... */
char title[100], *paragraph = d_("blah, blah, blah"), tab[100];
```
...so, for the above, xgettext finds the string, and I needed to add a new something for xgettext, in this case, was adding [d_](https://github.com/rpeyron/plugin-gimp-fourier/blob/main/po/Makefile.am#L52)
this lets you compile the code as expected...and then to make the second stage work, you still need gettext to work with printf, so it probably should look like this instead:
```
printf("%s\n\n%s\n\t%s\n", title, _(paragraph), tab);
```


> Btw, @JoesCat, feel free to poke me about RTL issues at designwithfontforge. That stuff is not new to me :)

Thanks! If you want to poke at designwithfontforge, go ahead, but I suspect your plate is pretty full as-is right now ;-P

At the moment, think I've managed to clean-up the 1st 1/3 to be all synced-up, but have yet to go through the last 2/3rd syncing and cleaning up. I also expect there's several "new" pages that need to be synced from Japanese to the other languages. Far too much to do all in one shot - and the break in between helps. ...however, 1st-thing-1st, I need to get some exercise first! Time for a walk....

iorsh:
Opened skef/fontforge#1 into this branch. @skef, could you please check if you have time? I can open PR directly into fontforge/master, if that's easier.

skef:
I'll check in a bit

iorsh (approved):
Combined PR from skef/fontforge#1 (opened by me, approved by @skef) and commit 340bf1920bb11691bb4922755d74d6b634e3bb30 (by @skef, reviewed by me). 

Looks good overall.

PR #5583, opened 2025-07-11
Be more clever about label text in gtextfield

### LINKED ISSUE #5580: Python scripting: askMulti segfault when missing optional default values
I was looking through the python scripting menu documentation and found a consistent crash when using File -> Execute Script.

From some quick testing, all of the non-"askMulti" variations of the menus seem to work as expected.

I found that if you do not include the optional "default" key or set it to None when using askString and askChoice in a python list in AskMulti there is a segfault. If it is any help at all I've attached an strace log of opening a font file and executing the askMulti with only 'string' and 'ques' (the first non-working example).

The documentation says default is optional and as with the non-askMulti versions, it indeed is. so I would expect that not including the 'default' key in an askMulti list would have no ill effects.

These work fine as expected

[fontforge-strace.txt](https://github.com/user-attachments/files/21090573/fontforge-strace.txt)

:
```
import fontforge
ans = fontforge.askString("title", "ques", "def" )
```

```
import fontforge
dict = fontforge.askMulti("title", [ { 'type': 'string', 'question': 'ques', 'tag': 3, 'default': 'def' } ])
```

```
import fontforge
dict = fontforge.askMulti("title", [ { 'type': 'string', 'question': 'ques', 'default': '' } ])
```

```
import fontforge
dict = fontforge.askMulti("title", [ { 'type': 'choice', 'question': 'What direction:',
  'answers': [ { 'name': 'Clockwise', 'tag': 'cw', 'default': True }, { 'name': 'Counter-Clockwise', 'tag': 'ccw' } ] } ] )
```

These crash immediately:
```
import fontforge
dict = fontforge.askMulti("title", [ { 'type': 'string', 'question': 'ques' } ] )

```
```
import fontforge
dict = fontforge.askMulti("title", [ { 'type': 'string', 'question': 'ques', 'default': None } ])
```

This crashes after pressing "Ok" in the dialog menu, but not when pressing cancel:
```
import fontforge
dict = fontforge.askMulti("title", [ {'type': 'choice', 'question': 'What direction:',
  'answers': [ {'name': 'Clockwise', 'tag': 'cw' }, { 'name': 'Counter-Clockwise', 'tag': 'ccw' } ] } ] )
```

skef:
Closes #5580

### DIFF
--- a/gdraw/gtextfield.c
+++ b/gdraw/gtextfield.c
@@ -2684,7 +2684,7 @@ static GTextField *_GTextFieldCreate(GTextField *gt, struct gwindow *base, GGadg
     _GGadget_Create(&gt->g,base,gd,data,def);
 
     gt->g.takes_input = true; gt->g.takes_keyboard = true; gt->g.focusable = true;
-    if ( gd->label!=NULL ) {
+    if ( gd->label!=NULL && gd->label->text!=NULL ) {
 	if ( gd->label->text_is_1byte )
 	    gt->text = /* def2u_*/ utf82u_copy((char *) gd->label->text);
 	else


frank-trampe (approved):
This looks obviously necessary and correct in the case that it handles.

PR #5584, opened 2025-07-15
Add minimal support for GDEF version 1.3

khaledhosny:
Instead of checking the full version number and rejecting any new versions, check for major version number since new minor versions add new fields but do not change existing ones.

This allows loading glyph classes etc from GDEF version 1.3, while ignoring item variation sore as OpenType variations in general are not supported by FontForge

### Type of change
- **Non-breaking change**


iorsh (approved):
Looks good to me.

PR #5589, opened 2025-07-21
Sanitize messages from python

### LINKED ISSUE #45: Interpolation crash
I open this SFD, https://gist.github.com/3506577, select the 'n' and 'o' and run the attached fontforge python script and get a segfault :)

Caellian:
Previously passing strings containing '%' would cause fontforge to crash, this commit makes fontforge emit a warning instead and escape those characters.

Closes #45.

### Testing

Tested with:
```py
import fontforge

fontforge.logWarning("This should be just fine as always...")
fontforge.logWarning("Ooops, I forgot to escape % so there might be trouble")
fontforge.logWarning("This case might be a bit different:%")
fontforge.logWarning("A % string with % very%many % unescaped %%% formats")
```

Output:
```
This should be just fine as always...
logError called with a string containing unescaped format character(s): 'Ooops, I forgot to escape % so there might be trouble'
Ooops, I forgot to escape % so there might be trouble
logError called with a string containing unescaped format character(s): 'This case might be a bit different:%'
This case might be a bit different:%
logError called with a string containing unescaped format character(s): 'A % string with % very%many % unescaped %%% formats'
A % string with % very%many % unescaped %% formats
```

### Type of change
- **Bug fix**
- **Non-breaking change**


fontforge/python.c
```diff
@@ -80,6 +80,7 @@
 #include "tottfgpos.h"
 #include "ttf.h"
 #include "ttfinstrs.h"
+#include "uiinterface.h"
```
Caellian, line 83: This was a transient include before, but LogError is directly used by this file.

fontforge/python.c
```diff
@@ -1241,12 +1243,62 @@ static PyObject *PyFF_scriptFromUnicode(PyObject *UNUSED(self), PyObject *args)
...
+
 static PyObject *PyFF_logError(PyObject *UNUSED(self), PyObject *args) {
     char *msg;
     if ( !PyArg_ParseTuple(args,"s", &msg) )
-return( NULL );
+        return( NULL );
+    sanitize_str_for_printf(&msg, "logError called with a string containing unescaped format character(s): \"%s\"");
     LogError(msg);
-Py_RETURN_NONE;
+    Py_RETURN_NONE;
```
Caellian, line 1300: As codebase is moving away from left margin returns I indented returns in touched functions.

fontforge/python.c
```diff
@@ -1241,12 +1243,62 @@ static PyObject *PyFF_scriptFromUnicode(PyObject *UNUSED(self), PyObject *args)
     return TagToPythonString(script, false);
 }
 
+static bool sanitize_str_for_printf(char** message, const char* warning_fmt) {
```
Caellian: I wasn't sure where to put this, there wasn't some clear 'string utils' place, so I placed it in python.c before it's used.

Let me know where and I can move it.

Caellian:
cc @skef

fontforge/python.c
```diff
@@ -1241,12 +1243,62 @@ static PyObject *PyFF_scriptFromUnicode(PyObject *UNUSED(self), PyObject *args)
...
+    // properly escaped.
+    sanitized = malloc(length + additional_characters + 1);
+    for (source_pos = 0; source_pos < length; source_pos++) {
+        sanitized[target_pos++] = (*message)[source_pos];
+        if ((*message)[source_pos] == '%') {
+            sanitized[target_pos++] = '%';
+            if (source_pos != length - 1 && (*message)[source_pos + 1] == '%') {
+                source_pos++;
+                continue;
+            } else if (!already_warned) {
```
iorsh: No need to keep `already_warned` flag, you can check `additional_characters > 0` and call `LogError()` before the loop

fontforge/python.c
```diff
@@ -1241,12 +1243,62 @@ static PyObject *PyFF_scriptFromUnicode(PyObject *UNUSED(self), PyObject *args)
...
+    *message = sanitized;
+    return TRUE;  // new message was allocated
+}
+
 static PyObject *PyFF_logError(PyObject *UNUSED(self), PyObject *args) {
     char *msg;
     if ( !PyArg_ParseTuple(args,"s", &msg) )
-return( NULL );
+        return( NULL );
+    sanitize_str_for_printf(&msg, "logWarning called with a string containing unescaped format character(s): \"%s\"");
```
iorsh: Check output from `sanitize_str_for_printf()` and release allocated message (also below)

fontforge/python.c
```diff
@@ -1256,18 +1308,20 @@ static bool showPythonErrors = 1;
 static PyObject *PyFF_postError(PyObject *UNUSED(self), PyObject *args) {
     char *msg, *title;
     if ( !PyArg_ParseTuple(args,"ss", &title, &msg) )
-return( NULL );
+        return( NULL );
+    sanitize_str_for_printf(&msg, "postError called with a message string containing unescaped format character(s): \"%s\"");
```
iorsh: Pass just the name of Python routine and make the message translatable inside `sanitize_str_for_printf()`:
```
LogError(_("Function %s() called with a message string containing unescaped format character(s): \"%s\""), py_routine_name, *message);
```

fontforge/python.c
```diff
@@ -1241,12 +1243,62 @@ static PyObject *PyFF_scriptFromUnicode(PyObject *UNUSED(self), PyObject *args)
...
+                continue;
+            }
+            // not escaped
+            additional_characters++;
+        }
+        // no need to escape
+    }
+    if (additional_characters == 0) {
+        // String is okay to pass to printf and friends
+        return FALSE;  // new message wasn't allocated
```
iorsh: By convention, please use lowercase `true`, `false`

iorsh:
> ### Testing
> Tested with:
> 
> ```python
> import fontforge
> 
> fontforge.logWarning("This should be just fine as always...")
> fontforge.logWarning("Ooops, I forgot to escape % so there might be trouble")
> fontforge.logWarning("This case might be a bit different:%")
> fontforge.logWarning("A % string with % very%many % unescaped %%% formats")
> ```

[Optional] Please add this code snippet to a unit test (see test*.py examples in [tests](https://github.com/fontforge/fontforge/tree/master/tests))

fontforge/python.c
Caellian: That string is managed by Python. I tried freeing it initially but it segfaults on `free`.

iorsh: No, I mean release message only when it was allocated by `sanitize_str_for_printf()`:
```
bool message_reallocated = sanitize_str_for_printf(&msg, "logWarning called with a string containing unescaped format character(s): \"%s\"");
LogError(msg);
if (message_reallocated) {
    free(msg);
}
```

Caellian: Oooh, hahahaha. Right. :man_facepalming:

tests/test1022.py
iorsh, line 1: Please give it a meaningful name, e.g. test_logging.py, and add the test to `tests/CMakeLists.txt`

iorsh (approved):
Looks good to me

PR #5592, opened 2025-07-30
Fix a crash caused by deleting a glyph with vertical kerning pairs.

### LINKED ISSUE #5591: Deleting a glyph with vertical kerning pairs leaves those vkerns pointing to the deleted glyph
In SFRemoveGlyph in encoding.c, the code loops over the glyph's horizontal kerning pairs and deletes those that point at the glyph we're removing. However, it doesn't delete any vertical kerning pairs that point to the glyph. This leaves the vkerns pointing at bad memory, and any attempts to access the vkerns can cause a crash.

I was hitting this in code that processed a Japanese font, removing glyphs outside the basic multilingual plane. I have an incoming pull request with the fix.

BradEdwards:
- SFRemoveGlyph removed the horizontal kerning pairs but left vertical kerning pairs pointing to the deleted glyph.
- This could cause crashes when then accessing the vkerns.

### Type of change

- **Bug fix**
- Fixes #5591

### DIFF
--- a/fontforge/encoding.c
+++ b/fontforge/encoding.c
@@ -2217,10 +2217,23 @@ return;
 
     /* Remove any kerning pairs that look at this character */
     for ( i=0; i<sf->glyphcnt; ++i ) if ( sf->glyphs[i]!=NULL ) {
+	/* Remove matching horizontal kerning pairs */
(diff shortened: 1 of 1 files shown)


iorsh (approved):
Looks good to me

PR #5593, opened 2025-08-01
Update po translations from Crowdin

iorsh:


frank-trampe:
So this just fills in a few gaps that you found?

po/es.po
```diff
@@ -6288,6 +6294,9 @@ msgstr "Pred."
 msgid "DefaultFont"
 msgstr "Fuente predeterminada"
 
+msgid "DefaultShaper"
```
frank-trampe, line 6297: Is the camel-casing (in place of a space) intentional?

iorsh:
It's mostly the updates to Spanish translation, which was finalized, and I think this effort warrants a special update. Earlier here was a big localization PR in #5575, reviewed by @skef. It cleaned up a lot of whitespaces, which confused both translators and the Crowdin system.

frank-trampe (approved):
This looks fine to me.

po/es.po
iorsh: Yes, this is how all preferences look - see https://github.com/fontforge/fontforge/blob/88de9b7dc76fbdef9069b7664ecc280f1c15ba2f/fontforgeexe/prefs.c#L304
Maybe I'll get to fix this eventually, but not now...

PR #5594, opened 2025-08-01
Upgrade to Unicode 16.0.0

### LINKED ISSUE #5472: FontForge does not support Unicode 15.1.0 and 16.0.0.

### Important

- [ ] Screenshot
- Currently, the version created when the command is executed is Unicode 15.0.0.
> cd Unicode
> python makeutype.py

An error occurs when you manually change the version to 15.1.0 or 16.0.0.

iorsh:
Fixes #5472.


ChiefMikeK:
@iorsh 👀  
Would be better to update for **17.0**
- [The Unicode Blog: Finalizing Unicode 17.0 in preparation for a September 9 release.)](https://blog.unicode.org/2025/07/highlights-from-utc-meeting-184.html)

jtanx (approved):

iorsh:
Thank you, @jtanx !

@ChiefMikeK, hopefully, further releases would be sufficiently frequent, and updating to Unicode 17 would not be delayed too much. For now, I'm merging it as is.

PR #5608, opened 2025-09-01
Fix for cmake installation in MacOS CI runner

iorsh:
`brew install cmake` fails in recent MacOS runners due to change in actions/runner-images#12791.

This is a know issue - see actions/runner-images#12912. Fixed per comment by @XuehaiPan. 

### Type of change
- **Non-breaking change**


iorsh:
Not needed. CI issue fixed in actions/runner-images#12934

PR #5612, opened 2025-09-05
Remove X11 and non-Cairo drawing backends

### LINKED ISSUE #3462: GDK and X11 size difference on X11
Currently, the GDK back-end is larger on OS X than the X11 when on a Retina display. This causes issues on my MacBook Pro with a smaller screen, with some windows being unusable or nigh-unusable.

- [X] Screenshot
screen shot 2019-02-06 at 11 06 46 am

Fontforge dad4fb502202e8b467d571f5ab0f07b7a3565f69, Mac OS X 10.14.3
The X11 and GDK back-ends being the same size; The Mac version is noticeably bigger than the X11 version.

### LINKED ISSUE #4346: We ought to consider removing the Xorg backend
GDK is very stable. Xorg is rotting more and more and is rarely tested. Two backends are a waste of our scarce resources and hold back the project. My COLR/CPAL work is just one thing being held back by this. So is #3874.

99% of those on current releases are using GDK. The only ones who aren't are those specifically building on their own and telling Cmake to. 

In the next release, Cmake ought to emit a deprecation flag that will take a special command line option to overcome. In the release after, we ought to remove both and all Xorg backend code.

### LINKED ISSUE #5611: Building with `ENABLE_X11` no longer works
https://github.com/fontforge/fontforge/pull/5546 adds an unconditional dependency on GTK and GDK, leading to an error when building with `ENABLE_X11=ON`:

```
CMake Error at fontforgeexe/gtk/CMakeLists.txt:16 (target_link_libraries):
  Target "ffgtk" links to:

    PkgConfig::GTKMM

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

I would not be opposed to removing the X11 backend entirely.

valadaptive:
Resolves #5611 by just removing support for the X11 backend entirely.

If the eventual goal is to implement everything in GTK, we can remove some X11-exclusive code and get rid of quite a bit of cruft. In my testing, the X11 backend seems a bit broken--it doesn't actually exit the process when you close the windows. The non-Cairo drawing backend renders handles *below* outlines, and randomly decides to zoom into an arbitrary point:

https://github.com/user-attachments/assets/b0ebb1d1-6e03-4df1-b7db-eb21b17fdf8e

This PR is split into a few different parts. I can split them into separate PRs, but as a new contributor, I don't want to spam this repo with a bunch of them at once:

1. Use [`unifdef`](https://dotat.at/prog/unifdef/) to remove the `_NO_LIBCAIRO` ifdefs. This flag was already never being set, and not much substantive code was gated behind it.
2. Unifdef `FONTFORGE_CAN_USE_GDK`. This removes all the gxdraw code, which was gated behind `#ifndef FONTFORGE_CAN_USE_GDK`.
3. Remove the `ENABLE_X11` CMake option, which now does nothing.
4. Remove the UseCairoDrawing preference, and corresponding command-line flag. These already have no effect if `ENABLE_X11` is off (which it is by default). Also remove the gxdraw source files (now just stubs) entirely.
5. Remove `GDrawHasCairo`, which now always returns true, and all code paths in the charview used for non-Cairo rendering. This gets rid of a lot of cruft. This should be reviewed carefully.

### Type of change

- Fixes #4346, fixes #3462, 
- **Breaking change** (? This doesn't affect the Python API, but could theoretically break someone's workflow somewhere)


iorsh:
I didn't review this thoroughly, but the idea looks reasonable. GDK backend has been enabled for at least 6 years, and I guess raw X11 is not really needed anymore.

Still, I'd like to seek additional opinions for such a bold move.

frank-trampe:
@valadaptive, this is something that @iorsh and I have discussed doing in the future and shall likely happen, but it is going to require a little consultation with downstream parties and then a deprecation notice in order to limit disruption. Our limited market power means that changes that break somebody's packaging script can lead to the version getting pinned for a long time.

valadaptive:
Which downstream packagers still enable the X11 backend?

valadaptive:
Also, as @iorsh mentioned in #5611:

> This wouldn't be a real fix though, as the new GTK window comes up and even works, but behaves rather erratically.

The latest master is *already* broken if the X11 backend is enabled. I definitely understand if you don't want this change being made by a new contributor though.

skef:
@frank-trampe Is the implication that #5546 either needs to be backed out or needs an X11-specific work-around?

iorsh:
@valadaptive, I was unable to reproduce the arbitrary zoom issue you encountered, but I guess it's reasonably common if you stumbled upon it.

@skef, a build fix for #5611 is possible. It would slowly degrade the user experience with X11 backend (with newer GTK windows working, but buggy) instead of shutting it off at once.

skef:
I remain skeptical of the inferences being made about the user base. What significant contact have we even had with the user base over the past, say, 4 years? How much script-level development has shifted to fontTools over the past, say, 7 years? What users who are relying on the X11 back-end specifically wouldn't be adequately served by simply sticking with an older version of the program?

valadaptive:
> What significant contact have we even had with the user base over the past, say, 4 years?

Perhaps relevant, I read [this article](https://chameth.com/making-a-font-of-my-handwriting/) on Hacker News, and it seems to show how people see FontForge nowadays.

frank-trampe:
I have no current specific information from downstream users/packagers, just memories of past large changes not going over very well. Having been a victim of a few abrupt removals of non-deprecated features in other products, I am sensitive to the general importance of giving notice of large changes.

FontForge has always been clunky, even when it had fewer features, and it was once a lot buggier than it is now. (Try a version from 2012.) Accordingly, evidence-based disdain for FontForge is nothing new. There are nonetheless a lot of people who still use FontForge, and it is important to avoid unnecessarily inconveniencing them.

iorsh:
I did some checks, and of all the major distros I could recall none seems to override the default option of `ENABLE_X11=OFF`

- Debian - https://salsa.debian.org/fonts-team/fontforge OK
- Homebrew - https://github.com/Homebrew/homebrew-core/blob/main/Formula/f/fontforge.rb OK
- MSYS2 - https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-fontforge OK
- Fedora - https://src.fedoraproject.org/rpms/fontforge/tree/rawhide OK
- FreeBSD - https://www.freshports.org/print/fontforge OK
- Ubuntu - adopts Debian package as is https://launchpad.net/~fontforge

frank-trampe:
Good. Our informal deprecation has been fruitful.

cmake/FontForgeConfigure.cmake
```diff
@@ -80,14 +76,10 @@ function(fontforge_generate_config template destination)
   _set_negated(_NO_XKB "${X11_Xkb_FOUND}")
   _set_negated(_NO_XINPUT "${X11_Xi_FOUND}")
 
-  if(NOT ENABLE_GUI OR NOT ENABLE_X11)
+  if(NOT ENABLE_GUI)
```
iorsh: I think this clause can go, together with `X_DISPLAY_MISSING` and `xkeysyms_unicode.c`

fontforgeexe/charview.c
```diff
@@ -3397,8 +3068,6 @@ static GWindow CharIcon(CharView *cv, FontView *fv) {
     bdf = NULL; bdfc = NULL;
     if ( sc->layers[ly_fore].refs!=NULL || sc->layers[ly_fore].splines!=NULL ) {
 	bdf = fv->show;
-	if ( sc->orig_pos>=bdf->glyphcnt || bdf->glyphs[sc->orig_pos]==NULL )
```
iorsh: Please explain why this code was removed. `fv->filled` belongs to `FontView`.

fontforgeexe/startui.c
```diff
@@ -264,11 +261,7 @@ static void SplashLayout() {
 #ifdef FONTFORGE_CONFIG_USE_DOUBLE
     uc_strcat(pt,"-D");
 #endif
-#ifdef FONTFORGE_CAN_USE_GDK
     uc_strcat(pt, "-GDK3");
```
iorsh: Remove this. Now, when GDK3 is manadatory, there is no need to mention it on the splash screen.

fontforgeexe/startui.c
```diff
@@ -657,9 +650,7 @@ int fontforge_main( int argc, char **argv ) {
 #ifdef FONTFORGE_CONFIG_USE_DOUBLE
 	        "-D"
 #endif
-#ifdef FONTFORGE_CAN_USE_GDK
             "-GDK3"
```
iorsh: Remove this too, unnecessary now.

po/ca.po
iorsh, line 1: *.po files are autogenerated, and any localized message which was deleted from the source will be automatically deleted here. Please remove them from the PR.

cmake/FontForgeConfigure.cmake
valadaptive: I've removed all of that now.

fontforgeexe/charview.c
valadaptive: Good catch; I only meant to remove `cv->filled` since Cairo draws the equivalent.

fontforgeexe/startui.c
valadaptive: I've removed the `-GDK3` display.

po/ca.po
valadaptive: I've reverted those changes.

valadaptive:
I've made the requested changes.

I've also noticed that `cv->back_img_out_of_date` and `cv->backimgs` seem to be completely unused by the Cairo version of the charview. I'll remove those in a follow-up PR; this one is big enough as-is.

iorsh (commented):
Please also remove a few more `define`s I missed earlier:

- _NO_XKB
- _NO_XINPUT
- BUILT_WITH_XORG
- _NO_LIBCAIRO (there are few remnants in startui.c)

The file `cmake/backports/3.15.7/FindX11.cmake` should probably also go.

Please don't force-push, it break the possibility to do incremental review. Thanks!

valadaptive:
I've removed those additional `define`s (seems like unifdef missed those `_NO_LIBCAIRO`s for some reason) and `FindX11.cmake`.

I think the entire CMake backports folder can probably go away in another PR--[even Debian oldoldstable ships cmake 3.18](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=cmake), and CMake prints a big ugly deprecation warning if we set `cmake_minimum_required` to anything less than 3.10. Would another PR for that be accepted? Not sure which CMake version to target.

iorsh:
> I've removed those additional `define`s (seems like unifdef missed those `_NO_LIBCAIRO`s for some reason) and `FindX11.cmake`.
> 
> I think the entire CMake backports folder can probably go away in another PR--[even Debian oldoldstable ships cmake 3.18](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=cmake), and CMake prints a big ugly deprecation warning if we set `cmake_minimum_required` to anything less than 3.10. Would another PR for that be accepted? Not sure which CMake version to target.

CMake was introduced by @jtanx, and I don't know what were his considerations when adding local `Find*.cmake` files to the build system. My guess is that they were too divergent or just not available on all platforms. I recently tried to remove `FindGDK3.cmake`, but it didn't go smoothly, so I gave up.

@jtanx, would it be safe to remove these files and rely on the standard CMake distribution?

In any case, I'd prefer to defer such PR until after the official release. We've got enough risks for this release already.

valadaptive:
> I recently tried to remove `FindGDK3.cmake`, but it didn't go smoothly, so I gave up.

`FindGDK3` is not part of the backports folder. [Everything in here should be provided by newer versions of CMake](https://github.com/fontforge/fontforge/tree/master/cmake/backports/3.15.7)--I tried removing the entire folder on my end and it builds just fine (Fedora 42).

frank-trampe:
Let's also leave any `libfontforge` changes for a future pull request.

iorsh:
> Let's also leave any `libfontforge` changes for a future pull request.

As far as I could understand, this was the issue with #5610. Do we have `libfontforge` headers here too? I see that the library defines two sets of headers: `FONTFORGE_NOINST_HEADERS` and `FONTFORGE_INST_HEADERS`, but they end up in the same list without any differentiation.

@frank-trampe, do you have specific changes in mind?

iorsh:
Also, it looks like @jtanx explicitly killed `libfontforge` in #3855:
> The only major exclusion compared to autotools is it does not install headers and pkgconfig details. While it is certainly _possible_ to mimic the old behaviour, I'm not seeing much reason to make libfontforge something that can be developed against. As seen in #3823, FontForge isn't well designed to be used in external projects, and we provide no stability guarantees over what is effectively internal interfaces. I doubt that we even export all required headers at the moment.

It was noticed by downstream maintainers - e.g. https://bugs.archlinux.org/task/65877.html, but some of them probably chose to keep maintaining it internally, hence the versioning `libfontforge2`, `libfontforge3`, `libfontforge4` etc.

Ubuntu, for example, slowly pushes out the 2019 version - https://launchpad.net/ubuntu/focal/amd64/libfontforge-dev

frank-trampe:
I somehow missed that at the time. It is certainly a problem that we need to fix at some point.

fontforgeexe/charview.c
```diff
@@ -519,215 +512,6 @@ static void CVDrawBB(CharView *cv, GWindow pixmap, DBounds *bb) {
     GDrawSetDashedLine(pixmap,0,0,0);
 }
 
-/* Sigh. I have to do my own clipping because at large magnifications */
-/*  things can easily exceed 16 bits */
-static int CVSplineOutside(CharView *cv, Spline *spline) {
```
frank-trampe, line 524: @valadaptive, are you satisfied that clipping happens efficiently in the Cairo pathway?

valadaptive:
> Ubuntu, for example, slowly pushes out the 2019 version - https://launchpad.net/ubuntu/focal/amd64/libfontforge-dev

Ubuntu 20.04 will *always* ship the 2019 version though. That won't change no matter what we do here because any Debian-based distro release will (AFAIK) ship the same major versions of packages for its entire lifecycle, backporting fixes from newer versions if enterprise customers pay them to do so. [In 22.04 onwards, `libfontforge-dev` has been completely dropped.](https://launchpad.net/ubuntu/jammy/+package/libfontforge-dev) So it's not as if distros are foregoing newer versions of FontForge in order to maintain support for libfontforge--Ubuntu 20.04 still has libfontforge because it ships old versions of *all* software as a matter of policy.

fontforgeexe/charview.c
valadaptive: Clipping has never happened in the Cairo pathway (on our side at least; Cairo probably does it internally). I haven't changed that in this PR.

frank-trampe:
@iorsh, I was originally proposing to leave out the `struct linelist` changes for a future pull request, but it seems pretty clear that no outside product would use it and that we can make those changes without a separate discussion.

frank-trampe:
Any volunteers to check for `valgrind` regressions?

fontforgeexe/charview.c
frank-trampe: I understand that. My question is whether efficient rendering requires clipping geometry before passing it to Cairo and, if so, whether we do that. I have done very little work on the rendering pipeline, so I do not know this off-hand.

valadaptive: I would assume it doesn't, considering we've never done so. I haven't delved too deeply into 2D vector graphics rendering either, but my understanding is that clipping is moreso a correctness thing than a performance one (e.g. if a shape is displayed clipping the left edge of the screen, you need to know to start the scanline with a fill). Cairo definitely handles that for us.

I may not fully understand the question--if you understand that the Cairo backend has never done clipping on the FontForge side, then why would it start being a performance issue now? Is it just something you noticed and are curious about?

frank-trampe: Yes. Just trying to maximize the value of whatever knowledge you have acquired working on this. It works fine, so there is no reason to dig into this further.

iorsh:
> Any volunteers to check for `valgrind` regressions?

I valground some simple actions, like opening font, opening char view, copy-paste. The logs appear similar, I couldn't see any new issues.

frank-trampe:
I sent the following message to the FontForge users mailing list last Saturday. There are no responses so far. Are there any other checks or adjustments that we need to make before merging?

> Hello, all.
> 
> FontForge has supported GDK for some time, and there are considerable difficulties in maintaining the legacy X11 interface. We are considering removing it and mandating GDK and Cairo. FontForge can still run in X11 through GDK even without the direct X11 interface. Downstream packagers and distributors may need to adjust scripts and other packaging. Discussion is currently on pull request #5612 on GitHub. We would normally offer a lengthy deprecation window, and we will do so if there is a compelling reason, but a lot of the work is already done, and we would like to move quickly if there are no deal-breakers.
> 
> Best wishes,
> 
> Frank

iorsh:
I'm clear with the merge. There is an unrelated CI failure on `master` (https://github.com/fontforge/fontforge/actions/runs/17700511538) due to recent OS upgrade to Microsoft Windows Server 2025, but I don't want it to delay this PR any further.

frank-trampe:
Okay. It sounds like everybody is satisfied. Let's squash and merge.

Issue #5086, opened 2022-08-17
Be forced 101 keyboard layout for inputting


karakufire:
Especially type signs(e.g. _"\), different sign is inputted from I expected.
I'm using Japanese 109 keyboard layout(reference to https://en.wikipedia.org/wiki/Keyboard_layout#Japanese).

I guess this issue could occur to all users who use keyboard kayout other than 101 keyboard layout.
Currently does Fontforge exist config to use favorite keyboard layout? If it's yes could you tell me how to configure it?
Thank you.

jtanx:
What operating system and version of FontForge is this?

ctrlcctrlv:
I think this is related to #5031.

karakufire:
OS: Windows 10
Fontforge version: 20220308

ctrlcctrlv:
Well now I'm even more sure this is highly related to #5031. :-)

Either GDK itself needs patched or how we're hooking into it needs patched.

Phroneris:
Same issue here.

- Windows 10 64-bit
- FontForge 20230101
- Japanese 106/109 Keyboard layout

Pressing the top (number) row from the left with Shift should result in `!"#$%&'()=~|` on my keyboard.
However, in FontForge text fields such as file names, it results in `!@#$%^&*()_"|`.
(Each string is different length, but not my typo; Shift+0 has no associated character in 106/109 layout.)

It's inconvenient when entering font name references like `^View.DefaultFont` in the Appearance Editor.
The `^` turns into `'`.

PerryWerneck:
This seens to be a gdk problem. Does anyone open an issue there?

Issue #5015, opened 2022-05-19
Element->Generate Fonts should turn to Files->Generate Fonts


Firestar-Reimu:
https://fontforge.org/docs/faq.html#faq-outline-conversion

I did not find `generate fonts` under `element` and find that it is under `files`

ctrlcctrlv:
- Closing as duplicate of #3638 and [Project №7](https://github.com/fontforge/fontforge/projects/7). Menu restructuring is a much wider issue than this.

PR #5622, opened 2025-09-19
Apply licenses to OtherSubrs

bgermann:
Separate out the proprietary Adobe code and add its license.

Adobe has licensed the OtherSubrs code under Apache 2 license in afdko. Apply the license to fontforge's copy.

afdko's t1write_flexothers.h covers othersubrs0, othersubrs1, othersubrs2, and othersubrs3.

afdko's t1write_gcothers.h covers othersubrs13 and cid_othersubrs.

Link: https://github.com/adobe-type-tools/afdko/commit/780164a43d86a782be479ad1e54cc44363ca082b

### Type of change
- **Non-breaking change**


iorsh:
The Apache 2.0 license is from 2004, while the relevant code is from 1990 or earlier. @bgermann, do you know what was the previous license? I'm not fully convinced that we must follow the re-licensing for the otherwise intact code.

bgermann:
The previous license is available with https://adobe-type-tools.github.io/font-tech-notes/pdfs/5015.Type1_Supp.pdf

see the Note on page 35 (cannot copy the text)

bgermann:
The previous license is proprietary and does not allow modification and allows use only in Type 1 font programs. 

The Apache license allows much more.

bgermann:
This solves a long-standing [Debian bug](https://bugs.debian.org/665334) at least partly.

skef:
I know that this stuff is handled sloppily in almost every project, open source and commercial, and that no one cares, but the copyright date change is questionable at best. Presumably, the copyright holder can choose to license under new terms, which is what is being claimed in the PR, and would be a power separate from when the copyright stems from. Maybe refer to the license but keep the dates as they were? 

Note: I am not offering legal advice, nor am making any claims about Adobe's views or actions in this situation. I'm taking the PR at face value in making this observation.

bgermann:
The Apache License requires to keep the original copyright notice, so changing it to the previous years is a license violation (even though this would be the accurate thing to do).

frank-trampe:
Does this patch result from copying and pasting a few lines or from copying the current Adobe file verbatim over the old copy that we have? The latter is fine, the former is questionable.

bgermann:
I have verified that the code matches except for whitespace. As the Apache License grants you the right to change the code, this is acceptable from my point of view.

skef:
OK, that makes sense.

bgermann:
One more note: I have tried to find the source for mmfindfont but failed to do so. It was added alongside makeblendedfont, so that procedure's license might apply as well.

jayaddison:
@bgermann FYI - there is a one-statement difference between the ADFKO open source subroutines and the originally-published T1 type specifications.  Adobe themselves have assured me it doesn't make a difference -- but I want to mention it anyway for your awareness.  See #5623 for some more details and references (GitHub's web UI should already display that issue here since it references this one, but I'm adding a comment reply also).

skef:
> One more note: I have tried to find the source for mmfindfont but failed to do so. It was added alongside makeblendedfont, so that procedure's license might apply as well.

Do you mean that you tried to find it in an Apache-licensed form?

bgermann:
I tried to find it generally.

skef:
> I tried to find it generally.

That is a pickle. Perhaps it was in #5086 but was no longer considered necessary by the time #5015 was released.

bgermann:
Hey. Any chance this will be integrated?

JoesCat:
Hi @bgermann,
I haven't been paying much attention to FontForge for a while, but this last forced-push caught my attention ATM ;-P

Okay... Reading through all this, I think we can clarify a few things here. You probably already know this, but I'm just writing it out here for other readers tuning-in that aren't aware:

1. Apache 2.0 is compatible with GPL3 (not GPL2), example, it can be within a GPL3 project. Example: GPL3{code,code,Apache2.0_code,code,code}
2. GPL3 is not compatible with Apache 2.0, example, it cannot be within an Apache 2.0 project (Apache license allows for closing source, GPL requires source be always available). Example: Apache2.0{code,code,~GPLn_code~,code,code}

In summary, FontForge (as a GPL3+ project) can accept the code ([More info](https://www.google.com/search?client=firefox-b-e&channel=entpr&q=is+Apache+2.0+compatible+with+GPL3%3F)).

Next....digging into the code, if in doubt, my usual course of action tends to take [20120731](https://github.com/fontforge/fontforge/tree/v20120731-b) tag-point as a reference (pretty much 99% GW code) and comparing what was [then](https://github.com/fontforge/fontforge/blob/v20120731-b/fontforge/othersubrs.c), vs [now](https://github.com/fontforge/fontforge/blame/master/fontforge/othersubrs.c). Looking at a diff between both, it has not changed much between the past and now, and in light of the discussion here, I really do think the recently added BSD3 header at the top only seems to muddle things up here. I think it needs to be removed. Removing the clause, basically reverts the file back to the original 20120731-b thoughts (intended direction), and back to the [debian problem mentioned above](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=665334). I think GW had a pretty good grasp of the copyright situation back then (example, keeping libspiro optional to try maintain some amount of BSD3 direction), and I think this is one of those problem situations, where you also notice that GW did not label this file with a BSD3 header. A major reason with Fontforge going towards GPL3+ was to make it easier to do future merges like this.

I like the direction you're suggesting...

> The previous license is proprietary and does not allow modification and allows use only in Type 1 font programs.
>
> The Apache license allows much more.
>
> This solves a long-standing [Debian bug](https://bugs.debian.org/665334) at least partly.

Seems to resolve 3+ problems.

We can pretty much say, this code is Adobe based in terms of authorship, the original circumstances back in the past allowed for using as-was without edits, the new circumstances is Apache license, and gives more flexibility, FontForge has a BSD past, but as a project is GPL3, so this, plus improvements, can be absorbed in.

Would you be okay with doing this?
- Pulling-out the BSD header (but to meet the terms of BSD3, add all the authors seen in git blames).
- Putting in the Apache 2.0 license as you intend (I would suggest using the 20120731-b code as a reference as to where you think the best place to put the license belong - probably at the top). [Would it make sense to add GPL3+ at the top](https://www.google.com/search?client=firefox-b-e&channel=entpr&q=can+be+have+a+GPL3+license+and+subset+apache2+license+in+the+same+file%3F), or would it make more sense to add a /*block*/ below saying this is part of Fontforge, which is a GPL3+ project?
- Leave othersubrs.c in one piece without stripping-out part of it into another file (multimastersubrs.c), this seems unnecessary, plus FontForge benefits keeping the existing [git blame](https://github.com/fontforge/fontforge/blame/master/fontforge/othersubrs.c).
- Keep erode, don't remove it. If you grep "erode" it shows up multiple times in file othersubstr.c
- To avoid further forced-pushes, just add all the fixes as a 4th patch, and let it get squash-merged
- Let the other participants in this thread provide feedback if they have any before starting this extra work yourself, if none, then let's go with this.

bgermann:
I have modified the PR to keep erode even though that subroutine is commented out.

I would very much like to move the proprietary code to a different file because then I can easily repackage the Debian source package to exclude it. The Debian Free Software Guidelines require a source package not to contain any proprietary code.

About relicensing the BSD-licensed code: From what I can see George Williams and maybe Julien Nabet would have to give their consent to relicense the non-Adobe code. The other contributions look like non-copyrightable changes to me. I would apply the Apache license myself if George Williams agrees. But as contacting him might take a long time, I would suggest making this a different PR. If you want to make the change without requesting consent you can apply it yourself after this is merged.

I do not think it makes much sense to add an additional GPL-3+ text anywhere in the file. That usually complicates things. Either way - the Adobe copyright notice in the copyright[] array must be kept so that it lands in the generated fonts and those fonts are subsequently compliant with the license.

bgermann:
By the way: the separated-out code was never touched in git history after adding it so you do not lose an easy git blame information. Only the comment for the erode subroutine was modified in 649bfd1e098cee48e6931e0965b93d31fdd40489 and 0e5717aaa60783ad05ae3d9ac8bfdd4b007f50b3

JoesCat:
@bgermann, @jayaddison provides another link in #5623, and following the breakcrumbs [message 5, the search for 1183615869](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1029555) which then references [afdko t1write_hintothers.h](https://github.com/adobe-type-tools/afdko/blob/develop/c/shared/t1write/t1write_hintothers.h), which then seems to match FontForge's othersubrs.c char *othersubrs3[].

Reading further on [bug 1029555](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1029555)  about the lintian bug trigger, do you think it could be an improvement to also add a comment above *othersubrs3[] referring to afdko's t1write_hintothers.h? Comments add no extra code, but they are a time saver for future use in avoiding a lot of hunting and searching for anyone following later in time.

JoesCat:
From reading and following these links... Adding this for reference...
`cd /usr/; find . -name *.pf[a,b] -exec t1disasm {} \;  | grep " Adobe Systems Inc" | wc`
(probably should include m, as in pf[a,b,m])

also in terms of makeblended.... afdko
```
t1disasm tests/tx_data/input/zx.pfb |grep "makeblendedfont"
t1disasm tests/tx_data/input/zy.pfb |grep "makeblendedfont"
```

bgermann:
Okay, I have added the sources for the arrays.

JoesCat:
Reading through some of the links, this has been an issue covering over several years - Thanks for the effort and work.

PR #5626, opened 2025-09-26
Fix hotkeys in BitmapView

### LINKED ISSUE #5150: No Hotkeys/Shortcuts in BitmapView

In contrast to Charview, FontView and MetricsView there are not hotkeys available in BitmapView (for no obvious reason). After tracking down some of the past PRs, I found that #3637 implemented hotkeys for BitmapView. @skef tried to revive the PR in #3638.
Charview, FontView and MetricsView seem to have gotten a menu overhaul, but BitmapView was left out for some reason (still old inline hotkey definition and no definitions in `share/default`).

I would love to see hotkeys in BitmapView. Let me know if there are any workarounds or any help that is needed to make this happen. (I have not fully understood how the current hotkey system works)

Tynach:
It turns out that in order to get BitmapView's hotkeys working, the top-level window (bv->gw) needs its 'Type Name' set (by calling `GDrawSetWindowTypeName(bv->gw, "BitmapView");` after it's created in fontforgeexe/bitmapview.c), and the hotkeys added to share/default.

None of the other views (such as CharView, FontView, and MetricsView) even set the hotkeys within the source code itself anymore, and doing so doesn't actually work it seems, so I could also add a commit to set all of those to 'No Shortcut' within the C code.. But in the meantime, I wanted to have the minimal number of changes to the code, so that this fix could be included as easily as possible.

I also noticed that the share/default file is kinda.. Haphazardly organized. It's not clear whether these should be added to the start or the end, since mostly (but not entirely) the views are organized alphabetically (CharView, FontView, MetricsView), but there's some CharView things added to the very top and very bottom, seemingly out of place.

To avoid messing with where existing hotkeys exist within the file I added the new BitmapView hotkeys to the bottom, but I can move them to the top if that's preferred.

### Type of change
- **Bug fix**
- Fixes #5150

### DIFF
--- a/fontforgeexe/bitmapview.c
+++ b/fontforgeexe/bitmapview.c
@@ -2434,6 +2434,7 @@ BitmapView *BitmapViewCreate(BDFChar *bc, BDFFont *bdf, FontView *fv, int enc) {
     wattrs.event_masks = -1;
     wattrs.cursor = ( bc->refs == NULL ) ? ct_pencil : ct_pointer;
     bv->v = GWidgetCreateSubWindow(gw,&pos,v_e_h,bv,&wattrs);
+    GDrawSetWindowTypeName(bv->v, "BitmapView");
 
     bv->height = pos.height; bv->width = pos.width;
     bv->b1_tool = ( bc->refs == NULL ) ? bvt_pencil : bvt_pointer; bv->cb1_tool = bvt_pointer;
--- a/fontforgeexe/bitmapview.c
+++ b/fontforgeexe/bitmapview.c
@@ -2381,6 +2381,7 @@ BitmapView *BitmapViewCreate(BDFChar *bc, BDFFont *bdf, FontView *fv, int enc) {
 
     bv->gw = gw = GDrawCreateTopWindow(NULL,&pos,bv_e_h,bv,&wattrs);
     free( (unichar_t *) wattrs.icon_title );
+    GDrawSetWindowTypeName(bv->gw, "BitmapView");
 
     GDrawGetSize(GDrawGetRoot(screen_display),&zoom);
     zoom.x = BVPalettesWidth(); zoom.width -= zoom.x-10;
--- a/share/default
+++ b/share/default
@@ -616,3 +616,75 @@ CharView.Menu.View.Show.Tab.Tab4:Alt+5
 CharView.Menu.View.Show.Tab.Tab5:Alt+6
 
 +CharView._ImmediateKeys.TogglePreview:`
+
+
+
+BitmapView.Menu.File.New:Ctl+N
+BitmapView.Menu.File.Open:Ctl+O
+BitmapView.Menu.File.Close:Ctl+Shft+Q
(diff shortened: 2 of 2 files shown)


fontforgeexe/bitmapview.c
```diff
@@ -2381,6 +2381,7 @@ BitmapView *BitmapViewCreate(BDFChar *bc, BDFFont *bdf, FontView *fv, int enc) {
 
     bv->gw = gw = GDrawCreateTopWindow(NULL,&pos,bv_e_h,bv,&wattrs);
     free( (unichar_t *) wattrs.icon_title );
+    GDrawSetWindowTypeName(bv->gw, "BitmapView");
```
iorsh, line 2384: It looks like other views also call `GDrawSetWindowTypeName()` for the drawing subwindow `v`, like `GDrawSetWindowTypeName(cv->v, "CharView");`

In Linux the code seems to work without it, but there could theoretically be focus-related issues on other platforms.

Tynach: I had initially included both, but after some testing I removed the one going to `bv->v` since it didn't seem necessary. I'll admit, I haven't tested on any other platforms.. I don't have easy access to a Windows computer right now (haven't booted into Windows in years and I know it'll try to install tons and tons of updates and take all day before I can actually use it), let alone a Windows computer with the necessary libraries and compiling tools necessary.

I can add it in a second commit though anyway; it didn't seem to cause any problems to be there, and if there's the possibility that *not* having it can cause problems on other platforms then it's probably best to include it.

iorsh: Ok, then add the call with `bv->v` please, just to be on the safe side.

iorsh (approved):
Looks good to me, tested on my machine (Ubuntu 24.04) and it works.

Tynach:
@iorsh I've also considered fixing up the menus' `H()` entries so that in the C code they all say 'No Shortcut', like the other views do. This would include me adding `H()` to the lines that lead to submenus.

I actually wasn't expecting submenus to work without doing the latter, but they do, so I didn't touch them. And yeah, I know that none of the submenu items have a hotkey assigned to them by default, but I tried it anyway by adding one to my local user-level 'hotkeys' file, and it did actually work.

I don't think the code for that would depend on the platform, but I don't know for sure? At any rate, I think I can do that if you think it'd be a good idea to do that at the same time; otherwise it could be a separate pull request.

Tynach:
@iorsh I don't have authorization to push things to master, so it tells me merging is blocked.

iorsh:
> @iorsh I've also considered fixing up the menus' `H()` entries so that in the C code they all say 'No Shortcut', like the other views do. This would include me adding `H()` to the lines that lead to submenus.

If custom submenu hotkeys already work, we should probably leave the code as is. I think these `H_()` clauses are outdated, and wouldn't be considered anyway.

iorsh:
> @iorsh I don't have authorization to push things to master, so it tells me merging is blocked.

I'll merge it in a few days if there are no further comments from other developers.

PR #5637, opened 2025-10-06
Fix iconv(3) arguments.

0-wiz-0:
iconv(3) expects an "iconv_t" as first argument, not an "iconv_t *".

### Type of change

- **Bug fix**


iorsh (approved):

iorsh:
@0-wiz-0
Is this critical to be merged before the release? The fix looks safe, but I prefer to avoid any destabilization right now.

0-wiz-0:
I sent this PR because it broke the build on NetBSD-current with gcc 14 (it was ok with gcc 12). It will be some time before this ends up in a NetBSD release, so no hurry from my side; it might however break in other environments with gcc 14.

PR #5404, opened 2024-04-18
Use PyConfig API on Python 3.8

vstinner:
The Py_SetProgramName() function used by fontforge is deprecated and was removed in Python 3.13 alpha 1.

Replace the deprecated function with the PyConfig API on Python 3.8. If the Python initialization fails, log an error to stderr and call exit(1).

### Type of change

- **Bug fix**
- **New feature**
- **Breaking change**
- **Non-breaking change**

PR #5643, opened 2025-10-10
Use importlib in plugin manager

### LINKED ISSUE #5502: Fontforge hardcodes dependency to deprecated Python module `pkg_resources`
As seen here in `fontforge/plugin.c`:

https://github.com/fontforge/fontforge/blob/9af60edefc61d1f9244601802067c33fd221db69/fontforge/plugin.c#L293-L295

Currently fontforge has a hard dependency to `pkg_resources` in Python.

As of 2024, according to https://setuptools.pypa.io/en/latest/pkg_resources.html :

> Use of pkg_resources is deprecated in favor of [importlib.resources](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources), [importlib.metadata](https://docs.python.org/3.11/library/importlib.metadata.html#module-importlib.metadata) and their backports ([importlib_resources](https://pypi.org/project/importlib_resources), [importlib_metadata](https://pypi.org/project/importlib_metadata)). Some useful APIs are also provided by [packaging](https://pypi.org/project/packaging) (e.g. requirements and version parsing). Users should refrain from new usage of pkg_resources and should work to port to importlib-based solutions.

It would be great if fontforge could migrate away from the usage of `pkg_resources`.

* * *

Downstream Debian bug report: https://bugs.debian.org/1083402

### LINKED ISSUE #5671: Missing pkg_resources on macOS
Setup: M1 Apple Silicon Mac, macOS Tahoe 26.1, no other Python installed than the one coming with Xcode 26.1.

Upon startup of the new October 2025 release, the splash screen appears together with a warning "Core python package 'pkg_resources' not found: Cannot discover plugins". After a couple of seconds the splash screen vanishes, only the warning window remains and clicking anywhere just produces beeps. The only way out is Command-Option-Escape to force-quit the app.

To be sure, I checked with the January 2023 release: no warning, the Open Font dialog appears and everything works OK.

I see the same warning appeared years ago [https://github.com/fontforge/fontforge/issues/4761](https://github.com/fontforge/fontforge/issues/4761). Comparing the two FontForge versions:

- January 2023
```
% ls -1 FontForge.app/Contents/Resources/opt/local/lib/python3.10/site-packages
_distutils_hack
distutils-precedence.pth
fontforge.so
libsvn
pip
pip-22.2.2.dist-info
pkg_resources
psMat.so
setuptools
setuptools-65.4.1.dist-info
svn
wheel
wheel-0.37.1.dist-info
```
- October 2025
```
% ls -1 FontForge.app/Contents/Resources/opt/local/lib/python3.13/site-packages 
fontforge.so
pip
pip-25.2.dist-info
psMat.so
wheel
wheel-0.45.1.dist-info
```

So a few things are missing, including setuptools and pkg_resources which seem to be involved according to that previous issue. I cannot say more, as I don't really know Python.

iorsh:
Replace the deprecated module `pkg_resources` with `importlib`.

Fixes #5502, fixes #5671

### DIFF
--- a/fontforgeexe/pluginui.c
+++ b/fontforgeexe/pluginui.c
@@ -522,6 +522,7 @@ static int PLUG_PluginOp(GGadget *g, GEvent *e) {
     } else if (cid == CID_Revert) {
         FigurePluginList(d);
     }
+    GGadgetRedraw(list);
     return true;
 }
 


iorsh:
@skef, I did some sanity check with one of your plugins `fontforge_featuremunge_skef`. It detects the installed plugin and does some stuff with the features. Please take a look at the code, I'm not fully confident about it.

skef:
What's our minimum python version for this upcoming release?

iorsh:
> What's our minimum python version for this upcoming release?

Python >= 3.8 (see #5404)

skef:
I thought we were still supporting at least one release older than 3.10. If that's the case, then either we should be trying to ensure that https://pypi.org/project/backports.entry-points-selectable/ is installed (which is likely tricky), or the pkg_resources code you removed should be wrapped up in a separate function that serves as the fallback to `GetPluginEntryPoints` (instead of just returning NULL). 

I'll have a closer look at the code soon but from a quick review it looks like it's doing the desired thing.

iorsh:
I didn't quite understand you. Python 3.8 is supported, and `importlib.metadata` should be available - see https://docs.python.org/3/library/importlib.metadata.html. `backports.entry-points-selectable` tries to provide it for versions under 3.6, but we are already using PyConfig C API, which doesn't exist before 3.8.

Am I missing something?

skef:
I'm looking at the notes directly *above* this link (there doesn't seem to be a direct link): https://docs.python.org/3/library/importlib.metadata.html#distribution-metadata . Seems like some of the interface the new code relies on was introduced in Python 3.10.

skef:
The python folks have had a bunch of this sort of problem in recent years. They've wanted to replace a lot of the packaging infrastructure but the replacements start out half-baked, get changed in response to feedback, and then the older stuff gets deprecated and retired too early.

fontforge/plugin.c
```diff
@@ -357,9 +400,9 @@ static bool DiscoverPlugins(int do_import) {
...
         PyObject *dist = PyObject_GetAttrString(entrypoint, "dist");
-        if (dist != NULL) {
-            tmp = PyObject_GetAttrString(dist, "PKG_INFO");
-            tmp2 = PyObject_CallMethodObjArgs(dist, getmetastr, tmp, NULL);
+        if (dist != NULL && PyObject_HasAttrString(dist, "metadata")) {
+            tmp = PyObject_GetAttrString(dist, "metadata");
+            tmp2 = PyObject_GetIter(tmp);
             Py_DECREF(tmp);
             if (PyIter_Check(tmp2)) {
                 while ((str = PyIter_Next(tmp2))) {
```
skef: Looks to me like this stuff will need to change too. Here's the pattern in a script:

```
#!/usr/bin/env python

from importlib.metadata import entry_points

eps = entry_points()

blah = eps.select(group='fontforge_plugin')

it = iter(blah)

for i in it:
    m = getattr(getattr(i, 'dist'), 'metadata')
    jt = iter(m)
    for j in jt:
        print(j, m[j])
```
I think it amounts to something like:

```
if (strncmp(metaline, "Home-page: ", 11) == 0) {
    if (pe->package_url != NULL) {
        free(pe->package_url);
    }
    pe->package_url = copy(metaline + 11);```
```

becoming

```
if (strcmp(metaline, "Home-page") == 0) {
    if (pe->package_url != NULL) {
        free(pe->package_url);
    }
    pe->package_url = copy(PyUnicode_AsUTF8(PyDict_GetAttr(dist, str)));
```

But presumably with more error checking. (I'm just writing this out by hand, no compilation or testing, just for reference). 

Unfortunately, it also appears that Home-page is now deprecated in favor of the Project URL list, which consists of key value pairs. These are the "well-known" keys: https://packaging.python.org/en/latest/specifications/well-known-project-urls/#well-known-labels

What a pain.

iorsh: I'm trying to run it on all Python versions 3.8, 3.9, 3.10, 3.11, 3.12, and there are still some cryptic bugs in 3.8-3.9. I need to gather some courage to attack it once again. Hopefully soon.

skef:
If we could figure out how to add tests for this stuff somehow, that would be nifty. I didn't figure out a clean way. Maybe a separate directory where you start by making an env and installing a package and then testing against that?

iorsh:
Test sequence:

1. `python3.<?> -m pip install fontforge_featuremunge_skef-0.0.1-py3-none-any.whl --break-system-packages`
2. Run FontfForge
3. Observe "Plugin Configuration" dialog with entry "FeatureMunge (New)"
4. Select the plugin and click "Enable -> Yes", click "OK" to close plugin dialog
5. Open any font with GPOS/GSUB lookups
6. Under "Tools" observe "Feature files" with a submenu
7. Apply any submenu item and check if it worked.
8. Close FontForge
9. Open FontForge again, check that plugin menu is available and works.
10. Go to "File --> Configure Plugins" and observe "FeatureMunge (On)"
11. Select the plugin and click "Disable"
12. Click "More Info" (?)
13. Close "Plugin Configuration" dialog and close FontForge
14. Open FontForge with any font, observe that under "Tools" no "Feature files" submenu is available
15. `python3.<?> -m pip uninstall fontforge_featuremunge_skef-0.0.1-py3-none-any.whl --break-system-packages`
16. Open FontForge with any font
17. Go to "File --> Configure Plugins" and observe "FeatureMunge (Off) [Not Found]" in red
18. Select the plugin and click "Delete"
19. Close the dialog and FontForge

iorsh:
Tested with Python 3.8, 3.9, 3.10, 3.11, 3.12, everything works. @skef, could you please review?

iorsh:
> If we could figure out how to add tests for this stuff somehow, that would be nifty. I didn't figure out a clean way. Maybe a separate directory where you start by making an env and installing a package and then testing against that?

I thought about it a bit, but things seem too complicated for now. We need to think how to make plugin capability more visible, to encourage participation.

skef (approved):
It's hard for me to trace all of this through the various versions in my head, but it takes into account all the various issues brought up with the different versions, and together with your testing I think this is in good shape. 

The Python folks really should have thought about all of this more from the start.

PR #5646, opened 2025-10-10
Python documentation errors

### LINKED ISSUE #5385: Errors in description of the math constants
Looks like one of descriptions of the math constants lack correctness in the "fontforge/mathconstants.c"

https://github.com/fontforge/fontforge/blob/master/fontforge/mathconstants.c#L67
The description referred above reads: "Minimum distance between the baseline of an upper limit and the **bottom** of the base operator.", when it should be: "Minimum distance between the baseline of an upper limit and the **top** of the base operator."

In any way the same function is described on https://learn.microsoft.com/en-us/typography/opentype/spec/math
as "Minimum distance between baseline of upper limit and **(ink) top** of the base operator."

### LINKED ISSUE #5471: gsub_reverseccchain misdocumented as gsub_revesechain
While I spelled 'reverse' correctly despite the documentation, I didn't realise I needed an extra `c` before `chain`, so this mistake really threw me.

https://github.com/fontforge/fontforge/blob/705433c295326df4a16e5491df917d0b48427242/doc/sphinx/scripting/python/fontforge.rst?plain=1#L4057

### LINKED ISSUE #5550: genericGlyphChange vMap component order
Cracking this problem took me the better part of three days and all because the documentation is incorrect.  This is what the documentation says about the vMap parameter to font.genericGlyphChange() (and presumably its glyph analogue):

If vCounterType is omitted, or is the string “mapped”, then certain zones on the glyph may be placed at new (or the same) locations – similar to BlueValues. So you can specify a zone for the baseline, one for the x-height and another for the top of capitals and ascenders (and perhaps a fourth for descenders). Each such zone is specified by the vMap argument which is a tuple of 3-tuples, each 3-tuple specifying a zone with: **Original location, original width, and final location.**

Note
No default value is provided for this argument you must figure out all the values yourself.

The emphasis is mine.  This order corresponds with the UI order under Element -> Style -> Change Glyph.  

However, to get genericGlyphChange() to do what you want, the second and third elements of the tuple need to be reversed: **Original location, final location, and original width.**

I have only tested this on the glyph object version, but I presume it is the same for the font objects with selection since I am pretty sure you just wrapped the same code.  But something needs to be changed: either the order of the arguments or the documentation itself.

iorsh:
Fixes #5385, fixes #5550, fixes #5471

### DIFF
--- a/doc/sphinx/scripting/python/fontforge.rst
+++ b/doc/sphinx/scripting/python/fontforge.rst
@@ -4114,7 +4114,7 @@ This type may not be pickled.
    * ``gsub_ligature``
    * ``gsub_context``
    * ``gsub_contextchain``
-   * ``gsub_revesechain``
+   * ``gsub_reversecchain``
    * ``morx_indic``
    * ``morx_context``
    * ``morx_insert``


chanicpanic (approved):

fontforge/mathconstants.c
```diff
@@ -64,7 +64,7 @@ struct math_constants_descriptor math_constants_descriptor[] = {
     MCDD(N_("SuperscriptBottomMaxWithSubscript:"),SuperscriptBottomMaxWithSubscript,SuperscriptBottomMaxWithSubscript_adjust,N_("The maximum level to which the (ink) bottom\nof superscript can be pushed to increase the\ngap between superscript and subscript, before\nsubscript starts being moved down."),0),
     MCDD(N_("SpaceAfterScript:"),SpaceAfterScript,SpaceAfterScript_adjust,N_("Extra white space to be added after each\nsub/superscript."),0),
     MCDD(N_("UpperLimitGapMin:"),UpperLimitGapMin,UpperLimitGapMin_adjust,N_("Minimum gap between the bottom of the\nupper limit, and the top of the base operator."),1),
-    MCDD(N_("UpperLimitBaselineRiseMin:"),UpperLimitBaselineRiseMin,UpperLimitBaselineRiseMin_adjust,N_("Minimum distance between the baseline of an upper\nlimit and the bottom of the base operator."),0),
+    MCDD(N_("UpperLimitBaselineRiseMin:"),UpperLimitBaselineRiseMin,UpperLimitBaselineRiseMin_adjust,N_("Minimum distance between the baseline of an upper\nlimit and the top of the base operator."),0),
```
frank-trampe, line 67: Verified per [specification](https://learn.microsoft.com/en-us/typography/opentype/spec/math).

doc/sphinx/scripting/python/fontforge.rst
```diff
@@ -4654,7 +4654,7 @@ This type may not be pickled.
    x-height and another for the top of capitals and ascenders (and perhaps a
    fourth for descenders). Each such zone is specified by the ``vMap`` argument
    which is a tuple of 3-tuples, each 3-tuple specifying a zone with: Original
-   location, original width, and final location.
+   location, final location, and original width.
```
frank-trampe, line 4657: Verified per [python.c](https://github.com/fontforge/fontforge/blob/afa46903e531080591835d0a0a753d7efb5d27cb/fontforge/python.c#L18920C19-L18924C27) as highlighted by @iorsh.

frank-trampe (approved):
This looks good to me.

PR #5650, opened 2025-10-15
Differentiate Supplementary PUA planes by name

### LINKED ISSUE #5560: Font Info Unicode Ranges: regression in Fontforge 2023-01-01
Font Info in FontForge 20230101 only shows some of the Unicode ranges containing glyphs in the font. FontForge 20190801 shows all the ranges that libuninameslist knows about. The font tested is IndonesianTexts-Regular.ttf from
https://drive.google.com/file/d/1_BBowgacuDHpXpkBwuFpY63vwROxXlIN/view

Tested in Ubuntu 20.04 & 24.04 and Windows 10 & 11.

FontForge 20190801:

Image

FontForge 20230101:

Image

iorsh:
Better distinction between **"Supplementary Private Use Area-A/B"** Unicode ranges and **"Supplementary Private Use Area Plane A/B"** Unicode planes

Fixes #5560

### DIFF
--- a/Unicode/makeutype.py
+++ b/Unicode/makeutype.py
@@ -244,8 +244,8 @@ class UnicodeData:
             (0x20000, 0x2FFFD, "Supplementary Ideographic Plane"),
             (0x30000, 0x3FFFD, "Tertiary Ideographic Plane"),
             (0xE0000, 0xEFFFD, "Supplementary Special-purpose Plane"),
-            (0xF0000, 0xFFFFD, "Supplementary Private Use Area Plane A"),
-            (0x100000, 0x10FFFD, "Supplementary Private Use Area Plane B"),
+            (0xF0000, 0xFFFFD, "Plane 15 (Supplementary Private Use Area-A)"),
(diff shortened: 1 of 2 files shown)


dscorbett:
The blocks are named “Supplementary Private Use Area-A” and “Supplementary Private Use Area-B” though.

iorsh:
> The blocks are named “Supplementary Private Use Area-A” and “Supplementary Private Use Area-B” though.

Yes, the block names are intact - see a few lines above. I'm changing only the names of `unicode_planes[]`, not of `unicode_blocks[]`, which are autogenerated.

dscorbett:
Those planes don’t have names, but if you want to name them distinctly from their coterminous blocks, Unicode refers to them as “Plane 15” and “Plane 16”. “Plane A” and “Plane B” aren’t anything.

iorsh:
@dscorbett , does this look better now, or you think "Plane 15/16" without elaboration would be preferable?

dscorbett:
I think this is good.

PR #5657, opened 2025-10-21
Document missing python module members

chanicpanic:
This PR documents some attributes, functions, and methods that are provided by the `fontforge` Python module but are missing from the module documentation.

For the most part, the descriptions are the docstrings from `python.c`.

### Type of change
- **Non-breaking change**


iorsh:
Generally looks good. My only concern is that some of the API points may have been omitted intentionally, being broken, unstable or otherwise unfit. @skef, what do you think?

PR #5663, opened 2025-10-26
Fix UI font on Windows for CJK locales

### LINKED ISSUE #4983: UI font problem (portions) for Chinese UI
- [/] Screenshot  
image
Notice the missing glyphs on the side bar and tabs. Also notice that the missing character in the tab (U+6807 `标`) is actually displayed in the settings section. (Top bar should be `上标/下标` for superscript/subscript)

- [/] The FontForge version and the operating system you're using
FontForge 2022-03-08, OS Windows 10 21H2 Build 19044.1586

- [/] The behavior you expect to see, and the actual behavior
The UI should display the words correctly, instead part of the display shows missisng character box.

- [/] Steps to reproduce the behavior 
1. Change locale to `zh_CN` in `fontforge.bat`.
2. Go into FontForge and open Font Info.

### LINKED ISSUE #5660: [BUG] Many characters cannot be displayed in v2025
The new version has a bug where text cannot be displayed properly.

| [FontForge October 2025 Release](https://github.com/fontforge/fontforge/releases/tag/20251009) | [FontForge January 2023 Release](https://github.com/fontforge/fontforge/releases/tag/20230101) |
| - | - |
| Image | Image |

**OS:** Win 10 x64 ver 10.0.19045.6456
**Language:** Simplified Chinese (zh-hans-CN)

iorsh:
The "system-ui" Pango font alias doesn't support CJK scripts on Windows - see Pango source https://gitlab.gnome.org/GNOME/pango/-/blob/main/pango/pangowin32-fontmap.c#L374

This PR detects dynamically the locale used by gettext() and overrides the generic UI font with proper Windows-based alternative.

Fixes #4983, fixes #5660.


skef:
I take there's some reason we can't just put these at the tail end of the Windows UI font fallback list?

iorsh:
> I take there's some reason we can't just put these at the tail end of the Windows UI font fallback list?

There would be a problem due to the [Han unification](https://en.wikipedia.org/wiki/Han_unification). If we list the fonts in certain order, e.g. "Microsoft YaHei UI,Microsoft JhengHei UI,Yu Gothic UI" (Simplified Chinese, Traditional Chinese, Japanese), then a Japanese user would have Unihan characters selected from "Microsoft YaHei UI", and other characters from "Yu Gothic UI". This would create a mess, since the same character (Unicode point) may have different incompatible shapes in different scripts.

frank-trampe:
This is fine with me, but, per my discussion with @iorsh, this is a work-around for what I consider to be a Pango bug. The alias "system-ui" ought to refer to the system's configured user interface font, not the first font in an arbitrary list that is installed.

skef:
Let me take a quick look at this in more depth.

skef:
I think this is a basically sensible approach. The locale string retrieval stuff makes me a little queasy but it will probably work.

In terms of nitpicking, the one concern I might raise is that we're adding magical substitution logic onto a font string that has a specific meaning, and therefore creating a situation that can't be worked around. One could imagine an alternative where we add `windows-cjk-workaround` (or whatever) to the head of Windows' `FONTFORGE_UI_FAMILIES` in `BuildUtils.cmake` and then do this substitution on that instead, so that someone who wants to avoid this logic (perhaps after Pango figures their shit out) can do so. As long as fallback is implemented correctly that would presumably have the same effect.

skef (approved):
Assuming this passes basic interactive testing, I think this is great

tuffnatty:
Things I don't like in this PR:
1. `char *gettext_locale()` prototype in intl.h does specify a return type, but the returned value is never used;
2. The same prototype does not specify any argument type or argument count, so the compiler can't control them in the places where the function is called;
3. The function copies an unverified number of bytes from the user-controlled environment to a stack-allocated buffer of fixed size;
4. `fix_CJK_UI_font()` then uses pointer gymnastics and strcpy(3) to construct another string, while the Unicode lib has `smprintf()` which seems to be a local implementation of `asprintf()` and that's what I'd use here for a bit cleaner and safer code.

Probably this is nitpicking, regarding the amount of legacy code in FontForge, but IMHO, we should at least avoid introducing new pieces of such code.

iorsh:
This is not a nitpicking, but a perfectly valid list of comments. The PR has already been merged, so I probably won't get to fix these issues any time soon. Please feel free to open a follow-up PR for these, or participate in future code reviews. I generally tend to keep PRs open for at least a few days, usually more, so that all the interested parties can comment.

PR #5667, opened 2025-10-29
Fix glyph name and Unicode value deduction for SVG fonts

### LINKED ISSUE #5349: Glyphs disappear when I load a file.
### Setup

I made an SVG font using Inkscape's font editor and saved it as [an SVG file](https://www.dropbox.com/scl/fi/u78tp8zmpumu022mgtcdl/abeceda-font.svg?rlkey=bsnoxqrd414ujexuj272y33ch&dl=0). I opened it in FontForge, and it had all glyphs. I saved it as [a FontForge file](https://www.dropbox.com/scl/fi/u77esafbxqamjymg1v69f/Greek-Serif.sfd?rlkey=bfu54njj2k7mtbnxwcpm7enah&dl=0), closed FontForge, opened FontForge and loaded the new file.

(The links lead to files with which that happens. I did not do anything else to the files except the described loading and saving. The files are on Dropbox, and, if it asks you to log in, you can close the pop-up and not log in.)

### Behavior

When I loaded the file, it had most of the glyphs missing, and their places were named like “38 (0x26) U+???? "NameMe.38"”. That can be seen in the image. The file is supposed to contain glyphs for all small Greek letters, greek punctuation, some letter variants, space and the null character.

Snímek obrazovky z 2024-01-24 11-23-43

I expect that it would show all the glyphs which it showed previously. I do not know what is wrong, but I suspect that FontForge wrongly saves or loads files, or the missing characters were corrupted, so they were silently discarded. *If they are discarded as corrupted, please, make it not be silent but show a warning.*

### System information

I use FontForge 2023-01-01 on Arch Linux with Linux 6.7.0 and Gnome 45.3 on Wayland.

### Important

- [x] Screenshot

iorsh:
Fixing the deduction of Unicode value and glyph name.

The SVG glyph name used to be adopted without proper checks. This is undesirable, since the SVG names are much more permissive that Type1 standard (which is used by FontForge). Now we are checking the name validity, and trying to set name from the Unicode value when applicable.

Fixes #5349.

### DIFF
--- a/fontforge/svg.c
+++ b/fontforge/svg.c
@@ -2922,7 +2922,6 @@ static int SVGGetUnicodeEnc(xmlChar *unicode, xmlChar *arabic_form, xmlChar *gly
 static char* SVGGetGlyphName(xmlChar *glyphname, xmlChar *orientation,
 	int unicodeenc, unsigned int glyph_idx, bool* glyphname_valid) {
     char* name = NULL;
-    char buffer[400];
     bool has_vert_suffix = false;
     unsigned int name_len = 0, vert_len = 0;
     *glyphname_valid = false;
@@ -2939,10 +2938,10 @@ static char* SVGGetGlyphName(xmlChar *glyphname, xmlChar *orientation,
 
     /* Get name from unicode value. */
     if (name == NULL) {
-        if (unicodeenc == -1) {
-            sprintf( buffer, "glyph%d", glyph_idx);
-            name = copy(buffer);
-        } else
+        char buffer[40];
+        if (unicodeenc == -1)
+            name = smprintf("glyph%d", glyph_idx);
+        else
             name = copy(StdGlyphName(buffer,unicodeenc,ui_none,NULL));
     }
 
@@ -2951,9 +2950,9 @@ static char* SVGGetGlyphName(xmlChar *glyphname, xmlChar *orientation,
     has_vert_suffix = (name_len >= vert_len) &&
                       (strcmp(".vert", name + name_len - vert_len) == 0);
     if (!has_vert_suffix && orientation != NULL && *orientation == 'v') {
-        sprintf(buffer, "%s.vert", name);
+        char* new_name = smprintf("%s.vert", name);
         free(name);
-        name = copy(buffer);
+        name = new_name;
     }
 
     return name;


iorsh:
@frank-trampe, question regarding the `SplineChar::glif_name`. When the user renames the glyph, do we need to discard the preserved UFO name? I don't see a way for user to remove it.

frank-trampe:
There is a "Clear Special Data" command in the "File" menu that clears all of the otherwise inaccessible UFO state. I think that changing the glyph name ought to affect the glif name.

fontforge/svg.c
```diff
@@ -2888,12 +2892,78 @@ static void SVGParseGlyphBody(SplineChar *sc, xmlNodePtr glyph,
...
+	unicodeenc = UniFromName((char *) glyphname,ui_none,&custom);
+
+    return unicodeenc;
+}
+
+/* This function should never return NULL or empty string. */
+static char* SVGGetGlyphName(xmlChar *glyphname, xmlChar *orientation,
+	int unicodeenc, unsigned int glyph_idx, bool* glyphname_valid) {
+    char* name = NULL;
+    char buffer[400];
```
frank-trampe: Is this guaranteed to be big enough?

fontforge/svg.c
```diff
@@ -2919,50 +2989,27 @@ static SplineChar *SVGParseGlyphArgs(xmlNodePtr glyph,int defh, int defv,
...
+	/* Preserve the original glyph name and use it for SVG export.  */
+	free(sc->glif_name);
+	sc->glif_name = copy((char *) glyphname);
     }
-    if ( glyphname!=NULL ) {
-	if ( sc->unicodeenc==-1 )
-	    sc->unicodeenc = UniFromName((char *) glyphname,ui_none,&custom);
-	sc->name = copy((char *) glyphname);
-	xmlFree(glyphname);
-    } else if ( orientation!=NULL && *orientation=='v' && sc->unicodeenc!=-1 ) {
```
frank-trampe, line 2947: With removal of this code, we no longer apply ".vert" to an unnamed glyph?

fontforge/svg.c
iorsh: Fixed `SVGGetGlyphName()` to use buffer just for call to `StdGlyphName()`, and now 40 bytes are definitely enough.

fontforge/svg.c
iorsh: At the end of `SVGGetGlyphName()`, ".vert" is applied to all glyph names, regardless of whether they came from named or unnamed glyphs.

PR #5669, opened 2025-11-03
Document `no-mac-names` flag for font.generate

MihailJP:
For `font.generate`, `no-mac-names` flag gets implemented several years ago, but does not get documented yet. This PR documents the flag.

### Type of change
- **Non-breaking change**

### DIFF
--- a/doc/sphinx/scripting/python/fontforge.rst
+++ b/doc/sphinx/scripting/python/fontforge.rst
@@ -4545,7 +4545,7 @@ This type may not be pickled.
    .. object:: no-mac-names
 
       Do not include Mac names used on Classic Mac OS. This option does not
-      affect native Mac OS X applications.
+      affect native macOS (formerly known as Mac OS X) applications.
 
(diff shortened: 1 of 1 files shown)


doc/sphinx/scripting/python/fontforge.rst
```diff
@@ -4542,6 +4542,10 @@ This type may not be pickled.
 
       Include a 'TeX ' table in an ttf/otf file
 
+   .. object:: no-mac-names
+
+      Do not include Mac names used on Classic Mac OS, not on current macOS
```
iorsh: "current macOS" is probably "Mac OS X"? Then maybe it would be better to say
"Do not include Mac names used on Classic Mac OS. This option doesn't affect native Mac OS X applications"

dscorbett: It’s called macOS now, not Mac OS X.

iorsh (approved):

PR #5672, opened 2025-11-08
At UFO export, use family name for styleMapFamilyName

MihailJP:
Currently UFO export uses full name for styleMapFamilyName. This causes issue if the resulting UFO files are used for sources of a variable font. Family name should be used for said UFO entry.

1. Export source fonts in UFO.
2. Generate a designspace file using fontTools.designspaceLib.
3. Generate a variable font using fontmake.

For italic, the resulting variable font contains 'Italic' in its family name where should not.

### Type of change

- **Bug fix**

### DIFF
--- a/fontforge/sfd.c
+++ b/fontforge/sfd.c
@@ -7740,12 +7740,12 @@ bool SFD_GetFontMetaData( FILE *sfd,
 
 	for ( english=sf->names; english!=NULL && english->lang!=0x409; english=english->next );
 	if ( english && d->lastStyleMapFamilyName ) {
-	    if ( sf->names->names[ttf_family] ) {
+	    if ( english->names[ttf_family] ) {
 		LogError(_("'StyleMapFamilyName' entry has been ignored") );
 		free(d->lastStyleMapFamilyName);
 	    }
 	    else {
-		sf->names->names[ttf_family] = d->lastStyleMapFamilyName;
+		english->names[ttf_family] = d->lastStyleMapFamilyName;
 	    }
 	    d->lastStyleMapFamilyName = NULL;
 	}


frank-trampe:
This looks correct to me, but it is a potentially breaking change for a workflow that relies on the past behavior. Could we keep the full name fallback?

iorsh:
@frank-trampe, from what I can see in [UFO spec](https://unifiedfontobject.org/versions/ufo2/fontinfo.plist/#notes-2), `styleMapFamilyName` should just go to OpenType Name ID 1. I suggest we kill `SplineFont::styleMapFamilyName` altogether, and in the probably rare cases where an existing SFD contains both OpenType Name ID 1 and a different `"StyleMapFamilyName"`, we can drop it with a warning.

Right now it's just stuck there to confuse the user, who wouldn't be able to get rid of it in any reasonable way.

frank-trampe:
That seems reasonable. I can't think of any cases in which the two fields should diverge in a way that we would preserve.

iorsh (changes requested):
@MihailJP
Please observe that in the "Font Info" --> "TTF Names" dialog some names are red, which means they are identical to the PostScript names. However there is an option to detach them with a right click. This is how we should address the UFO attribute issue.

Following Frank's comment, please update the logic as follows:

- `SplineFont::styleMapFamilyName` goes away
- UFO `familyName` attribute goes to the PostScript name `SplineFont::familyname` (this already happens)
- UFO `styleMapFamilyName` attribute (if it is different from `familyName`) goes to `SplineFont::names`, which hold all the TTF Names grouped by language. When it happens, the TTF "Family" string ID 1 becomes detached from the PostScript name
- Conversely, when exporting to UFO, `styleMapFamilyName` attribute should be filled with the English language `SplineFont::names[???].names[ttf_family]`, if present.
- When reading SFD file, the `StyleMapFamilyName` entry should either go to `SplineFont::names` as appropriate, or should be dropped with warning if there is already TTF string id 1.

fontforge/sfd.c
```diff
@@ -7691,13 +7689,28 @@ bool SFD_GetFontMetaData( FILE *sfd,
     }
     else if ( strmatch(tok,"StyleMapFamilyName:")==0 )
     {
-    sf->styleMapFamilyName = SFDReadUTF7Str(sfd);
+	struct ttflangname *names;
```
iorsh: `StyleMapFamilyName` is dumped before the `LangName` entries, and read before them too. I suggest you just put it into a new field in `SFD_GetFontMetaDataData` and check collision later.

fontforge/sfd.c
```diff
@@ -7738,7 +7751,29 @@ bool SFD_GetFontMetaData( FILE *sfd,
     }
     else if ( strmatch(tok,"LangName:")==0 )
     {
+	struct ttflangname *names, *prev;
+
 	sf->names = SFDGetLangName(sfd,sf->names);
+
+	if ( sf->names->lang == 0x409 ) {  /* if StyleMapFamilyName was loaded */
```
iorsh: `SFDGetLangName()` appends records to the end of `sf->names` list, so when English is not the first `LangName` entry (which is what happens with e.g. some of my fonts), the condition `sf->names->lang == 0x409` will never be satisfied.

fontforge/ufo.c
```diff
@@ -1170,19 +1170,12 @@ static int UFOOutputFontInfo(const char *basedir, SplineFont *sf, int layer, int
...
-            if (sf->styleMapFamilyName[0]!='\0')
-                styleMapFamily = sf->styleMapFamilyName;
-        } else if (preferredFamilyName != NULL && preferredSubfamilyName != NULL) {
-            styleMapFamily = malloc(strlen(preferredFamilyName)+strlen(preferredSubfamilyName)+2);
-            strcpy(styleMapFamily, preferredFamilyName);
-            strcat(styleMapFamily, " ");
-            strcat(styleMapFamily, preferredSubfamilyName);
-        } else if (sf->fullname != NULL) styleMapFamily = sf->fullname;
+        if (familyName != NULL) {
+            styleMapFamily = malloc(strlen(familyName)+1);
```
iorsh: Why do we need to copy a string here? It seems like simple
```
char* styleMapFamily = fetchTTFAttribute(sf,ttf_family);
```
would be enough.

fontforge/sfd.c
```diff
@@ -7751,28 +7734,20 @@ bool SFD_GetFontMetaData( FILE *sfd,
...
-		}
-		prev->next = names->next;
-		for ( int i = 0; i < ttf_namemax; ++i ) {
-		    if ( names->names[i] )
-			free(names->names[i]);
-		}
-		free(names);
+	for ( english=sf->names; english!=NULL && english->lang!=0x409; english=english->next );
+	if ( english && d->lastStyleMapFamilyName ) {
+	    if ( sf->names->names[ttf_family] ) {
```
iorsh: `if ( english->names[ttf_family] )`

Issue #5615, opened 2025-09-11
Questions about ABI stability and dropping old features


valadaptive:
I don't know of a better way to word this and don't want it to come across as hostile, but I'm a bit confused about whether this project has a stable ABI, if it's a goal for it to have one, and whether contributors should consider the ABI when making changes.

It seems changes that could affect the ABI are frowned upon due to concerns about libfontforge users, which I became aware of through my PR https://github.com/fontforge/fontforge/pull/5610:

> People do reportedly link against libfontforge.

> One example I found from casual digging is https://github.com/coolwanglu/pdf2htmlEX. You can search the entire GitHub for libfontforge, and perhaps there are more.

> Trying to gauge who are users of libfontforge is similar, lack of issues doesn't necessarily mean noboy uses it.

There also seem to have been concerns that downstream packagers could pin FontForge to old versions if we change things too fast for them; see https://github.com/fontforge/fontforge/pull/5612:

> Our limited market power means that changes that break somebody's packaging script can lead to the version getting pinned for a long time.

> I have no current specific information from downstream users/packagers, just memories of past large changes not going over very well.

> Ubuntu, for example, slowly pushes out the 2019 version - https://launchpad.net/ubuntu/focal/amd64/libfontforge-dev

In that PR, two things were discovered that are relevant to these concerns:

1. As I mentioned in that thread, the Launchpad link is to Ubuntu 20.04, which will *always* ship the 2019 version of FontForge. That's just the way that Ubuntu's (and Debian's) release lifecycle works--they pick one set of major versions of packages for a release, then continue maintaining them (if there's volunteer power or enterprise money) until it hits EOL. So there's nothing FontForge could do to get Ubuntu 20.04 to update the version that they ship. And [Ubuntu 22.04 *did* drop libfontforge-dev in order to upgrade to the latest FontForge release](https://launchpad.net/ubuntu/jammy/+package/libfontforge-dev). Is this where the "version getting pinned for a long time" thing comes from?
2. We stopped shipping FontForge headers in https://github.com/fontforge/fontforge/pull/3855, which was pretty much exactly *six years ago*, and all distros seem to have followed suit. It seems some of this project's own maintainers did not notice, in those six full years, that we're no longer shipping the headers.

[...]

PR #5673, opened 2025-11-09
Python API: Add font.markSets

tuffnatty:
This patch adds support for `font.markSets` Python API, most similar in syntax to the already existing `font.markClasses`.

### Type of change
- **New feature**


fontforge/python.c
```diff
@@ -14651,35 +14651,37 @@ return( -1 );
 return( 0 );
 }
 
-static PyObject *PyFF_Font_get_mark_classes(PyFF_Font *self, void *UNUSED(closure)) {
+static PyObject *_get_mark_classes_or_sets(int start, int count, char **names, char **values) {
```
iorsh: Please keep naming style for functions

fontforge/python.c
```diff
@@ -14734,15 +14738,63 @@ static int PyFF_Font_set_mark_classes(PyFF_Font *self,PyObject *value, void *UNU
...
+    char **names, **sets;
+
+    if ( CheckIfFontClosed(self) )
+        return (-1);
+    sf = self->fv->sf;
+    cnt = _parse_mark_classes_or_sets(value, true, &names, &sets);
+    if ( cnt==-1 )
+        return (-1);
+
+    MarkSetFree(sf->mark_set_cnt,sf->mark_sets,sf->mark_set_names);
```
iorsh: `MarkSetFree()` and `MarkClassFree()` are identical. Please combine them under new name (e.g. `MarkCollectionFree()`, but feel free to change)

fontforge/python.c
```diff
@@ -15619,6 +15671,9 @@ static PyGetSetDef PyFF_Font_getset[] = {
     {(char *)"style_set_names",
      (getter)PyFF_Font_get_style_set_names, (setter)PyFF_Font_set_style_set_names,
      (char *)"A tuple, each entry of which is a 3-element tuple containing the language name (e.g. \"English (US)\"), the style set tag (e.g. \"ss01\") and the style set name.", NULL},
+    {(char *)"markSets",
```
iorsh, line 15671: Please put this section right after "markClasses"

tuffnatty: Done

fontforge/python.c
tuffnatty: Done

fontforge/python.c
tuffnatty: Done

tuffnatty:
@iorsh P.S. Actually, now I think that merging MarkSetFree() and MarkClassFree() breaks libfontforge API compatibility and should be reverted.

iorsh:
@tuffnatty, you are right, I forgot about it. We don't really guarantee its stability, but we don't break it either without necessity. See #5610, #5615.

tuffnatty:
Reverted API/ABI breakage.

PR #5675, opened 2025-11-09
FindMatchingHVEdge: report problematic glyph (#5087)

### LINKED ISSUE #5087: FindMatchinHVEdge does not report which glyph has the problem
- [ ] Screenshot
- [x ] \(optional) Possible solution/fix/workaround

git pull from today.

Problem: certain kinds of bad contour issues (which can be invisible at first glance) cause an error from FindMatchinHVEdge(), which unfortunately gives no clue as to which glyph caused the problem.  
This can cause many hours of headaches for a user.

It is very easy to fix.  See attached diff

[stemdb.c.diff.gz](https://github.com/fontforge/fontforge/files/9376300/stemdb.c.diff.gz)

tuffnatty:
This PR is a tiny bit reformatted version of the original patch attached to #5087.

### Type of change

- **Non-breaking change** (Fixes #5087)

### DIFF
--- a/fontforge/stemdb.c
+++ b/fontforge/stemdb.c
@@ -811,7 +811,10 @@ return( 0 );
 	winding += nw;
     }
     if ( space[i]==NULL ) {
-	fprintf( stderr, "FindMatchinHVEdge didn't\n" );
+	if ( gd->sc )
+	    fprintf( stderr, "FindMatchingHVEdge didn't: 0x%X, '%s'\n", gd->sc->unicodeenc, gd->sc->name );
+	else
+	    fprintf( stderr, "FindMatchingHVEdge didn't: unknown glyph\n" );
 return( 0 );
     }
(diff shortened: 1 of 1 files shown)


fontforge/stemdb.c
```diff
@@ -811,7 +811,10 @@ return( 0 );
 	winding += nw;
     }
     if ( space[i]==NULL ) {
-	fprintf( stderr, "FindMatchinHVEdge didn't\n" );
+	if ( gd->sc )
+	    fprintf( stderr, "FindMatchinHVEdge didn't: 0x%X, '%s'\n", gd->sc->unicodeenc, gd->sc->name );
```
iorsh: typo: `FindMatchingHVEdge`

tuffnatty: Fixed

PR #5679, opened 2025-11-13
Clean up #5663 a little bit

tuffnatty:
This is a PR in the follow-up of discussion in #5663. It fixes a potential buffer overrun issue in #5663 and simplifies the surrounding code, hopefully making it a little easier to understand.

### Type of change
- **Non-breaking change**

### DIFF
--- a/gdraw/gresource.c
+++ b/gdraw/gresource.c
@@ -219,7 +219,7 @@ static char* replace_CJK_font_alias(const char* resource) {
 }
 
 void GResourceAddResourceString(const char *string,const char *prog) {
-    char *ept, *value=NULL, *fixed_CJK=NULL;
+    char *ept, *value=NULL;
     const char *pt, *next;
     int cnt, plen;
     struct _GResource_Res temp;
@@ -277,11 +277,13 @@ return;
 	    while ( isspace( *pt ) && pt<next ) ++pt;
 	    value = copyn(pt,next-pt);
 #if defined(__MINGW32__)
-		fixed_CJK = replace_CJK_font_alias(value);
-		if (fixed_CJK != NULL) {
-			free(value);
-			value = fixed_CJK;
-		}
+        {
+            char* fixed_CJK = replace_CJK_font_alias(value);
+            if (fixed_CJK != NULL) {
+                free(value);
+                value = fixed_CJK;
+            }
+        }
 #endif
         temp.val = value;
 	    temp.new = true;


tuffnatty:
Well, I am not that well-versed in Windows CJK support, but I have run:
```
set LANGUAGE=zh_CN
fontforge.bat
```
It runs in Chinese but shows a lot of missing character glyphs, both with my patch and without it, sort of like in the first screenshot in #4983. Probably one also needs to install missing CJK support in Windows or something like that.

skef:
@tuffnatty did you reinstall from scratch when you were testing, and do you have a custom resources file? The fix relies on having the right entry in the resources.

There is a general problem with new releases of fontforge and existing custom resources files from older versions. In the past we've sort of punted on any kind of systemic solution and just told people who run into problems that they might need to start over again with their customizations.

tuffnatty:
I've never run FontForge in this Windows VM before.

tuffnatty:
So I've installed the Chinese language pack, changed my user's default UI language to Chinese (Simplified, China), edited the resources file to explicitly specify Microsoft YaHei UI.
When I run FontForge by clicking fontforge.bat or run_fontforge.exe, the UI is still English.
When I use cmd.exe to `set LC_ALL=zh_CN` and then run FontForge, the UI is Chinese and misses lots of glyphs.
With or without my patch, makes no difference. I use the GitHub Actions artifacts on a Windows 10 22H2 VM.

tuffnatty:
@iorsh , If you have a working testing environment for #5663, could you please test these artifacts?
https://github.com/fontforge/fontforge/actions/runs/19311309209/job/55231975836 (a CI build of #5663)
https://github.com/fontforge/fontforge/actions/runs/19343230233/job/55336736313 (a CI build of this PR)

iorsh:
Looks like my fix works in development environment, when compiled and run, but not when run from artifact or installed. I need to take a look.

iorsh:
I think I found the root cause here. The font resource mechanism has two paths:

- `GResFont` structure which is responsible for default font names, and the fonts user defined through the Appearance Editor
- `_GResource_Res` list which holds data from resource files.

The actual selection happens in https://github.com/fontforge/fontforge/blob/ce3693d8bb8c1f0d1c22641df3bf2d608bba7387/gdraw/gresource.c#L311

I fixed the first path, and it worked when the binary loaded without any resource files. But the installed or live binary has the resources, specifically `fontforge.View.DefaultFont: 12pt windows-cjk-workaround,system-ui,Deja Vu Sans,Calibri`, which is not replaced with the actual CJK UI font, but still overrides the defaults.

iorsh:
@tuffnatty, may I push the fix into your branch? We are going to have a merge conflict otherwise.

tuffnatty:
@iorsh sure

iorsh:
@tuffnatty, could you please test? Should be working now.

tuffnatty:
> @tuffnatty, could you please test? Should be working now.

Confirming, the build artifact is working, no missing glyphs.

PR #5681, opened 2025-11-14
Quickfix a collision between ttf_flags and ps_flags (#5678)

### LINKED ISSUE #5678: Generated TTF has no outlines
Consider the following workflow:
1. Open a font
2. Choose File -> Generate Fonts
3. As the generated font type, choose OpenType (CFF)
4. Choose Options
5. In the Options dialog, uncheck PostScript Hints and choose Ok
6. Now, change your mind and choose TrueType as the generated font type.
7. Choose Generate.

The generated font does not contain any outlines and can't be opened with FontForge.
The same result is reproducible by passing `('no-hints',)` to `font.generate()` from Python.

I've bisected the problem to be in  https://github.com/fontforge/fontforge/commit/387146a241b36bcdf6ce229c5a3fe367ed3854a1 (by the way, IMHO when merging a change of this size one shouldn't squash all the commits together, it makes software archaeology much harder, at least global refactoring changes should be separate).
This commit has introduced a `ttf_flag_no_outlines` enum value into `enum ttf_flags`. Apparently, this and neighbouring values were put in a hole originally reserved between bit 15 (16?) and bit 29 for ORing with values from `enum ps_flags`. And `ps_flag_nohints` is equal to the newly introduced `ttf_flag_no_outlines` and is misinterpreted as such.
Probably, the neighbouring flags (starting with `ttf_native_kern` which was introduced even earlier) also have collisions with other `ps_flags` values.

If it was meant that `ps_flags` don't carry any useful info when generating a TTF, they should probably be filtered out before application.
If `ps_flags` may still be useful while generating a TTF, the collision can be addressed in several ways.
1. Get rid of the collision by assigning new positions to the new flags. If 32-bit support is to be dropped, the change could be trivial: moving the new flags to 1 << 31 .. 1 << 33. If 32-bit support is to be kept, however, the flags need to be changed to a uint64_t value across the whole code base, enum usage is to be dropped (as enums are restricted to an int), and support for uint64_t "preferences" is to be introduced. I would not consider dropping 32-bit support because of such a minor issue, however, implementing all these changes can easily introduce new bugs and make code clumsier.
2. Get rid of the collision by storing `ps_flags` and `ttf_flags` separately. E.g., moving `ttf_flags` out of the `flags` would require to review most of TTF generation code and add an extra int "preference" for the TTF part of `old_sfnt_flags`.
3. Consider using other places for flags which are needed by new code, e.g. add some `ufo_flags` and `harfbuzz_flags`. In this case, the only code that needs to be updated is the new code, so the chance of introducing bugs in the old functionality is minimal.

tuffnatty:
This change replaces `ttf_flag_fake_map` and `ttf_flag_no_outlines` introduced  in https://github.com/fontforge/fontforge/commit/387146a241b36bcdf6ce229c5a3fe367ed3854a1 and colliding with `enum ps_flags` values, with a single flag to which the last available bit in the flags int on 32-bit systems is being assigned. It changes the sign of the 32-bit int but will hopefully never be serialized, as it's only used internally.

`ttf_native_kern` colliding with `ps_nohintsubs` since the beginning of times does not need to be moved, as `ps_nohintsubs` seems to be no longer used anywhere in the code.

Probably there are better fixes, see the discussion in #5678, but this one seems to fix the problem at hand.

### Type of change
- **Bug fix**
Fixes #5678

### DIFF
--- a/fontforge/splinefont_enums.h
+++ b/fontforge/splinefont_enums.h
@@ -84,6 +84,7 @@ enum bitmapformat {
 };
 
 enum ttf_flags {
+    // WARNING: These flags must not collide with ps_flags
     ttf_flag_shortps = 1 << 0,
     ttf_flag_nohints = 1 << 1,
     ttf_flag_applemode = 1 << 2,
@@ -102,12 +103,11 @@ enum ttf_flags {
     ttf_flag_dummyDSIG = 1 << 15,
     ttf_native_kern = 1 << 16,  // This applies mostly to U. F. O. right now.
     ttf_flag_no_outlines =
-        1 << 17,  // HarfBuzz: Skip "glyf" and "CFF " tables for performance
+        1 << 19,  // HarfBuzz: Skip "glyf" and "CFF " tables for performance
     ttf_flag_oldkernmappedonly =
-        1 << 18,  // Allow only mapped glyphs in the old-style "kern" table,
+        1 << 29,  // Allow only mapped glyphs in the old-style "kern" table,
                   // required for Windows compatibility
-    ttf_flag_nomacnames = 1 << 19  // Don't autogenerate mac name entries
-    // WARNING: These flags must not collide with ps_flags
+    ttf_flag_nomacnames = 1 << 30  // Don't autogenerate mac name entries
 };
 
 enum ttc_flags { ttc_flag_trymerge = 0x1, ttc_flag_cff = 0x2 };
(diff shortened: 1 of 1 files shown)


iorsh:
`ps_flag_restrict256` is fortunately also not in use, and it gives us one last bit :)
I'd suggest to wipe out `ps_flag_restrict256` and `ps_flag_nohintsubs`, 19 years are enough time.

And we probably need to do something with this bit array, the 32 bit restriction becomes impractical

iorsh:
@tuffnatty, could you please finalize this PR? I think it would be enough to wipe `ps_flag_restrict256` and `ps_flag_nohintsubs`, and let `ttf_flag_fake_map` and `ttf_flag_no_outlines` occupy their slots. Any further improvements can wait.

fontforge/splinefont_enums.h
```diff
@@ -120,25 +121,23 @@ enum openflags {
...
-    ps_flag_nohintsubs = 0x10000,
-    ps_flag_noflex = 0x20000,
-    ps_flag_nohints = 0x40000,
-    ps_flag_restrict256 = 0x80000,
-    ps_flag_afm = 0x100000,
-    ps_flag_pfm = 0x200000,
-    ps_flag_tfm = 0x400000,
-    ps_flag_round = 0x800000,
+    // WARNING: These flags must not collide with ttf_flags
+    ps_flag_noflex = 1 << 17,
```
iorsh, line 125: Same values, expressed as shift to harmonize with `ttf_flags`

iorsh:
After the cancellation of `ttf_flag_fake_map` in #5736, this PR only fixes the collision of `ttf_flag_no_outlines` and removes unused flags `ps_flag_nohintsubs` and `ps_flag_restrict256`.

skef (approved):
LGTM

PR #5682, opened 2025-11-14
GTK-based "Find Problems" dialog

### LINKED ISSUE #3031: Find problems dialogue and message size to be improved
- [X] Screenshot

1. The 'Find Problems" screen is too small in height and some of the tick boxes are truncated.

problem not resized

2. The message 'Not sure if this is an error" does not translate well in other language because the length of the message are controlled by original text and the width of the screen is controlled by the length of the buttons.

validationerrormessage

Fontforge 20:55 AWST 4-Oct-2016 on Windows but also present on latest build on Ubuntu.

1. All the 9 options of the points tabs should be visible when the window pops up.  Currently, only 7 and an half show.
2. The text should be displayed on 3 lines not 5 and the message box should be able to support longer text than the English version.

Create a new font.  Just draw one with a line that intersect.  Then click on Element>Validation>Find Problems.  Also click on Element>Validation>Validate.

1.  In .\fontforgeexe\problems.c

   ```
    wattrs.utf8_window_title = _("Find Problems");
    pos.x = pos.y = 0;
    pos.width = GGadgetScale(GDrawPointsToPixels(NULL,218));
    - pos.height = GDrawPointsToPixels(NULL,294);
    + pos.height = GDrawPointsToPixels(NULL,324);
```
2. Not sure if I should just change the text in fr.po or do this:

```
  if ( ff_ask(_("Not sure if this is an error..."),(const char **) buts,0,1,
-           _("This font contains non-integral coordinates. That's OK\n"
-           "in PostScript and SVG but causes problems in TrueType.\n"
-            "Should I consider that an error here?"))==0 ) {
+             _("This font contains non-integral coordinates. That's OK in PostScript and SVG but causes problems in TrueType.\n"
+             "Should I consider that an error here?"))==0 ) {
```
Is there a better way of making the text wrap no matter what length it is?

### LINKED ISSUE #3899: Dialog window width and height
dialogs

MacOS 10.11.6
FontForge version 20190413 based on sources from 26-Jul-2019 git:279e69163e4da2be6240d1b35fdbb6438e792344

Problem:
Dialog opens, but not all text shows (marked red). There is also no scrollbar to suggest there might be more items in the dialog.
Go to: Element > Validation > Find Problems
Screenshot **left** side: when the dialog opens
Screenshot **middle**: dragging the size of the dialog down and right
Screenshot **right** side: dragging the size of the dialog only down in another tab

What should happen?
Dialog window should initially expand to such a size, that all items display fully (also all items within other tabs, if present). Also independent of language!

I haven't checked all dialog windows, so maybe a general way should be ensured for all dialogs of FontForge (this is just an idea).

### PR DESCRIPTION
**Pull request by iorsh**
This PR replaces legacy "Find Problems" dialog with a GTK-based one. The fix improves dialog layout and arbitrary line breaks in long messages, which can't be properly localized.

Fixes #3899, fixes #3031.

New UI:
findprobs

Old UI:
prob_dflt_sys

### DIFF
--- a/fontforgeexe/gtk/find_problems.cpp
+++ b/fontforgeexe/gtk/find_problems.cpp
@@ -178,17 +178,9 @@ ProblemRecordsOut FindProblemsDlg::show(GWindow parent,
             const auto& [checkbox, entry] = dialog.widget_map_[record.cid];
             if (!checkbox.get_active()) continue;
 
-            ProblemRecordValue new_value = record.value;
-            if (std::holds_alternative<int>(record.value)) {
-                auto int_entry = dynamic_cast<widgets::IntegerEntry*>(entry);
-                if (int_entry) {
-                    new_value = int_entry->get_value();
-                }
-            } else if (std::holds_alternative<double>(record.value)) {
-                auto double_entry = dynamic_cast<widgets::DoubleEntry*>(entry);
-                if (double_entry) {
-                    new_value = double_entry->get_value();
-                }
+            NumericalValue new_value;
+            if (entry) {
+                new_value = entry->get_num_value();
             }
             records_out.emplace(record.cid, new_value);
         }
--- a/fontforgeexe/gtk/widgets/num_entry.hpp
+++ b/fontforgeexe/gtk/widgets/num_entry.hpp
@@ -26,14 +26,17 @@
  */
 #pragma once
 
-#include <iostream>
+#include <variant>
 #include <gtkmm.h>
 
 namespace ff::widgets {
 
+using NumericalValue = std::variant<std::monostate, int, double>;
+
 class NumericalEntry : public Gtk::Entry {
  public:
     NumericalEntry();
+    virtual NumericalValue get_num_value() const = 0;
 
  protected:
     virtual bool validate_text(const Glib::ustring& text) const = 0;
@@ -46,6 +49,7 @@ class IntegerEntry : public NumericalEntry {
  public:
     void set_value(int val);
     int get_value() const;
+    NumericalValue get_num_value() const override { return get_value(); };
 
  private:
     bool validate_text(const Glib::ustring& text) const override;
@@ -57,6 +61,7 @@ class DoubleEntry : public NumericalEntry {
 
     void set_value(double val);
     double get_value() const;
+    NumericalValue get_num_value() const override { return get_value(); };
 
  private:
     std::string decimal_point_;
--- a/fontforgeexe/gtk/find_problems_shim.hpp
+++ b/fontforgeexe/gtk/find_problems_shim.hpp
@@ -34,15 +34,15 @@
 
 namespace ff::dlg {
 
-using ProblemRecordValue = std::variant<std::monostate, int, double>;
-using ProblemRecordsOut = std::map<short /*cid*/, ProblemRecordValue>;
+using NumericalValue = std::variant<std::monostate, int, double>;
+using ProblemRecordsOut = std::map<short /*cid*/, NumericalValue>;
 
 struct ProblemRecord {
     short cid;
     L10nText label;
     L10nText tooltip;
     bool active;
-    ProblemRecordValue value;
+    NumericalValue value;
     short parent_cid;
     bool disabled;
 };

### CODE REVIEW

**Comment by skef** [comment]
Since this has some movement, what does the new window do when it needs to wrap its tabs (e.g. if localization makes those names longer)?

**Comment by iorsh** [comment]
> Since this has some movement, what does the new window do when it needs to wrap its tabs (e.g. if localization makes those names longer)?

Fixed, the tabs should look like this now (note the arrows):
image

I'd be grateful if you could review the entire PR, Frank is quite busy recently.

**Comment by skef** [comment]
It's a lot but I'll try to take a look.

**Comment by skef** [comment]
Independent of the review, it might be good if you could sort of fuzz for analogous formatting problems. That question was based on a pure guess on my part; there could be other things.

**Comment by iorsh** [comment]
Actually, it was a direct hit. I did consider the other stuff, but couldn't see any way to wrap tab labels, so I just left them concise. Found the solution now by re-reading the documentation for `Gtk::Notebook`

There are also item labels which I intentionally unabbreviated, none of them seems to span more than a monitor width (even with very generous margins). Tooltips can be very long, but I removed all newlines from them, and GTK wraps them automatically:
image

**Comment by iorsh** [comment]
Btw, this one is big indeed, but you can pick any of my recent PRs (the latest six are all unreviewed) if you have some time for that. Thanks!

**Comment by skef** [comment]
Been looking at this a bit more. I added a branch called "modules-5682, which is:

1. The modules branch
2. With this branch ported on top
3. and problems.c renamed to problems.cpp and ported to minimal c++

Once the modules code is committed, this will make it easier for you to factor out a bunch of the C->C++->C bridging code, which I think would be an overall win.

File: fontforgeexe/gtk/find_problems.cpp [hunk 1]
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    for (const ProblemTab& tab : pr_tabs) {
+        for (const ProblemRecord& record : tab.records) {
+            const auto& [checkbox, entry] = dialog.widget_map_[record.cid];
+            if (!checkbox.get_active()) continue;
+
+            ProblemRecordValue new_value = record.value;
+            if (std::holds_alternative<int>(record.value)) {
+                auto int_entry = dynamic_cast<widgets::IntegerEntry*>(entry);
```
skef: This is (arguably) a bit smelly. What about pushing the definition of std::variant<monostate, int, double> to somewhere common (either a new .hpp in this hierarchy or elsewhere) and having a virtual method on NumericalEntry return that? Then you can just assign here?

fontforgeexe/gtk/widgets/num_entry.hpp
```diff
@@ -0,0 +1,67 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ * WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+ * OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+ * ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#pragma once
+
+#include <iostream>
```
skef: Claude says this can go.

iorsh: Done

fontforgeexe/gtk/find_problems.cpp
iorsh: Fixed. I took the liberty to define `NumericalValue` twice for now; I'll try to think about some common header when there are a bit more examples needing it, to see a bigger picture.

skef: I have no problem with the double definition.

skef:
I don't know if there's any MID stuff in here, but why don't you merge this first and I'll figure out in #5763 after.

PR #5683, opened 2025-11-14
Metrics view: Fix crash on Up/Down keypress while in the feature list

tuffnatty:
#19 has introduced a crash when first clicking a feature in the feature list, thus setting the focus on it, and then pressing an Up or Down arrow key.
Then #420 removed the buggy code.
#451 re-introduced the code, and the bug as well.

### Type of change
- **Bug fix**


iorsh (approved):

iorsh:
Looks good to me.

PR #5694, opened 2025-11-22
Implement `with` statement for Python font object

### LINKED ISSUE #2673: fontforge.font should have __exit__ attribute
I think `fontforge.font` class should have `__exit__` attribute so that we can write like:

``` python
# assuming 'somefont.sfd' exists
with fontforge.open('somefont.sfd') as font:
    # do something
    font.save('destfont.sfd')
    # Here we need not font.close()
```

Actually, this code abends with `AttributeError: __exit__`.

Since the class is built in, attributes cannot be added from a script.

iorsh:
Support context protocol to allow Python syntax like
```
with fontforge.open('somefont.sfd') as fnt:
    # do something
    fnt.generate('somefont.ttf')
```
Fixes #2673

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -19568,7 +19568,7 @@ PyMethodDef PyFF_Font_methods[] = {
     { "validate", (PyCFunction)PyFFFont_validate, METH_VARARGS, "Check whether a font is valid and return True if it is." },
     { "reencode", (PyCFunction)PyFFFont_reencode, METH_VARARGS, "Reencodes the current font into the given encoding." },
     { "clearSpecialData", (PyCFunction)PyFFFont_clearSpecialData, METH_NOARGS, "Clear special data not accessible in FontForge." },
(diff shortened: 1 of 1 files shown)


fontforge/python.c
```diff
@@ -19564,6 +19568,8 @@ PyMethodDef PyFF_Font_methods[] = {
     { "validate", (PyCFunction)PyFFFont_validate, METH_VARARGS, "Check whether a font is valid and return True if it is." },
     { "reencode", (PyCFunction)PyFFFont_reencode, METH_VARARGS, "Reencodes the current font into the given encoding." },
     { "clearSpecialData", (PyCFunction)PyFFFont_clearSpecialData, METH_NOARGS, "Clear special data not accessible in FontForge." },
+    { "__enter__", (PyCFunction) PyFFFont_enter, METH_NOARGS, "Frees up memory for the current font. Any python pointers to it will become invalid." },
```
skef: Documentation string is wrong here.

skef (approved):
Other than the doc string, looks good1

PR #5698, opened 2025-11-29
Ignore bit 1 of head.flags for CFF fonts

### LINKED ISSUE #5697: Does FontForge fail to fully support editing the STIXTwoMath-Regular.otf font?
Does FontForge fail to fully support editing the STIXTwoMath-Regular.otf font?
When I open STIXTwoMath-Regular.otf in FontForge and generate a new font file without making any modifications — I name it myStixTwoMath.otf — I find that myStixTwoMath.otf is not identical to the original STIXTwoMath-Regular.otf. The generated font has metric discrepancies.

A very typical example concerns the glyphs uni23A7, uni23A8, uni23A9, and uni23AA, which are components of the extensible braceleft. Their horizontal positions shift, so they are no longer aligned. To further investigate, I inspected other glyphs in the font and found that almost all glyph outlines have shifted horizontally to some degree — and the amount of shift is inconsistent.

However, when I use FontForge to open other math fonts such as texgyretermes-math.otf or NewCMMath-Regular.otf, this issue does not occur.

So, does this indicate a problem with STIXTwoMath-Regular.otf itself, or is it caused by FontForge’s incomplete compatibility with this font?

As shown in the image below, this is the braceleft generated using myStixTwoMath.otf：

Image

Image

Image

Image

Image

khaledhosny:
This bit is “relevant only for TrueType rasterizers” per the spec: https://learn.microsoft.com/en-us/typography/opentype/spec/head#flags

If the bit is unset, FontForge shifts glyph outlines, but this is wrong for CFF fonts.

### Type of change
- **Bug fix**: Fixes #5697

### DIFF
--- a/fontforge/parsettf.c
+++ b/fontforge/parsettf.c
@@ -5594,9 +5594,8 @@ return( 0 );
 	readttfglyphs(ttf,info);
     } else if ( info->cff_start!=0 ) {
 	info->to_order2 = (loaded_fonts_same_as_new && new_fonts_are_order2);
-	// Relevant only for TrueType: 
+	// Relevant only for TrueType (Font Header Table specification): 
 	// https://learn.microsoft.com/en-us/typography/opentype/spec/head#flags
-	// https://github.com/fontforge/fontforge/issues/5697
 	info->apply_lsb = false;
 	if ( !readcffglyphs(ttf,info) ) {
 	    switch_to_old_locale(&tmplocale, &oldlocale); // Switch to the cached locale.


fontforge/parsettf.c
```diff
@@ -5594,6 +5594,8 @@ return( 0 );
 	readttfglyphs(ttf,info);
     } else if ( info->cff_start!=0 ) {
 	info->to_order2 = (loaded_fonts_same_as_new && new_fonts_are_order2);
+	// relevant only for TrueType: https://github.com/fontforge/fontforge/issues/5697
```
iorsh: Please comment with spec name / chapter / link. We can't expect GitHub to stay around forever.

iorsh (commented):
Looks good to me

Finii:
Excellent fix, thank you!

PR #5699, opened 2025-12-01
Fix Python config initialization

### LINKED ISSUE #4835: SEGV when running fontforge -script -lang=py blah
I happened to run with the above arguments and got a SEGV. `blah` didn't exist.

One is supposed to do `-lang=py -script` so those are bad args but the program shouldn't crash.

Just filing this to track ...

### LINKED ISSUE #5062: "fontforge -script --help" segfault and coredump'ed
20220308 on fedora 36. Problem as in title.

I haven't used it for some years and needed a reminder (of how it sets python path, etc, to look up other python modules). So I added "--help" to it. Well, it gave me the information I wanted, but segfaulted and core dump at the end. An oversight somewhere, i guess.

iorsh:
Fixes #5062, fixes #4835

### DIFF
--- a/fontforge/python.c
+++ b/fontforge/python.c
@@ -20826,7 +20826,7 @@ void FontForge_FinalizeEmbeddedPython(void) {
 }
 
 /* This is called to start up the embedded python interpreter */
-void FontForge_InitializeEmbeddedPython(void) {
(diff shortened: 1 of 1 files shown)


skef (approved):
LGTM

PR #5708, opened 2025-12-19
GTK-based "Bitmap Strikes" dialog

### LINKED ISSUE #3346: Conflicting interpretation of comma with some locales (ex. French locale)

### Important

- [x] Screenshot

**Screenshot with French locale:**

screenshot

**Fontforge Version:**

         Based on sources from 00:15 UTC 31-Jul-2017-ML-D.
         Based on source from git with hash: b9149c13e8f9464fc21473f1f676b36a2130775d
        no xdefs_filename!
        TESTING: getPixmapDir:/usr/share/fontforge/pixmaps
        TESTING: getShareDir:/usr/share/fontforge
        TESTING: GResourceProgramDir:/usr/bin
        trying default theme:/usr/share/fontforge/pixmaps/resources
        fontforge 00:15 UTC 31-Jul-2017
        libfontforge 20170731

**Expected behaviour:**

When I enter a pixel size in “*Bitmap strikes available*”, I expected any optional decimal separator to not be interpreted a list separator. With EN locale, the decimal separator is a dot, and the comma is a list separator for multiple point sizes. With FR locale, the decimal separator is unfortunately the comma too, which makes Fontforge believe there are multiple values, while there is only one. This is especially an issue since Fontforge additionally temporarily generates a zero decimal part even when the value is an integer. This zero decimal is then interpreted as a second point size equal to zero, which prevents font generation at a later stage (the font generator complains there is an unavailable point size and treat as an error).

**Steps to reproduce the behavior:**

See above (steps are implied).

**Possible solution/fix/workaround:**

Use dot as a decimal separator, whenever the locale says it’s a comma. At least in France, we are even more used to the dot decimal separator than the comma, so it won’t be an issue (the comma as decimal separator is rather pedantic in France, although I can’t telle for other countries).

iorsh:
This PR replaces legacy "Find Problems" dialog with a GTK-based one. Fixes #3346.

**Changes:**
New UI:
bs_proj

Old UI:
bs_sys

- List of numbers is locale-sensitive, uses semicolon as list separator when the comma is used as decimal point.
- OS-based monitor resolutions are gone, showing the typical 96 PPI resolution, and the user's actual resolution.
- Point sizes are read-only now, the bitmap strikes should be defined by pixel height only.
- "Use FreeType" option is gone. FreeType is always available, and the rasterization uses it.
- Error indication is now inline: 
bs_proj_err

### DIFF
--- a/fontforgeexe/gtk/bitmaps_dlg.cpp
+++ b/fontforgeexe/gtk/bitmaps_dlg.cpp
@@ -47,7 +47,7 @@ static std::string FormatBitmapSize(int px_size) {
 static std::string FormatBitmapSize(int px_size, double scale) {
     char decimal_point = ui_utils::get_decimal_point();
     char buffer[10];
-    sprintf(buffer, "%.1f", px_size * scale);
+    snprintf(buffer, sizeof(buffer), "%.1f", px_size * scale);
     return buffer;
 }
 


fontforgeexe/gtk/bitmaps_dlg.cpp
```diff
@@ -0,0 +1,278 @@
...
+static const uint16_t kSizeError = 0xffffu;
+
+static std::string FormatBitmapSize(int px_size) {
+    return std::to_string(px_size);
+}
+
+static std::string FormatBitmapSize(int px_size, double scale) {
+    char decimal_point = ui_utils::get_decimal_point();
+    char buffer[10];
+    sprintf(buffer, "%.1f", px_size * scale);
```
skef: Might as well snprintf here.

PR #5709, opened 2025-12-20
Fix glyph.codepoint attribute

### LINKED ISSUE #5191: In python with Fontforge-2023-01, glyph.codepoint always gives `U+%04X` for any glyph
FontForge-2023-01-01-Windows.exe

- [x] Descption
- **Expected**: get right codepoint of any glyph by `glyph.codepoint`, e.g. the glyph `0` has a codepoint `'U+0030'`
- **Actual**: get wrong codepoint of any glyph by `glyph.codepoint`, e.g. the glyph `0` has a codepoint `'U+%04X'`

- [x] Code to reproduce the behavior 
```python
import fontforge

font_File_Path = r"arial.ttf"
font = fontforge.open(font_File_Path)

any_character = '0'
glyphs = font.selection.byGlyphs.select(any_character)

for glyph in glyphs:
	print('glyph.codepoint = ' + str(glyph.codepoint)) # always 'U+%04X'
```

iorsh:
Format Unicode code point in U+XXXX format for the Basic Multilingual Plane, or U+XXXXXX for the supplementary planes, or ``None``.

Fixes #5191.

### DIFF
--- a/doc/sphinx/scripting/python/fontforge.rst
+++ b/doc/sphinx/scripting/python/fontforge.rst
@@ -1704,7 +1704,7 @@ must be created through the font.
 .. attribute:: glyph.codepoint
 
    Unicode code point for this glyph in U+XXXX format for the Basic Multilingual
-   Plane, or U+XXXXXX for the supplementary planes, or ``None``. (readonly)
+   Plane, and up to U+XXXXXX for the supplementary planes, or ``None``. (readonly)
 
 .. attribute:: glyph.color
 


dscorbett:
This pull request uses `U+%06X` for non-BMP code points, but the usual format for referencing a Unicode code point is always `U+%04X` regardless of plane. For example, U+12345 is referred to as U+12345, but this pull request would add an extra leading zero: U+012345.

iorsh:
> This pull request uses `U+%06X` for non-BMP code points, but the usual format for referencing a Unicode code point is always `U+%04X` regardless of plane. For example, U+12345 is referred to as U+12345, but this pull request would add an extra leading zero: U+012345.

Fixed

JoesCat:
> > This pull request uses `U+%06X` for non-BMP code points, but the usual format for referencing a Unicode code point is always `U+%04X` regardless of plane. For example, U+12345 is referred to as U+12345, but this pull request would add an extra leading zero: U+012345.
> 
> Fixed

I ran into some older (or maybe other) compilers ~2012 in the past and had to include "U+%05X", you might find a couple of these %05X with grep. Just an FYI, or sanity check if you're able to check with another compiler (VS, MAC, other?).

iorsh:
@JoesCat, I'm relying on CI checks to get a better picture of compilation problems. Windows and Mac builds seem to pass fine, MSVC compilation is unfortunately a project on its own

PR #5711, opened 2025-12-21
Apply width when copying chars in bitmap-only font

### LINKED ISSUE #5620: Copied glyphs in a bitmap font cannot be exported
This issue has also been reported in this comment, but I think it deserves its own issue, since the comment is unrelated to the issue it's posted under: https://github.com/fontforge/fontforge/issues/5033#issuecomment-1383045564

FontForge version: 20230101

To reproduce:

1. Create a new font.
2. Go to Element -> Bitmap Strikes Available and enter a pixel size (e.g. 16), converting the font to a bitmap font.
3. Edit some glyph.
4. Copy and paste the glyph to another glyph. The glyph will be visible, and will save correctly, but the faint "X" in the background is still present over it. This persists even after saving the font (.sfd) and reopening it.
5. Use File -> Generate Fonts and save the font in any format. Using BDF is convenient because it's easy to verify with a text editor that the copied glyph isn't present in the output, but I also tried exporting as OTB and verified with my system's font viewer that only one glyph is visible in the output.

Here's one possible outcome of the above steps (copied capital "A" to lowercase "a"): [repro.zip](https://github.com/user-attachments/files/22326981/repro.zip)

It's worth noting that in the .sfd file, the uppercase "A" (original glyph) has `Flags: W` but the lowercase "a" (copied glyph) does not:

```
StartChar: A
Encoding: 65 65 0
Width: 1000
Flags: W
LayerCount: 2
EndChar

StartChar: a
Encoding: 97 97 1
Width: 1000
LayerCount: 2
EndChar
```

Manually editing the file to add `Flags: W` to the lowercase "a" makes everything work correctly (the "X" is no longer present behind the glyph in the UI, and it is exported correctly).

iorsh:
Fixes #5620

### DIFF
--- a/fontforge/bitmapchar.c
+++ b/fontforge/bitmapchar.c
@@ -864,7 +864,7 @@ return( NULL );
 	bdf->glyphcnt = sf->glyphcnt;
     }
     if ( (bc = bdf->glyphs[gid])==NULL ) {
-	if ( use_freetype_to_rasterize_fv ) {
+	if ( use_freetype_to_rasterize_fv && SCDrawsSomething(sc) ) {
 	    void *freetype_context = FreeTypeFontContext(sf,sc,NULL,ly_fore);
 	    if ( freetype_context != NULL ) {
 		bc = SplineCharFreeTypeRasterize(freetype_context,


iorsh:
This fix only affects bitmap-only fonts. Tested copy-paste scenarios in Font View, Bitmap View - behavior looks ok, no visible regressions.

Issue #5706, opened 2025-12-12
CVE vulnerabilities metaissue


zdi-disclosures:
Hello,

We have submitted 12 vulnerability tickets to your FontForge account on https://sourceforge.net. We have not received any updates since July 2025. Please provide a private/secure channel where we can submit the report confidentially.

We reserve the right to publish the reports in accordance with our Disclosure Policy, which you can read here: https://www.zerodayinitiative.com/advisories/disclosure_policy/

Thank you and best regards,
Zero Day Initiative
Trend Micro
ZDI-DISCLOSURES@trendmicro.com

iorsh:
Hello, you can contact me at iorsh@users.sourceforge.net

FontForge account on https://sourceforge.net/ is not really monitored, as far as I know.

JoesCat:
@zdi-disclosures - thanks for contacting us here.
If you are also monitoring libuninameslist, or libspiro, can you also update your list to point here too.

Agreed, it appears sourceforge is not watched by the maintainers listed on that site.

Looking at the site, some issues were posted 3 months ago.
https://sourceforge.net/projects/fontforge/
[activity](https://sourceforge.net/p/fontforge/activity/?page=0&limit=100#6931cea738f9ed8cdff962d6)

@davelab6 - are you able to grant access?

skef:
@zdi-disclosures Please read https://github.com/fontforge/fontforge/wiki/Community-guidelines#D1 and confirm that your security problems are relevant relative to our policy. FontForge is riddled with potential security problems and the project is generally speaking not interested in security reports. If you would like to make PRs with fixes, they will be reviewed.

skef:
@zdi-disclosures You are of course free to publish any reports that you would like.

skef:
@zdi-disclosures I went ahead and looked at the reports you filed on the SourceForge site. There are indeed the sort of reports we do not welcome unless they come with test-passing PRs that fix the problems. 

Publish what you want, Lucas -- let it serve as a healthy warning to anyone intending to run FontForge without a good idea of where their input files come from.

ahmetfurkankavraz:
I was reviewing these CVE issues and I could only find reproduces for 4 of them in the https://sourceforge.net/p/fontforge/patches/. Is there any chance that some of these are private/forbidden for everybody? 

[...]

PR #5721, opened 2026-01-08
Fix CVE-2025-15275: Heap buffer overflow in SFD image parsing

ahmetfurkankavraz:
Validate clutlen parameter (0-256) before use to prevent heap buffer overflow when writing to fixed-size clut array.

Fixes: 
CVE-2025-15275  
ZDI-CAN-28543
[ZDI-25-1189](https://www.zerodayinitiative.com/advisories/ZDI-25-1189/)

Type of change:
- **Bug fix**


ahmetfurkankavraz:
Reproducer can be found in https://sourceforge.net/p/fontforge/patches/37/

PR #5720, opened 2026-01-07
Fix CVE-2025-15279: Heap buffer overflow in BMP RLE decompression

ahmetfurkankavraz:
The readpixels() function reads RLE count values from BMP files without validating buffer bounds. A malicious BMP can specify excessive counts causing heap buffer overflow during pixel decompression, potentially leading to remote code execution.

Add bounds checking after each count read to ensure ii + cnt does not exceed the allocated buffer size (head->height * head->width). Return 0 on validation failure to trigger error handling.

CVE-2025-15279
CVSS: 7.8 (High)
ZDI-CAN-27517
[ZDI-25-1184](https://www.zerodayinitiative.com/advisories/ZDI-25-1184/)

Type of change:
- **Bug fix**

This PR is related to the issue #5706

PR #5723, opened 2026-01-09
Fix CVE-2025-15279: Move bounds check inside cnt >= 3 block

ahmetfurkankavraz:
Move the bounds check to inside the 'if (cnt >= 3)' block. This fixes the issue where cnt == 0, cnt == 1, and cnt == 2 require different ii calculations (end-of-line, end-of-bitmap, delta) and the bounds check before the conditional would incorrectly reject valid operations.

CVE-2025-15279
ZDI-CAN-27517
[ZDI-25-1184](https://www.zerodayinitiative.com/advisories/ZDI-25-1184/)

Type of change:
- **Bug fix**

### DIFF
--- a/gutils/gimagereadbmp.c
+++ b/gutils/gimagereadbmp.c
@@ -190,10 +190,10 @@ static int readpixels(FILE *file,struct bmpheader *head) {
 		    head->byte_pixels[ii++] = ch;
 	    } else {
 		cnt = getc(file);
-		if (cnt < 0 || ii + cnt > head->height * head->width) {
-		    return 0;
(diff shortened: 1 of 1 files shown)


ahmetfurkankavraz:
A colleague of mine warned me that this commit might cause some problems when the cnt value in the second if statement equals to 0, 1 or 2 which acts in another logic compared to cnt value is greater than or equals to 3.

This fixes an issue appeared with #5720 .

hasimzc:
wow ahmetfurkankavraz you are great !!!

PR #5730, opened 2026-01-16
Add CMake target to autoformat source files

iorsh:
`ninja format` now automatically formats all C/C++ source files listed in `.clang-files`.

- `.clang-files` entries may include shell wildcards
- All new C/C++ files should be added to `.clang-files`
- Existing files should generally not be re-formatted, since even format-only commit renders `git blame` feature useless.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,3 +202,10 @@ endif()
 
 setup_cpack()
 print_build_options()
+
+add_custom_target(
+    format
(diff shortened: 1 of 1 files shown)


skef:
This seems fine to me. Are we doing CI on whether they're formatted on check-in?

iorsh:
No, it's only manual at the moment. Later maybe I'll setup CI to run it and fail if any files are changed. Theoretically, formatting changes can be also auto-committed, but it ruins code provenance.

skef (approved):
This is a sensible approach. We can add CI later if we feel like it.

PR #5736, opened 2026-01-27
Simplify fake encoding trick in Metrics View

iorsh:
The current fake encoding maps unencoded glyphs to an empty Private Use Area. This approach mostly works, but it's cumbersome, and suffers from limitations:

- The font may occupy all the private areas
- The font may have too many unencoded glyphs for the selected PUA.

This PR simplifies this approach by mapping unencoded glyphs beyond the upper limit of Unicode. Such mapping is possible with HarfBuzz - see comments by @khaledhosny in harfbuzz/uharfbuzz#22. In the legacy shaper this mapping was also tested and shown to work.

### DIFF
--- a/fontforgeexe/metricsview.c
+++ b/fontforgeexe/metricsview.c
@@ -3937,7 +3937,6 @@ static ShaperContext* MVMakeShaperContext(MetricsView *mv) {
     context->sf = mv->sf;
     context->mv = mv;
     context->apply_ticked_features = ApplyTickedFeatures;
-    context->fake_unicode = MVFakeUnicodeOfSc;
     context->get_enc_map = SFGetMap;
     context->get_char_width = MVCharWidth;
     context->get_metrics = MVGetMetrics;
--- a/fontforgeexe/shapers/shaper_shim.hpp
+++ b/fontforgeexe/shapers/shaper_shim.hpp
@@ -58,9 +58,6 @@ typedef struct shaper_context {
         SplineFont* sf, uint32_t* flist, uint32_t script, uint32_t lang,
         bool gpos_only, int pixelsize, SplineChar** glyphs);
 
-    // Map glyphs without unicode value to a private area
-    int (*fake_unicode)(MetricsView* mv, SplineChar* sc);
-
     // Get encoding map
     EncMap* (*get_enc_map)(SplineFont* sf);
 


fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -70,14 +76,23 @@ HarfBuzzShaper::HarfBuzzShaper(std::shared_ptr<ShaperContext> context)
         hb_blob_create(blob, blob_size, HB_MEMORY_MODE_WRITABLE, NULL, NULL);
 
     hb_ttf_face = hb_face_create(hb_ttf_blob, 0);
+    hb_font_t* hb_ttf_raw_font = hb_font_create(hb_ttf_face);
```
skef: Does the redeclaration make this local?

fontforgeexe/shapers/harfbuzz.cpp
```diff
@@ -70,14 +76,23 @@ HarfBuzzShaper::HarfBuzzShaper(std::shared_ptr<ShaperContext> context)
...
 
-    hb_ttf_font = hb_font_create(hb_ttf_face);
+    // To access unencoded glyphs with HarfBuzz, we need to assign them fake
+    // encodings above the canonic Unicode range. We also need to create a
+    // subfont with custom encoding resolution callback.
+    hb_ttf_font = hb_font_create_sub_font(hb_ttf_raw_font);
+    hb_font_funcs_t* funcs = hb_font_funcs_create();
```
skef, line 85: Usually in harfbuzz creates are paired with destroys at at given level. Is the reference counting right here?

fontforgeexe/metricsview.c
```diff
@@ -1923,8 +1912,7 @@ return;					/* Nothing changed */
 
     missing = 0;
     for ( tpt=pt; tpt<ept; ++tpt )
-	if ( mv->fake_unicode_base>0 && *tpt>=mv->fake_unicode_base &&
-		*tpt<=mv->fake_unicode_base+mv->sf->glyphcnt )
+	if ( *tpt>=FAKE_UNICODE_BASE && *tpt<=FAKE_UNICODE_BASE+mv->sf->glyphcnt )
```
skef: The guard above uses `tpt<...`. If it's a normal count are we going one off the end here?

fontforgeexe/shapers/harfbuzz.cpp
iorsh: Fixed

fontforgeexe/shapers/harfbuzz.cpp
iorsh: Fixed

fontforgeexe/metricsview.c
iorsh: `*tpt >= FAKE_UNICODE_BASE+mv->sf->glyphcnt` should never happen, and we would crash if it does. Fixed it for consistency. Keeping only `*tpt>=FAKE_UNICODE_BASE` would be more robust, but would also hide possible problems.

PR #5743, opened 2026-01-30
Fix CVE-2025-15270: Heap buffer overflow in SFD kern class parsing

ahmetfurkankavraz:
Add validation to check for negative kern class name length values before using them as array indices. This fixes a heap buffer overflow where malicious SFD files could supply negative length values (e.g., -1, -38, -139), causing writes to memory before the allocated buffer through `kc->firsts[i][temp] = '\0'` and `kc->seconds[i][temp] = '\0'` operations.

CVE-2025-15270  
ZDI-CAN-28563  
[ZDI-25-1194](https://www.zerodayinitiative.com/advisories/ZDI-25-1194/)

Type of change:
- **Bug fix**


ahmetfurkankavraz:
Reproducer can be found in https://sourceforge.net/p/fontforge/patches/41/

PR #5744, opened 2026-01-30
Fix import of multiple images in UI

### LINKED ISSUE #5738: Cannot import multiple SVGs for glyphs at one time
Cannot import multiple SVG files to be glyphs at one time

1. Select multiple glyph speces to import for
2. Open the import dialog and select multiple files to import
Image
3. Click OK
4. See the error dialog
Image
5. Sigh and import them one at a time.

P.S. I'm a new user of FontForge, but what's with the absolutely ancient UI elements from when I used to sneak into the graduate computing lab to use the colour X terminals in the early 1990s?

iorsh:
The file open dialog supports multiple selection, which it delivers in the form
```<dir>/<file_name>; <file_name>; <file_name>; ...; <file_name>```

Only the first filename contains the full path, and it is expected to be extracted and prepended to each of the semicolon-separated file names. This preprocessing does happen in `_FVMenuOpen()` when we try to open multiple font files, but the image import dialog ignores this convention in `FVImportImages()`, and subsequently fails for all the images after the first one.

Fixes #5738

### DIFF
--- a/gdraw/gfilechooser.c
+++ b/gdraw/gfilechooser.c
@@ -1838,6 +1838,7 @@ char** GFileChooserGetMultipleFiles(const char* multipath) {
 
         if (fpt == NULL) fpt = file + strlen(file);
         full = malloc(c_strlen(path) + 1 + (fpt - file) + 1);
+	full[0] = '\0';
 
         if (path) {
             strcpy(full, path);
@@ -1848,5 +1849,6 @@ char** GFileChooserGetMultipleFiles(const char* multipath) {
         file = fpt + 2;
     } while (*fpt != '\0');
 
+    free(path);
     return path_list;
 }

### CODE REVIEW

File: gdraw/gfilechooser.c [hunk 1]
```diff
@@ -1811,3 +1811,42 @@ return;
...
+
+    /* Count delimiting semicolons and allocate the NULL-terminated array of
+     * strings */
+    for (sc_count = 0; spt[sc_count];
+         spt[sc_count] == ';' ? sc_count++ : *spt++);
+    path_list = calloc(sc_count + 2, sizeof(char*));
+
+    eod = strrchr(multipath, '/');
+    if (eod != NULL) {
+        path = copyn(multipath, eod - multipath);
```
skef, line 1832: What frees this?

gdraw/gfilechooser.c
```diff
@@ -1811,3 +1811,42 @@ return;
...
+        fpt = strstr(file, "; ");
+
+        if (fpt == NULL) fpt = file + strlen(file);
+        full = malloc(c_strlen(path) + 1 + (fpt - file) + 1);
+
+        if (path) {
+            strcpy(full, path);
+            strcat(full, "/");
+        }
+        strncat(full, file, fpt - file);
```
skef, line 1847: Since you have the guard on path, full might be garbage at this point.

gdraw/gfilechooser.c
iorsh: Fixed

gdraw/gfilechooser.c
iorsh: Fixed.

skef (approved):
SGTM

PR #5746, opened 2026-01-31
[TEST] MSVC Python wheel compatibility - DO NOT MERGE

skef:
## Purpose

**This is a TEST PR to identify MSVC compilation issues. DO NOT MERGE.**

We are exploring building FontForge's Python modules with MSVC for compatibility with Python.org Windows Python installations (which use MSVC ABI, incompatible with our MinGW builds).

## What This Tests

The CI workflow has been temporarily simplified to build **Python modules only** (no GUI) on:
- Linux (Python 3.11)
- macOS (Python 3.11)
- **Windows MSVC** (Python 3.11, vcpkg dependencies)

## Changes

- `inc/basics.h`: Add `_MSC_VER` handling for UNUSED macro
- `CMakeLists.txt`: Add MSVC compiler options, `BUILD_PYTHON_WHEEL_ONLY` option
- `fontforge/CMakeLists.txt`: Use `WINDOWS_EXPORT_ALL_SYMBOLS` for MSVC
- `vcpkg.json`: Dependency manifest for Windows MSVC builds
- `.github/workflows/main.yml`: Simplified for Python module testing (TEMPORARY)

## Expected Outcome

The Windows MSVC job will likely fail, revealing compilation issues to fix iteratively. Linux and macOS should pass.

## Next Steps

1. Analyze CI failures
2. Fix MSVC compatibility issues
3. Restore full CI workflow
4. Keep MSVC job as addition for wheel builds

---
🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/PYTHON_MODULE_PLAN.md
+++ b/PYTHON_MODULE_PLAN.md
@@ -1,0 +1,335 @@
+# FontForge Python Module / PyPI Distribution Plan
+
+This document outlines the plan for making FontForge's Python bindings installable via `pip` and distributable through PyPI.
+
+## Goals
+
+1. Enable `pip install fontforge` from PyPI with pre-built wheels
+2. Support source distribution (sdist) for users who can build from source
+3. Maintain compatibility with distro packaging and full application builds
+4. Use a single CMake codebase for all build scenarios
+
+## Module Structure
+
+Two top-level modules (historical, not changing):
+- `fontforge` - Main module with font manipulation API
+- `psMat` - PostScript matrix operations
+
+## Build Scenarios
+
+| Scenario | Detection | GUI | `fontforge.version()` | libfontforge Location | Module Location |
+|----------|-----------|-----|----------------------|----------------------|-----------------|
+| **Wheel** | `SKBUILD` defined | OFF | CalVer (`2025.10.9`) | Wheel root | Wheel root |
+| **Distro module-only** | Manual cmake | OFF | CalVer (`2025.10.9`) | System libdir | site-packages |
+| **App + module** | Manual cmake | ON | Numeric (`20251009`) | System libdir | site-packages |
+
+The version format allows detection of context: CalVer = standalone module, Numeric = embedded in app.
+
+## Dependencies
+
+### Required (always linked)
+| Library | Purpose |
+|---------|---------|
+| libxml2 | XML/SVG parsing |
+| zlib | Compression |
+| libiconv | Character encoding |
+
+### Include in wheels (bundle where possible)
+| Library | Purpose | Notes |
+|---------|---------|-------|
+| FreeType | Bitmap rasterization | Preferred; falls back to internal rasterizer if unavailable |
+| libpng | PNG image support | Include |
+| libjpeg | JPEG image support | Include |
+| libtiff | TIFF image support | Include |
+| libspiro | Spiro curve support | Include |
+| woff2 | WOFF2 format support | Include |
+
+### Not needed for wheels
+| Library | Purpose | Notes |
+|---------|---------|-------|
+| GLIB | GObject, GIO | Only needed for GUI builds |
+| Gettext/Intl | Internationalization | Only needed for GUI builds |
+| Readline | CLI editing | Only needed for interactive scripting |
+
+## Versioning
+
+### Current System
+- `PROJECT_VERSION = 20251009` (YYYYMMDD integer)
+- Used for app builds and distro module installs
+
+### Python/PyPI Version
+- CalVer format: `2025.10.9` (YYYY.M.D, no leading zeros)
+- Derived automatically from PROJECT_VERSION
+- Used only for wheel builds
+
+### Transformation (in CMake)
+```cmake
+string(SUBSTRING "${PROJECT_VERSION}" 0 4 _FF_YEAR)
+string(SUBSTRING "${PROJECT_VERSION}" 4 2 _FF_MONTH)
+string(SUBSTRING "${PROJECT_VERSION}" 6 2 _FF_DAY)
+math(EXPR _FF_MONTH_INT "${_FF_MONTH}")
+math(EXPR _FF_DAY_INT "${_FF_DAY}")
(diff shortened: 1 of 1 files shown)


iorsh:
Hi @skef, just heads-up that I'm planning to move `shapers` module from `fontforgeexe` to `fontforge` and make Python wheel dependent on it. The rationale is to use it as a backend for `font.printSample()`

iorsh:
Accordingly, this would introduce HarfBuzz dependency

skef:
HarfBuzz shouldn't have any big problem with windows compilation. Does anything problematic come with it? Lots of GLib use? (That's what I'm trying to get rid of in the core.)

skef:
And, I guess, how does printSample() behave in a python module context?

skef:
Oh, I guess it doesn't actually print. Most of the image processing stuff is in the core code and should be OK. It's probably not a big deal.

iorsh:
I think it can perform actual printing with `fontforge.printSetup()`, but all this stuff is so ancient, it's hard for me to believe it can actually do modern OS-specific printing job. The useful part is creating PS/PDF files, and I want to use Cairo for that. The internet says that GLib is an optional dependency of Cairo, but standard Cairo builds may very well include it.

iorsh:
https://github.com/microsoft/vcpkg/issues/11640

skef:
Printing from Python seems -- unlikely. I was just checking to see why you wanted this moved to a context where that was the case. 

I'm less happy about moving Cairo down here than HarfBuzz. Is that really necessary?

iorsh:
I'm working on this: https://github.com/fontforge/fontforge/discussions/5633, and it's entirely Cairo-based. We can keep the existing Python printing intact, but the users will very soon start to complain about it being very limited comparatively to the Print dialog.

I can consider some non-Cairo graphics backend, but that doesn't sound trivial.

skef:
OK, but that is separate from the question of moving it into fontforge. You're simultaneously saying "let's make this more of a good UI thing" and "let's move this down into the core code". Maybe only one of those things makes sense?

skef:
Like, are we going to eventually link GTK down there too because it makes something easier?

iorsh:
Some GTK-based software seems to work somehow with MSVC (e.g. https://developer.gimp.org/core/setup/build/windows/), maybe it's worth to check their approach.

skef:
Working with MSVC is one thing, compiled into a non-UI python module is another.

skef:
I'm trying to get the modules pip-installable and working in the standard Windows python, something that's been desired for ages.

skef:
And so, yeah, I'm not thrilled about "let's throw fontforgeexe into fontforge", even if that's a hyperbolic description of your plans.

iorsh:
One approach is to disallow printing in that configuration. Printing is not something inherently necessary in font scripting. The ability to output bitmap with a font sample for regression testing can be useful though, and Cairo-based print would be able to do that.

Shaper is already a separate module, and I'm trying to keep it that way.

skef:
what is cairo buying us over freetype + harfbuzz?

iorsh:
I don't have immediate answer for that. Maybe that could be actually the way to go, I need to check.

iorsh:
The main advantage of Cairo over FreeType is its ability to create PDFs. If we wish to keep Cairo off the Python module, I think we can do the following:

- `shapers` target has optional dependency on HarfBuzz (you can specify `ENABLE_HARFBUZZ OFF` and have only built-in shaper available)
- Make `fontforge` target dependent on `shapers`
- Refactor printing modes `chars`, `waterfall`, `fontsample` etc. and separate them from the existing printing code into some page layout module. Later it would serve the Cairo-based Print UI too.
- To perform Python printing, compute glyph positions with `shapers` module (either with built-in shaper or HarfBuzz), arrange the text with the new layout module, and **feed it into the existing in-house PS/PDF creation code** instead of Cairo.
- In future, MSVC could potentially allow to add Windows-specific capabilities, such as Uniscribe shaping.

Does it make sense?

skef:
I've been thinking about this too. I think the below mostly aligns with what you're saying, but is a bit more of a "policy".

The main thing we're experiencing is that FontForge used to be a monolith, with very little in the way of dependencies (basically X stuff). That meant that having "everything" in "every context" was very cheap. Now it isn't a monolith, nor should it be. But that calls the everything/every-context goal into question. 

The prospect of Uniscribe shaping makes that all the more obvious: However useful that might be, it's only going to happen on some of our Python installations that happen to be on Windows. So clearly the functionality at the Python level would be different in different contexts. 

You want a unified shaper module, which is reasonable. I don't want to have to link "everything" into the Python context, especially because I think the use-cases are, well, corner cases. I think that is reasonable. 

I'm starting to think that the way out of this is to be more granular:

1. Start accepting that some Python-level functionality will differ in different contexts. We can document these cases.
2. Allow what is available at the Python level to be influenced by the build. So, if you want better shaping in the modules, you can add HarfBuzz. If you want to be able to "Print" (I'm still skeptical this actually works at the module level because I have no idea where Python would send the output), you can add Cairo. 
3. Be very liberal about what is available in Python in the application by default, and be pretty conservative about what is available in the modules by default. 

This means that some of our modules will be trickier to test because what is there will change. However, we've already sort of been accepting that, we've just been thinking of it as a "what the build found" thing, rather than a "what makes sense to put where" thing.

iorsh:
Ok, I think I understand your approach. Python modules with different capabilities would require more separation of code, which requires work, but eventually contributes to better modular architecture. I'll go along these lines.

For now, I'll start with moving `shapers` under `fontforge` target. Shapers can be compiled with or without HarfBuzz, so I think we are reasonably safe here.

Wanted to ask you to review #5736 on this occasion. I addressed your comments there.

skef:
I think I would even be OK with heading towards two flavors of wheel on PyPI, a minimal module and a maximal module, with separate release version patterns. So it doesn't need to be a matter of "application gets everything, modules get substantially less". I just don't want to move to a place where we're *requiring* a lot of new libraries at the module level.

PR #5749, opened 2026-02-08
Make fontforge module dependent on shapers

iorsh:
- `shapers` are now a submodule of `fontforge`, in the sense that `fontforge` depends on it.
- `shapers` have optional HarfBuzz dependency, controlled with `ENABLE_HARFBUZZ`
- Most changes are due to shaper code using the inner structure of `SplineChar`, which is not directly available, and provided with accessors.

The last item mostly reflects the necessity to return shaping results as a sequence of `SplineChar` pointers. This approach will be eventually eliminated when we further modernize the MetricsView and other capabilities related to shaping.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,7 @@
 name: CI
 on:
   push:
-    branches: [ master, releases/**, move_shapers ]
+    branches: [ master, releases/** ]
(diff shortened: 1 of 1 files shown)


skef (approved):
I don't see anything in here I have an issue with.

PR #5755, opened 2026-02-15
Fix crash in LoadPluginMetadata when distribution not found

### LINKED ISSUE #5754: LoadPluginMetadata crashes
If fontforgeVF plugin is installed with `--user` option, LoadPluginMetadata function crashes while calling Python API PyObject_HasAttrString. As result, Fontforge does not start. Also, `skippyplug` option does not work.

This occurs at b2789ba, on Linux (openSUSE 16.0), with Python 3.13.

gdb backtrace:
```
#46 0x00007ffff79b16d3 in LoadPluginMetadata (pe=0xbc1070) at ~/git/fontforge/fontforge/plugin.c:403
#47 DiscoverPlugins (do_import=1) at ~/git/fontforge/fontforge/plugin.c:475
#48 PyFF_ImportPlugins (do_import=1) at ~/git/fontforge/fontforge/plugin.c:521
#49 0x00007ffff79fc6f7 in PyFF_ProcessInitFiles (do_inits=do_inits@entry=1, do_plugins=do_plugins@entry=1) at ~/git/fontforge/fontforge/python.c:21168
#50 0x00000000005a7fce in fontforge_main (argc=<optimized out>, argv=<optimized out>) at ~/git/fontforge/fontforgeexe/startui.c:831
```

skef:
- Don't DECREF borrowed reference from PyDict_GetItemString
- Fix memory leak: DECREF globals and locals dictionaries
- Handle Py_None return (not just NULL) when no distribution matches
- Print original Python exception if present, then set formatted error

Fixes #5754

(Well, maybe).

### DIFF
--- a/fontforge/plugin.c
+++ b/fontforge/plugin.c
@@ -390,13 +390,20 @@ static void LoadPluginMetadata(PluginEntry* pe) {
     function_args = PyTuple_Pack(1, pe->entrypoint);
     if (function_args != NULL) {
         dist = PyObject_Call(func, function_args, NULL);
+        Py_DECREF(function_args);
     }
-    Py_XDECREF(func);
-    Py_XDECREF(function_args);
+    /* Note: func is a borrowed reference from PyDict_GetItemString, don't DECREF */
 
-    if (dist == NULL) {
-        PyErr_Print();
-        PyErr_SetString(PyExc_RuntimeError, "Could not retrieve distribution");
+    Py_DECREF(globals);
+    Py_DECREF(locals);
+
+    if (dist == NULL || dist == Py_None) {
+        if (dist == NULL) {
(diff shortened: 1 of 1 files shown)


iorsh:
I took a look at this issue too. There was a crash in `PyErr_Print()`, I think I exacerbated the issue by calling `PyErr_Print()` without prior check of `PyErr_Occurred()`. Do you think it would be worth to fix that too?

skef:
Probably wouldn't hurt

iorsh:
On the second thought, `PyErr_Occurred()` check did eliminate the crash, but it also masked the root cause, so let's better keep it as is.

iorsh (approved):
Looks good to me.

PR #5757, opened 2026-02-16
Fix SFGenerateFont() for Mac family flags

### LINKED ISSUE #5752: Generate Mac Family dialog silently enables Apple checkbox, causing subsequent font generation to fail
Hey there, and thanks a lot for this great tool 🙏

So, I clicked on "File > Generate Mac Family" by accident, which displayed this warning without allowing to go further:

> To generate a Mac family file, the current font must have plain (Normal, Regular, etc.) style, and there must be other open fonts with the same family name.

Then when I generated as usual with "File > Generate Fonts" then "Generate" button, it displayed the following error :

> Internal Error Attempt to output -2147483648 into a 16-bit field. It will be truncated and the file may not be useful.

And the resulting file was kind of broken.

I later realized that having clicked "File > Generate Mac Family" sneakily ticked the "Apple" box in "File > Generate Fonts > Options", and it took me a long while to figure out that I had to go into Options and untick that box.

I'm not sure what the best behavior would be, maybe at least to not tick that box, especially if the "Generate Mac Family" fails with this warning.

iorsh:
Mac-specific flags should not be remembered in the global preference of font generation flags, unless the user actually performed the generation.

This behavior change is consistent with the current behavior of "Generate" dialog, when the user changes font generation options.

Fixes #5752

### DIFF
--- a/fontforgeexe/savefontdlg.c
+++ b/fontforgeexe/savefontdlg.c
@@ -2007,11 +2007,6 @@ int SFGenerateFont(SplineFont *sf,int layer,int family,EncMap *map) {
 	    bitmaptypes[i].text = (unichar_t *) _((char *) bitmaptypes[i].text);
     }
 
-    if ( family==gf_macfamily ) {
-	old_sfnt_flags |=  ttf_flag_applemode;
-	old_sfnt_flags &= ~ttf_flag_otmode;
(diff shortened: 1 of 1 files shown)


skef (approved):
LGTM

PR #5763, opened 2026-02-27
Put app mids in include files

skef:
I'm working on some things that benefit from more centralization, and it is helpful to be able to deal with MIDs from more places. So this PR puts the MIDs in include files and also gives them distinct ranges.


iorsh (approved):
Looks good to me.

PR #5767, opened 2026-03-01
Fix Build Accented Glyph when only spacing accent exists

### LINKED ISSUE #4892: Wrong building of Uppercase + hungarumlaut
I think I have found a bug in building accented glyphs using the Element - Build - Build Accented Glyphs function. It relates to capital letters with Hungarumlaut (| O | and | U |). In lowercase letters, the construction occurs with a vowel (| o | or | u |) + | hungarumlautcmb |. If I then create the glyph | Hungarumlautcmb | (as I created | Acute |, | Grave |, | Circumflex | etc. for capital letters), - unlike the other cases where the construction procedure is correct - here the glyph is not used. In its place, the system uses the simple | Acute |. I believe that an incorrect pairing is programmed.

### LINKED ISSUE #5351: Combining accents with suffixes are ignored when building accented uppercase letters
While using "spacing" accents (e.g. `grave`) works as expected (e.g. `Grave`, `grave.cap`, see [Building Accented and other Composite Glyphs](https://fontforge.org/docs/techref/accented.html)) to build accented uppercase letters (e.g. `Agrave`), combining accents (e.g. `gravecomb`) only work when using a glyph name starting with a capital letter (e.g. `Gravecomb`). Suffixed combining accents (e.g. `gravecomb.cap`) do not work. I believe `gravecomb.cap` should work too, since the combining accents are more common today than spacing accents.

### Steps to reproduce
1. Open a new font in FontForge and change the encoding to *Unicode, BMP*.
2. Create the glyphs `U+0041` (`A`), `U+0300` (`gravecomb`) and `gravecomb.cap` and add a contour to each.
3. Build the glyph `U+00C0` (`Agrave`) (Element -> Build -> Build Accented Glyph)

This bug also occurs when using the Python API:
```py
import fontforge
font = fontforge.font()
font.reencode("UnicodeBMP")

A = font.createChar(0x0041)
gravecomb = font.createChar(0x0300)
gravecomb_cap = font.createChar(-1, "gravecomb.cap")
Agrave = font.createChar(0x00C0)

for g in [A, gravecomb, gravecomb_cap]:
  c = fontforge.contour()
  c.moveTo(0, 0).lineTo(0, 100).lineTo(100, 100).lineTo(100, 0).lineTo(0, 0)
  c.closed = True
  g.foreground += c

Agrave.build()
print(Agrave.references) # (('gravecomb', (1.0, 0.0, 0.0, 1.0, -50.0, 160.0), False), ('A', (1.0, 0.0, 0.0, 1.0, 0.0, 0.0), False))
```
Note that `'gravecomb'` is referenced instead of `'gravecomb.cap'`.

### Expected behavior
`Agrave` should be built from `A` and `gravecomb.cap` if it is defined.

### Actual behavior
`Agrave` is built from `A` and `gravecomb`, even though `gravecomb.cap` is defined.

### Versions
Tested on:
- Ubuntu 22.04.3 LTS with FontForge build from main branch (d7c68971)
- Windows 10 Pro with FontForge 20230101

### Possible solution/fix
This seems to be a bug in `fvcomposite.c` in the `GetGoodAccentGlyph` function:
The condition on [line 1845](https://github.com/fontforge/fontforge/blob/d7c6897111e61c1a305099cab312b090d2aeb830/fontforge/fvcomposite.c#L1845) includes `(PreferSpacingAccents || !haschar(sf,uni,NULL))`. Why should `gravecomb.cap` only be applied to uppercase letters if there is no `gravecomb` (assuming we don't `PreferSpacingAccents`)? I think `gravecomb.cap` should always be used for uppercase glyphs if it is available (this is checked a few lines below).

This extra condition was added [here](https://github.com/fontforge/fontforge/pull/3774/commits/cf85d53afa634237793804fec6a3c88cbdd63aac#diff-6a3219cfd73addb4297856f352be3d50714556427090ce5d50edb0903a721318R1840) as part of PR #3774. I don't understand the code 100%, but I think the condition should be used to decide whether we want to check the name of `uni` with the suffixes, or whether we want to loop over `accents[uni-BottomAccent]` and check their names appended with the suffixes.

If you remove the `!` in [line 1845](https://github.com/fontforge/fontforge/blob/d7c6897111e61c1a305099cab312b090d2aeb830/fontforge/fvcomposite.c#L1845), `gravecomb.cap` is used to build `Agrave` (in the GUI, the `gravecomb.cap` reference is shown and the above Python script prints `(('gravecomb.cap', `...). However, this quick workaround most likely breaks the `PreferSpacingAccents` preference and possibly other things.

skef:
## Summary

I ran into a problem testing menu stuff and Claude and I fixed it, and while we were there we fixed the other two things below (or tried to). 

fixes #4892
fixes #5351

- Fixes "Build Accented Glyph" being incorrectly disabled when a font has spacing accents (e.g., U+00B4 ACUTE ACCENT) but lacks combining accents (e.g., U+0301 COMBINING ACUTE ACCENT)

## Problem

When `PreferSpacingAccents` is false (the default), `GetGoodAccentGlyph()` would return NULL even after finding a valid spacing accent in the lookup table. This caused "Build Accented Glyph" to be grayed out for fonts that only have spacing accents.

The bug was in this logic:
```c
if ( *apt!='\0' && apt<end && PreferSpacingAccents)
    ach = *apt;
else if ( haschar(sf,uni,dot) ... )
    ach = uni;
```

The first condition only used the found spacing accent if `PreferSpacingAccents` was true. The fallback tried to use the combining accent, but if it doesn't exist (which is why we entered this branch in the first place), nothing was set.

## Fix

Remove the `PreferSpacingAccents` check from the condition. We enter this branch when EITHER `PreferSpacingAccents` is true OR the combining accent doesn't exist. In both cases, if we found a valid spacing accent, we should use it.

## Test plan

1. Open a font that has U+00B4 (ACUTE ACCENT) but NOT U+0301 (COMBINING ACUTE ACCENT)
2. Select U+00DD (LATIN CAPITAL LETTER Y WITH ACUTE) or similar
3. Before fix: "Build Accented Glyph" is disabled
4. After fix: "Build Accented Glyph" is enabled and works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/fontforge/fvcomposite.c
+++ b/fontforge/fvcomposite.c
@@ -168,7 +168,7 @@ static const char *uc_accent_names[] = {
     "Diaeresis",
     NULL,
     "Ring",
-    "Hungarumlaut",  /* was incorrectly "Acute" - fixes #4892 */
+    "Hungarumlaut",
     "Caron"
 };
 


fontforge/fvcomposite.c
```diff
@@ -168,7 +168,7 @@ static const char *uc_accent_names[] = {
     "Diaeresis",
     NULL,
     "Ring",
-    "Acute",
+    "Hungarumlaut",  /* was incorrectly "Acute" - fixes #4892 */
```
iorsh: No need for comment here

skef: Removed

iorsh (approved):

PR #5777, opened 2026-03-18
Fix some compilation warnings

iorsh:


fontforge/fvcomposite.c
```diff
@@ -1154,8 +1154,7 @@ return( greekalts );
 return( greekalts );
     }
 
-    if ( sf->uni_interp==ui_adobe && base>=0xf600 && base<=0xf7ff &&
-	    adobes_pua_alts[base-0xf600]!=0 )
+    if ( sf->uni_interp==ui_adobe && base>=0xf600 && base<=0xf7ff )
```
skef: Why are we not changing these checks to `adobes_pua_alts[base-0xf600][0] != 0` ?

iorsh: Agreed, fixed it.

PR #5792, opened 2026-04-06
Initialize legacy resources before converting them to CSS

iorsh:
Fixes issue in https://github.com/fontforge/fontforge/issues/5786#issuecomment-4188757586


frank-trampe (approved):
I can't readily verify whether this fixes anything, but it looks coherent and non-harmful.

PR #5794, opened 2026-04-07
Automatic Windows changelog from latest git commits

iorsh:
The current changelog file in Windows is outdated, and keeping it up-to-date is a useless burden. Now as only properly reviewed PRs enter master, its git log is sufficient as a changelog.


frank-trampe (approved):
Given that the oldest entry in the changelog came long after the origin of the current git repository, this change seems good. I also tested the `git` command.

PR #5796, opened 2026-04-10
GTK-based "Show Histogram" dialog

iorsh:
This PR replaces legacy Histogram dialog with a GTK-based one.

New UI: 
histogram

Old UI: 
image

**UX changes**

1. Small selection warning is now inline:
inline_warning_proj

instead of popup window
inline_warning_sys

2. User input is validated
bad_input_proj

3. Aggregate value for "Sum around" replaced with average, which is a more standard way to filter data. Averages are plotted over the histogram and do not override the original values
averaging


iorsh:
Add warning triangle to the offending input

iorsh:
Run valgrind

iorsh:
- Added warning icon to the highlighted input error. 
   image
- Valground the dialog and fixed a memory leak.

frank-trampe (approved):

PR #5803, opened 2026-04-20
Update CI build

rettinghaus:
### Type of change
- **Bug fix**
  - update used actions in CI
  - remove brew installs for pre-installed software in macos runner (see: https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md)
  - silence warnings for macos brew

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,7 +94,7 @@ jobs:
       - name: Install dependencies
         run: |
           brew install pkg-config ruby cmake ninja
-          brew install cairo coreutils fontconfig gettext giflib gtk+3 gtkmm3 jpeg libpng libspiro libtiff libtool libuninameslist python@3 wget woff2
+          brew install cairo coreutils fontconfig gettext giflib gtk+3 gtkmm3 jpeg libpng libspiro libtiff libtool libuninameslist python@3 wget woff2 --quiet
           PREFIX=$GITHUB_WORKSPACE/target
           echo "PREFIX=$PREFIX" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -88,13 +88,12 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           path: repo
       - name: Install dependencies
         run: |
-          brew install pkg-config ruby cmake ninja
-          brew install cairo coreutils fontconfig gettext giflib gtk+3 gtkmm3 jpeg libpng libspiro libtiff libtool libuninameslist python@3 wget woff2
+          brew install cairo coreutils fontconfig gettext giflib gtk+3 gtkmm3 jpeg libpng libspiro libtiff libtool libuninameslist python@3 wget woff2 --quiet
```
iorsh, line 96: Please keep the installation loud. Homebrew packages break a lot, so logging is important here.

.github/workflows/wheels.yml
```diff
@@ -37,18 +37,18 @@ jobs:
           # macOS x86_64
           - os: macos
             arch: x86_64
-            runner: macos-13
+            runner: macos-15-intel
```
iorsh: Please revert the runner changes. We intentionally run CI on the oldest system available to make sure the compiled binaries are compatible with as much platform versions as possible.

iorsh (changes requested):

.github/workflows/main.yml
rettinghaus: Errors will still be shown, it just silences the warnings about already installed packages.

iorsh (approved):

PR #5805, opened 2026-04-23
fix pyhook install dir for /usr/local

nadalaba:
On my machine:
`$ python3 -c "import sysconfig as sc; print(sc.get_path('platlib'))"`
/usr/local/lib/python3.13/dist-packages
`$ python3 -c "import sysconfig as sc; print(sc.get_config_var('platbase'))"`
/usr
So making `platbase=.` only removes `/usr/` and keeps `local/`, making `${CMAKE_INSTALL_PREFIX}/${PYHOOK_INSTALL_DIR}`:
`/usr/local/local/lib/python3.13/dist-packages/`, when the install prefix is the default `/usr/local`.

**[EDIT]:** looking at CI, it seems that this behavior is intended. I wonder why not install the modules to be useable by the existing python installation.

### DIFF
--- a/pyhook/CMakeLists.txt
+++ b/pyhook/CMakeLists.txt
@@ -44,10 +44,11 @@ else()
       OUTPUT_VARIABLE _platlib_path_abs
       OUTPUT_STRIP_TRAILING_WHITESPACE)
     if(NOT _platlib_path_abs OR _platlib_path_abs_result)
-      message(FATAL_ERROR "Could not determine the path to install pyhook binaries to, got '${PYHOOK_INSTALL_DIR}' with return value ${_pyhook_install_dir_result}")
+      message(FATAL_ERROR "Could not determine the path to install pyhook binaries to, got '${_platlib_path_abs}' with return value ${_platlib_path_abs_result}")
     endif()
     # Convert platlib dir from absolute to prefix-relative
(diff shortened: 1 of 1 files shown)


iorsh:
I remember these paths required a lot of tuning to achieve satisfactory performance on all platforms. I did some of it, and then it still wasn't good enough, so @jtanx stepped in in #5423.

@nadalaba, could you please clarify what is not working as intended?

nadalaba:
@iorsh I'm not really sure what's intended. Here is what I expected:

When installing fontforge from source, installed python modules will be accessible to the system's python executable [source](https://fontforge.org/docs/scripting/python.html#fontforge-as-a-python-extension).

So I expect it to look for my system's python library path `'platlib'`, and install the modules there (on my machine this is on /usr/local/lib/python3.13/dist-packages).

However, currently it is installing the modules on a new path inside `CMAKE_INSTALL_PREFIX` (on my machine it is installing them to `/usr/local/local/lib/python3.13/dist-packages` -double `local/`-), where there could be no python installation, making the installed modules importable from nowhere.

nadalaba:
After further examining the build system, it seems that I misunderstood the purpose of the "install" target.
Maybe what I needed (move python modules to system's python) can be added as a different cmake target (e.g., "py_install") intended to be used after "install"?

nadalaba:
superseded by #5841

PR #5818, opened 2026-05-22
Fix segfault in Python `font.italicize` function

gucci-on-fleek:
Previously, trying to use any of the `XX_condense` arguments in the Python `font.italicize` function would cause a segfault because `SquashParse` incorrectly passed the arguments of `PyArg_ParseTuple` by value instead of using a pointer. This commit fixes the issue by taking the address of the arguments.

This issue appears to have been present since the initial commit that added the code (4b87646), which I guess suggests that nobody has tried to use the `XX_condense` arguments in Python before, or that I've missed something obvious and my fix is actually incorrect.

### Type of change

- **Bug fix**


iorsh (approved):
Looks good to me

PR #5819, opened 2026-05-22
Conditionally use Python3::Module Target Linking in CMake

justdan96:
Add conditional linking for Python3::Module target in CMakeLists.txt, fixes an issue where some Python installations wouldn't expose `Python3::Python` but they would expose `Python3::Module`. I've tested that with this code `poetry` can successfully pull in `fontforge` as a git dependency.

### Type of change

- **Bug fix**


iorsh:
Looks reasonable to me. @skef, do you have any comments?

skef (approved):
I vaguely remember some trouble choosing between these targets, but if the current heuristic isn't working we can deal with further issues downstream -- this is configured as a backup.

PR #5820, opened 2026-05-24
Expose the `xheight_percent` parameter to Python

gucci-on-fleek:
Previously, the Python `font.italicize()` function had no interface to change the x-height of the font, which meant that the function unconditionally used the default value of 95%. This commit exposes this parameter to Python via a new keyword argument.

### Type of change

- **New feature**


iorsh (approved):
Looks good to me

PR #5822, opened 2026-05-24
Fix pyhook initialization

### LINKED ISSUE #5811: Cannot open a font from ffpython code if it contains pickled data in font.persistent
I am not sure whether the present problem is directly related to [https://github.com/fontforge/fontforge/pull/5787] or not, but I now need to open a font from ffpython code outside of Fontforge proper.  If there is any pickled data stored in `font.persistent`, I get the following type of error: 

```
Q:\My Drive\FontForge\Pseudofonts>ffpython realfactory.py
Loading SFD: C:\Users\mpugh5\Documents\U2460_Regular.sfd
Fatal Python error: PyImport_AppendInittab: PyImport_AppendInittab() may not be called after Py_Initialize()
Python runtime state: initialized

Current thread 0x000086d8 (most recent call first):
  File "Q:\My Drive\FontForge\Pseudofonts\realfactory.py", line 39 in main
  File "Q:\My Drive\FontForge\Pseudofonts\realfactory.py", line 59 in <module>

Extension modules: __FontForge_Internals___, fontforge (total: 2)
```

The data that is stored there is perfectly accessible from inside FontForge using the tools I am building.  And the font opens fine in Fontforge. Line 39 in the code is just `pf = fontforge.open(SFD_PATH)`.

iorsh:
Fix embedded Python initialization in pyhook mode, so that it doesn't call `PyImport_AppendInittab()` after `Py_Initialize()`. This issue was specific to pickling, as pickle module triggered `RegisterAllPyModules()` while opening a font with persistent objects.

Fixes #5811


fontforge/python.cpp
```diff
@@ -20819,6 +20820,13 @@ static void FontForge_InitializeEmbeddedPythonEx(int argc, wchar_t **argv) {
     if ( python_initialized )
 	return;
 
+    /* If Python is already running (pyhook/module mode), do not perform
+     * embedded initialization steps such as inittab registration or
+     * Py_InitializeFromConfig().
+     */
+    if ( Py_IsInitialized() )
```
skef, line 20827: If we're calling this in multiple places now, it seems like `python_initialized` could use a better name to differentiate whatever it tracks from the state that python is tracking.

iorsh: Renamed to `embedded_python_initialized`

skef (approved):
LGTM

PR #5823, opened 2026-05-26
Fix crash on GTK initialization

### LINKED ISSUE #5786: Fontforge crashing when I try to add a new encoding slot
Hello. I've been attempting to make a script for a conlang which requires a lot of ligatures, but I've been unable to add new encoding slots for a while now. Every time I try to add them, no menu pops up, it freezes, and then it crashes. Does anyone know how to fix this?

iorsh:
Fixes #5786


skef (approved):
LGTM

PR #5826, opened 2026-05-29
Allow multiple contextual subtable rules

### LINKED ISSUE #5825: Python API cannot create multi-rule class contextual GPOS/GSUB subtables
## Problem

FontForge's Python API can create contextual subtables with `type="class"` via `font.addContextualSubtable()`, and FontForge correctly compiles those subtables to OpenType contextual subformat 2. However, the API only accepts a single rule string per subtable, so a script that needs many class contextual rules must create one OpenType subtable per rule.

This is a problem for fonts with large generated contextual positioning systems. My concrete use case is a Byzantine music notation font. The font needs to raise some above-note marks to avoid collisions with other above-note marks. A simplified example is:

- raise an ison mark when it follows a gorgon mark on certain base glyphs
- ignore unrelated intervening marks by using a mark attachment/filtering class
- use the same technique for other frequent collisions, such as koronis over gorgon/fthora and fthora over klasma/apli/gorgon

The generated rules are naturally class-context rules:

```text
baseClass gorgonClass | isonClass @<raise_020> |
baseClass fthoraClass | isonClass @<raise_040> |
...
```

For my current narrow ison/gorgon case, there are about 938 contextual rules in one font. For the broader collision-solving case, there are about 13,834 contextual rules in one font. If each rule becomes a separate subtable, the lookup becomes a very large flat list of subtables. In shaping engines such as HarfBuzz, that means the engine has to scan many subtables/rules once the current glyph is a candidate. It also risks approaching implementation limits: HarfBuzz's has a 16,384 subtable sanitizer cap.

Class contextual subformat 2 is the OpenType representation that should address this: one subtable can hold shared class definitions and multiple rules, grouped by the current/input class. FontForge's UI and SFD model appear to support this, but the Python API cannot currently create it.

## Expected Behavior

FontForge's Python API should provide a way to create one class contextual subtable containing multiple rules, producing one OpenType ChainContextPos/ChainContextSub Format 2 subtable with multiple ClassRule records.

For example, an API like this would be sufficient:

```python
font.addContextualSubtable(
    "ctx",
    "ctx-1",
    "class",
    (
        "1 2 | 1 @<raise_020> |",
        "1 3 | 1 @<raise_040> |",
    ),
    bclasses=(None, ("baseA", "baseB"), ("markA",), ("markB",)),
    mclasses=(None, ("target",)),
)
```

Expected compiled result:

- one contextual lookup
- one contextual subtable
- `ChainContextPos Format 2` or `ChainContextSub Format 2`
- two rules inside that one subtable

## Actual Behavior

`font.addContextualSubtable()` accepts only one rule string and creates an FPST with `rule_cnt = 1`.

As a result, the only currently available Python workaround is to call `addContextualSubtable()` once per rule:

```python
font.addContextualSubtable(..., "class", "1 2 | 1 @<raise_020> |", ...)
font.addContextualSubtable(..., "class", "1 3 | 1 @<raise_040> |", ...)
```

That produces two OpenType subtables, not one subtable with two rules. Scaling this pattern to thousands of generated rules creates thousands of subtables.

This is especially unfortunate because FontForge already has the pieces internally:

- The Python API supports `type="class"`.
- The OpenType writer emits class contextual subtables as subformat 2.
- The UI path can store multiple class rules in one FPST.
- The SFD format can represent `ChainPos2: class` / `ChainSub2: class` with `rule_cnt > 1`.

## Minimal Reproducible Example

Save this as `repro.py` and run it with `fontforge -script repro.py`.

```python
import fontforge
from fontTools.ttLib import TTFont


def make_glyph(font, name, codepoint):
    glyph = font.createChar(codepoint, name)
    pen = glyph.glyphPen()
    pen.moveTo((0, 0))
    pen.lineTo((500, 0))
    pen.lineTo((500, 500))
    pen.lineTo((0, 500))
    pen.closePath()
    glyph.width = 600
    return glyph


font = fontforge.font()
font.encoding = "UnicodeFull"
font.fontname = "ContextClassRepro"
font.familyname = "Context Class Repro"
font.fullname = "Context Class Repro"

make_glyph(font, "baseA", 0xE000)
make_glyph(font, "baseB", 0xE001)
make_glyph(font, "markA", 0xE002)
make_glyph(font, "markB", 0xE003)
make_glyph(font, "target", 0xE004)

features = (("mark", (("DFLT", ("dflt",)),)),)

font.addLookup("raise_020", "gpos_single", (), ())
font.addLookupSubtable("raise_020", "raise_020-1")
font["target"].addPosSub("raise_020-1", 0, 20, 0, 0)

font.addLookup("raise_040", "gpos_single", (), (), "raise_020")
font.addLookupSubtable("raise_040", "raise_040-1")
font["target"].addPosSub("raise_040-1", 0, 40, 0, 0)

font.addLookup("ctx", "gpos_contextchain", (), features, "raise_040")

# This is the currently required workaround: one addContextualSubtable call per
# rule. Each call creates a separate class contextual subtable.
font.addContextualSubtable(
    "ctx",
    "ctx-1",
    "class",
    "1 2 | 1 @<raise_020> |",
    bclasses=(None, ("baseA", "baseB"), ("markA",), ("markB",)),
    mclasses=(None, ("target",)),
)
font.addContextualSubtable(
    "ctx",
    "ctx-2",
    "class",
    "1 3 | 1 @<raise_040> |",
    "ctx-1",
    bclasses=(None, ("baseA", "baseB"), ("markA",), ("markB",)),
    mclasses=(None, ("target",)),
)

font.generate("ContextClassRepro.otf")

tt = TTFont("ContextClassRepro.otf")
gpos = tt["GPOS"].table
ctx_lookup = gpos.LookupList.Lookup[-1]

print("Lookup type:", ctx_lookup.LookupType)
print("Subtable count:", len(ctx_lookup.SubTable))
for index, subtable in enumerate(ctx_lookup.SubTable, start=1):
    inner = getattr(subtable, "ExtSubTable", subtable)
    print("Subtable", index, type(inner).__name__, "Format", inner.Format)
```

Actual output:

```text
Lookup type: 8
Subtable count: 2
Subtable 1 ChainContextPos Format 2
Subtable 2 ChainContextPos Format 2
```

The desired output for this example is:

```text
Lookup type: 8
Subtable count: 1
Subtable 1 ChainContextPos Format 2
Rule count inside the class set: 2
```

The currently desired API is not available. For example, there is no working equivalent of:

```python
font.addContextualSubtable(
    "ctx",
    "ctx-1",
    "class",
    (
        "1 2 | 1 @<raise_020> |",
        "1 3 | 1 @<raise_040> |",
    ),
    bclasses=(None, ("baseA", "baseB"), ("markA",), ("markB",)),
    mclasses=(None, ("target",)),
)
```

## Motivation

For small hand-authored features, one contextual subtable per rule is tolerable. For generated fonts, it becomes a large performance and size problem.

In my real font:

- narrow current case: about 938 contextual rules
- broader collision-solving case: about 13,834 contextual rules

With the current Python API, those become about 938 or 13,834 subtables. With multi-rule class subtables, these could be grouped into a much smaller number of subformat 2 subtables, sharing class definitions and letting shaping engines dispatch through class rule sets instead of scanning a flat pile of subtables.

## Solution

A backwards-compatible solution would be to extend `font.addContextualSubtable()` so that the `rule` argument may be either:

- a string, preserving current behavior
- a sequence of strings, creating one FPST with `rule_cnt = len(rules)`

Implementation sketch:

1. In the Python binding, detect whether `rule` is a string or a sequence.
2. For a string, keep the current behavior.
3. For a sequence, allocate `fpst->rules = calloc(rule_count, sizeof(struct fpst_rule))`.
4. Call `FPSTRule_From_Str()` once per rule.
5. Preserve the existing `bclasses`, `mclasses`, `fclasses`, and class-name arguments as shared class definitions for the subtable.

An alternative API would also work:

```python
font.addContextualSubtable(...)
font.addContextualRule("ctx-1", "1 3 | 1 @<raise_040> |")
```

However, allowing a sequence of rule strings is probably the smallest backwards-compatible change.

## Alternatives Considered

- Creating one class contextual subtable per rule works functionally but creates too many subtables.
- `font.mergeLookupSubtables()` is not a workaround because contextual lookup types are rejected by `SFSubTablesMerge()`.
- Adobe feature-file import does not appear to solve this because FontForge's feature parser chooses glyph or coverage contextual formats for contextual sequences, not class contextual format with shared ClassDef data.
- Direct SFD post-processing may be possible because the SFD format can represent `ChainPos2: class` with multiple rules, but that is fragile compared to a supported Python API.

## Testing Plan

1. Add support for passing a sequence of rule strings to `font.addContextualSubtable()` when `type="class"`.
2. Modify the repro above to replace the two `addContextualSubtable()` calls with one call whose `rule` argument is a tuple of two rule strings.
3. Generate the OTF.
4. Inspect the GPOS table with fontTools.
5. Verify:
   - the contextual lookup has exactly one subtable
   - the subtable is `ChainContextPos Format 2`
   - the relevant `ChainPosClassSet` contains two `ChainPosClassRule` records
   - shaping still applies the correct nested lookup for both class rules

This should also be tested for GSUB contextual chaining, and for non-chaining contextual GPOS/GSUB lookups if the same API path is shared.

- [ ] Screenshot

FontForge version:

```text
fontforge 20251009
build date: 2026-01-16 00:00 UTC
```

basil:
### Type of change
- **New feature**
- **Non-breaking change**

See #5825. Allow `font.addContextualSubtable()` to accept a sequence of rule strings for glyph and class contextual subtables, creating multiple rules in one subtable while preserving existing single-string behavior. We include API documentation updates and a regression test covering multi-rule class/glyph subtables plus validation for invalid rule sequences.

Fixes #5825


fontforge/python.cpp
```diff
@@ -17389,6 +17392,78 @@ return( NULL );
 	}
     }
 
+    if ( PyUnicode_Check(rule_obj) || PyBytes_Check(rule_obj)) {
```
iorsh: Bytes object is generally not allowed for string input. For consistency, please don't allow it here too.

fontforge/python.cpp
```diff
@@ -17389,6 +17392,78 @@ return( NULL );
...
+	    PyObject *rule_item = PySequence_Fast_GET_ITEM(rule_sequence,i);
+	    const char *rule_str;
+	    Py_ssize_t rule_len;
+	    if ( PyUnicode_Check(rule_item)) {
+		rule_str = PyUnicode_AsUTF8AndSize(rule_item,&rule_len);
+		if ( rule_str==NULL ) {
+		    Py_DECREF(rule_sequence);
+return( NULL );
+		}
+	    } else if ( PyBytes_Check(rule_item)) {
```
iorsh: No bytes input.

fontforge/python.cpp
```diff
@@ -17389,6 +17392,78 @@ return( NULL );
 	}
     }
 
+    if ( PyUnicode_Check(rule_obj) || PyBytes_Check(rule_obj)) {
+	const char *rule_str;
+	Py_ssize_t rule_len;
+	if ( PyUnicode_Check(rule_obj)) {
+	    rule_str = PyUnicode_AsUTF8AndSize(rule_obj,&rule_len);
```
iorsh: Please use PyUnicode_AsUTF8() for consistency, we don't expect NULL characters in user input. NULL char check below not needed too.

fontforge/python.cpp
```diff
@@ -17485,19 +17560,22 @@ return( NULL );
...
-	free(msg);
-	msg = NULL;
-    }
-    if ( msg!=NULL ) {
-	PyErr_Format(PyExc_TypeError, "%s", msg );
-	free(msg);
+    for ( size_t i=0; i<rules.size(); ++i ) {
+	msg = FPSTRule_From_Str( sf,fpst,&fpst->rules[i],rules[i].data(),&is_warning);
+	if ( is_warning ) {
+	    LogError("%s",msg);
```
iorsh: Please mention rule number in the error message:
`LogError(_("Warning in rule %d: \"%s\"",i,msg);`

fontforge/python.cpp
```diff
@@ -17485,19 +17560,22 @@ return( NULL );
...
+	    free(msg);
+	    msg = NULL;
+	}
+	if ( msg!=NULL ) {
+	    PyErr_Format(PyExc_TypeError, "%s", msg );
```
iorsh: Same here - rule number

tests/test_contextual_multirule.py
```diff
@@ -0,0 +1,132 @@
...
+    (
+        "target @<raise_020>",
+        "baseA @<raise_040>",
+    ),
+)
+
+assert font.getLookupSubtables("glyph_ctx") == ("glyph_ctx-1",)
+
+with tempfile.TemporaryDirectory() as tmpdirname:
+    sfd_path = os.path.join(tmpdirname, "ContextClassMultiRule.sfd")
```
iorsh, line 114: Instead of examining SFD contents, please close font, load SFD, dump feature file and examine its contents.

basil:
@iorsh Thank you, addressed in commit 0292d0c60.

fontforge/python.cpp
```diff
@@ -17389,6 +17392,56 @@ return( NULL );
 	}
     }
 
+    if ( PyUnicode_Check(rule_obj)) {
+	const char *rule_str;
+	rule_str = PyUnicode_AsUTF8(rule_obj);
+	if ( rule_str==NULL )
+return( NULL );
+	rules.push_back(std::string(rule_str));
+    } else if ( PyBytes_Check(rule_obj)) {
```
iorsh: This branch not needed, bytes will fall into PySequence_Fast() check

basil:
Thanks @iorsh, fixed in commit 288ea9fa7783b50a8075c2796db0368153724b6c

iorsh (approved):
Looks good to me

basil:
@iorsh Thank you very much!

PR #5833, opened 2026-06-06
Optimize `SCOrderAP` for large input

andiwand:
I noticed that a small input PDF to https://github.com/pdf2htmlEX/pdf2htmlEX took seconds to load which was quite unexpected. After quick profiling it pointed to `SCOrderAP` in fontforge. I only briefly checked the code myself and was surprised by the nested loop for sorting the APs. This special PDF contains a font that has characters with >100k APs which hits the $N^2$ loop quite hard.

I then sent off a coding agent to optimize this function which now detects if the input is already sorted, sorted in reverse, or defaults to using `qsort`. https://github.com/fontforge/fontforge/commit/2ea680cb6c39750f7d594bd379398f38000815ea

After that the agent also suggested to reverse the order of the APs already during reading as they are stored in memory in the opposite direction, resulting in the initial poor sorting performance. https://github.com/fontforge/fontforge/commit/78ba37d0e2bafe6e699c59d4ad6780a6fbc001ea

I think https://github.com/fontforge/fontforge/commit/78ba37d0e2bafe6e699c59d4ad6780a6fbc001ea would now actually do the trick by itself but maybe `SCOrderAP` is used in other places and still benefits from the changes.

### Type of change
- **Non-breaking change**

### DIFF
--- a/fontforge/parsettfatt.c
+++ b/fontforge/parsettfatt.c
@@ -867,7 +867,8 @@ static void MarkGlyphsProcessLigs(FILE *ttf,int baseoffset,
 	AnchorClass **classes) {
     int basecnt,compcnt, i, j, k, kbase;
     uint16_t *loffsets, *aoffsets;
-    SplineChar *sc;
+	SplineChar *sc;
+	AnchorPoint **tail, *ap;
 
     fseek(ttf,baseoffset,SEEK_SET);
     basecnt = getushort(ttf);
@@ -880,9 +881,14 @@ return;
     for ( i=0; i<basecnt; ++i )
 	loffsets[i] = getushort(ttf);
     for ( i=0; i<basecnt; ++i ) {
-	sc = info->chars[baseglyphs[i]];
-	if ( baseglyphs[i]>=info->glyph_cnt || sc==NULL )
+		if ( baseglyphs[i]>=info->glyph_cnt )
     continue;
+		sc = info->chars[baseglyphs[i]];
+		if ( sc==NULL )
+    continue;
+		tail = &sc->anchor;
+		while ( *tail!=NULL )
+	    tail = &(*tail)->next;
 	fseek(ttf,baseoffset+loffsets[i],SEEK_SET);
 	compcnt = getushort(ttf);
 	if ( feof(ttf)) {
@@ -895,9 +901,11 @@ return;
 	    aoffsets[k] = getushort(ttf);
 	for ( k=kbase=0; k<compcnt; ++k, kbase+=classcnt ) {
 	    for ( j=0; j<classcnt; ++j ) if ( aoffsets[kbase+j]!=0 ) {
-		sc->anchor = readAnchorPoint(ttf,baseoffset+loffsets[i]+aoffsets[kbase+j],
-			classes[j], at_baselig,sc->anchor,info);
-		sc->anchor->lig_index = k;
+		ap = readAnchorPoint(ttf,baseoffset+loffsets[i]+aoffsets[kbase+j],
+			classes[j], at_baselig,NULL,info);
+		ap->lig_index = k;
+		*tail = ap;
+		tail = &ap->next;
 	    }
 	}
 	free(aoffsets);
--- a/fontforge/splinechar.c
+++ b/fontforge/splinechar.c
@@ -1054,35 +1054,69 @@ void AltUniAdd_DontCheckDups(SplineChar *sc,int uni) {
     }
 }
 
+struct aporder {
+    AnchorPoint *ap;
+    int order;
+};
+
+static int SCOrderAPCmp(const void *a, const void *b) {
+    const struct aporder *ap_a = a, *ap_b = b;
+
+    if ( ap_a->ap->lig_index < ap_b->ap->lig_index )
+return( -1 );
+    if ( ap_a->ap->lig_index > ap_b->ap->lig_index )
+return( 1 );
+return( ap_a->order - ap_b->order );
+}
+
 void SCOrderAP(SplineChar *sc) {
-    int lc=0, cnt=0, out=false, i,j;
-    AnchorPoint *ap, **array;
+    int cnt=0, ascending=true, descending=true, first=true, prev=0, i;
+    AnchorPoint *ap, *next, *rev = NULL;
+    struct aporder *array;
     /* Order so that first ligature index comes first */
 
     for ( ap=sc->anchor; ap!=NULL; ap=ap->next ) {
-	if ( ap->lig_index<lc ) out = true;
-	if ( ap->lig_index>lc ) lc = ap->lig_index;
+	if ( first ) {
+	    prev = ap->lig_index;
+	    first = false;
+	} else {
+	    if ( ap->lig_index < prev )
+		ascending = false;
+	    if ( ap->lig_index > prev )
+		descending = false;
+	    prev = ap->lig_index;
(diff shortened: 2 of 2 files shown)


fontforge/parsettfatt.c
```diff
@@ -895,9 +901,11 @@ return;
 	    aoffsets[k] = getushort(ttf);
 	for ( k=kbase=0; k<compcnt; ++k, kbase+=classcnt ) {
 	    for ( j=0; j<classcnt; ++j ) if ( aoffsets[kbase+j]!=0 ) {
-		sc->anchor = readAnchorPoint(ttf,baseoffset+loffsets[i]+aoffsets[kbase+j],
-			classes[j], at_baselig,sc->anchor,info);
-		sc->anchor->lig_index = k;
+		ap = readAnchorPoint(ttf,baseoffset+loffsets[i]+aoffsets[kbase+j],
```
iorsh, line 904: Please explain this change. Is it necessary?

fontforge/splinechar.c
```diff
@@ -1054,35 +1054,69 @@ void AltUniAdd_DontCheckDups(SplineChar *sc,int uni) {
...
+return( -1 );
+    if ( ap_a->ap->lig_index > ap_b->ap->lig_index )
+return( 1 );
+return( ap_a->order - ap_b->order );
+}
+
 void SCOrderAP(SplineChar *sc) {
-    int lc=0, cnt=0, out=false, i,j;
-    AnchorPoint *ap, **array;
+    int cnt=0, ascending=true, descending=true, first=true, prev=0, i;
```
iorsh, line 1073: It is known that `qsort` doesn't work well with already sorted data. We usually ignore that fact, so let's keep the code consistent for readability. Please just use qsort unconditionally.

fontforge/splinechar.c
```diff
@@ -1054,35 +1054,69 @@ void AltUniAdd_DontCheckDups(SplineChar *sc,int uni) {
     }
 }
 
+struct aporder {
+    AnchorPoint *ap;
+    int order;
```
iorsh, line 1059: `order` is an arbitrary element meant to preserve existing order when `lig_index` is same. Is it necessary?

iorsh:
Please also provide a sample PDF and the instructions for reproduction of your issue with standalone FontForge application. I understand your analysis, but it would still be useful to verify that your change indeed improves FontForge functionality.

andiwand:
> Please also provide a sample PDF and the instructions for reproduction of your issue with standalone FontForge application. I understand your analysis, but it would still be useful to verify that your change indeed improves FontForge functionality.

Thanks for the review @iorsh ! The only sample PDF we have contains some sensitive information. We would be willing to mail it to you privately if that is acceptable

fontforge/splinechar.c
andiwand: yes this is supposed to keep the stability of the sort. since `lig_index` should be unique I don't think it is necessary

fontforge/splinechar.c
andiwand: I fear that would bring us back to the $N^2$ complexity and result in a similar runtime like before

fontforge/parsettfatt.c
andiwand: this change here links the APs in `lig_index` order instead of reverse. I think it could be made a bit prettier. the other option would be to append with this new ordering already inside `readAnchorPoint` but that would effect other potential calling sites

iorsh: This code is critical, since it is reading the binary file, and `SCOrderAP` is a kind of post-processing. I would highly prefer to leave it as is, and make all changes in `SCOrderAP`. Would you fix still be possible if the change is limited to `SCOrderAP`?

fontforge/splinechar.c
iorsh: Ok, in that case let's sort by `lig_index` without insisting on stability. It wouldn't contribute much, I suspect.

fontforge/splinechar.c
iorsh: Agreed. Please add a comment about this and a `TODO: use std::sort() in future`.

iorsh:
Please email PDF to iorsh@users.sourceforge.net.

iorsh:
I added a printout to `SCOrderAP()`: 
```
    printf("Reordering %d anchor points for %s\n", cnt, sc->name);
```

and opened your PDF with an offending font. 
The result is as follows:
```
Reordering 4 anchor points for uniFEFB
Reordering 4 anchor points for uniFEFC
Reordering 4 anchor points for uniFEF7
Reordering 4 anchor points for uniFEF8
Reordering 4 anchor points for uniFEF5
Reordering 4 anchor points for uniFEF6
Reordering 4 anchor points for uniFEF9
Reordering 4 anchor points for uniFEFA
Reordering 6 anchor points for uniFEDF_uniFEE0_uniFEEA.liga
Reordering 4 anchor points for uniFEDF_uni0672.fina.rlig
Reordering 4 anchor points for uniFEE0_uni0672.fina.rlig
Reordering 4 anchor points for uniFEDF_uni0673.fina.rlig
Reordering 4 anchor points for uniFEE0_uni0673.fina.rlig
Reordering 4 anchor points for uniFEDF_uni0675.fina.rlig
Reordering 4 anchor points for uniFEE0_uni0675.fina.rlig
Reordering 4 anchor points for uni06B5.init_uniFE8E.rlig
Reordering 4 anchor points for uni06B5.medi_uniFE8E.rlig
Reordering 4 anchor points for uni06B6.init_uniFE8E.rlig
Reordering 4 anchor points for uni06B6.medi_uniFE8E.rlig
Reordering 4 anchor points for uni06B7.init_uniFE8E.rlig
Reordering 4 anchor points for uni06B7.medi_uniFE8E.rlig
Reordering 4 anchor points for uni06B8.init_uniFE8E.rlig
Reordering 4 anchor points for uni06B8.medi_uniFE8E.rlig
Reordering 8 anchor points for uniFDF2
Reordering 8 anchor points for uniFDFC
Reordering 4 anchor points for uni076A.medi_uniFE8E.rlig
Reordering 4 anchor points for uni076A.init_uniFE8E.rlig
Reordering 6 anchor points for uniFEDF_uniFEE0_uniFEEA.liga
Reordering 4 anchor points for uniFEDF_uni0773.fina.rlig
Reordering 4 anchor points for uniFEE0_uni0773.fina.rlig
Reordering 4 anchor points for uniFEDF_uni0774.fina.rlig
Reordering 4 anchor points for uniFEE0_uni0774.fina.rlig
Reordering 4 anchor points for uni08A6.init_uniFE8E.rlig
Reordering 4 anchor points for uni08A6.medi_uniFE8E.rlig
Reordering 8 anchor points for uni0627_uni0627_uniFEDF_uniFEE0_uniFEEA.liga
Reordering 8 anchor points for uniFED3_uniFEDF_uniFEE0_uniFEEA.liga
Reordering 8 anchor points for uniFED3_uniFEDF_uniFEE0_uniFEEA.liga
```
Please note that no glyphs has more than 8 points in the original PDF. Does your tool modify the font or the PDF before feeding it into FontForge?

andiwand:
I am not aware that there are any modifications on the fly. To my understanding the font is read from the PDF via poppler and directly provided to FontForge without modifications. I will try get the same `printf` in there and run it again. I will also try to extract the font as a file.

Issue #5700, opened 2025-12-02
Support ARM and Intel hardware on Mac


Wampler-R:
I downloaded the 2025-10-09 disk image for MacOS.  When I try to open it, the icon flashes, but nothing happens.  There is no error shown.

iorsh:
Please try the following:

1. Install the latest image https://github.com/fontforge/fontforge/actions/runs/19876277967/artifacts/4744336608
2. Check if other releases (like [2023-01-01](https://github.com/fontforge/fontforge/releases/tag/20230101)) install correctly
3. Try to run the installation from terminal and see if it produces errors.

GwynethLlewelyn:
I have exactly the same issue, but, in my case, it was easy to figure out: I have an Apple Silicon (M1) Mac, and I don't really wish to install Rosetta 2, preferring to wait for content developers to distribute their apps in the Universal Binary format, which was introduced in 2020 — but FontForge isn't one of them, so, either this gets compiled locally, or I'll have to be patient and wait 5–10 years, after which Apple will remove Rosetta 2 support forever and everybody will start complaining! 😆

Jokes aside, I haven't managed to get a compilation to properly finish (yet), so I can understand that there may be good reasons why a Universal Binary version is not forthcoming. macOS (at least Tahoe) doesn't check if an application is Intel only/ARM only/both — it just 'installs' it and leaves the decision to run it under Rosetta 2 if the user wants that.

Other releases are also compiled for the Intel architecture only, so they exhibit the same issue (I've tried).

I've also tried the MacPorts version, which is still using 2023-01-01. And, indeed, it _has_ an ARM binary to download! Alas, surprisingly, it behaves exactly like @Wampler-R reported. Even when calling from the terminal shell to see the error messages, there are none to show.

@Wampler-R didn't say, but it's likely they're running macOS Tahoe. The issue of running X11/Xquartz under Tahoe is known: https://discussions.apple.com/thread/256157657?sortBy=rank

It also helps to use the newest version of XQuartz, 2.8.5 (which is, well, two years old), as opposed to the archaic version that comes with (previous) versions of macOS, 1.20.11 (or so). This can be downloaded directly from the [XQuartz](https://www.xquartz.org/) official website, which comes with a version of `quartz-wm` (the 'default' X window manager for XQuartz) that even works (unlike the version for 1.20.11, which simply refuses to start).

[...]

PR #5834, opened 2026-06-07
macOS: ad-hoc code sign the app bundle, with a Developer ID path

skef:
The macOS CI build moved to an arm64 runner (#5642). Unsigned arm64 Mach-O code is rejected by the kernel ("Code Signature Invalid"), and lddx's install_name_tool rewrites invalidate the linker's ad-hoc signatures, so the resulting bundle fails to launch on Apple Silicon.

Re-sign the bundle in ffosxbuild.sh after lddx, defaulting to an ad-hoc signature so the app loads on arm64 (still un-notarized, so it needs the usual Gatekeeper bypass, matching the prior x86 experience). Setting CODESIGN_IDENTITY switches to a hardened-runtime, notarizable build using osx/entitlements.plist; disable-library-validation lets the embedded Python interpreter load third-party C extensions under the hardened runtime. main.yml gains two no-op-by-default steps (cert import, notarize + staple) that activate when the corresponding secrets exist.

Add a workflow_dispatch-only rehearsal that exercises the full signing path (minus notarization, which self-signed certs cannot do) with a throwaway self-signed cert, verifying the hardened-runtime bundle and that disable-library-validation works end to end.

### Type of change

- **Bug fix**
- **New feature**
- **Breaking change**
- **Non-breaking change**

### DIFF
--- a/.github/workflows/mac-signing-rehearsal.yml
+++ b/.github/workflows/mac-signing-rehearsal.yml
@@ -1,0 +1,162 @@
+# Manual rehearsal of the macOS Developer ID signing path, using a throwaway
+# self-signed certificate generated on the runner.
+#
+# Purpose: exercise everything the real signing path does EXCEPT notarization,
+# which a self-signed cert cannot do (Apple's notary service requires a genuine
+# Developer ID Application certificate). In particular this validates the one
+# thing the ad-hoc path can never test: that under the hardened runtime, the
+# `disable-library-validation` entitlement actually lets FontForge's embedded
+# Python interpreter load a third-party (foreign-signed) C extension.
+#
+# This workflow is workflow_dispatch only: it never runs on push and never
+# produces a release artifact. The bundle it builds is SELF-SIGNED and must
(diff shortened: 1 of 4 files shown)


iorsh:
Is this PR ready for review?  Looks like @bgvoisin confirmed that it works on ARM in https://github.com/fontforge/fontforge/issues/5700#issuecomment-4644447074

bgvoisin:
@iorsh I can confirm this new build works on ARM not only on macOS 26 Tahoe (the current version, runs on both Intel and ARM) but also on macOS 27 Golden Gate (the next version, ARM-only, in beta since Monday). I tried basic things like opening font files, browsing through the glyphs, saving the fonts to other formats, all works OK.

skef:
I marked it ready. Keep in mind I don't have much of any specialized knowledge about this, and there's quite likely slop mixed in. However, it's just CI, not the code.

iorsh (approved):
I don't understand much about it either, but it works, so let it be.

Issue #5836, opened 2026-06-09
RFE: Change scaling behavior to allow dragging points instead of scaling about origin


erco77:
This request for enhancement changes the scale tool so when a vertex is dragged, it tracks the mouse, locking points on the opposite side of the object, allowing the dragged edge to be controlled more easily, similar to many modern applications (e.g. gimp, word, libreoffice).  Holding down SHIFT during dragging constrains interactive scaling to a single dimension, depending on which way the mouse is dragged. (Used in the below screen grab video during precise edge alignments)

I have a [working fork with a branch 'scaling-fix'](https://github.com/erco77/fontforge/tree/scaling-fix) which implements the new scaling behavior; [this short video](https://github.com/user-attachments/assets/b1f37126-c2da-470a-aa5d-5d6ae3dcd665) demonstrates the described scaling behavior, editing two glyphs (2MB .webm).

I should add I used the assist of codex (chatgpt 5.5) to implement this, and then hand crafted the coding style/comments and clarified indenting. Followed the suggestions in CONTRIBUTING.md for normal "return" (and break) indenting, and use of bool instead of ints.

[Attached patch](https://github.com/user-attachments/files/28768040/scaling-mods.patch) implements the behavior described as a single commit, including sphinx docs modifications. The above-mentioned fork 'scaling-fix' branch is basically the same thing.

Docs mod:

<img width="480" height="459" alt="Image" src="https://github.com/user-attachments/assets/c4157cc1-ca95-42db-9f9f-8e82558db29b" />

This was useful for me, not sure if it's useful for everyone. In my case it was critical to be able to scale fonts with fine control over the glyph's outer perimeter.

iorsh:
The idea of making scale tool more standard is good. Your approach is a step in right direction, but in its current form it replaces a very counterintuitive tool with a reasonably good, but insufficiently visualized tool. Inkscape deals with scaling on a visual level: 

<img width="238" height="269" alt="Image" src="https://github.com/user-attachments/assets/1f91cbb0-8dd7-4e2a-b6d3-dc7347ec9bc8" />

so that the user knows immediately what is expected from them, and what would be the result.

I looked at the code, and Codex seems to write quite a lot of that (as expected, since it's paid by word count). I didn't review thoroughly, but I wouldn't expect a replacement of quite a small piece of math to take so much. Most of the added code probably already exists elsewhere in a form of functions, which can be reused instead of duplication.

[...]

PR #5837, opened 2026-06-10
"Outline Window" Improved interactive scaling + docs

erco77:
This changes the behavior of the glyph editor's scaling tool to use a resizable bounding box with resize handles for improved scale control, as per fontforge issue #5836, my own RFE (request for enhancement). Includes docs and screenshot.

These changes depended on a strong mix of AI (codex/chatgpt 5.5) and human (myself) source code development cycle. While I'm a well seasoned C/C++ programmer, I leaned heavily codex to do the heavy lifting; admittedly I never would have taken this on had it not been available.

Tested and developed only on Linux (Ubuntu 24.x).

Did not heavily test, but I can probably revisit within a week if fixes are needed. Font editing is not something I normally do, and only first picked up fontforge a few days ago to add some codepage 437 box drawing and geometric shapes to an existing font.


doc/sphinx/ui/mainviews/charview.rst
```diff
@@ -519,14 +519,28 @@ The scale tool
...
+
+* Dragging a corner resize handle scales the selection while keeping the opposite corner fixed.
+
+* Dragging a left/right edge handle scales only horizontally while keeping the opposite edge fixed.
+
+* Dragging a top/bottom edge handle scales only vertically while keeping the opposite edge fixed.
+
+Clicking or dragging anywhere other than a bounding box handle does not start a
+scale operation, and is ignored. To change the selected points to scale, switch
+to the "pointer tool" to make a new selection, the come back to the scale tool
```
iorsh: *then* come back

fontforge/views.h
```diff
@@ -1037,6 +1037,10 @@ extern void BVChar(BitmapView *cv, GEvent *event );
 extern void CVMouseDownPoint(CharView *cv,GEvent *event);
 extern void CVMouseMovePoint(CharView *cv,PressedOn *);
 extern void CVMouseUpPointer(CharView *cv );
+extern bool CVScaleSelectionBounds(CharView *cv, DBounds *bounds);
+extern void CVScaleHandlePoint(const DBounds *bb, enum expandedge edge, BasePoint *pt);
+#define CVSCALE_HANDLE_FUDGE 7.5    // #pixels for fudging mouse proximity
```
iorsh: Please use `snapdistance` instead, it's already configurable.

fontforgeexe/charview.c
```diff
@@ -3780,8 +3832,8 @@ return;
 return;
 
     if ( cv->active_tool==cvt_scale ) {
-	xdiff = 100.0 + (cv->info.x-cv->p.cx)/(4*tab->scale);
-	ydiff = 100.0 + (cv->info.y-cv->p.cy)/(4*tab->scale);
+	xdiff = cv->expandwidth==0  ? 100.0 : 100.0*(cv->info.x-cv->expandorigin.x)/cv->expandwidth;
```
iorsh, line 3835: What is `cv->expandwidth`? What does it mean when it's zero?

erco77: cv->expandwidth is in the original fontforge code.

It's the horiz distance between the locked scale origin and the dragged handle of a bounding box.

Value of zero would mean the bounding box is either zero width in X, or there is no horizontal span.. a situation to avoid for various reasons, including avoiding div by zero and scaling to zero.

fontforge/views.h
erco77: Hmm, snapdistance seems related, but might not be the same concept here.

I think snapdistance causes moved points to snap to nearby interesting features (like other geometry or guides).

Whereas the CVSCALE_HANDLE_FUDGE is how close the mouse can be to the bounding box handles before it recognizes the mouse being over them and "clickable"; the concept doesn't snap the mouse to geometry.

Codex suggests if you're wanting fewer special constants, a compromise could be:

`real fudge = 2*snapdistance/tab->scale;`

That still uses the existing preference, avoids a totally separate magic constant, and keeps the effective default close to the current behavior, so that it's easier to grab the handles. But if someone lowers the snapdistance, it would affect that too.

[EDIT: Or, perhaps better, making the extra area around the handles for the mouse hotbox to be e.g. 1.5x the size of the handles themselves. That's probably better.]

doc/sphinx/ui/mainviews/charview.rst
erco77: My doc typo; fixed in 7a4565e17.

fontforge/views.h
iorsh: Yes, we already have a similar `fudge` computation in `CVMoveSelection()` https://github.com/fontforge/fontforge/blob/416e773cd0432ce92c787b9700c48d95b9f4937e/fontforgeexe/cvpointer.c#L1310

Let's make it `real fudge = ???*snapdistance/tab->scale;`. Please set the constant (2 or else) at your discretion.

erco77: Hmm, I've explored this approach for the last hour or so, and it just doesn't work; it works fine if I leave snapdistance at the default, but as soon as I change it to e.g. '1', the handles become almost unselectable, and if I make it large, say '10', the handles can be grabbed from too far away.

The only way to make the snapdistance influence work is to have two magic numbers that min/max the snapdistance calculations to keep it within reasonable range, but then it really isn't effectively different from just using the single 7.5 value.

I really don't think the snapdistance makes sense here; when the user changes this, they won't realize its affect on the ability to select the mouse sensitive scale handles.

If you have other suggestions I can try them. Perhaps something like a "mouse proximity" fudge factor in the prefs, if there is such a thing, which softens how close one has to be to a mouse-sensitive point on the screen for the mouse cursor to change and the object to be selectable.

I really have to underline 'snapdistance' seems to be for an entirely different purpose; while dragging a point, the point snaps to nearby geometry. However the fudge factor with the 7.5 is not related to snapping during dragging, it's how far away the mouse hover can be within the proximity of the handles to change the mouse icon into one of the "handle drag" icons so clicking down on the handle can then drag it. Nothing to do with how it drags or what it should snap to - the "snap to" effect for where the handles snap to is already in the code.

erco77: Just pushed a small change to the comment for CVSCALE_HANDLE_FUDGE, e31efe008, to clarify this value is related to mouse hover proximity, so as not to mislead one into thinking it might also be related to 'snapping'.

erco77: Of interest, I asked codex to find a similar concept to my situation in the original codebase where a fudge factor affects mouse hover proximity, and it found one:

- Apparently in "Pointer" mode in the glyph editor, when the mouse gets close to a vertical guide (the "advance width line" or "right bearing line"), the cursor changes from a "pointer arrow" into a "left/right arrow", indicating clicking will drag the guide left and right. Here it found this hard coded 3.5 constant fudge value in the CVCheckResizeCursors() function (in cvpointer.c) to transition the mouse cursor icon from the pointer arrow icon to a left/right arrow, which reads:
```
real fudge = 3.5/tab->scale;
```
This 3.5 value in the original repo is similar to the 7.5 value for my scaling branch.

Apparently though, codex warns the Pointer code uses a /different/ fudge value than the one it uses for the mouse icon transition, using 'snapdistance' instead of the 3.5 constant, causing a UI discrepancy (bug) between what the mouse icon shows, and the function that occurs during on click. The mouse click detection occurs in charview.c, whereas mouse icon transition happens in cvpointer.c

_(I ran into this same issue during development of the new scale code, and the codex solution was to make the constant a global macro for 7.5 value, so both modules could access the same constant.)_

I tested the original code base to see if there was a UI issue between mouse hover and click in Pointer mode over the "right bearing line", and found if moved to the right of the guide until the mouse icon changed from left/right back to an arrow and then clicked, I could still drag the guide. I found this discrepancy was affected by the "snapdistance" preference; if I set it to 10, I could be 5 pixels away from the transition and still slip into "guide drag" mode. Seems like the use of snapdistance here is perhaps not what's desired, the 3.5 constant seems like the better choice..? In which case it too might benefit from a global constant, perhaps the same one I use.

If you want, if we resolve the current issue with a global, I can then open a new issue for the Pointer mode problem and solve it with a similar solution, changing the global macro name CVSCALE_HANDLE_FUDGE to something more general, like MOUSE_PROXIMITY_FUDGE or some such.

iorsh: Perhaps I misunderstood your code, but I was thinking you are applying fudge as an additional area around the scaling handlers. What I mean, is even for the zero fudge, the entire handler (4x7 pixels?) should be clickable and draggable, and if we increase `snapdistance`, coarser clicking would suffice. I'm ok with a new preference, but it feels like these preferences would be basically responsible for the same thing based on user's mouse and dexterity.

erco77: > Perhaps I misunderstood your code, but I was thinking you are applying fudge as an additional area around the scaling handlers.

Yes, that's the correct.

>  and if we increase snapdistance, coarser clicking would suffice

Only if I replace the 7.5 constant with a formula that includes `snapdistance`, yes it affects this behavior.

But I spent several hours trying to make that work, and it just doesn't; `snapdistance` really shouldn't affect hover distance from handles and grid lines. The prefs description of `snapdistance` has more to do with magnetizing points to nearby geometry:
```
  When the mouse pointer is within this many pixels
  of one of the various interesting features (baseline,
  width, grid splines, etc.) the pointer will snap 
  to that feature.
```

People will want to change this magnetic distance to nearby geometry without affecting mouse proximity to handles and guides.

The `fudge` value seems like something that really should only be a constant, around +/- 5 pixels; changing it below 3 makes things unselectable, and above 9 makes the selection area too wide.

Mixing the two concepts causes problems that can be seen in the existing code base (i.e. not my fork) as a UI bug:
1. Set the Preference -> Editing -> SnapDistance to 10, and "SnapToint" to "ON".
2. Edit a character and choose the "Pointer" tool
3. Hover over the right hand guide ("Right Bearing Line") and a left/right arrow icon appears; you can move left or right of the guide (+/- 3 pixels), and a left click while that icon shows allows you to drag the guide.
4. IMPORTANT: While hovering over the grid line, move the mouse to the right until the cursor changes back to the pointer icon.
5. BUG: Now left click and drag. What should happen is you drag out a selection. But what actually happens is you end up dragging the guide.

I found that with `snapdistance` set to 10, there's an additional 5 pixel area to the right of when the pointer changes to normal where it still drags the guide. Higher `snapdistance` makes this area larger.

This is a mild UI bug, but relevant here because of the mix of fudge (for selecting the mouse icon) vs using `snapdistance` (for the mouse click),  causing the mouse click to not represent what the mouse cursor is showing.

The fact these two terms are mixed in the code base seems to indicate there might be a disagreement in the code about what these two terms do.

I'd offer using a constant is the better method, and leave `snapdistance` to only affect how actively dragging a point snaps to nearby geometry.

If it's still unclear, I think if I make a video it'll help a lot, as I think the more I write, the less clear I'm being. :/

iorsh: Ok, since we can't see any easy change, let's just keep it this way. It's a small comment, and the functionality  works fine, that's what matters most.

iorsh (approved):
Looks good to me. Since it's a change in behavior, I'll keep it open for a 3-4 more days before merging, to allow more input for people.

iorsh:
Merging with bypassed requirements, since `mac (macos-14, arm64)` CI requirement was added after this PR was finalized. It does have a passing `mac` req.

PR #5845, opened 2026-06-28
Modernized GTK-based Print dialog

### LINKED ISSUE #2171: Consider removing print support
FontForge's "print" command is disproportionately expensive to maintain.  The file print.c file alone is over 3000 lines, and those lines of code cost more to maintain, per line, than most lines of code in the rest of the system.  Much of the material in that file is OS-specific, because it needs to talk to the unique interfaces of each system where FontForge runs; maintaining such code requires developers who have and are familiar with the individual systems.  Testing it in any automated way is nearly impossible.  Another large section of the file consists of sample text in many different languages, most of which are unfamiliar to the average developer; that data doesn't require a lot of maintenance, but what maintenance it requires is highly specialized and again can't be done by the average developer.   There are thousands of lines of code in sflayout.c and langfreq.c that also seem to exist only to support the printing feature.

Printing requires expensive support elsewhere in the system for things like character set conversion, library stuff to do things like write PDF files (used very little, or not at all, elsewhere), and it has indirect costs too, for instance (by necessitating low-level access to OS interfaces) by breaking the possibility of abstracting file-writes that might be valuable elsewhere.  In summary, having a printing feature in FontForge costs a lot.

The feature does not work well, either.  On my Linux system, it does not pop up the standard printing dialogue, nor talk to CUPS.  I haven't tested other systems carefully, but it does not appear to *properly* use the native interfaces anywhere else either.  Instead it pops up its own unfamiliar GUI print configurator, and then behind the scenes talks to the standard interfaces in a nonstandard way.  I have no idea whether I can expect it to work; I certainly can't expect it to work the same way as printing in other GUI programs.  Making it work the same way as printing in other GUI programs, would obviously be a large amount of work for the development team both up-front and in ongoing maintenance.  *Not* deleting the current feature also has a very significant cost in ongoing maintenance.

Even assuming that talking to the printer worked perfectly, FontForge's printouts offer a selection of canned layouts that may not reflect what users will actually want to print out, and cannot easily be modified or controlled because they are defined entirely by C code.  I, and I think most other people likely to want to generate printouts from fonts, would prefer to define my own test printout layouts using a word processor or other external software.  I could also easily print a PDF file externally if FontForge would generate one; the fact that FontForge tries to be capable of also sending it directly to the printer is not a particularly valuable minor detail in relation to how much that minor detail costs.

So is it worth it?  Do people use this feature at all, much less use it enough to justify its cost to the project?

In FontAnvil the answer is pretty clearly "no," and I'm removing printing from FontAnvil.  FontForge's considerations are different, and maybe having a printing feature does make sense for FontForge.  But my suggestion, as in the title of this issue, is that FontForge should seriously *consider* removing print support.

iorsh:
# Modernize the Print dialog

This PR introduces a new GTK-based Print dialog. Fixes #2171.

## What's new
- OS integration with GTK (native Windows print dialog, standard Unix-compatible dialog in Linux and Mac)
- Harfbuzz support
- Formatted sample text (weight, slant, width, size according to the availability of open fonts)
- Sample text can be saved and loaded with formatting
- Several scaling options for full-page glyphs

## What is dropped
- Antialiasing option (now always on)
- Rendering mode (now everything is with Cairo)
- Screen DPI setting (not applicable anymore, the preview is identical to printed page)

**Development notes:** https://github.com/iorsh/fontforge/wiki/Print-dialog

Screenshots:
SampleTextEditor
SampleText
MultisizeGlyph
FullFont
FullPage


skef:
@iorsh I'm going to be busy before July 5th but I will try to get to this soon after.

.github/workflows/main.yml
```diff
@@ -1,7 +1,7 @@
 name: CI
 on:
   push:
-    branches: [ master, releases/** ]
+    branches: [ master, releases/**, gtk_print2 ]
```
skef, line 4: Should we clean this up?

.vscode/gdb_ff_printers.py
```diff
@@ -0,0 +1,39 @@
+import gdb
```
skef, line 1: What about these? Should we be keeping .vscode stuff?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1262 @@
...
+#include <array>
+#include <charconv>
+#include <regex>
+#include <sstream>
+#include <set>
+
+extern "C" {
+#include "../fffreetype.h"
+
+extern int SCRightToLeft(SplineChar* sc);
```
skef, line 39: Is there a reason we can't include these from their headers outside of the extern "C" block (and in the case of MakeGlyphTTFMap add it to tottf.h)?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1262 @@
...
+                                   const char* name);
+extern SplineFont** FVCollectFamily(SplineFont* sf);
+extern SplineCharTTFMap* MakeGlyphTTFMap(SplineFont* sf);
+extern char* SFGetFullName(SplineFont* sf);
+SplineChar** FVGetSelection(FontViewBase* fv);
+}
+#include "gutils.h"
+#include "ustring.h"
+
+namespace ff::utils {
```
skef, line 59: This name seems to imply a more general bucket than ff:layout, but everything in it now seems to be consistent with layout. What is its role, and why put FullGlyphPrinter and its siblings into it?

fontforge/layout/cairo_painter.hpp
```diff
@@ -0,0 +1,333 @@
...
+
+#include "layout_shim.hpp"
+#include "i_printer.hpp"
+#include "i_shaper.hpp"
+#include "shaper_shim.hpp"
+
+typedef struct splinechar SplineChar;
+typedef struct fontviewbase FontViewBase;
+
+using ff::layout::SplineFontProperties;
```
skef, line 42: Why are these at global scope in a .hpp file?

skef:
Going through the multi-family stuff in this has soured me a bit on the state of this PR. Do we really need this complexity? If we do really need it, shouldn't we think through the implementation more? E.g. how does the interface work if only bold and bold-italic are open? What if there are two informally related fonts of different families you want to test together? (And wasn't that possible with the dialog being removed?)

iorsh:
Yes, the multi-family stuff is complicated, I'd rather define it as multi-font, because I explicitly limited it to a single family. I think it would be good to discuss, because I'm not totally settled on this too, and would be glad to consider changes. 

Indeed it was possible to create a sample with several unrelated fonts. I don't think it was in real use ever, because the legacy dialog didn't have any capability to reuse a formatted sample, so every time the user wanted to compare two fonts, they had to load or paste a plain text sample and format it from scratch.

Now the formatted samples can be saved as XML and reused, and I wanted this to be as general as possible. My intended use case is about maintaining several font families, each having e.g. regular, bold, italic, bold-italic. When I want to add a new feature, e.g. a new currency sign, I can create a sample text which has this sign in all four styles, but without font name mentioned. Then I load each family separately, and the same sample can be used to test any family. If the sample contained explicit font or family name, this would be impossible. I myself have a lot of that when I work of Hebrew diacritics, and I have an extensive testing suite which should be displayed in each of many font families.

The above intent clearly breaks when the saved samples contain font names. If you can think of an option to settle this, I would be glad to implement it.

For limited family of fonts, the interface blocks unused options. In your example (only bold and bold-italic) only the "Italic" button would be active, and the bold weight would be applied automatically because it's the only one available. If the user loads a sample which has more styles, e.g. light/bold, the missing styles will fallback to something reasonably close among the available fonts.

skef:
What if you're developing a font to be compatible with a different font (say in a different script) and want to test both together? 

I don't think I agree that this new interface is "as general as possible" -- it seems more like it addresses a specific use case you have in mind while eliminating others. Just on its face: why does the particular use case you describe correspond to generality?

skef:
If you wanted something like what you describe without limiting the UI to it, I would think the approach would be to save some metadata about styles with the font identifiers such that you could "apply" the sample to a new group of fonts, with the system doing the best match possible and then maybe popping up a dialog if there are questions.

iorsh:
Thinking about it, indeed my notion of generality is quite subjective. I'll think how this can be settled. Maybe we can let the user have all open fonts, but my flavor of generality will kick in when all the fonts actually belong to the same family...

skef:
I hope this transitions back to a more specific-font related UI, but if not I don't like the "merely disabled" aspect of the styles. If only bold and bold italic are open, for example, the bold button should somehow indicate "can't be toggled but is enabled".

skef:
(There's also stuff like math fonts. Or slant as opposed to italic? Can we actually enumerate and capture in UI all relevant font "styles"?)

skef:
I guess on another topic -- I know it isn't in the current system but should we have something for entering less usual/typeable codepoints? Like the MetricsView has? (Ideally the same as the MetricsView has?)

davelab6:
There are many apps and projects for creating 'proofing' PDFs... these may
be inspiring :)

https://ohnotype.co/blog/proof-it

https://idiotproofed.com/

https://hellbox.com/

https://fontproofer.com/

iorsh:
> There are many apps and projects for creating 'proofing' PDFs... these may be inspiring :) https://ohnotype.co/blog/proof-it https://idiotproofed.com/ https://hellbox.com/ https://fontproofer.com/

Look interesting, but probably for future. For now I prefer not to expand too much beyond the existing capability. We have too many font editing gaps, like color fonts etc, so I'm only doing some minimal UI cleanups and enhancements.

iorsh:
> I guess on another topic -- I know it isn't in the current system but should we have something for entering less usual/typeable codepoints? Like the MetricsView has? (Ideally the same as the MetricsView has?)

Fixed, with same approach as in MetricsView

image

iorsh:
@skef, I think I addressed your main concerns.

- As noted, unencoded glyphs can be entered as `\glyph2395`
- The rich text editor now has _specific mode_, which is triggered by fonts from unrelated family, and allows the user to select the exact font for any text segment. 
image

- In the _generic mode_ (the existing one) disabled controls now show correct values, e.g. if all fonts are bold, the "Bold" button will be disabled, but visibly pressed.

iorsh:
I do think the UI needs more polishing, but can't fully wrap my head around it now. The users can do a lot of things, and I think the existing capability is reasonably preserved.

Hopefully we will get more user input for this dialog, and have better perspective on it. Maybe a proofing view can be introduced in a long term, with ideas from @davelab6's examples.

@skef, please kindly continue the review.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+                             (ft_ascender - ft_descender)
+                       : 1000;
+    double sf_ascent = sf_properties.ascent * scale;
+    double sf_descent = -sf_properties.descent * scale;
+
+    return {sf_ascent, sf_descent};
+}
+
+// Returns [x_position, y_position, scale]
+std::array<double, 3> FullGlyphPrinter::calculate_full_glyph_location(
```
skef, line 180: Why std::pair for the above and not std::tuple for this?

fontforge/layout/layout_shim.cpp
```diff
@@ -0,0 +1,103 @@
...
+    // rule, we prefer exact match, of course, but if there is no exact match we
+    // prefer to preserve expansion or condensing.
+    static const std::map<int16_t, std::vector<int16_t>> width_mapper = {
+        {1, {1, 2, 3, 4, 5, 6, 7, 8, 9}}, {2, {2, 1, 3, 4, 5, 6, 7, 8, 9}},
+        {3, {3, 2, 4, 1, 5, 6, 7, 8, 9}}, {4, {4, 3, 5, 2, 1, 6, 7, 8, 9}},
+        {5, {5, 4, 6, 3, 7, 2, 8, 1, 9}}, {6, {6, 7, 5, 8, 9, 4, 3, 2, 1}},
+        {7, {7, 8, 6, 9, 5, 4, 3, 2, 1}}, {8, {8, 9, 7, 6, 5, 4, 3, 2, 1}},
+        {9, {9, 8, 7, 6, 5, 4, 3, 2, 1}},
+    };
+    const std::vector<int16_t>& m = width_mapper.at(os2_width);
```
skef, line 90: [C] `width_mapper` is keyed 1-9, but this `.at()` runs unconditionally on `os2_width`, which comes from `sf->pfminfo.width`. That field is zero-initialized and only populated when `pfminfo.pfmset` is true, and nothing on the print path calls `SFDefaultOS2Info()` or `SFDefaultOS2Simple()`. So a File > New font, or anything loaded from a format carrying no OS/2 data, reaches here with `os2_width == 0` and this throws `std::out_of_range`.

There's no try/catch anywhere in cairo_painter.cpp, print_preview.cpp or layout_shim.cpp, and the handlers in application.cpp sit well above the draw callback — so this is a hard terminate when you open Print on a new font.

Note it also runs *before* the `full_name` early-return below, so an exact name match doesn't save it.

Separately and non-fatally: when `other.os2_width` is 0 the `std::find` misses, `width_dist` becomes 9, and the resulting 900-point penalty makes an OS/2-less font effectively unselectable whenever any properly tagged font is open.

Probably cleanest to normalize unset width/weight to 5/400 in `SFGetProperties()` rather than guarding here.

fontforge/splinefont.c
```diff
@@ -392,6 +393,22 @@ void SplineFontSetUnChanged(SplineFont *sf) {
...
 
+char* SFGetFullName(SplineFont *sf) {
+    return sf->fullname ? sf->fullname : sf->fontname;
+}
+
+cpp_SplineFontProperties* SFGetProperties(SplineFont* sf) {
+    return make_SplineFontProperties(
+    sf->ascent,
+    sf->descent,
+    (sf->italicangle != 0.0),
```
skef, line 404: [C] Deriving italic purely from `italicangle` means an italic face with a zero italic angle reads as upright, and an oblique reads as italic. `pfminfo.stylemap` (the OS/2 fsSelection subset) is the more reliable signal and is already parsed.

This matters more than it looks: the (italic, os2_weight, os2_width) triple is the *only* thing distinguishing faces downstream in `SplineFontProperties::distance()`, so getting it wrong makes a face unreachable rather than merely mislabelled.

This is also the natural place to normalize unset `pfminfo.weight`/`pfminfo.width` to 400/5 — see my comment on layout_shim.cpp:90.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+    // Desired properties are derived from the default ones, with
+    // segment-specific tags overriding them when applicable.
+    SplineFontProperties text_props =
+        SplineFontProperties::from_tags(parsed_tags);
+    SplineFontProperties desired_properties = default_properties;
+    desired_properties.merge(text_props);
+
+    // Find the face with properties closest to the desired properties.
+    auto closest_face =
+        std::min_element(cairo_family_.begin(), cairo_family_.end(),
```
skef, line 926: [C] `std::min_element` returns the *first* minimum, so when two faces tie on distance the one at index 0 always wins — that's `current_sf`, whichever FontView happened to invoke the dialog.

The tie isn't hypothetical. "Type" here is just the (italic, os2_weight, os2_width) triple, so two same-family faces with matching OS/2 data are indistinguishable: different optical sizes, a Text vs Display cut, two builds of the same font.

When that happens the second font becomes completely unreachable. `family_has_multiple()` returns false for every property, so all four style controls are disabled; and because the fonts share a family name `generic` stays true, so there's no font-name combo either. The only way to reach it is a `<font value="...">` tag, which the generic UI gives you no way to author.

At minimum the tie should be broken deliberately rather than by iteration order.

fontforgeexe/gtk/print_preview.cpp
```diff
@@ -0,0 +1,762 @@
...
+
+    // Fonts outside the family being proofed are underlined for distinction.
+    rt_props.underline = (font_props.family_name == default_family)
+                             ? Pango::UNDERLINE_NONE
+                             : Pango::UNDERLINE_SINGLE;
+    return rt_props;
+}
+
+void PrintPreviewWidget::build_sample_text_editor() {
+    bool generic = true;
```
skef, line 344: [C] Mode selection is automatic and can't be overridden — a single open family means generic mode, with no way to ask for specific mode even when the style controls can't distinguish the open faces (see my comment on cairo_painter.cpp:926).

Since specific mode is the only way out of that dead end, letting the user pick the mode rather than inferring it from family names seems worth doing.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+
+    for (const auto& [current_tags, raw_text] : parsed_text) {
+        std::vector<ParsedTag> parsed_tags;
+        for (const std::string& tag : current_tags) {
+            parsed_tags.emplace_back(parse_tag(tag));
+        }
+        size_t font_idx = select_face(parsed_tags, default_properties);
+        Cairo::RefPtr<Cairo::FtFontFace> font_face =
+            cairo_family_[font_idx].face;
+        select_face(parsed_tags, default_properties);
```
skef, line 690: [C] This second `select_face()` call discards its result and duplicates line 687 — looks like leftover from the unencoded-glyph change (fbba0fc0).

fontforgeexe/gtk/widgets/rich_text_editor.cpp
```diff
@@ -0,0 +1,972 @@
...
+    std::map<std::string /*id*/, Glib::RefPtr<Gtk::TextTag>> tag_map;
+    std::vector<std::pair<std::string /*id*/, std::string /*label*/>> labels;
+    std::string default_id;
+
+    for (const auto& [font_name, properties] : font_list) {
+        std::string tag_id = "font|" + font_name;
+        if (default_id.empty()) {
+            default_id = tag_id;
+        } else {
+            auto tag = text_view_.get_buffer()->create_tag(tag_id);
```
skef, line 620: [C] If two of the open fonts share a full name, `create_tag()` gets called twice with the same tag id. GTK refuses a duplicate tag name and returns NULL, and `tag->property_weight()` on the next line then dereferences it.

Reachable whenever specific mode is active and two open fonts have the same full name — e.g. two builds of the same font open at once alongside anything from another family.

fontforge/layout/cairo_painter.hpp
```diff
@@ -0,0 +1,346 @@
...
```
skef, line 42: [C] Following up on my earlier question here: `ParsedTag` isn't just at global scope, it's a verbatim redeclaration of the alias already declared at global scope in layout_shim.hpp:49, which this file includes. It compiles because identical alias redeclarations are legal, but the definition now lives in two places.

Line 43 can just go, and line 42 should move inside `ff::utils`. Both currently leak into every TU that includes this header — print_dialog.cpp, rich_text_editor.cpp and print_preview.hpp, plus anything that includes print_preview.hpp in turn.

fontforge/layout/cairo_painter.hpp
```diff
@@ -0,0 +1,346 @@
...
+using ParsedTag =
+    std::pair<std::string /*tag name*/, std::string /*tag value*/>;
+
+namespace ff::utils {
```
skef, line 46: [C] Following up on my earlier comment about this name: `git grep 'namespace ff::utils' master` returns nothing, so this isn't an existing convention being followed — the namespace is introduced by this PR and used only by print-dialog code.

Meanwhile `IPrinter` and `PageContext` live in `ff::layout` while all four of their implementations live here in `ff::utils`, in a directory called layout/. I don't see a defensible line between the two. Suggest folding the lot into `ff::layout`.

fontforge/layout/cairo_painter.hpp
```diff
@@ -0,0 +1,346 @@
...
+                  const ff::layout::PageContext* context) override;
+
+ private:
+    // TODO(iorsh): Storing reference is dangerous, consider other non-copying
+    // storage options.
+    // Currently active font face (for example, whose FontView invoked the Print
+    // dialog).
+    Cairo::RefPtr<Cairo::FtFontFace> cairo_face_;
+
+    // All the other currently open faces from the same family.
```
skef, line 209: [C] This comment is wrong in two ways, and it's repeated verbatim at line 310.

"All the other" — element 0 *is* the current face, as the comment at lines 57-59 explicitly states.

"from the same family" — no longer true as of this PR, which renamed `FVCollectFamily` to `FVCollectFonts` (fontviewbase.c:2188); it now returns every open font regardless of family. That rename is the entire basis of specific mode, so this comment contradicts the feature it's part of.

fontforge/layout/cairo_painter.hpp
```diff
@@ -0,0 +1,346 @@
...
+    size_t page_count() const override { return pagination_list_.size(); }
+    void add_page(size_t page_number,
+                  const ff::layout::PageContext* context) override;
+
+ private:
+    // TODO(iorsh): Storing reference is dangerous, consider other non-copying
+    // storage options.
+    // Currently active font face (for example, whose FontView invoked the Print
+    // dialog).
+    Cairo::RefPtr<Cairo::FtFontFace> cairo_face_;
```
skef, line 207: [C] `cairo_face_` is initialized in the constructor (cairo_painter.cpp:579) and never read anywhere. Looks like it can go.

fontforge/layout/cairo_painter.hpp
```diff
@@ -0,0 +1,346 @@
...
+    // All the other currently open faces from the same family.
+    CairoFontFamily cairo_family_;
+
+    // Sorted with encoded glyph first (ordered by encoding), unencoded glyphs
+    // second (ordered by glyph index).
+    PrintGlyphVec print_map_;
+
+    std::unique_ptr<ff::layout::IPrinter> active_printer_;
+
+    // Cached layout data, which should be invalidated whenever the painter
```
skef, line 319: [C] A few small things in this header, collected here rather than as separate threads:

- **This comment** describes a member that no longer exists; it now floats above `sort_glyphs()`, which is unrelated. Worth confirming the invalidation contract it documented didn't get dropped along with the member.
- **Lines 219-224**: the two doc comments are orphaned and mismatched. "Pagination list for `full_layout_`…" sits after `full_layout_` followed by a blank line and so describes nothing, and "Vector of output lines with calculated height" then attaches to `pagination_list_`, which is a `vector<size_t>` of line indices. They want swapping.
- **Line 72**: `class CairoPainter;` is a dead forward declaration — the definition is at line 255 and nothing in between refers to it.
- **Lines 216-217**: commented-out members left in place.
- **Line 260**: `default_rec()` indexes `cairo_family_[0]` with no bounds check, while `family_has_multiple()` immediately below does check `empty()`. `create_cairo_family()` always pushes `current_sf` first so it can't actually be empty — but pick one and state the invariant.
- **Line 338**: the free `draw_line()` (draws a rule) shares a name with `SampleTextPrinter::draw_line()` at line 242 (draws a text line buffer). Unrelated operations, same name, same header.
- **Lines 97-99**: `kScaleToPage`/`kScaleEmSize`/`kScaleMaxHeight` as `static const std::string` is a string-typed enum. The project is C++17 (CMakeLists.txt:53), so `inline constexpr std::string_view` would drop the out-of-line definitions and the runtime init — or make it an actual enum.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+//                             FullGlyphPrinter                             //
+//////////////////////////////////////////////////////////////////////////////
+
+const std::string FullGlyphPrinter::kScaleToPage = "scale_to_page";
+const std::string FullGlyphPrinter::kScaleEmSize = "scale_to_em_size";
+const std::string FullGlyphPrinter::kScaleMaxHeight = "scale_to_max_height";
+
+size_t FullGlyphPrinter::page_count() const { return print_map_.size(); }
+
+void FullGlyphPrinter::add_page(size_t page_number,
```
skef, line 76: "add_page" for this case is confusing. Why isn't it "draw_page"?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+
+    // We are already in point units. Further rescale surface, exclude the top
+    // 12 pt for glyph title
+    cr->translate(0, full_glyph_top_margin_);
+    Cairo::Rectangle shifted_printable_area{
+        0, 0, printable_area.width,
+        printable_area.height - full_glyph_top_margin_};
+
+    // Retrieve the font metrics in normalized size
+    auto [sf_ascent, sf_descent] = get_splinefont_metrics(cr);
```
skef, line 107: get_splinefont_metrics seems like it's called a lot for something that doesn't change unless the font changes.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+        scale = std::min(x_scale, y_scale);
+    } else if (scaling_option == kScaleEmSize) {
+        x_min = std::min(0.0, text_extents.x_bearing);
+        y_min = sf_descent;
+        double y_max = sf_ascent;
+
+        double y_scale = printable_area.height / (y_max - y_min);
+        scale = y_scale;
+    } else if (scaling_option == kScaleMaxHeight) {
+        // Collect all glyphs to determine maximum y-extents
```
skef, line 208: memoize?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+    // SplineFont)
+    FT_Face face = ft_scaled->lock_face();
+    FT_Short ft_ascender = face->ascender;
+    FT_Short ft_descender = face->descender;
+    ft_scaled->unlock_face();
+
+    // Fontforge SplineFont::ascent doesn't always become the real ascent value
+    // in the font. OS/2 metrics can modify that. Convert the FontForge ascent
+    // value from font units to Cairo units:
+    double scale = (ft_ascender - ft_descender) != 0
```
skef, line 169: is the 1000 default reasonable for Cairo units?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+    double pointsize = 0;
+    if (scaling_option == kScaleToPage) {
+        x_min = std::min(0.0, text_extents.x_bearing);
+        double x_max = std::max(text_extents.x_advance,
+                                text_extents.width + text_extents.x_bearing);
+        y_min =
+            std::min(sf_descent, -text_extents.y_bearing - text_extents.height);
+        double y_max = std::max(sf_ascent, -text_extents.y_bearing);
+
+        double x_scale =
```
skef, line 195: Is 1e-5 a reasonable default here and below?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+
+    cr->scale(glyph_scale, glyph_scale);
+    cr->translate(-x_min, shifted_printable_area.height / glyph_scale + y_min);
+    Cairo::Rectangle na{shifted_printable_area.x / glyph_scale + x_min,
+                        shifted_printable_area.y / glyph_scale -
+                            shifted_printable_area.height / glyph_scale - y_min,
+                        shifted_printable_area.width / glyph_scale,
+                        shifted_printable_area.height / glyph_scale};
+
+    double scaled_size = glyph_scale * normalized_size;
```
skef, line 129: Dead?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+    cr->set_font_size(normalized_size);
```
skef, line 130: We did this above?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+    auto glyph_it = print_map_.begin() + page_number;
+
+    // Print page title for glyph
+    std::string page_title(SCGetName(glyph_it->second));
+    page_title += " from ";
+    page_title += SFGetFullName(font_rec_.sf);
+    init_document(cr, printable_area, page_title, full_glyph_top_margin_);
+
+    // We are already in point units. Further rescale surface, exclude the top
+    // 12 pt for glyph title
```
skef, line 100: 12 pt?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+    cr->translate(0, full_glyph_top_margin_);
+    Cairo::Rectangle shifted_printable_area{
+        0, 0, printable_area.width,
+        printable_area.height - full_glyph_top_margin_};
+
+    // Retrieve the font metrics in normalized size
+    auto [sf_ascent, sf_descent] = get_splinefont_metrics(cr);
+
+    // Glyph metrics in normalized size
+    static const double normalized_size = 1.0;
```
skef, line 110: This smells funny.

fontforgeexe/gtk/print_preview.cpp
```diff
@@ -0,0 +1,762 @@
...
+    {500, Pango::WEIGHT_MEDIUM}, {600, Pango::WEIGHT_SEMIBOLD},
+    {700, Pango::WEIGHT_BOLD},   {800, Pango::WEIGHT_ULTRABOLD},
+    {900, Pango::WEIGHT_HEAVY},  {1000, Pango::WEIGHT_ULTRAHEAVY},
+};
+
+// Preserve the dialog state between invocations
+static struct {
+    std::string radio_selection = SAMPLE_TEXT;
+    size_t full_display_size = 20;
+    std::string full_page_scaling = utils::FullGlyphPrinter::kScaleToPage;
```
skef, line 83: [C] This initializes a file-scope static in this TU from `FullGlyphPrinter::kScaleToPage`, which is a `static const std::string` defined in a *different* TU (cairo_painter.cpp:70). Both are dynamically initialized and the order between TUs is unspecified, so if `persistent` is constructed first it captures an empty string.

That failure is silent and it chains: `set_active_id("")` at line 183 fails, `get_active_id()` at line 639 returns `""`, and `FullGlyphPrinter::calculate_full_glyph_location()` then matches none of its three `scaling_option ==` branches and falls through to `return {x_min, y_min, scale}` with the initial `scale = 1` — a 1pt glyph, i.e. a blank page the first time Full Page Glyph mode is opened.

I can't say whether it actually fires on any current build; it may well be fine today and break under a different linker or with LTO. But it's silent when it does, and there are two independent fixes worth having regardless:

- Make the constants `inline constexpr std::string_view` (the project is C++17, CMakeLists.txt:53). That makes them constant-initialized and removes the ordering hazard entirely. I'd raised this in my comment on cairo_painter.hpp:97-99 as a tidiness point — this is the real reason for it.
- Give `calculate_full_glyph_location()` an explicit default or assert instead of falling through. An unrecognized scaling option silently producing a blank page is a bad failure mode however it's reached, and it's the general cost of using strings where an enum belongs.

fontforgeexe/gtk/print_preview.cpp
```diff
@@ -0,0 +1,762 @@
...
+
+    scaling_option_ = Gtk::make_managed<Gtk::ComboBoxText>();
+    scaling_option_->append(utils::FullGlyphPrinter::kScaleToPage,
+                            "Scale each glyph to page size");
+    scaling_option_->append(utils::FullGlyphPrinter::kScaleEmSize,
+                            "Scale glyphs to em size");
+    scaling_option_->append(utils::FullGlyphPrinter::kScaleMaxHeight,
+                            "Scale glyphs to maximum height");
+    scaling_option_->set_active_id(persistent.full_page_scaling);
+    scaling_option_->set_valign(Gtk::ALIGN_START);
```
skef, line 184: [C] Should this combo have a `signal_changed()` connection to `preview_area.queue_draw()`? `size_entry_` gets one twelve lines up, and `script_lang_combo_` and the feature list get theirs at lines 215, 296 and 299, but I don't see one for `scaling_option_`.

Because `activate_cairo_printer()` re-reads the combo on every draw, the rendering should be correct once a redraw happens for any other reason — so my reading is that picking a different scaling mode does nothing visible until you scroll or resize the preview, rather than that it renders wrongly. Is that what you see, or is the redraw coming from somewhere I've missed?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+
+//////////////////////////////////////////////////////////////////////////////
+//                             FullGlyphPrinter                             //
+//////////////////////////////////////////////////////////////////////////////
+
+const std::string FullGlyphPrinter::kScaleToPage = "scale_to_page";
+const std::string FullGlyphPrinter::kScaleEmSize = "scale_to_em_size";
+const std::string FullGlyphPrinter::kScaleMaxHeight = "scale_to_max_height";
+
+size_t FullGlyphPrinter::page_count() const { return print_map_.size(); }
```
skef, line 74: [C] This gives every entry in `print_map_` its own page, unfiltered, so glyphs with no outline get a page each. Combined with the `1e-5` scale fallback at line 195, an empty glyph with zero advance produces a page that is blank with no indication of why.

That's not a corner case for the workflow this dialog is for: a font in development has runs of empty glyphs, and this PR specifically added unencoded glyphs to `print_map_`, which is where you'd expect to find them. Printing a partly-built font in Full Page Glyph mode could be mostly blank pages.

Should empty glyphs be skipped here, and does that match what the legacy dialog did?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+                       : 1000;
+    double sf_ascent = sf_properties.ascent * scale;
+    double sf_descent = -sf_properties.descent * scale;
+
+    return {sf_ascent, sf_descent};
+}
+
+// Returns [x_position, y_position, scale]
+std::array<double, 3> FullGlyphPrinter::calculate_full_glyph_location(
+    const std::string& scaling_option, const Cairo::RefPtr<Cairo::Context>& cr,
```
skef, line 181: [C] Why is `scaling_option` a parameter? It's a member (`scaling_option_`), this is a `const` member function of the same class, and the only call site (line 118) passes exactly `scaling_option_`. Reading it directly would drop the parameter and remove the possibility of the two disagreeing.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+    // TODO(iorsh): extend lines by tiny dashes
+    draw_line(cr, na, 0, false);
+    draw_line(cr, na, text_extents.x_advance, false);
+    draw_line(cr, na, 0, true);
+    draw_line(cr, na, -sf_ascent, true);
+    draw_line(cr, na, -sf_descent, true);
+}
+
+std::pair<double, double> FullGlyphPrinter::get_splinefont_metrics(
+    const Cairo::RefPtr<Cairo::Context>& cr) const {
```
skef, line 144: [C] This is `const`, but it mutates the Cairo context it's handed — `set_font_face()` and `set_font_size()` on the next two lines. The constness is honest about the object and silent about the context, which is the part a caller would care about.

It matters here rather than being pedantic, because the mutation is load-bearing: `add_page()` calls this at line 107 and then sets the same face and size again at lines 111-112, so the redundancy is only visible if you open the helper. Worth either renaming it so the side effect is expected, or having it save and restore the context state so callers don't have to reason about it.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+    }
+
+    return height;
+}
+
+size_t MultiSizePrinter::page_count() const {
+    if (print_map_.empty()) {
+        return 1;
+    } else {
+        return (print_map_.size() - 1) / lines_per_page_ + 1;
```
skef, line 306: Claude thinks lines_per_page_ can be zero.  And maybe can go negative in some circumstances?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+}
+
+double MultiSizePrinter::draw_line_multisize(
+    const Cairo::RefPtr<Cairo::Context>& cr, int glyph_index,
+    double y_start) const {
+    // Compute line height by maximum glyph size
+    double maximum_size =
+        (pointsizes_.empty())
+            ? 1
+            : *std::max_element(pointsizes_.begin(), pointsizes_.end());
```
skef, line 282: [C] This computes the line height from the *maximum* of `pointsizes_`, but the constructor sizes lines from `pointsizes_[0]` — line 241 uses it for `extravspace`, and line 245 divides by it to get `lines_per_page_`. So pagination is derived from one size and drawing from another.

Those are the same number today only because `kMultiPointsizes` happens to be written in descending order (print_preview.cpp:55-57). Nothing states that as a requirement and nothing enforces it: reorder that list, or construct a `MultiSizePrinter` with any other vector, and you get a `lines_per_page_` computed from a size the lines aren't actually drawn at. Lines then overrun the page bottom with no error — `add_page()` accumulates `y_start` without ever checking against the printable area.

Worth making one of the two canonical. If the maximum is what actually governs line height, the constructor should use `max_element` too (or the class should sort/validate on entry); if the vector really is required to be descending, that belongs in a comment on the member and ideally an assert in the constructor.

Related to the `lines_per_page_` question on line 306 — same variable, different failure. That one is about the value being zero or negative; this one is about it being computed from the wrong size in the first place.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+                                   const CairoFontRec& font_rec,
+                                   const std::vector<double>& pointsizes)
+    : print_map_(print_map), font_rec_(font_rec), pointsizes_(pointsizes) {
+    if (pointsizes_.empty()) {
+        return;
+    }
+
+    double extravspace = pointsizes_[0] / 6;
+
+    double char_area_height = printable_area.height - margin_ - top_margin_;
```
skef, line 243: [C] This uses the constructor-captured `printable_area` to decide pagination, but `add_page()` ignores that copy and reads `cairo_ctx->printable_area_` instead (line 259) for `init_document`. So within one class, how many lines fit on a page and how the page is actually drawn are keyed off two different objects that merely happen to agree.

`FullDisplayPrinter` has the same split — `split_to_lines()`/`paginate()` from the constructor copy at lines 322-330, per-page area in `add_page()` at line 512. `FullGlyphPrinter` doesn't capture one at all and takes it per page, which I think is the right model: page setup can change under a live printer, and the context is the authority for the page being drawn right now.

Two unrelated things keep the copies in agreement today, neither of them a guarantee:

- The preview rebuilds the printer on every redraw (`activate_cairo_printer()` is called from `draw_preview_area()`, print_preview.cpp:717), so the constructor copy is never more than one frame old.
- The print path derives both from the same `context->get_page_setup()`, so `begin_print()` and each `draw_page_cb()` see the same numbers.

The print path is the one to worry about if that ever changes: `begin_print()` activates the printer once (print_preview.cpp:483) and GTK then calls `draw_page_cb` per page *without* re-activating, so a stale constructor copy would survive an entire spool rather than being corrected on the next frame.

Suggest picking one source. Either take the area per-page everywhere and compute pagination lazily, or capture it once and use the captured copy for drawing too — but not one of each.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+        std::cerr << "Invalid page context passed to MultiSizePrinter"
+                  << std::endl;
+        return;
+    }
+
+    auto cr = cairo_ctx->cr_;
+    auto printable_area = cairo_ctx->printable_area_;
+
+    init_document(cr, printable_area,
+                  std::string("Sample Sizes of ") + SFGetFullName(font_rec_.sf),
```
skef, line 261: [C] These page titles aren't translated. `cairo_painter.cpp` contains zero `_()` calls, while print_preview.cpp has 14 — so the dialog chrome around the preview is localized and the text printed on the page is not.

Four sites, all user-visible in printed output:

- line 261 — `"Sample Sizes of "`
- line 95 — `" from "` in `FullGlyphPrinter` (joins glyph name to font name)
- line 515 — `"Font Display for "` in `FullDisplayPrinter`
- line 602 — `"Sample Text from "` in `SampleTextPrinter`

Filing this once here rather than four times since it's one decision. The `" from "` case needs a little care — it's a sentence fragment assembled by concatenation, so it wants to become a format string with positional arguments rather than a translated fragment, otherwise translators get a bare preposition with no context.

Is the omission deliberate — printed output intentionally left in English — or just not gotten to yet?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+        glyph_lines_.empty() ? 16 : glyph_lines_[0].indexes.size();
+
+    static const std::array<std::string, 16> slot_labels = {
+        "0", "1", "2", "3", "4", "5", "6", "7",
+        "8", "9", "A", "B", "C", "D", "E", "F"};
+    for (size_t i = 0; i < line_length; ++i) {
+        Cairo::Rectangle slot{margin_ + left_code_area_width_ + extrahspace_ +
+                                  i * (extrahspace_ + pointsize_),
+                              top_margin_, pointsize_, top_code_area_height_};
+        draw_centered_text(cr, slot, slot_labels[i]);
```
skef, line 530: [C] This can index past the end of `slot_labels`. The array is `std::array<std::string, 16>` (line 523), but `line_length` on line 520 comes from `glyph_lines_[0].indexes.size()`, and `split_to_lines()` picks that from two different ladders (lines 349-363): powers of two capped at **16** when encoded glyphs are present, and **20**/10/5/2/1 when they are not. With no encoded glyphs and `max_slots >= 20`, `line_length` is 20 and this reads `slot_labels[16..19]`. `std::array::operator[]` is unchecked, so that's a garbage `std::string` handed to `show_text()`.

Reachable with a selection of 17 or more unencoded glyphs at a small enough point size. `max_slots` is `floor(char_area_width / (pointsize * 4/3))`, and on A4 portrait `char_area_width` is around 453 pt, so it clears 20 at roughly 16 pt and below. An unencoded-only selection is a normal thing to have mid-development.

There's a design problem underneath the overrun: this header row is drawn on every page regardless of what's on it, but unencoded rows are labelled with decimal running indices (`std::to_string(i)`, line 414), not hex offsets. A `0`-`F` column header over an unencoded section doesn't correspond to anything even when the index happens to be in range.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+
+    for (size_t j = 0; j < glyph_line.indexes.size(); ++j) {
+        int codepoint = glyph_line.indexes[j];
+        if (codepoint == -1) {
+            continue;
+        }
+
+        unichar_t glyph_unistr[2] = {0, 0};
+        glyph_unistr[0] = (unichar_t)codepoint;
+        char* glyph_utf8 = u2utf8_copy(glyph_unistr);
```
skef, line 488: [C] `glyph_utf8` is never freed — `u2utf8_copy()` allocates and nothing releases it.

It's also allocated unconditionally but only used in the `encoded` branch below, so every unencoded slot allocates a string and drops it unread.

This is once per slot, per line, per page draw. Since `draw_preview_area()` calls `activate_cairo_printer()` and redraws on every expose event (print_preview.cpp:717), it leaks continuously for as long as the dialog is open, not just once per print job.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+
+        Cairo::Rectangle glyph_slot{margin_ + left_code_area_width +
+                                        extrahspace +
+                                        j * (extrahspace + pointsize),
+                                    y_start, pointsize, pointsize};
+        if (glyph_line.encoded) {
+            draw_centered_text(cr, glyph_slot, glyph_utf8);
```
skef, line 495: [C] The two halves of this grid resolve glyphs by different mechanisms. Encoded slots go through `draw_centered_text()` → `cr->show_text()`, the Cairo *toy* API, which looks the codepoint up in the font's cmap. Unencoded slots go through `draw_centered_glyph()` → `show_glyphs()` with the TTF index carried in `print_map_`.

For a well-formed font these agree. Where they don't — several glyphs mapped from one codepoint, or a selection that doesn't match what the cmap resolves to — the encoded half shows the cmap's choice rather than the glyph the user actually selected in the FontView, with no indication. Given this dialog is for proofing a font under development, when the cmap is exactly the thing that might be wrong, I'd expect both halves to go through the glyph index from `print_map_`.

Is there a reason the encoded path needs `show_text()` rather than `show_glyphs()` with `.first`?

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+    extrahspace_ = pointsize_ / 3;
+
+    double char_area_width =
+        printable_area.width - margin_ * 2 - left_code_area_width_;
+    double char_area_height =
+        printable_area.height - margin_ - top_margin_ - top_code_area_height_;
+    int max_slots = static_cast<int>(
+        std::floor(char_area_width / (extrahspace_ + pointsize_)));
+
+    glyph_lines_ = split_to_lines(max_slots);
```
skef, line 329: [C] The constructor does the entire layout: `split_to_lines()` walks all of `print_map_` calling `SCGetEncoding()` per glyph and building a `std::map<int, std::vector<int>>`, then `paginate()` runs on the result.

That would be fine once per print job, but `draw_preview_area()` calls `activate_cairo_printer()` on *every* redraw (print_preview.cpp:717), which constructs a fresh printer each time — so the whole font is re-bucketed and re-paginated on every expose event: every scroll, resize and `queue_draw`.

Same shape as the `kScaleMaxHeight` extents pass in `FullGlyphPrinter`. Both want caching, and in both cases it can't go in the printer constructor for exactly this reason — it has to live on `CairoPainter` or be memoized against the font.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+    }
+
+    return glyph_lines;
+}
+
+void FullDisplayPrinter::paginate(double char_area_height, double pointsize,
+                                  double extravspace) {
+    int max_lines = static_cast<int>(
+        std::floor(char_area_height / (extravspace + pointsize)));
+    if (max_lines <= 0) {
```
skef, line 427: [C] Worth noting this guard exists here but the equivalent is missing in `MultiSizePrinter`: `page_count()` at line 306 divides by `lines_per_page_` with no check, and that value is computed the same way from page geometry. So the degenerate-page-size case is handled in this class and faults in the other one.

On the guard itself: `glyph_line_pagination_ = {0}` gives `page_count() == 1`, and `add_page()` then takes `end_line_it = glyph_lines_.end()` and draws every line onto that single page, running off the bottom. That's better than a fault, but it's silent. Clamping `max_lines` to 1 and letting normal pagination proceed would at least keep the output readable.

fontforge/layout/cairo_painter.cpp
```diff
@@ -0,0 +1,1294 @@
...
+    cr->set_source_rgb(0, 0, 0);
+
+    size_t line_length =
+        glyph_lines_.empty() ? 16 : glyph_lines_[0].indexes.size();
+
+    static const std::array<std::string, 16> slot_labels = {
+        "0", "1", "2", "3", "4", "5", "6", "7",
+        "8", "9", "A", "B", "C", "D", "E", "F"};
+    for (size_t i = 0; i < line_length; ++i) {
+        Cairo::Rectangle slot{margin_ + left_code_area_width_ + extrahspace_ +
```
skef, line 527: [C] These header labels rely on the font state left behind by `init_document()`, which ends by selecting `"times"` bold at 12 pt (lines 1238-1240) to draw the page title. Nothing here sets a face or size.

`draw_glyphs_line()` sets it explicitly for the row labels (lines 472-474), so the row labels and the column headers — which are meant to match — get their styling from two different places. Change what `init_document()` leaves on the context and the column header silently changes with it. Worth setting it explicitly here too.

skef (commented):
A set of initial comments covering the start of the review.

iorsh:
# Summary of 2026-Aug-17 discussion

## XML Manifest

The exported XML file shall have a manifest-like section at the beginning which lists all the fonts which were used in the text sample creation. The listed fonts shall carry family, font name, attributes such as weight / stretch etc.

The XML body shall refer to the fonts listed in the maifest.

## Font substitution logic

The substitution logic is responsible for matching a font from XML manifest to an actually available font. This logic shall be separate from the printing module. The printing module shall accept fonts which are already resolved in the current session.

## Print UI

The UI shall not have generic and specific modes. The fonts shall be grouped by family (first drop-down list), and once the family is selected the user should be able to specify font by name (second drop-down list) or attributes. The controls shall update to reflect consistently the user's choice. For example, if the user chose a bold font from the list of fonts, the "Bold" button should be pressed. Conversely, the active font in the list of fonts shall follow the selected attributes.

iorsh:
I'm planning the following XML structure:
```xml
<ff_root>
  <manifest>
    <font key="FrankRuehlCLM-Medium" default="true">
      <family>Frank Ruehl CLM</family>
      <name>FrankRuehlCLM-Medium</name>
      <weight>regular</weight>
    </font>
    <font key="FrankRuehlCLM-BoldOblique">
      <family>Frank Ruehl CLM</family>
      <name>FrankRuehlCLM-BoldOblique</name>
      <weight>bold</weight>
      <italic>true</italic>
    </font>
    <font key="NachlieliCLM-Light">
      <family>Nachlieli CLM</family>
      <name>NachlieliCLM-Light</name>
      <weight>light</weight>
    </font>
  </manifest>
  <body>
    <segment font="FrankRuehlCLM-Medium">Sample text in the default font</segment>
    <segment font="FrankRuehlCLM-BoldOblique">Sample text in bold oblique font</segment>
    <segment font="NachlieliCLM-Light">Sample text in different font</segment>
  </body>
</ff_root>
```
@skef, please take a look.

skef:
@iorsh I think this will work well. The way I see it is that there are two natural approaches, one of which is this one where the name of the font is the "index" or "key", and the other where the font section is a list and the offset into the list is the "index" or "key". The main advantages to this approach are that it is more "self-documenting" in the file and seems like better XML practice.

The trade-off is that when one hands off a font remapping to the lower-level facilities (presumably along with a parse tree of the body), that remapping will be in terms of font name. But I don't see any problem with that. The important thing (in my view) is to keep the remapping *logic* at a higher level, so that what happens at the lower-level is deterministic. The fact that it is in terms of a name seems fine.

You'll have to decide whether the remapping is optional, so that if there isn't a map entry for a font the lower-level will try to grab one with that name (and if it doesn't find it that's presumably an error), or if the mapping is always present and often just maps same-to-same. The latter seems cleaner to me, both in that it removes more logic from the lower level and because the other side of the mapping can be the actual font structure instead of a string.

iorsh:
I prefer the remapping to happen at the higher level, so that the low-level printing module receives data which is guaranteed to be supported by available fonts. I understand that this approach would be in line with your preference too.

The decision to use font name in `<segment font="FrankRuehlCLM-Medium">` is strictly for readability of XML files if the user wants to open them with plain text editor. A font name mentioned in the `segment` should not carry any meaning besides being a text-based key into the manifest. If the `body` contains something like `<segment font="Arial-Bald">`, but Arial-Bald (sic.) is not described in the manifest, that XML file would be considered broken. The program will not attempt to guess the intention behind it.

PR #5847, opened 2026-07-01
Simplify the scalable icon

linkmauve:
The original file had its black circle drawn using a total of 2048 straight lines, whereas SVG already has a primitive for circles, which was already used for the white disc inside.

By replacing the black circle with the outline of the white disc, and simplifying a bunch of other things (such as removing unneeded matrix transforms), this file went down from 33 KiB to just 1.5 KiB, or gzipped from 13 KiB to just 820 B.  This should also help with parsing and rendering time, as found out using GTK’s iconview demo.

Thanks @benjamin-otte for finding this bug!

### Type of change

- **Bug fix**
- ~**New feature**~
- ~**Breaking change**~
- **Non-breaking change**


iorsh (approved):
Looks good to me

PR #5850, opened 2026-07-09
Improve menu interactions (hover, sticky menus)

### LINKED ISSUE #2064: Strange menu behavior
I'm new in FontForge and have faced its non-intuitive style of navigation through the main menu (in both Main window and Glyph window).
<ol>
<li>Entries are highlighted only after being <b>clicked</b>&mdash;not when hovered. But if I click an <i>inactive</i> entry in any submenu then entries become highlighted when just <b>hovered</b>.</li>
<li>If I navigate with keyboard arrows, when I reach "Tools" which is inactive, I cannot move to another entry as the highlighting gets stuck to the "Tools".</li>
<li>I failed to find any hotkey that brings me to the main menu (like Alt+F10 in many programs).</li>
</ol>
Are (1) and (2) bugs or that's a reflection of unique and inimitable design?

FontForge version 20141230 from ppa:fontforge/fontforge
OS Ubuntu 14.04

nicoverbruggen:
I've been using FontForge a lot recently on both Fedora KDE and macOS, and I've noticed that the menu behaves oddly. Menus are sticky, i.e. you need to click and sometimes hovering over menu entries doesn't seem to work properly. I never know if the click I'm about to make will actually make it through.

A clicked main menu item also doesn't look properly selected when it's active and there was obvious padding missing as well.

I've tried to address these things in this PR, which was created with some assistance from LLMs, I should note. I've checked the code myself and tested the fixes on macOS, but not on Linux (yet).

So, this PR:

- Ensures that clicks outside of the window consistently dismiss menus
- Addresses sticky menus (hover + clear-on-leave)
- The active title now draws a padded, bar-height highlight instead of a cramped text-tight fill, which looked off

### Changes

| Area | Files |
|------|-------|
| Quartz first-click dismiss | `gdraw/ggdkdraw.c`, `gdraw/ggdkdrawP.h` |
| Sticky hover + menubar switching + highlight padding | `gdraw/gmenu.c` |
| Menubar highlight color | `fontforgeexe/pixmaps/tango/resources.in` |

For comparison, here's a demonstration to illustrate how much smoother things become (with Tango theme on):

### Before

https://github.com/user-attachments/assets/d6b47d73-8a5b-4406-a24a-10a2faced5a9

This is especially frustrating when I have a menu open and I want to navigate to another menu; clicking on it toggles the menu and immediately dismisses it.

### After

https://github.com/user-attachments/assets/60eda7b7-4dff-474c-9d07-662f7cafb2f2

Navigating with the mouse has been clunky, and this improves things significantly.

### Type of change

- **Bug fix** - fixes #2064
- **Non-breaking change**

If any further changes are required, I'd be happy to make them, or if you the original menu click behaviour was as intended, feel free to close this PR. I think it's a significant improvement either way, so I wanted to at the very least open a PR to hear if there's any interest in merging this in.

### Notes

- If necessary, the comments can be stripped if they're too descriptive.
- I haven't tested on Windows and/or Linux. Can someone check?
- For a better user experience, the highlight color for disabled entries could be a different color?

### DIFF
--- a/fontforgeexe/pixmaps/tango/resources.in
+++ b/fontforgeexe/pixmaps/tango/resources.in
@@ -440,7 +440,7 @@ Gdraw.GGroup.Box.Padding: 0
 Gdraw.GMenuBar.Box.BorderOuter: False
 Gdraw.GMenuBar.Box.GradientStartCol: #e5e4e3
 Gdraw.GMenuBar.Box.BorderInnerCol: #edeceb
-Gdraw.GMenuBar.Box.ActiveBorder: #ededed
+Gdraw.GMenuBar.Box.ActiveBorder: #88b2de
 Gdraw.GMenuBar.Box.BorderShape: rect
 Gdraw.GMenuBar.Box.BorderWidth: 1
 Gdraw.GMenuBar.Box.Radius: 2
--- a/gdraw/ggdkdraw.c
+++ b/gdraw/ggdkdraw.c
@@ -1010,10 +1010,20 @@ static void _GGDKDraw_DispatchEvent(GdkEvent *event, gpointer data) {
             Log(LOGDEBUG, "Button %7s: [%f %f]", evt->type == GDK_BUTTON_PRESS ? "press" : "release", evt->x, evt->y);
 
 #ifdef GDK_WINDOWING_QUARTZ
-            // Quartz backend fails to give a button press event
+            // Quartz backend fails to give a button press event when the click
+            // lands outside a grabbed window (e.g. clicking off an open menu to
+            // dismiss it): only the release is delivered, never the press.
             // https://bugzilla.gnome.org/show_bug.cgi?id=769961
+            //
+            // Synthesize the missing press when an out-of-bounds release arrives
+            // and no real press is currently held down (which is what tells us
+            // the press was swallowed, rather than this being the release of a
+            // genuine press-inside/drag-outside gesture). Gating on the button
+            // actually being down fixes dismissal on the *first* click; the old
+            // heuristic (release_w == gw) only became true after a first,
+            // wasted click had already set release_w to this window.
             if (!evt->send_event && evt->type == GDK_BUTTON_RELEASE &&
-                    gdisp->bs.release_w == gw &&
+                    !gdisp->bs.is_pressed &&
                     (evt->x < 0 || evt->x > gw->pos.width ||
                      evt->y < 0 || evt->y > gw->pos.height)) {
                 evt->send_event = true;
@@ -1040,12 +1050,19 @@ static void _GGDKDraw_DispatchEvent(GdkEvent *event, gpointer data) {
                     gdisp->bs.cur_click = 1;
                 }
                 gdisp->bs.last_press_time = gevent.u.mouse.time;
+                // Track only genuine presses; a synthetic press (see the Quartz
+                // workaround above) is marked send_event and must not register as
+                // a real button-down, or the paired release would look unbalanced.
+                if (!evt->send_event) {
+                    gdisp->bs.is_pressed = true;
+                }
             } else {
                 gevent.type = et_mouseup;
                 gdisp->bs.release_w = gw;
                 gdisp->bs.release_x = evt->x;
                 gdisp->bs.release_y = evt->y;
                 gdisp->bs.release_button = evt->button;
+                gdisp->bs.is_pressed = false;
             }
             gevent.u.mouse.clicks = gdisp->bs.cur_click;
         }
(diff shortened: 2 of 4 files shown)


iorsh (commented):
The changes look reasonable, but this code supports a very old and stable behavior of a crucial UI element. Changing it bears obvious risk of instability or new severe bugs. 

I'm fine with approving it, but testing must be done on all platforms, preferably in various window environments.

Please elaborate on your testing, which platforms / windows managers you use. Please test popup menus too.

iorsh:
Please also note that there is a major project underway to transition UI to GTK3. You can see #5545, which introduces behavior similar to the one you pursue. Your change is still welcome though, as GTK3 transition may take years, and your proposed menu improvement is immediate.

nicoverbruggen:
Exactly, I had seen that GTK3 UI changes were underway and suspected an interim fix might be nice.

I'm currently using a build with these patches for my own font projects on my Apple silicon Mac, but I will definitely take time to do a more thorough pass and also test on Windows and Linux to make sure everything behaves as expected.

Is testing on both KDE and GNOME desktop environments sufficient for Linux, or should I test more DE's too? Should I also aim to test on older versions of Windows (e.g. Windows 10)? 

I'm happy to take the time to test this as extensively as necessary to make sure this breaks nothing as menu functionality should be rock solid.

iorsh:
For Linux it would be necessary to test Gnome / KDE on X11 / Wayland (4 combinations total). I've had issues with each one of them when working on GTK, even if GTK3 is supposed to be well settled and platform-agnostic. No need for other window managers, I think. We shall address their specific problems later if such problems arise.

For Windows, I guess one environment is ok, since I'm not aware of any major window management overhaul Microsoft did in the recent years. Windows 10 would be a nice addition, if you have it readily available.

For Mac, I have no idea how different its OS versions are in terms of window management. Please check if your system is sufficiently representative.

nicoverbruggen:
Okay, I can definitely do that with virtual machines + my own setup which is mostly Wayland + KDE and GNOME.

For macOS, I can easily test the last three versions of the OS, as I have various virtual machines that I use for testing my own native Mac apps. I hope that's sufficient, but if anyone can advise on what the minimum supported version of macOS is for FontForge, I'd be happy to make a best effort to validate that, too.

iorsh:
A more important comment (see #2064): it looks like the desired behavior is already available, but somehow not manifested until you click an inactive menu item. Could you please check if you can ride on this finding and get what you want with less changes? Perhaps this "inactive click" triggers something that can be activated right away?

skef:
Note: I have a big menu PR coming down the pike and I might want to hold off merging this until that can be sorted out. (Although it looks quite modest.)

nicoverbruggen:
> A more important comment (see #2064): it looks like the desired behavior is already available, but somehow not manifested until you click an inactive menu item. Could you please check if you can ride on this finding and get what you want with less changes? Perhaps this "inactive click" triggers something that can be activated right away?

After looking at #2064 it seems that the inactive click leaves the menu’s `pressed` state set because the disabled mouse-up path returns before clearing it, which explains why hover starts working afterward. I’ve pushed a follow-up commit that makes hover handling depend on the active menu selection instead of button state, which should be better.

iorsh:
> Note: I have a big menu PR coming down the pike and I might want to hold off merging this until that can be sorted out. (Although it looks quite modest.)

Does your PR directly clash with these changes? I was under impression that you are going to mostly change and regroup the menu items arrangement, but maybe I missed something.

skef:
I can take a look at that later, but I think probably not?

nicoverbruggen:
Okay, I'm going to be testing over the next week. I've been using this build w/ the updated menus heavily so far and I feel like it's a massive improvement. I haven't encountered any bugs so far either, from what I can see.

I've tested on:

- [x] macOS Tahoe
- [x] macOS Sequoia
- [x] CachyOS (KDE + Wayland)
- [x] Windows
- [x] Fedora Workstation (GNOME + Wayland)

I'm not running X11 anymore, so will require VMs later:

- [ ] KDE + X11
- [ ] GNOME + X11

skef:
I've only glanced at this but consider my concerns non-blocking. Iorsh can review and merge as he sees fit.

nicoverbruggen:
I'll resume testing on X11 today, I'll be spinning up some VMs to do this. I just tested on Windows, which seems to be OK for this change!

Have any of you been able to give this a spin to see if the menus work fine on your end as well?

Edit: Whoops, apparently Fedora only supports Wayland now. I was only able to try GNOME + Wayland today, which seemed to work fine as well.

iorsh:
I'll try to check a few linux configurations tonight.

iorsh (approved):
I tested on Ubintu 24: Plasma and Gnome, each with X11 and Wayland. Everything works fine, so I think we are good to go with it.

PR #5852, opened 2026-07-15
macenc.c patch (3 bugs + some improvements)

Yukinoroh:
This patch addresses three bugs:
* Presence of Windows language 0xFFFF generating Kurdish Mac names
* No mac names generated for AAT feature 22 option 1, and bogus option 3 generated
* Mistake in French name for feature 103 option 0

Also added:
* Some mac->windows language mappings
* otf->aat mapping of pkna feature

P.S.: The setting and name structures are not easy to edit. Appending stuff at the end is for now the less risky solution.

=====

_WinLangFromMac[] structure:
	- added some mappings from Microsoft's MS-LCID version 16.0 (2024)
	- changed 0xFFFF mappings to 0x1000 (Microsoft's specification: undefined)

macfeat_otftag[] structure:
	- fixed feature 25 option 1 to table "pkna"
	- fixed feature 22: options 0,2,3 were present but their names suggested 0,1. Added translations for option 2
	- fixed French translation of feature 103 option 0 from "Pleine Taille" to "Demi-taille"

WinLangToMac function:
	- added conditional statement to avoid lookup of 0xFFFF and 0x1000 languages.


Yukinoroh:
Any ideas why the macos checks fail?

Yukinoroh:
Part of my fixes were about fixing win to mac conversion, but I notice that there isn't any check done the other way around. If mac to win conversion gave language 0xFFFF, now it will give 0x1000, which is at least more in line with Microsoft's specification.

FontForge loads language 0xFFFF without any issues - it just shows it as "65535". So I guess language 0x1000 will show as "4095". I am not sure how it will react in presence of multiple 0x1000 languages, though. (Theorically possible, just as until now it was theorically possible to end up with many 0xFFFF languages from mac conversion.)

Giving a better feedback when loading 0x1000 and 0xFFFF languages might be more helpful than showing integer value. Something like 0xFFFF = Unknown, and 0x1000 = Undefined, both localized in user's language. I don't know.

fontforge/macenc.c
```diff
@@ -881,68 +881,68 @@ static uint16_t _WinLangFromMac[] = {
...
-/*120*/	0xffff,
-	0xffff,
-	0xffff,
-	0xffff,
-	0xffff,
-	0xffff,
-	0xffff,
-	0xffff,
+/*90*/	0x487,		/* Kinyarwanda/Ruanda */
+	0x1000,		/* Rundi/Kirundi */
```
iorsh: Please make `0x1000` a named constant

Yukinoroh: Allright. Give me some time.

iorsh (approved):
Looks good to me.

Yukinoroh:
I used a define statement; let me know if it works.

iorsh:
> I used a define statement; let me know if it works.

That's fine with me.

PR #5853, opened 2026-07-16
Numerically stabilize InterpolateVal() in Style transformations

### LINKED ISSUE #2647: Italic filter breaks some characters than crashing
On some characters (such as μ), the italic filter seems to get "confused" and encounters several NaNs in its calculations. This would be fine in and of itself (I'm sure the "italicizing" process is quite complicated) if the filter backed out the changes and left the remaining work to the user, but instead it leaves the NaN coordinates in:

A screenshot of the problem

To make matters worse, opening the glyph in a new window or clicking in the existing outline window causes FontForge to crash. To reproduce, italicize the single character in [this font](https://github.com/fontforge/fontforge/files/181113/bugdemo.sfd.zip) with default settings.

Core dump:
[fontforge.dump.zip](https://github.com/fontforge/fontforge/files/181105/fontforge.dump.zip)

### LINKED ISSUE #5812: Bug: FontForge crashes when trying to convert the LoveYaLikeASister font to italics
Hi!

I just want to report this issue.

I'm trying to use FontForge to convert this Google Font to italics: https://fonts.google.com/specimen/Love+Ya+Like+A+Sister

I'm using the latest version of FontForge for Windows (I also tested a previous version from 2 years ago and this bug is also present):
```
FontForge Windows build (64-bit)
Thu, Oct  9, 2025  7:07:22 PM +0000
c41bdb922285f35defd1e1385adfd13bde1ab32a [master]
Based on master: c41bdb922285f35defd1e1385adfd13bde1ab32a
```

There are the steps to reproduce the crash:
* open FontForge and load the font file `LoveYaLikeASister-Regular.ttf`
* select all the elements with CTRL+A (or just the `e` letter)
* go to `Element > Style > Italics`
* press `OK`

Then FontForge performs the conversion for the first characters and crashes on the `e` letter.

iorsh:
Fixes #2647, fixes #5812

### DIFF
--- a/fontforge/scstyles.c
+++ b/fontforge/scstyles.c
@@ -138,7 +138,7 @@ return ((val < (&psp->me.x)[xdir] && val < (&nsp->me.x)[xdir]) ||
 }
 
 /* This is not exactly an interpolation. We are positioning a middle point
- * between two new values in the same proportion as the the old middle point was
+ * between two new values in the same proportion as the old middle point was
  * between two old values. */
 static double InterpolateVal(double a, double b, double a1, double b1,
                              double val) {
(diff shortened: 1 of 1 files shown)


skef (approved):
You've got a double "the" but looks good otherwise.

PR #5857, opened 2026-07-18
Warning in AI_TestOrdering() upon reverse anchor order

### LINKED ISSUE #5717: Anchors on RTL Ligatures
When editing anchors on a ligature glyph that is meant for use in an RTL script, you can get a modal dialog box complaining that the anchors are out of order.  What's worse is that once you hit "OK", fontforge apparently tries to fix the problem by incrementing the ligature-element index of the anchor, which only causes the same error again, and no matter what you try you aren't given a chance to fix the problem; the program goes into an endless loop that you have to kill it out of.  (This may actually not be specifically due to the ligature being RTL, but it's what's causing the problem for me, since the anchors kind of need to be "out of order" in such a case.  There's no way to tell FF that the direction of writing is RTL.)

Note that I mean a ligature glyph that is _not_ a normal Unicode-encoded glyph, but an unencoded glyph (or one in the PUA, which is universally LTR per Unicode).  Working on cooking up an example; there are a lot of moving parts to this bug.

Maybe make the error non-modal, or not try to "correct" the problem?  I've managed some workaround because if you order your activities right, you can enter an "invalid" location and have it not get around to checking it.

iorsh:
Fixes #5717

### DIFF
--- a/fontforgeexe/cvgetinfo.cpp
+++ b/fontforgeexe/cvgetinfo.cpp
@@ -1052,7 +1052,7 @@ static void AI_TestOrdering(GIData *ci,real x) {
 		( aps->lig_index>ap->lig_index &&
 		    (( isr2l && aps->me.x>x) ||
 		     (!isr2l && aps->me.x<x))) ) {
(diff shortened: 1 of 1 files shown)


skef (approved):
LGTM

PR #5861, opened 2026-07-20
Support glyph export for bitmap-only fonts

### LINKED ISSUE #5526: Glyph.Export() makes empty PNG images
I can open ter-x16b.pcf in FontForge UI and I can export all glyph one by one to PNG format.
But when I try do the same thing with Python script:

I've very simple script:
```
#!/usr/bin/env python3

import fontforge
import pdb

font = fontforge.open("ter-x16b.pcf")

for glyph in font.glyphs():
    # pdb.set_trace()

    if glyph.isWorthOutputting():
        glyph.export(f"{glyph.glyphname}.png")
```

I've only blank white PNG.
A

It's reproduced at Linux, MacOS.
fontforge version is 20230101

iorsh:
Support `glyph.export()` from bitmap-only fonts (e.g. PCF). Current code always attempts to rasterize `SplineChar` object, which is obviously empty when the font is bitmap-only.

### Type of change

- **New feature** - fixes #5526


skef:
This needs to bake a little longer. Two problems:

1) The base.clut assignment is given a pointer to something about to go out of scope.
2) test_bitmap_export.py uses "pointsize" where it probably means "pixelsize".

iorsh:
> This needs to bake a little longer. Two problems:
> 
>     1. The base.clut assignment is given a pointer to something about to go out of scope.
> 
>     2. test_bitmap_export.py uses "pointsize" where it probably means "pixelsize".

Fixed.

fontforge/splinefill.c
```diff
@@ -702,22 +702,8 @@ static void InitializeHints(SplineChar *sc, EdgeList *es) {
...
-	free(bdfc->bitmap);
-	bdfc->bitmap= bitmap;
-	bdfc->bytes_per_line = bpl;
-    }
-}
-
-void BCRegularizeGreymap(BDFChar *bdfc) {
-    int bpl = bdfc->xmax-bdfc->xmin+1;
+void BCRegularizeBitmap(BDFChar *bdfc, int bitsperpixel) {
+    int bpl =(bdfc->xmax-bdfc->xmin)*bitsperpixel/8+1;
```
skef, line 706: Can bitsperpixel only be 1 or 8? This changes the logic at intermediate values (e.g. 4).

fontforge/cvexport.c
```diff
@@ -531,102 +532,101 @@ return(0);
...
-	/* Sigh. Bitmaps use a different defn of set than images do. make it consistent */
-	tot = bdfc->bytes_per_line*(bdfc->ymax-bdfc->ymin+1);
-	for ( pt = bdfc->bitmap, end = pt+tot; pt<end; *pt++ ^= 0xff );
+    if (sc->parent->onlybitmaps) {
+        BDFFont* bdf;
+        for (bdf = sc->parent->bitmaps; bdf != NULL; bdf = bdf->next) {
+            bdfc = bdf->glyphs[sc->orig_pos];
+            if (bdfc != NULL) break;
+        }
+        if (bdfc->byte_data) {
```
skef, line 568: Is this safe?

iorsh: Probably not. Fixed.

fontforge/splinefill.c
iorsh: Fixed.

Officially only 1 or 8 - see https://fontforge.org/docs/scripting/python/fontforge.html#fontforge.glyph.export. In fact, 2/4 are not blocked and work for BMP. For PNG - they produce bad  file (I fixed it in https://github.com/fontforge/fontforge/pull/5861/commits/41a937acd95b8cc226bc870812a6195ea1ced0f1) and for some formats they are silently replaced with 1.

I think we can now allow 2/4 bits in the documentation too.

fontforge/cvexport.c
```diff
@@ -531,102 +532,106 @@ return(0);
...
-		(int) rint(sc->width*pixelsize/emsize),
-		(int) rint(sc->parent->ascent*pixelsize/emsize));
-
-	/* Sigh. Bitmaps use a different defn of set than images do. make it consistent */
-	tot = bdfc->bytes_per_line*(bdfc->ymax-bdfc->ymin+1);
-	for ( pt = bdfc->bitmap, end = pt+tot; pt<end; *pt++ ^= 0xff );
+        if (bdfc->byte_data) {
+            int depth = BDFDepth(bdf);
+            bdfc = BCScaleGrey(bdfc, bdf->pixelsize, depth, pixelsize, depth);
+            bitsperpixel = 8;
```
skef: Given that variable grayscales work now, is this right?

iorsh: Fixed with warning, conversion between depths in bitmap characters doesn't work.

skef (approved):
LGTM

PR #5875, opened 2026-08-09
Do not require Python distutils module

bgermann:
The distutils module is deprecated and not in the stdlib from 3.10 on. Make the code able to run without distutils.

### Type of change
- **Non-breaking change**

### DIFF
--- a/pycontrib/FontCompare/setup.py
+++ b/pycontrib/FontCompare/setup.py
@@ -15,7 +15,6 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 #!/usr/bin/env python
-from distutils.core import setup
 from setuptools import setup 
 setup(name='Font Compare',
       version='1.0',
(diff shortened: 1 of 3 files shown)


pycontrib/FontCompare/setup.py
```diff
@@ -15,7 +15,6 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 #!/usr/bin/env python
-from distutils.core import setup
```
iorsh, line 18: The contributed scripts are usually not maintained, and not tested.

The change is ok if you tested it, otherwise you can just leave the script as is, the dependencies of this script are responsibility of its users.

iorsh:
I'm not sure I understand the merit of keeping and maintaining our own (probably very outdated) copies of CMake recipes. Maybe we can just drop them altogether and rely on CMake distribution to contain updated `.cmake` stuff, at least for Python?

@bgermann, I don't expect you to actually do that, only if you want to give it a shot, and it turns out easy. But we do support Python 3.8 and later, so these changes must be tested for all these versions.

@skef, what do you think?

PR #5877, opened 2026-08-09
Deep-copy MATH and variant data in SplineCharCopy to fix use-after-free in MergeFonts

### LINKED ISSUE #5876: use-after-free caused by SplineCharCopy
Please see https://bugs.debian.org/948876. Citing the most accurate part of the bug:

I tried to look into this issue without being involved
in packaging fontforge.
I found it most reproducible when building with
"-fsanitize=address", and then always failing on accessing
the same address. [1]

As far as I see this is what happens:

- Address 0x60400008a210 gets returned by the allocator [2],
  and stored in "sf->glyphs[49391]->vert_variants".

- Memory gets freed below SplineFontFree while still
  stored below "sf->..." [3].

- Address 0x60400008a210 gets returned a second time.
  This is returned as the previous allocation by AddressSanitizer [1].

- And freed again.

- The first pointer gets further copied around (See attached file.)

- Now in gv_len this address is again accessed and causes the crash. [1]

(Is there a way to force AddressSanitizer to return unique memory addresses?)
The line numbers of the AddressSanitizer outputs do not
completely match because of some added fprintf's.

A temporary workaround could be to disable the call to
SplineFontFree in _MergeFont. Then no crash happens.

bgermann:
SplineCharCopy copied the source SplineChar's vert_variants, horiz_variants, italic_adjusts, top_accent_adjusts, mathkern, user_decomp and glif_name pointers instead of owning copies. When fonts were merged with MergeFonts the source font was freed (SplineFontFree in _MergeFont), leaving the merged glyphs referencing freed memory. Reading or writing MATH data afterwards (for example gv_len while dumping the MATH table, or freeing the merged font) hit freed memory, crashing fontforge.

This is the bug reported as Debian bug #948876 (kodi FTBFS caused by a fontforge segfault in gv_len after merging fonts with MATH data).

Also complete GlyphVariantsCopy, which did not copy the italic_correction field.

Add a regression test that merges two fonts carrying MATH data, verifies the merged data, generates a TTF, and closes the font; it fails deterministically without the fix.

Fixes: #5876

### Type of change

- **Bug fix**
- **Non-breaking change**


tests/test_merge_math.py
```diff
@@ -0,0 +1,75 @@
...
+    p.moveTo(0, 0)
+    p.lineTo(100, 0)
+    p.lineTo(100, 100)
+    p.lineTo(0, 100)
+    p.closePath()
+    return g
+
+# The MATH data must be copied into the merged font, not shared with the
+# (freed) source font.
+srcdir = os.environ.get("FONTFORGE_TEST_FONT_DIR", "")
```
iorsh: Please use `tempfile` - see https://github.com/fontforge/fontforge/blob/aea51e80b428af232ec0a22d139ca3ee58715aa9/tests/test_persistent_payload.py#L39

bgermann: Done.

iorsh (approved):
Looks good to me

PR #5893, opened 2026-09-09
Remove unused icons

iorsh:
Some icons are probably artifacts from ancient times, and they were superseded by newer designs with different names. Most of them are not in use, one was still in use by mistake alongside the newer design.


skef (approved):
LGTM

PR #5895, opened 2026-09-14
fvimportbdf: bounds-check PCF glyph bitmap offsets (heap OOB read)

eunho87:
### What

`PcfReadBitmaps()` in `fontforge/fvimportbdf.c` reads per-glyph bitmap data from
a PCF font using a file-controlled `offsets[]` array without validating it
against the allocated bitmap buffer size (`sizebitmaps`):

```c
memcpy(bc->bitmap, bitmap+offsets[i],
       bc->bytes_per_line * (bc->ymax-bc->ymin+1));
```

`offsets[i]` comes straight from the file (`getformint32`) and is never checked.
A crafted `.pcf` therefore causes a heap out-of-bounds read (CWE-125): a negative
offset reads before the buffer, a large one past it. The non-1 glyph-pad branch
has the same issue.

### Reachability

`.pcf` files are auto-detected by magic (`\x01fcp`) in `_ReadSplineFont()` and
routed to `SFFromBDF(...,2,...)` -> `SFImportBDF` -> `PcfParse` -> `PcfReadBitmaps`,
so simply opening a crafted font reaches it.

### Reproduction

Take a valid PCF (e.g. `6x13.pcf`) and set the first BITMAPS-table `offsets[]`
entry to `0xFFFFFFFF` (-1). Opening it under AddressSanitizer reports:

```
heap-buffer-overflow READ of size 1, 1 bytes before a 214292-byte region
  PcfReadBitmaps  fvimportbdf.c:1710
  PcfParse        fvimportbdf.c:1831
  SFImportBDF / SFFromBDF / _ReadSplineFont (splinefont.c:1144)
  bitmap allocated at PcfReadBitmaps fvimportbdf.c:1683
```

### Fix

Validate `offsets[i]` and the full glyph extent against `sizebitmaps` before each
memcpy in both glyph-pad branches; skip the glyph otherwise. This is the same
class of bug recently fixed in libXfont2's `pcfReadFont()` (CVE-2026-56002).

Impact is out-of-bounds read only (crash / adjacent-heap disclosure into the
glyph bitmap); not a write, so no code execution.


eunho87:
@iorsh — PCF bitmap parser OOB read, adjacent to your recent bitmap-font work (#5861). Minimal fix, ASAN repro is in the description. Would you be the right person to review this?

iorsh (approved):
Looks good to me.

PR #5899, opened 2026-09-19
Allow 0-len mult subs in glyph info dialog

chanicpanic:
This PR removes the 0-len restriction for multiple substitutions from the Glyph Info dialog.

#4619 allowed 0-len multiple substitutions to be created via the Font Info dialog. However, a similar restriction remained on multiple substitutions created via the Glyph Info dialog. As a consequence, a glyph that is given a 0-len multiple substitution via Font Info can no longer be edited via the UI's Glyph Info dialog because of the "You must specify a glyph name for subtable %s" error (at least not without temporarily removing the 0-len mult sub and adding it back via Font Info). 

### Type of change

- **Non-breaking change**


iorsh (commented):
Looks good to me

iorsh (approved):

